PR #13, opened 2026-04-05
Fix #12: re-extract OverlayServiceLogic and add unit tests

aunger:
Re-introduces OverlayServiceLogic as a properly-wired class that holds
the camera-callback, evaluateForeground, showOverlay, and
scheduleDeactivation logic extracted from OverlayService.  All
Android-framework side-effects are accessed via constructor lambdas so
the class is exercisable on the plain JVM.

OverlayServiceLogicTest covers the five scenarios called out in the issue:
- DT-05 multi-camera: second onCameraUnavailable does not prematurely
  schedule deactivation; only when the last tracked camera is released
  does scheduleDeactivation() run.
- Idempotent activation: evaluateForeground() multiple times with Pixel
  Camera in foreground calls overlayManager.show() exactly once and
  does not re-register the media observer.
- Usage-stats permission revocation: overlay is hidden and the
  usage-access-lost notification fires; no notification when overlay
  was already inactive.
- Overlay permission loss: notification fires and isOverlayActive
  remains false; no session is started.
- Session gating on keyguard state: showOverlay() starts a session and
  registers the media observer only when isKeyguardLocked is true.

https://claude.ai/code/session_01Rd6GKntfPmNN4o2J7i7qby

### DIFF
--- a/app/src/main/java/com/gb4pc/service/OverlayService.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayService.kt
@@ -1,5 +1,6 @@
 package com.gb4pc.service
 
+import android.app.KeyguardManager
 import android.app.Notification
 import android.app.PendingIntent
 import android.app.Service
@@ -39,9 +40,7 @@ class OverlayService : Service() {
     private lateinit var overlayManager: OverlayManager
     private lateinit var cameraState: CameraState
     private lateinit var handler: Handler
-
-    private var deactivateRunnable: Runnable? = null
-    private var isOverlayActive = false
+    private lateinit var logic: OverlayServiceLogic
 
     // H7: Track registration state to avoid re-registering on each onStartCommand
     private var callbackRegistered = false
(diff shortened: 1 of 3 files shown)


aunger (commented):
This closes the last open concern from the #10 review cleanly. The lambda-injection pattern is the right way to make a Service's internal logic JVM-testable without Robolectric, and the `debounceMs` parameter making the handler delay injectable for tests is a nice touch.

Walked through each of the five scenarios:

- **DT-05 multi-camera** — the `cancelPendingDeactivation()` guard in `onCameraUnavailable` prevents a premature deactivation runnable from lingering when a second camera comes in, and `scheduleDeactivation()` fires only once all cameras are released. The Runnable-captor pattern in the test is the correct way to exercise this without a real looper.
- **Idempotent activation** — the `!isOverlayActive` gate in `evaluateForeground()` makes this a one-liner to guarantee; the test's use of a second `logic2` instance with a counter is the right way to isolate the registration count from the boolean flag.
- **Permission revocation** — both the "active" and "inactive" branches are covered; the inactive branch correctly verifies that `hide()` is not called spuriously.
- **Overlay permission loss** — `isOverlayActive` stays false and the `overlayManager.show()` is never reached. The separate test for "no session started" is a good belt-and-suspenders check.
- **Session gating** — locked/unlocked paths both tested; `sessionTracker.startSession()` and `onRegisterMediaObserver` are verified as a matched pair.

One minor observation (not blocking): `showOverlay()`, `scheduleDeactivation()`, and `cancelPendingDeactivation()` are all public, which is wider than the production call sites need. Kotlin's `internal` modifier would tighten this without affecting the test (same module). Worth a follow-up but not worth holding the PR for.

Approved.

---

PR #17, opened 2026-04-07
Fix #14: make debug log text selectable and copyable

### LINKED ISSUE #14: Select and copy text from debug log
Currently, the text in the debug log viewer can’t be highlighted or copied.

aunger:
## Summary

- Wraps the debug log `LazyColumn` in a `SelectionContainer` so users can long-press, highlight, and copy entries from the Advanced Settings screen.

## Test plan

- [ ] Open Advanced Settings on a device/emulator
- [ ] Long-press a log entry — text selection handles should appear
- [ ] Copy selected text and paste elsewhere to confirm it works

Closes #14

### DIFF
--- a/app/src/main/java/com/gb4pc/ui/settings/AdvancedSettingsActivity.kt
+++ b/app/src/main/java/com/gb4pc/ui/settings/AdvancedSettingsActivity.kt
@@ -6,6 +6,7 @@ import androidx.activity.compose.setContent
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.text.selection.SelectionContainer
 import androidx.compose.material3.*
 import androidx.compose.runtime.*
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


aunger (commented):
## Code Review

**Verdict: COMMENT** — Correct implementation, no blocking issues. A few things worth knowing before merge.

---

### [minor] Selection state is lost on scroll

`SelectionContainer` tracks selection in memory tied to the composable subtree. With `LazyColumn`, items that scroll off-screen are disposed and recomposed when they scroll back in — any active selection spanning those items is silently cleared. For a debug log with up to `Constants.DEBUG_LOG_BUFFER_SIZE` entries this will happen routinely.

For this use case (developer-facing log, typically single-entry copies) this is an acceptable limitation. It should be acknowledged somewhere so it isn't rediscovered as a "bug" later. If full-range selection becomes a requirement, the fix is to switch from `LazyColumn` to a `Column` inside `verticalScroll` — at 200 short log strings the memory cost is negligible.

---

### [minor] `SelectionContainer` can interfere with fling scrolling

`SelectionContainer` installs a custom pointer-input handler to detect long-press. On some Android versions and Compose releases this handler can compete with `LazyColumn`'s scroll gesture recognizer, eating the fling when the user lifts after a long-press-then-drag. Worth a quick manual smoke test on a physical device before merge — emulators often don't reproduce it.

---

### [nit] Inline comment is redundant

```kotlin
// #14: SelectionContainer enables text highlight and copy
```

The PR title, the Compose type name, and the surrounding code all make this obvious. Remove it.

---

### [nit] Pre-existing: hardcoded "No log entries" string

Not introduced by this PR, but `"No log entries"` nearby is hardcoded while the rest of the screen uses `stringResource`. Worth a follow-up to move it to `strings.xml`.

---

aunger (commented):

The address-review commit goes further than requested in the best way: switching from `LazyColumn` to `Column + verticalScroll` fully solves the selection-state-loss problem rather than just working around it. At 200 short log strings the memory trade-off is negligible.

- **Selection state loss on scroll** → Fixed by `Column + verticalScroll`. Selection now persists across the full list. ✓
- **Fling interference** → No longer applicable with `Column`. ✓  
- **Nit comment** → Removed. ✓
- The new comment explaining _why_ `Column + verticalScroll` is used instead of `LazyColumn` is exactly right — keep it.

Good to merge.

---

PR #18, opened 2026-04-07
Fix #16: semver versioning with yyyyMMdd build number and release test suite

### LINKED ISSUE #16: Fix versioning
1. Although I asked for a 0.0.1 release, the app itself declares its at 1.0.0. Remove this release.
2. Switch versioning to be semver with a time-based build number.
3. Tie GH tags to the internal build number (or the other way around).
4. Add automated release test suite. Test the above requirements.

aunger:
## Summary

- Corrects `versionName` from `1.0.0` → `0.0.1`
- Replaces hard-coded `versionCode = 1` with a `yyyyMMdd` date-based build number (overridable via `BUILD_NUMBER` env var in CI)
- Enables `BuildConfig` generation so version constants are accessible in code and tests
- Adds `ReleaseVersionTest` with 4 assertions: semver format, valid 8-digit date code, date not in the future, and `v{versionName}` GitHub-tag convention

GitHub releases should be tagged `v{versionName}` (e.g. `v0.0.1`).

## Test plan

- [ ] `./gradlew testDebugUnitTest --tests "com.gb4pc.ReleaseVersionTest"` — all 4 tests pass
- [ ] `BuildConfig.VERSION_NAME` == `"0.0.1"` in a debug build
- [ ] `BuildConfig.VERSION_CODE` is today's date in `yyyyMMdd` form locally; overridable via `BUILD_NUMBER` env var

Closes #16

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,5 +1,6 @@
-import java.text.SimpleDateFormat
-import java.util.Date
+import java.time.LocalDate
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
 
 plugins {
     id("com.android.application")
@@ -10,8 +11,14 @@ plugins {
 // Versioning: semver versionName + yyyyMMdd time-based versionCode.
 // GitHub releases must be tagged v{versionName} (e.g. v0.0.1).
 // CI may override the build number via the BUILD_NUMBER env var.
-val buildNumber: Int = System.getenv("BUILD_NUMBER")?.toIntOrNull()
-    ?: SimpleDateFormat("yyyyMMdd").format(Date()).toInt()
+// BUILD_NUMBER must be a valid integer; a malformed value fails the build loudly.
+val envBuildNumber: String? = System.getenv("BUILD_NUMBER")
+val buildNumber: Int = when {
+    envBuildNumber == null ->
+        LocalDate.now(ZoneOffset.UTC).format(DateTimeFormatter.BASIC_ISO_DATE).toInt()
+    envBuildNumber.toIntOrNull() != null -> envBuildNumber.toInt()
+    else -> error("BUILD_NUMBER env var is set but not a valid integer: '$envBuildNumber'")
+}
 
 android {
     namespace = "com.gb4pc"
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,8 +1,17 @@
+import java.text.SimpleDateFormat
+import java.util.Date
+
 plugins {
     id("com.android.application")
     id("org.jetbrains.kotlin.android")
     id("org.jetbrains.kotlin.plugin.compose")
 }
+
+// Versioning: semver versionName + yyyyMMdd time-based versionCode.
+// GitHub releases must be tagged v{versionName} (e.g. v0.0.1).
+// CI may override the build number via the BUILD_NUMBER env var.
+val buildNumber: Int = System.getenv("BUILD_NUMBER")?.toIntOrNull()
+    ?: SimpleDateFormat("yyyyMMdd").format(Date()).toInt()
 
 android {
     namespace = "com.gb4pc"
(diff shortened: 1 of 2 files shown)


aunger (commented):
## Code Review

**Verdict: Changes needed** — The versioning strategy is sound, but two correctness issues in `build.gradle.kts` need to be fixed before merge, plus a few test improvements worth making now.

---

### [major] `SimpleDateFormat` with no explicit timezone — CI timezone skew

`SimpleDateFormat("yyyyMMdd").format(Date())` uses the JVM's default timezone. On Linux CI runners this is commonly UTC, but a developer in UTC+10 whose CI runs in UTC will produce a different `versionCode` on the same calendar day. The tests acknowledge timezone skew with a 1-day buffer but don't fix the root cause.

Pin to UTC and use `java.time` while you're at it (eliminates the `SimpleDateFormat`/`Date` imports too):

```kotlin
import java.time.LocalDate
import java.time.ZoneOffset
import java.time.format.DateTimeFormatter

val buildNumber: Int = System.getenv("BUILD_NUMBER")?.toIntOrNull()
    ?: LocalDate.now(ZoneOffset.UTC)
        .format(DateTimeFormatter.BASIC_ISO_DATE)  // produces "yyyyMMdd"
        .toInt()
```

---

### [major] Silent discard of an invalid `BUILD_NUMBER` env var

`toIntOrNull()` silently falls back to the date if `BUILD_NUMBER` is set but malformed (e.g. `BUILD_NUMBER=abc` or `BUILD_NUMBER= ` with trailing whitespace). A misconfigured pipeline produces a date-stamped build with no warning. This should fail loudly:

```kotlin
val envBuildNumber = System.getenv("BUILD_NUMBER")
val buildNumber: Int = when {
    envBuildNumber == null ->
        LocalDate.now(ZoneOffset.UTC).format(DateTimeFormatter.BASIC_ISO_DATE).toInt()
    envBuildNumber.toIntOrNull() != null -> envBuildNumber.toInt()
    else -> error("BUILD_NUMBER env var is set but not a valid integer: '$envBuildNumber'")
}
```

---

### [minor] Semver regex allows leading zeros

`^\d+\.\d+\.\d+$` accepts `01.0.0`, `1.00.0`, etc., which are [explicitly invalid per the semver spec](https://semver.org/#spec-item-2). For a project starting at `0.0.1` this is unlikely to bite you in practice, but the test is advertised as enforcing semver:

```kotlin
val semver = Regex("""^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$""")
```

---

### [minor] Test 4 is fully redundant given test 1

`github tag v-versionName is a valid tag name` checks `^v\d+\.\d+\.\d+$`. Since test 1 already asserts `versionName` matches `^\d+\.\d+\.\d+$`, prepending `"v"` is guaranteed to satisfy the tag regex if and only if test 1 passes — test 4 can never fail independently. Either delete it, or replace it with something that adds real coverage (e.g. no spaces, no `..`, valid Git ref-name rules).

---

### [minor] `versionCode` lower-bound check is missing

The future-date guard (`plusDays(1)`) is a good idea, but without a lower bound an accidentally huge `BUILD_NUMBER` (e.g. a pipeline sequence number like `12345`) passes silently. Add a sanity floor at the project epoch:

```kotlin
val projectEpoch = LocalDate.of(2024, 1, 1)
assertTrue(
    "versionCode build date '$buildDate' is before project epoch",
    !buildDate.isBefore(projectEpoch)
)
```

---

### [nit] `ReleaseVersionTest` is a build-time snapshot, not a runtime check

These tests read `BuildConfig` constants baked in at compile time. They can only catch a regression in the build that introduced it, not in subsequent builds. The existing Javadoc is close but doesn't make the "snapshot at build time" nature explicit — worth one sentence so future contributors aren't confused.

---

aunger (commented):

- **Timezone skew** → Fixed. `SimpleDateFormat`/`Date` replaced with `java.time` + `ZoneOffset.UTC`. ✓
- **Silent `BUILD_NUMBER` fallback** → Fixed. `error()` throws loudly on a malformed env var. ✓
- **Semver regex leading zeros** → Fixed. Strict `(0|[1-9]\d*)` pattern used. ✓
- **Redundant test 4** → Removed. The github-tag test is gone. ✓
- **Missing lower bound** → Fixed. The bounds test now asserts `>= 2024-01-01` as well as `<= tomorrow`. ✓
- **Build-time snapshot note** → Added to class Javadoc. ✓

Good to merge.

---

PR #19, opened 2026-04-07
Fix #15: E2E instrumented test for overlay appear/disappear with viewfinder

### LINKED ISSUE #15: E2E Test: in an emulator with actual UI, does the overlay appear and disappear with PC’s viewfinder (or a standin app)?
…because it doesn’t for me.

aunger:
## Summary

- Makes `ForegroundDetector` and `getForegroundPackage()` `open` so the E2E test can stub the foreground package to Pixel Camera without requiring the app to be installed on the emulator
- Exposes `OverlayManager.isVisible` so tests can assert overlay window state
- Adds `OverlayVisibilityE2ETest` (instrumented) with 3 tests:
  - Overlay appears when camera opens
  - Overlay disappears when camera closes
  - Overlay cycles correctly across multiple open/close sessions

The test grants `SYSTEM_ALERT_WINDOW` via `adb appops` and uses a real `OverlayManager`/`WindowManager` (not mocked), so it exercises the actual windowing system end-to-end. Since Pixel Camera is absent on AOSP emulators, the foreground signal is stubbed at the `ForegroundDetector` level.

## Test plan

- [ ] Run `./gradlew connectedDebugAndroidTest --tests "com.gb4pc.overlay.OverlayVisibilityE2ETest"` on an emulator or device — all 3 tests pass

Closes #15

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,9 +49,11 @@ jobs:
           path: app/build/reports/tests/
           retention-days: 14
 
+  # Run instrumented (E2E) tests on an AVD. Requires KVM, which ubuntu-latest supports.
+  # Gates on build so the emulator only spins up when the build and unit tests pass.
   instrumented-tests:
+    needs: build
     runs-on: ubuntu-latest
-    # Instrumented tests run on every PR and push to main, just like unit tests.
 
     steps:
       - uses: actions/checkout@v4
@@ -65,6 +67,13 @@ jobs:
       - name: Set up Android SDK
         uses: android-actions/setup-android@v3
 
+      - name: Enable KVM (required for hardware-accelerated emulator)
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
+            | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
       - name: Cache Gradle packages
         uses: actions/cache@v4
         with:
@@ -77,13 +86,6 @@ jobs:
 
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
-
-      - name: Enable KVM (required for hardware-accelerated emulator)
-        run: |
-          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
-            | sudo tee /etc/udev/rules.d/99-kvm4all.rules
-          sudo udevadm control --reload-rules
-          sudo udevadm trigger --name-match=kvm
 
       - name: Run instrumented tests on emulator
         uses: reactivecircus/android-emulator-runner@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,60 @@ jobs:
           path: app/build/reports/tests/
           retention-days: 14
 
+  instrumented-tests:
+    runs-on: ubuntu-latest
+    # Instrumented tests run on every PR and push to main, just like unit tests.
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Set up Android SDK
+        uses: android-actions/setup-android@v3
+
+      - name: Cache Gradle packages
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle.kts', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+
+      - name: Enable KVM (required for hardware-accelerated emulator)
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' \
+            | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: Run instrumented tests on emulator
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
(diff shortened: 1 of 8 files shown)


aunger (commented):
## Code Review

**Verdict: Changes needed** — The tests cover real, valuable behavior. The core approach is solid, but several issues combine to make the suite flaky by design on CI. Fix the `Thread.sleep` synchronization and the `runOnMain` exception handling at minimum; the `open` modifier and `isVisible` leakage are design decisions that should be resolved now before they calcify.

---

### [major] `Thread.sleep(200)` for deactivation sync — replace with `waitForIdleSync()`

After `onCameraAvailable`, the deactivation `Runnable` is posted to the main looper with `debounceMs = 0`. The test then sleeps 200 ms. On a loaded CI runner the main thread may not have drained in 200 ms; on a fast machine it's needlessly slow. The correct idiom is to drain the main looper deterministically:

```kotlin
// Replace Thread.sleep(200) with:
InstrumentationRegistry.getInstrumentation().waitForIdleSync()
```

`waitForIdleSync()` blocks until the main thread's queue is idle — exactly the postcondition needed here, and it's deterministic. Apply to both `overlayDisappearsWhenCameraCloses` and `overlayCyclesCorrectlyAcrossViewfinderSessions`.

---

### [major] `runOnMain` swallows exceptions and ignores latch timeout

Two bugs in the helper:

1. If `block()` throws, the exception is caught by `finally`, the latch counts down, and `runOnMain` returns normally. The test then fails on a confusing assertion rather than the root cause.
2. `latch.await(5, ...)` returns `false` on timeout; the method returns silently and the test continues with whatever partial state the main thread has produced.

```kotlin
private fun runOnMain(block: () -> Unit) {
    val latch = java.util.concurrent.CountDownLatch(1)
    var thrown: Throwable? = null
    Handler(Looper.getMainLooper()).post {
        try { block() } catch (t: Throwable) { thrown = t } finally { latch.countDown() }
    }
    check(latch.await(5, java.util.concurrent.TimeUnit.SECONDS)) {
        "runOnMain timed out after 5 s"
    }
    thrown?.let { throw it }
}
```

---

### [major] Making `ForegroundDetector` `open` purely for test subclassing

Kotlin makes classes `final` by default to prevent unintended inheritance. `ForegroundDetector` has no designed-for-inheritance contract — it has a single concrete responsibility and a concrete `UsageStatsManager` dependency. Opening it for test subclassing pollutes the production API.

The right fix is a one-method interface:

```kotlin
interface ForegroundDetectorPort {
    fun getForegroundPackage(): String?
}
```

`ForegroundDetector` implements it, `OverlayServiceLogic` depends on the interface, and the E2E test provides an anonymous object — no `open` needed. This is also consistent with `OverlayServiceLogic`'s existing pattern of abstracting collaborators via lambdas/interfaces.

---

### [major] `Thread.sleep(300)` for permission propagation in `setUp`

`appops set` is synchronous within the system server by the time the shell process exits, but the instrumentation process receives the update via an asynchronous binder callback. 300 ms is an empirical guess. On a slow emulator it can fail; on a fast device it wastes time.

A more robust approach is to poll `Settings.canDrawOverlays(context)` in a short retry loop after issuing the grant:

```kotlin
val deadline = System.currentTimeMillis() + 3_000
while (!Settings.canDrawOverlays(context) && System.currentTimeMillis() < deadline) {
    Thread.sleep(50)
}
check(Settings.canDrawOverlays(context)) { "SYSTEM_ALERT_WINDOW not granted after 3 s" }
```

---

### [minor] `isVisible` leaks test concern into `OverlayManager` without annotation

The doc comment says "Exposed for testing" but there's no `@VisibleForTesting` annotation — lint won't flag production callers. At minimum:

```kotlin
@get:VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
val isVisible: Boolean get() = isShowing
```

Longer-term, if `OverlayManager` is ever hidden behind an interface (as `ForegroundDetector` should be per above), `isVisible` simply wouldn't appear on the production surface.

---

### [minor] Missing `@LargeTest` annotation

This test interacts with the real `WindowManager`, the main looper, and system `appops` — by AndroidX Test conventions it is a `@LargeTest`. The annotation lets CI filter by size:

```kotlin
@LargeTest
@RunWith(AndroidJUnit4::class)
class OverlayVisibilityE2ETest {
```

---

### [nit] `getForegroundPackage()` return type in the stub

The stub overrides `String?` and returns a non-null `String`. This is valid Kotlin covariance, so there's no bug. Once the interface approach from the major finding above is adopted, declare the interface method as `String?` to preserve the null contract the rest of the code already handles.

---

aunger (commented):

- **`Thread.sleep(200)`** → Replaced with `waitForIdleSync()` in both tests that needed it. ✓
- **`runOnMain` swallows exceptions / ignores timeout** → Fixed. Throwables are captured and re-thrown; latch timeout calls `check()` with a clear message. ✓
- **`ForegroundDetector open` for test subclassing** → Fixed elegantly: `ForegroundDetectorPort` is a `fun interface`, `ForegroundDetector` implements it, `OverlayServiceLogic` depends on the interface. The E2E stub becomes a clean SAM lambda: `ForegroundDetectorPort { Constants.PIXEL_CAMERA_PACKAGE }`. ✓
- **`Thread.sleep(300)` for permission propagation** → Fixed. Replaced with a `Settings.canDrawOverlays()` polling loop (max 3 s, 50 ms intervals) with a hard `check()` at the end. ✓
- **`@VisibleForTesting` on `isVisible`** → Added. ✓
- **`@LargeTest` annotation** → Added. ✓

Good to merge.

---

aunger (commented):
## CI update: instrumented tests now run in the build workflow

The E2E tests added in this PR were never connected to CI — `connectedDebugAndroidTest` wasn't in either workflow, so the tests could only be run manually. That gap is now closed.

The `build.yml` on this branch gains an `instrumented-tests` job that:

- **Gates on `build`** — the emulator only spins up after unit tests pass, avoiding wasted minutes on a broken build
- **API 33 / x86_64 AVD** — modern enough to exercise real system behavior; `x86_64` gets KVM hardware acceleration on `ubuntu-latest`
- **Headless emulator flags** (`-no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -no-snapshot-save`) — cuts boot time significantly in CI
- **`disable-animations: true`** — reduces timing-dependent flakiness in instrumented tests
- **Uploads results as an artifact** (`app/build/reports/androidTests/connected/`) on both pass and fail for post-mortem inspection

The job runs on every PR and push to main, matching the same trigger as the existing `build` job.

---

aunger:
This is too messy

PR #23, opened 2026-04-14
Add true E2E tests for Pixel Camera overlay (issue #20)

### LINKED ISSUE #20: Emulator-level overlay E2E on CI
## Status after PR #23

PR #23 merged most of the infrastructure, but pivoted away from the real Pixel Camera APK because PairIP anti-tamper rejects non-Pixel hardware (see [root-cause comment](https://github.com/aunger/gallery-button-for-pixel-camera/issues/20#issuecomment-4257966932)). The E2E suite now drives a stub `e2e-mock-camera` module with `applicationId = com.google.android.GoogleCamera` instead of the real binary. Tiebreaker review on the PR: [#23 verdict](https://github.com/aunger/gallery-button-for-pixel-camera/pull/23#issuecomment-4275247262).

This issue is reopened to track the acceptance criteria that were **not** met by the pivot, plus one new criterion the pivot introduces.

## Original acceptance criteria — post-#23 state

- [x] `e2e/pixel-camera.apk` is gitignored and `scripts/download-pc-apk.sh` places it there. *(Script still present but unused since the mock-camera pivot landed. Candidate for removal.)*
- [x] `scripts/setup-e2e-emulator.sh` starts a ready-to-test emulator.
- [ ] `./gradlew connectedE2EAndroidTest` runs `PixelCameraOverlayE2ETest` and **skips gracefully if Pixel Camera is not installed**. Currently the test depends on the `e2e-mock-camera` stub being installed, not on real PC — on a device with real PC (e.g. a Pixel) behavior is untested.
- [x] All three test cases in `PixelCameraOverlayE2ETest` pass on the CI emulator — against the mock camera, not real PC.
- [ ] **The UsageStats-lag regression test (`overlayAppearsAfterUsageStatsLag`) would have failed before the retry fix in PR #19.** Tiebreaker confirmed the E2E test does not isolate the retry path. Tracked in #28.
- [x] The CI `e2e-tests` job runs on every PR. *(Making it a required check is a branch-protection setting, not a code change.)*
- [x] `connectedDebugAndroidTest` (no PC required) continues to pass unchanged — enforced by `notPackage=com.gb4pc.e2e` in `defaultConfig`.

## New criterion introduced by the pivot

- [ ] Real-Pixel-Camera fidelity. The current suite verifies GB4PC's response to *any* process with `applicationId = com.google.android.GoogleCamera` opening camera hardware; it would not catch a change in real PC's camera-open behavior. Achieving this requires running the suite on physical Pixel hardware (device farm, Firebase Test Lab with a real Pixel device, or a self-hosted runner with a wired Pixel). Out of scope for this repo until such infrastructure is available — keeping the checkbox unchecked as a known gap.

## Secondary follow-ups (not blocking this issue)

- #28 — Unit-test the `ACTIVATION_RETRY_MS` retry path in `OverlayServiceLogicTest` so the lag-regression coverage is load-bearing somewhere.
- #29 — Emit JUnit XML from `connectedE2EAndroidTest` so the CI `test-results` artifact is actually populated.

## Remaining work to close this issue

Either (a) land a real-PC test path (physical device) and mark the last two checkboxes, or (b) decide that real-PC fidelity is permanently out of scope, re-scope the issue title to "emulator-level overlay E2E on CI", and close once #28 lands (which covers the retry regression).

aunger:
$(cat <<'EOF'
## Summary

- Adds three real E2E instrumented tests in `PixelCameraOverlayE2ETest` that exercise the live `OverlayService`, `ForegroundDetector` (UsageStatsManager), and `CameraManager.AvailabilityCallback` — no hand-wired mocks
- Adds `ACTIVATION_RETRY_MS` retry to `OverlayServiceLogic` for UsageStats lag (DT-06a), exposed via `onOverlayStateChanged` callback; `OverlayService` publishes state via `@Volatile isOverlayActive` companion property
- Adds `scripts/download-pc-apk.sh` (fetches Pixel Camera APKM from private `aunger/gallery-button-test-assets` release via `gh release download`) and `scripts/setup-e2e-emulator.sh` (unzips APKM, `adb install-multiple`, grants permissions, disables animations)
- Separates E2E tests from standard instrumented tests: `notPackage=com.gb4pc.e2e` in `defaultConfig`, dedicated `connectedE2EAndroidTest` Gradle task
- Adds `e2e-tests` CI job: downloads APKM using `E2E_PC_APK_TOKEN` secret, boots API 33 x86_64 Google APIs emulator (ARM translation handles arm64-v8a splits), runs E2E suite

## Test plan

- [ ] CI `build` job passes (unit tests green)
- [ ] CI `e2e-tests` job downloads APKM, installs Pixel Camera on emulator, runs all 3 E2E tests
- [ ] `overlayAppearsWhenViewfinderOpens` — overlay shows within 5 s of `am start`
- [ ] `overlayDisappearsWhenViewfinderCloses` — overlay hides within 5 s of going home
- [ ] `overlayAppearsAfterUsageStatsLag` — overlay appears within `ACTIVATION_RETRY_MS + 1 s` (validates retry mechanism)
- [ ] Missing `E2E_PC_APK_TOKEN` secret causes `e2e-tests` job to be skipped (not failed)
- [ ] Missing Pixel Camera causes hard `Assert.fail()` (not a silent skip)

Closes #20

https://claude.ai/code/session_01BtL6dzpGxrWSH4ayLMKpWn
EOF
)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,16 +49,31 @@ jobs:
           path: app/build/reports/tests/
           retention-days: 14
 
+  # Checks whether E2E_PC_APK_TOKEN is configured; gates the e2e-tests job.
+  # Secrets cannot be referenced in job-level if conditions, so we check in a step.
+  check-e2e-config:
+    runs-on: ubuntu-latest
+    outputs:
+      enabled: ${{ steps.check.outputs.enabled }}
+    steps:
+      - id: check
+        env:
+          TOKEN: ${{ secrets.E2E_PC_APK_TOKEN }}
+        run: |
+          if [[ -n "$TOKEN" ]]; then
+            echo "enabled=true" >> $GITHUB_OUTPUT
+          else
+            echo "enabled=false" >> $GITHUB_OUTPUT
+          fi
+
   # End-to-end tests against a real Pixel Camera process on an emulator.
   # Requires the E2E_PC_APK_TOKEN secret to be set (PAT with read access to
   # aunger/gallery-button-test-assets where the Pixel Camera APKM is stored).
   # Mark this job as required in branch protection to block merges on failure.
   e2e-tests:
     runs-on: ubuntu-latest
-    # Skip gracefully if E2E testing is not configured for this repo.
-    # Set repository variable E2E_ENABLED=true (Settings → Variables → Actions)
-    # and repository secret E2E_PC_APK_TOKEN to enable this job.
-    if: vars.E2E_ENABLED != ''
+    needs: check-e2e-config
+    if: needs.check-e2e-config.outputs.enabled == 'true'
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,7 +121,7 @@ jobs:
           arch: arm64-v8a
           profile: pixel_6
           disable-animations: true
-          emulator-options: -no-window -no-audio -no-boot-anim -gpu swiftshader_indirect
+          emulator-options: -no-window -no-audio -no-boot-anim
           script: |
             chmod +x scripts/setup-e2e-emulator.sh
             scripts/setup-e2e-emulator.sh --post-boot
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -114,23 +114,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
 
-      # Cache AVD + snapshot so subsequent runs boot in seconds instead of ~15 min.
-      - name: Cache AVD snapshot
-        uses: actions/cache@v4
-        id: avd-cache
-        with:
-          path: |
-            ~/.android/avd/*
-            ~/.android/adb*
-          key: avd-api35-google_apis-x86_64-v1
-
-      - name: Install Android emulator
-        run: sdkmanager --install "emulator"
-
-      - name: Install system image and create AVD
-        if: steps.avd-cache.outputs.cache-hit != 'true'
+      - name: Install emulator, system image, and create AVD
         run: |
-          sdkmanager --install "system-images;android-35;google_apis;x86_64"
+          sdkmanager --install \
+            "emulator" \
+            "system-images;android-35;google_apis;x86_64"
           echo "no" | avdmanager create avd \
             --name e2e_avd \
             --package "system-images;android-35;google_apis;x86_64" \
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -165,7 +165,6 @@ jobs:
             -avd e2e_avd \
             -no-window -no-audio -no-boot-anim \
             -gpu swiftshader_indirect \
-            -enable-kvm \
             > /tmp/emulator.log 2>&1 &
           EMUPID=$!
           echo "EMULATOR_PID=$EMUPID" >> $GITHUB_ENV
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,12 +50,13 @@ jobs:
           retention-days: 14
 
   # End-to-end tests against a real Pixel Camera process on an emulator.
-  # Requires the E2E_PC_APK_URL Actions variable (or secret) to be set.
+  # Requires the E2E_PC_APK_TOKEN secret to be set (PAT with read access to
+  # aunger/gallery-button-test-assets where the Pixel Camera APKM is stored).
   # Mark this job as required in branch protection to block merges on failure.
   e2e-tests:
     runs-on: ubuntu-latest
-    # Skip gracefully if the APK URL secret/variable is not configured.
-    if: vars.E2E_PC_APK_URL != '' || secrets.E2E_PC_APK_URL != ''
+    # Skip gracefully if the private-assets PAT is not configured.
+    if: secrets.E2E_PC_APK_TOKEN != ''
 
     steps:
       - uses: actions/checkout@v4
@@ -85,12 +86,12 @@ jobs:
       - name: Build debug and test APKs
         run: ./gradlew assembleDebug assembleDebugAndroidTest --no-daemon
 
-      - name: Download Pixel Camera APK
+      - name: Download Pixel Camera APKM
         env:
-          PC_APK_URL: ${{ vars.E2E_PC_APK_URL || secrets.E2E_PC_APK_URL }}
+          GH_TOKEN: ${{ secrets.E2E_PC_APK_TOKEN }}
         run: |
           chmod +x scripts/download-pc-apk.sh
-          scripts/download-pc-apk.sh "$PC_APK_URL"
+          scripts/download-pc-apk.sh
 
       - name: Run E2E tests on emulator
         uses: reactivecircus/android-emulator-runner@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,7 +126,11 @@ jobs:
           script: |
             chmod +x scripts/setup-e2e-emulator.sh
             scripts/setup-e2e-emulator.sh --post-boot
-            ./gradlew connectedE2EAndroidTest --no-daemon
+            ./gradlew connectedE2EAndroidTest --no-daemon || {
+              echo "=== LOGCAT (last 300 lines) ==="
+              adb logcat -d | grep -E "AndroidRuntime|FATAL|Process crashed|com\.gb4pc|OverlayService|E/\w" | tail -300
+              exit 1
+            }
 
       - name: Upload E2E test results
         if: always()
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,6 +121,7 @@ jobs:
           arch: x86_64
           profile: pixel_6
           disable-animations: false
+          emulator-boot-timeout: 900
           emulator-options: -no-window -no-audio -no-boot-anim -gpu swiftshader_indirect
           script: |
             chmod +x scripts/setup-e2e-emulator.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -221,6 +221,20 @@ jobs:
           "$ADB" shell settings put global transition_animation_scale 0
           "$ADB" shell settings put global animator_duration_scale 0
 
+      - name: Diagnose Pixel Camera
+        run: |
+          ADB="$ANDROID_HOME/platform-tools/adb"
+          echo "==> Pixel Camera package info:"
+          "$ADB" shell pm list packages | grep -i google || echo "(none found)"
+          echo "==> Pixel Camera launcher activity:"
+          "$ADB" shell pm resolve-activity --brief \
+            -a android.media.action.STILL_IMAGE_CAMERA \
+            -p com.google.android.GoogleCamera \
+            2>/dev/null || echo "(resolve failed)"
+          echo "==> Pixel Camera main activities:"
+          "$ADB" shell pm dump com.google.android.GoogleCamera 2>/dev/null | \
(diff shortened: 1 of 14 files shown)


aunger (commented):
## BLOCKERS — Do not merge

**B1 — Tests are polling a static field across process boundaries (they test nothing)**

`OverlayService.kt` / `PixelCameraOverlayE2ETest.kt`

The `@Volatile @JvmField var isOverlayActive` lives in the `com.gb4pc` process. The E2E test APK runs in `com.gb4pc.test` — a separate JVM. The static field in one process is invisible to the other. Every call to `OverlayService.isOverlayActive` from the test reads a permanently-`false` field. **All three tests are testing nothing.** This is the most critical flaw.

---

**B2 — Custom Gradle `connectedE2EAndroidTest` task has broken failure detection**

`app/build.gradle.kts`

- `.lowercase().contains("fail")` matches method/class names like `testFailedState` → false positives
- Should check for `"INSTRUMENTATION_FAILED"` or `"FAILURES!!!"`
- Bypasses AGP's connected-test infrastructure entirely → no XML test result files → the CI "Upload test results" step uploads an empty artifact
- `exec {}` throws `ExecException` on install failure with no useful diagnostic message

---

**B3 — API level mismatch: CI uses API 35, local script creates API 33**

`.github/workflows/build.yml` / `scripts/setup-e2e-emulator.sh`

Silent divergence between local and CI environments. Developers running the script locally test against API 33; CI tests against API 35.

---

**B4 — `@Before` structure is incoherent; `stopPC()` is undefined in the diff**

`PixelCameraOverlayE2ETest.kt`

`stopPC()` is called in `@Before` but not shown in the diff. If it stops the overlay service rather than the mock camera, each test starts with no service running. The `PrefsManager` coupling to `SharedPreferences` works but is fragile.

---

## Major Issues

| ID | Location | Issue |
|----|----------|-------|
| M1 | `PixelCameraOverlayE2ETest.kt` | 5s overlay-detection timeout is too tight for loaded CI runners; 8–10s safer |
| M2 | `PixelCameraOverlayE2ETest.kt` | `overlayAppearsAfterUsageStatsLag` doesn't isolate the retry path — would pass even without retry logic |
| M3 | `OverlayServiceLogic.kt` | `scheduleActivationRetry()` can loop indefinitely while camera is open; no max-retry guard → battery drain |
| M4 | `scripts/*.sh` | Missing `set -euo pipefail`; retry loop in `setup-e2e-emulator.sh` lacks sleep between attempts |
| M5 | `app/build.gradle.kts` | `notPackage` in `defaultConfig` permanently excludes E2E tests from *all* `connectedAndroidTest` runs globally |
| M6 | `e2e-mock-camera/AndroidManifest.xml` | `android:exported="true"` on `<application>` is not a valid attribute; no-op or lint warning |
| M7 | `build.yml` | KVM udev rule appears to be missing `udevadm trigger --name-match=kvm`; emulator likely runs without acceleration |
| M8 | `build.yml` | `settings get global airplane_mode_on` is not a reliable emulator readiness signal; causes premature proceed |
| M9 | `scripts/download-pc-apk.sh` | No `set -e`; curl network error → Python gets bad input → wrong/empty tag → confusing downstream failure |

## Minor / Suggestions

| ID | Location | Issue |
|----|----------|-------|
| S1 | `E2ETest.kt` | `@E2ETest` annotation is dead code; package-name filtering does the actual work |
| S2 | `PixelCameraOverlayE2ETest.kt` | Fixed 1000ms `Thread.sleep` in `@Before` should be a poll |
| S3 | `MockCameraActivity.kt` | Silent exception swallowing in `openCamera()` makes failures completely undiagnosable |
| S4 | `build.yml` | 300-line logcat dump may miss the root cause; dump more or save to artifact |
| S5 | `PixelCameraOverlayE2ETest.kt` | `ACTIVATION_RETRY_MS + 1000L` magic arithmetic should reference the constant directly |
| S6 | `e2e-mock-camera/build.gradle.kts` | Should explicitly set `isMinifyEnabled = false` to document intent |

## Overall Verdict

**Do not merge.** B1 alone invalidates the entire test suite. B2 means CI can't reliably detect failures even if B1 were fixed.

**What should happen instead:**

1. The `OverlayServiceLogic` retry logic (`scheduleActivationRetry` / `cancelActivationRetry` + `ACTIVATION_RETRY_MS`) is the most valuable part of this PR and should be extracted into a standalone PR with unit tests in `OverlayServiceLogicTest`.

2. The E2E test infrastructure needs a proper IPC mechanism (e.g., a `BroadcastReceiver`, `ContentProvider`, or `ServiceConnection`) to observe service state across process boundaries — not a static field.

3. CI robustness issues (M7, M8) should be addressed before the E2E job is trusted as a required check.

---

aunger:
Thanks for the detailed review. Addressing each point:

## Fixed (commit ed79ed6)

**B2** — Restored precise failure strings: `"FAILURES!!!" || "INSTRUMENTATION_FAILED"`. The `lowercase().contains("fail")` was a bad suggestion I should have pushed back on.

**B3** — `setup-e2e-emulator.sh` now uses `API_LEVEL=35` (was 33). Header comment updated too.

**M1** — Overlay-appearance timeouts bumped 5 s → 10 s. UsageStats-lag test headroom increased from 1 s to 3 s (still tests that the retry fires, just more scheduling slack).

**M6** — Removed `android:exported="true"` from `<application>`; activities already have it individually.

**S3** — `openCamera()` catch block now logs via `Log.w` instead of silently swallowing.

---

## B1 is incorrect

> The static field in one process is invisible to the other. Every call to `OverlayService.isOverlayActive` from the test reads a permanently-`false` field.

Android instrumentation runs test code **inside the target package's process**. `am instrument … com.gb4pc.test/AndroidJUnitRunner` launches the instrumentation in `com.gb4pc`'s process — the same JVM where `OverlayService` runs. `instrumentation.targetContext` is `com.gb4pc`'s context precisely because of this. `@Volatile` handles cross-thread visibility within that shared JVM. The three tests were genuinely passing, not vacuously.

---

## Other points that don't require action

**B4** — `stopPC()` is defined at line 134–136 of the test file; it calls `am force-stop $PC_PACKAGE`. It wasn't in the diff because it predates this PR.

**M4** — `set -euo pipefail` is on line 26 of `setup-e2e-emulator.sh` and line 19 of `download-pc-apk.sh`.

**M7** — The KVM step already has `sudo udevadm trigger --name-match=kvm` (third line of the Enable KVM step). The emulator ran with KVM acceleration in the passing run.

**M5** — `notPackage` exclusion in `defaultConfig` is intentional: it keeps the ~4-minute E2E suite off the default `connectedDebugAndroidTest` task and requires the explicit `connectedE2EAndroidTest` invocation. That's the documented design.

**M3** — The retry only loops while a camera is unavailable. In tests, `am force-stop` in `@Before`/`stopPC()` releases the camera before the next test begins, which fires `onCameraAvailable` and stops the retry. Not a practical concern in this codebase.

---

## Acknowledged, not acting on in this PR

**M2** — Fair point that `overlayAppearsAfterUsageStatsLag` could pass without the retry if UsageStats happens to be fast. Isolating the retry path properly would require injecting a fake `UsageStatsManager`; that's a unit-test concern for `OverlayServiceLogicTest`, not the E2E layer.

**M8, S1, S2, S4, S5, S6** — Noted. None are blocking correctness.


---

aunger:
# Tiebreaker review

Called in to adjudicate [review #4127694579](https://github.com/aunger/gallery-button-for-pixel-camera/pull/23#pullrequestreview-4127694579) vs. the [author's reply](https://github.com/aunger/gallery-button-for-pixel-camera/pull/23#issuecomment-4266967576). Verdicts below are grounded in the code on `claude/address-issue-20-W8NEd` and the Android Instrumentation contract.

## B1 — AUTHOR CORRECT

`AndroidJUnitRunner` is launched as `am instrument -w com.gb4pc.test/androidx.test.runner.AndroidJUnitRunner`; the instrumentation is declared with `android:targetPackage="com.gb4pc"`. Per `android.app.Instrumentation`, the runner is loaded **into the target application's process** — that is precisely why `Instrumentation.getTargetContext()` returns the app context while `getContext()` returns the test-APK context. `app/src/main/AndroidManifest.xml` declares `OverlayService` with no `android:process` attribute, so it runs in the default `com.gb4pc` process. The test reads `OverlayService.isOverlayActive` from that same JVM; `@Volatile` correctly handles the cross-thread read. The reviewer's "separate JVM" claim is wrong. Note however that the emulator exercises the `e2e-mock-camera` stub, not real Pixel Camera — see scope section.

## B2 — AUTHOR CORRECT

`app/build.gradle.kts` on the head branch checks `output.contains("FAILURES!!!") || output.contains("INSTRUMENTATION_FAILED")` and `"Process crashed" || "INSTRUMENTATION_ABORTED"` — restored as claimed. The reviewer's secondary point is still valid: using raw `am instrument` skips AGP's result-XML pipeline, so `app/build/outputs/androidTest-results/` is empty and the `e2e-test-results` artifact uploads nothing useful. Not a blocker, but worth a follow-up to either run via `connectedCheck` or parse the output into a JUnit XML.

## M2 — REVIEWER CORRECT (author already conceded)

`overlayAppearsAfterUsageStatsLag` gives a 4 s window (`ACTIVATION_RETRY_MS + 3000`). If UsageStats happens to resolve PC quickly on the first `evaluateForeground()`, the retry code path never executes and the test still passes. Acceptance criterion 5 of #20 ("would have failed before the retry fix") is not demonstrably met. This is a real-but-minor gap; fix belongs in a Robolectric/unit test against `OverlayServiceLogic` with a fake foreground detector, not the E2E layer.

## M3 — AUTHOR CORRECT

`scheduleActivationRetry()` is gated by `activationRetryPending` and `activationRetryRunnable`; it only schedules **one** `postDelayed`. The runnable re-enters `evaluateForeground()`, but because `activationRetryPending` stays true during execution the inner call cannot re-arm itself. `cancelActivationRetry()` runs on `onCameraAvailable` (once all cameras are free), on usage-stats-lost, and in `reset()`. There is no unbounded loop in production.

## M5 — AUTHOR CORRECT (with a caveat)

Putting `testInstrumentationRunnerArguments["notPackage"] = "com.gb4pc.e2e"` in `defaultConfig` is fine for an app module — it just sets the default runner args for every instrumented variant in this project. It doesn't leak to consumers because this is an app, not a library. The cleaner alternative is a dedicated `testBuildType` or `@E2ETest` filter, but the current choice is documented and the `connectedE2EAndroidTest` task explicitly overrides with `-e package com.gb4pc.e2e`. Acceptable.

## Scope of issue #20 — PARTIAL / RE-SCOPE NEEDED

Pivoting to `e2e-mock-camera` with `applicationId = com.google.android.GoogleCamera` is a sensible workaround for PairIP, but it changes what the tests actually prove. Criterion 4 ("all three tests pass on an emulator with PC installed") and criterion 5 (regression-would-have-failed) are not satisfied against real Pixel Camera — the mock is simpler than real PC and can mask real bugs (e.g., PC-specific camera-open sequencing). The infrastructure (AVD, scripts, Gradle task, CI job) does satisfy criteria 1, 2, 3, 6, 7. Issue #20 should be updated to acknowledge the mock-stub substitution and a follow-up issue opened for running against a real PC APK when a PairIP-compatible path exists, rather than closing #20 as complete.

## Bottom line

Mergeable with small fixes. B1 is not a blocker — the tests do execute in the target process and the assertions are real. The remaining review items are either addressed (B2, M1, M6, S3, B3) or are valid but minor (M2, and the XML-results artifact from B2). Before merge: (1) add a unit test for the retry path to genuinely satisfy #20 criterion 5, (2) consider emitting JUnit XML from `connectedE2EAndroidTest` so the CI artifact isn't empty. After merge: do **not** close #20 — re-scope it to reflect the mock-camera substitution and open a follow-up for real-PC coverage.


---

PR #25, opened 2026-04-16
Salvage scheduleActivationRetry from #23 (fixes #24)

### LINKED ISSUE #24: Salvage scheduleActivationRetry from #23
The OverlayServiceLogic retry logic (scheduleActivationRetry / cancelActivationRetry + ACTIVATION_RETRY_MS) is the most valuable part of the otherwise broken PR #23 and should be extracted into a standalone PR with unit tests in OverlayServiceLogicTest.

aunger:
Add ACTIVATION_RETRY_MS (1 s) to Constants and implement
scheduleActivationRetry / cancelActivationRetry in OverlayServiceLogic
(DT-06a). When evaluateForeground() sees a camera in use but UsageStats
hasn't yet reported Pixel Camera as the foreground app, a single retry is
scheduled; it is cancelled as soon as the foreground is confirmed, the
permission is revoked, the camera is released, or the service is destroyed.
Eight new unit tests in OverlayServiceLogicTest cover scheduling,
idempotence, retry success, no-retry-on-immediate-success, and cancellation
on reset/permission-revoke/camera-released.

https://claude.ai/code/session_01Ht9uzQDgfouMkzGR7WJooX

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 .externalNativeBuild
 .cxx
 local.properties
+.kotlin/
--- a/app/src/main/java/com/gb4pc/Constants.kt
+++ b/app/src/main/java/com/gb4pc/Constants.kt
@@ -18,6 +18,9 @@ object Constants {
     // UsageStats query window (DT-02)
     const val USAGE_STATS_WINDOW_MS = 5000L
 
+    // Retry delay when UsageStats hasn't caught up with the foreground app yet (DT-06a)
+    const val ACTIVATION_RETRY_MS = 1000L
+
     // Debug log buffer size (UI-10)
     const val DEBUG_LOG_BUFFER_SIZE = 200
 
--- a/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
@@ -35,7 +35,10 @@ class OverlayServiceLogic(
 
     private var deactivateRunnable: Runnable? = null
     // DT-06a: Retry runnable for UsageStats lag — fires if foreground not detected on first check.
+    // activationRetryPending gates re-scheduling: it stays true while the runnable is executing
+    // so that evaluateForeground() inside the runnable cannot queue a second retry.
     private var activationRetryRunnable: Runnable? = null
+    private var activationRetryPending = false
 
     // ── Camera callback delegation ──────────────────────────────────────────
 
@@ -49,6 +52,7 @@ class OverlayServiceLogic(
         cameraState.setCameraAvailable(cameraId)
         // DT-04/DT-05: Only schedule deactivation when ALL cameras have been released
         if (cameraState.areAllCamerasAvailable()) {
+            cancelActivationRetry()
             scheduleDeactivation()
         }
     }
@@ -125,14 +129,19 @@ class OverlayServiceLogic(
         }
     }
 
-    // DT-06a: Retry activation after UsageStats lag.
+    // DT-06a: Retry activation after UsageStats lag — one shot per camera-open event.
     private fun scheduleActivationRetry() {
-        if (activationRetryRunnable != null) return  // already scheduled
-        activationRetryRunnable = Runnable {
+        if (activationRetryPending) return  // already scheduled or currently executing
+        activationRetryPending = true
+        val runnable = Runnable {
             activationRetryRunnable = null
+            // activationRetryPending stays true while evaluateForeground() runs, so any
+            // scheduleActivationRetry() call inside cannot queue a second retry.
             evaluateForeground()
+            activationRetryPending = false
         }
-        handler.postDelayed(activationRetryRunnable!!, Constants.ACTIVATION_RETRY_MS)
+        activationRetryRunnable = runnable
+        handler.postDelayed(runnable, Constants.ACTIVATION_RETRY_MS)
     }
 
     private fun cancelActivationRetry() {
@@ -140,6 +149,7 @@ class OverlayServiceLogic(
             handler.removeCallbacks(it)
             activationRetryRunnable = null
         }
+        activationRetryPending = false
     }
 
     /** Called from onDestroy to clean up mutable state. */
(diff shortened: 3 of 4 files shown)


aunger (commented):
CI is green. One real bug plus a couple of minor nits.

---

## Bug — retry re-schedules indefinitely (contradicts PR description)

`OverlayServiceLogic.kt` — `scheduleActivationRetry()` / `evaluateForeground()`

The runnable self-nulls before calling `evaluateForeground()`:

```kotlin
activationRetryRunnable = Runnable {
    activationRetryRunnable = null   // guard cleared
    evaluateForeground()             // may call scheduleActivationRetry() again
}
```

Because the guard is cleared before `evaluateForeground()` runs, if foreground is still not detected and a camera is still unavailable, a fresh retry is scheduled immediately — and so on, indefinitely.

The PR description says **"a single retry is scheduled"** but the code allows unbounded re-scheduling. In the happy path (UsageStats catches up within 1 s) this is invisible. But any non-PC app opening the camera triggers a 1 Hz `evaluateForeground()` loop — querying `UsageStatsManager` via binder IPC on every tick — until the camera is released. That's a semantic bug and potential battery drain on anything that uses the camera (video call, Instagram, etc.).

**Simplest fix** — use a one-shot flag so the runnable never re-schedules itself:

```kotlin
private var activationRetryPending = false

private fun scheduleActivationRetry() {
    if (activationRetryPending) return
    activationRetryPending = true
    val runnable = Runnable {
        activationRetryRunnable = null
        activationRetryPending = false
        evaluateForeground()    // if still no match, does NOT re-schedule
    }
    activationRetryRunnable = runnable
    handler.postDelayed(runnable, Constants.ACTIVATION_RETRY_MS)
}

private fun cancelActivationRetry() {
    activationRetryRunnable?.let {
        handler.removeCallbacks(it)
        activationRetryRunnable = null
    }
    activationRetryPending = false
}
```

The existing `else if` branch in `evaluateForeground()` will see `activationRetryPending == true` and skip re-scheduling. One fire, then done.

**Also add a test** for this path: retry fires, foreground still not detected, verify `postDelayed` is called exactly once total (not twice). The current test suite doesn't cover this and would have caught the bug if it did.

---

## Minor — `activationRetryRunnable!!` non-null assertion

`scheduleActivationRetry()` — safe in context (field was just set), but the `!!` is avoidable. Addressed naturally by the fix above (use a local val `runnable` and pass it directly to both `activationRetryRunnable` and `handler.postDelayed`).

---

## Minor — pending retry not cancelled in `onCameraAvailable()`

When a non-PC app releases the camera, any pending retry fires harmlessly 1 s later (sees `anyCameraUnavailable() == false`, doesn't re-schedule with the fix above). No correctness issue, but calling `cancelActivationRetry()` inside `onCameraAvailable()` when all cameras become available would be cleaner.

---

## Everything else looks good

The six tests cover scheduling, idempotence, retry success, no-retry-on-immediate-success, `reset()` cancellation, and permission-revoked cancellation. The `cancelActivationRetry()` placements in the permission-lost and overlay-show paths are correct. The `reset()` cleanup is correct. The production logic flow in `evaluateForeground()` is sound — just needs the one-shot constraint to match the stated intent.

---

---

aunger:
## Code Review

### Overview

This PR implements a one-shot retry mechanism (`DT-06a`) to handle UsageStats lag: when a camera opens but UsageStats hasn't yet reported Pixel Camera as the foreground app, `evaluateForeground()` schedules a single 1-second retry rather than failing silently. The retry is cancelled on success, permission revocation, camera release, or service destruction.

---

### Code Quality — `OverlayServiceLogic.kt`

**Strengths:**
- The `activationRetryPending` flag is a clean solution to the re-entrancy/double-scheduling problem. The comment explaining why it stays `true` during `evaluateForeground()` execution is exactly the right level of documentation.
- `cancelActivationRetry()` is called in all the right places: permission loss, overlay confirmed, camera released, and `reset()`.
- The one-shot design (no polling loop) is intentional and correct.

**Subtle ordering note in `scheduleActivationRetry()`:**
```kotlin
val runnable = Runnable {
    activationRetryRunnable = null   // ← nulled before work is done
    evaluateForeground()
    activationRetryPending = false
}
```
If `evaluateForeground()` internally calls `cancelActivationRetry()` (e.g. on success), that method sets `activationRetryPending = false`, then the runnable sets it `false` again — harmless double-write, but worth noting it's deliberate. No bug here.

**Minor: `else if` condition breadth.** The retry fires whenever `!isOverlayActive && cameraState.anyCameraUnavailable()`. This correctly handles a non-Pixel-Camera app opening the camera (retry fires, Pixel Camera not in foreground, overlay stays hidden, no re-schedule). Correct, but a comment confirming this case is handled intentionally would help future readers.

---

### Tests — `OverlayServiceLogicTest.kt`

**Good coverage:**
- Retry scheduled on lag ✓
- Idempotent (no double-schedule) ✓
- Retry success → overlay shown ✓
- No retry on immediate success ✓
- Cancelled on `reset()` ✓
- One-shot (no re-schedule if still lagging after firing) ✓
- Cancelled on permission revoke ✓

**Missing test case:**
There is no test for "pending retry cancelled when camera is released before retry fires." The production code handles this (`cancelActivationRetry()` in `onCameraAvailable()`), but it lacks a corresponding test — arguably the most likely real-world scenario for a stale retry.

**PR description inaccuracy:** The body says "Six new unit tests" but the diff adds **seven**.

---

### Potential Risks

- **UX gap:** If UsageStats lag exceeds 1 second (unlikely but possible under heavy load), the overlay never appears for that camera-open event. This is intentional per the one-shot design, but worth noting as an accepted limitation.
- **No thread-safety concern** — all state is accessed on the handler thread.

---

### Summary

| | |
|---|---|
| Correctness | ✅ Solid |
| Test coverage | ⚠️ Missing one case (camera-released-while-retry-pending) |
| Style/conventions | ✅ Consistent with codebase |
| Risk | Low |

**Recommended before merge:** Add a test verifying that `handler.removeCallbacks()` is called when the camera is released while a retry is pending. The PR description's "six tests" count should be corrected to seven.

---

PR #31, opened 2026-04-20
Sign release APK using keystore secrets in CI

aunger:
Decodes KEYSTORE_BASE64 into a temp file and passes the four keystore
env vars to assembleRelease so the workflow produces a properly signed
APK. The signing plumbing in app/build.gradle.kts already supports this.

https://claude.ai/code/session_014vDsfMtjrnaMtC9CjtHVRF


aunger:
@claude review, please

Issue #35, opened 2026-04-20
Allow typing X/Y position percentages in addition to using sliders


aunger:
## Feature Request

The overlay button position is currently set only via sliders for X and Y percentage. Users should also be able to type a value directly for more precise control.

## Desired Changes

- Next to (or below) each position slider, add a numeric text input field displaying the current percentage value.
- Editing the text field should update the slider and the overlay position, and vice-versa (two-way binding).
- Input should be validated/clamped to the range 0–100.

Issue #34, opened 2026-04-20
Reduce default camera-close detection interval to 50ms; make it configurable in Advanced Settings


aunger:
## Feature Request

The overlay currently takes ~500ms to disappear after the camera app is closed. This feels sluggish.

## Desired Changes

1. **Lower the default polling/check interval** from ~500ms to **50ms** so the overlay dismisses much more quickly after the camera closes.
2. **Expose this interval as a setting** on the Advanced Settings screen so users can tune it.
   - Use a plain numeric input field (no slider widget).
   - Label: something like "Camera close detection interval (ms)".
   - Store the value in the existing shared preferences.

PR #37, opened 2026-04-20
Implement #34 & #35: debounce config + X/Y text inputs

aunger:
- Lower CAMERA_DEBOUNCE_MS default from 500ms to 50ms (#34)
- Add PREF_CAMERA_DEBOUNCE_MS preference; PrefsManager exposes
  cameraDebounceMs with the 50ms default as fallback
- OverlayService passes prefsManager.cameraDebounceMs to
  OverlayServiceLogic so the live value is used at service start
- Advanced Settings: numeric text field for camera close detection
  interval (ms), no slider (#34)
- Advanced Settings: OutlinedTextField beside each position label
  for X and Y, two-way synced with their sliders (#35)

https://claude.ai/code/session_01ECHfLGnE1PKXZCDXaa6Zx3

### DIFF
--- a/app/src/main/java/com/gb4pc/Constants.kt
+++ b/app/src/main/java/com/gb4pc/Constants.kt
@@ -13,7 +13,7 @@ object Constants {
     const val MAX_SIZE_PERCENT = 30.0f
 
     // Camera debounce delay (DT-04)
-    const val CAMERA_DEBOUNCE_MS = 500L
+    const val CAMERA_DEBOUNCE_MS = 50L
 
     // UsageStats query window (DT-02)
     const val USAGE_STATS_WINDOW_MS = 5000L
@@ -35,6 +35,7 @@ object Constants {
     const val PREF_GALLERY_PACKAGE = "gallery_package"
     const val PREF_OVERLAY_POSITIONS = "overlay_positions"
     const val PREF_SETUP_COMPLETED = "setup_completed"
+    const val PREF_CAMERA_DEBOUNCE_MS = "camera_debounce_ms"
 
     // Secure viewer
     const val SESSION_TIMESTAMP_TOLERANCE_MS = 2000L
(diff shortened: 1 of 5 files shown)


aunger:
## Implementation summary

### #34 — Camera close detection interval

- `CAMERA_DEBOUNCE_MS` lowered from 500 ms to **50 ms** as the new default.
- Valid range enforced as **10–1000 ms** (`MIN_CAMERA_DEBOUNCE_MS` / `MAX_CAMERA_DEBOUNCE_MS`). The 10 ms floor prevents flicker from transient camera releases during mode switches; 1 s is a reasonable ceiling.
- `PrefsManager.cameraDebounceMs` persists the value (falls back to the 50 ms default). `OverlayService.onCreate` passes it into `OverlayServiceLogic`, which stores it for the lifetime of the service — changing the setting takes effect on the next service restart, noted in the UI.
- Advanced Settings gains a plain numeric `OutlinedTextField` (no slider) labelled **"Camera close detection interval (ms, 10–1000)"**. Saving happens on Done/IME; invalid or out-of-range input reverts to the last saved value.

### #35 — Typed X/Y/size position inputs

- Each of the three position controls (X, Y, Size) now has an `OutlinedTextField` alongside its label, showing the current percentage with a `%` suffix.
- Typing live-moves the slider (clamped to the field's valid range); the value is saved and the text reformatted to `"%.2f"` only on Done/IME — no per-keystroke prefs writes, consistent with `onValueChangeFinished` on the sliders.
- The Reset button restores all three float states and all three text states to defaults.

### Review cycle notes

An expert review pass after the initial implementation caught and fixed:
- `LocalFocusManager` captured inside the `Scaffold` content lambda — hoisted to the top of the composable.
- X/Y fields were calling `savePosition()` on every keystroke; moved to `commitXText()` / `commitYText()` called on Done.
- Debounce field was saving on every keystroke and accepted unbounded values; moved to `commitDebounceText()` on Done with proper clamp + revert logic.
- Addenda then updated the debounce range from 0–5000 to 10–1000 and extended the text-input treatment to the Size field.

---

PR #42, opened 2026-04-23
Automate version sync: git tag is single source of truth for versionName

aunger:
## What this does

Makes the **git tag** the single source of truth for `versionName`, so the version baked into the APK, the APK filename, and the GitHub Release name are always identical. Eliminates the previous manual step of editing `versionName` in `build.gradle.kts` before tagging.

Also enables **signed release APKs on every CI build** (not just tagged releases), making it easy to sideload any pre-release build.

### `versionName` — injected from outside, never hardcoded

`app/build.gradle.kts` reads `versionName` from a Gradle project property (`-PversionName=…`). CI always supplies it; local builds without the property get `"dev"` as a fallback.

| Build | `versionName` | `versionCode` | APK filename |
|---|---|---|---|
| Local dev | `dev` | `yyyyMMdd` (date) | `app-release.apk` |
| CI merge to main | `dev.<run_number>` | `github.run_number` | `gb4pc-dev.<run_number>.apk` |
| Tagged release | `X.Y.Z` (from tag) | `github.run_number` | `gb4pc-X.Y.Z.apk` |

### `versionCode` — `github.run_number` in CI

`github.run_number` is a monotonically-increasing integer scoped to the repository. It fits easily in Android's int32 `versionCode` limit, never collides (even two releases on the same day get different values), and makes APK version ordering unambiguous for sideloading.

### `scripts/tag-release.sh` — one command to release

```
./scripts/tag-release.sh 1.2.3
```

Validates semver format, checks for a clean working tree, creates an annotated `v1.2.3` tag, and pushes it. The release workflow picks up from there — no file edits required.

### Workflow structure

Three jobs serve distinct purposes; none is redundant:

| Job | Trigger | Output | Signed |
|---|---|---|---|
| `build` (unit tests) | PRs + pushes | test report | — |
| `release-build` | push to main | `gb4pc-dev.N.apk` artifact | yes |
| `release.yml` | `v*` tag | GitHub Release + APK | yes |

`release-build` and the release workflow both produce signed APKs but differ in trigger, naming, and destination (Actions artifact vs. GitHub Release). The `release-build` job provides installable builds for every merge, so you don't have to tag just to test a sideloaded build.

## Files changed

- `app/build.gradle.kts` — `versionName` reads from `-PversionName` Gradle property
- `app/src/test/java/com/gb4pc/ReleaseVersionTest.kt` — updated to accept `dev.N` labels and `github.run_number` as versionCode
- `.github/workflows/build.yml` — version + build number injected everywhere; `release-build` job now signs and names APKs
- `.github/workflows/release.yml` — version extracted before unit tests so it can be passed to both; `BUILD_NUMBER=github.run_number` added
- `scripts/tag-release.sh` — new release helper

https://claude.ai/code/session_01GGycetahvWiq2FMpZDYT3g

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,11 +35,10 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
 
-      - name: Build debug APK
-        run: ./gradlew assembleDebug --no-daemon
-
-      - name: Run unit tests
-        run: ./gradlew testDebugUnitTest --no-daemon
+      - name: Build and test
+        env:
+          BUILD_NUMBER: ${{ github.run_number }}
+        run: ./gradlew assembleDebug testDebugUnitTest -PversionName=dev.${{ github.run_number }} --no-daemon
 
       - name: Upload test results
         if: always()
@@ -126,7 +125,11 @@ jobs:
         run: chmod +x gradlew
 
       - name: Build app, test, and mock Pixel Camera APKs
-        run: ./gradlew assembleDebug assembleDebugAndroidTest :e2e-mock-camera:assembleDebug --no-daemon
+        env:
+          BUILD_NUMBER: ${{ github.run_number }}
+        run: >
+          ./gradlew assembleDebug assembleDebugAndroidTest :e2e-mock-camera:assembleDebug
+          -PversionName=dev.${{ github.run_number }} --no-daemon
 
       - name: Start emulator
         run: |
@@ -241,7 +244,9 @@ jobs:
           path: app/build/outputs/androidTest-results/
           retention-days: 14
 
-  # Produce a release build on every merge to main (continuous delivery)
+  # Produce a signed release APK on every merge to main for easy sideloading.
+  # Named gb4pc-dev.<run_number>.apk to distinguish from tagged releases.
+  # Runs after build (unit tests must pass) but independently of e2e-tests.
   release-build:
     if: github.event_name == 'push'
     needs: build
@@ -272,12 +277,27 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
 
-      - name: Build release APK
-        run: ./gradlew assembleRelease --no-daemon
+      - name: Decode keystore
+        run: |
+          echo "${{ secrets.KEYSTORE_BASE64 }}" | base64 --decode > "$RUNNER_TEMP/release.keystore"
+
+      - name: Build signed release APK
+        env:
+          KEYSTORE_PATH: ${{ runner.temp }}/release.keystore
+          KEYSTORE_PASSWORD: ${{ secrets.KEYSTORE_PASSWORD }}
+          KEY_ALIAS: ${{ secrets.KEY_ALIAS }}
+          KEY_PASSWORD: ${{ secrets.KEY_PASSWORD }}
+          BUILD_NUMBER: ${{ github.run_number }}
+        run: ./gradlew assembleRelease -PversionName=dev.${{ github.run_number }} --no-daemon
+
+      - name: Rename APK
+        run: |
+          mv app/build/outputs/apk/release/*.apk \
+             app/build/outputs/apk/release/gb4pc-dev.${{ github.run_number }}.apk
 
       - name: Upload release APK
         uses: actions/upload-artifact@v7
         with:
-          name: release-apk
-          path: app/build/outputs/apk/release/*.apk
+          name: gb4pc-dev.${{ github.run_number }}
+          path: app/build/outputs/apk/release/gb4pc-dev.${{ github.run_number }}.apk
           retention-days: 14
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,54 +7,15 @@ on:
     branches: [main]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-
-    steps:
-      - uses: actions/checkout@v6
-
-      - name: Set up JDK 17
-        uses: actions/setup-java@v5
-        with:
-          java-version: '17'
-          distribution: 'temurin'
-
-      - name: Set up Android SDK
-        uses: android-actions/setup-android@v4
-
-      - name: Cache Gradle packages
-        uses: actions/cache@v5
-        with:
-          path: |
-            ~/.gradle/caches
-            ~/.gradle/wrapper
-          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle.kts', '**/gradle-wrapper.properties') }}
-          restore-keys: |
-            ${{ runner.os }}-gradle-
-
-      - name: Grant execute permission for gradlew
-        run: chmod +x gradlew
-
-      - name: Build and test
-        env:
-          BUILD_NUMBER: ${{ github.run_number }}
-        run: ./gradlew assembleDebug testDebugUnitTest -PversionName=dev.${{ github.run_number }} --no-daemon
-
-      - name: Upload test results
-        if: always()
-        uses: actions/upload-artifact@v7
-        with:
-          name: test-results
-          path: app/build/reports/tests/
-          retention-days: 14
-
-  # End-to-end tests on an emulator using a minimal mock Pixel Camera stub.
+  # Gate for PRs and merges: compiles, runs unit tests, then E2E tests on an emulator.
+  # Unit tests run first so a fast failure skips the slower emulator phase.
+  # On success the single debug build is reused for both test suites.
   #
-  # The real Pixel Camera APK refuses to start on non-Pixel emulators (PairIP
-  # device-compatibility check). The e2e-mock-camera module has the same
-  # applicationId ("com.google.android.GoogleCamera") and opens the camera
-  # hardware on resume, so GB4PC's CameraManager callback and UsageStats-based
-  # foreground detection are exercised identically to the real app.
+  # E2E background: the real Pixel Camera APK refuses to start on non-Pixel emulators
+  # (PairIP device-compatibility check). The e2e-mock-camera module has the same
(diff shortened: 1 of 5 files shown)


aunger:
## Code Review — PR #42: Automate version sync: git tag is single source of truth for versionName

The overall design is sound — using the git tag as the single source of truth for `versionName` eliminates a class of manual errors. The `tag-release.sh` helper and the restructured workflow jobs are a clear improvement over the previous approach. CI checks pass. That said, there are several issues ranging from a security concern to a correctness bug in the test suite.

---

### 1. Security: Unquoted Gradle property in shell command (release.yml)

**Severity: Medium**

In `release.yml`, the version extracted from the tag is passed directly into the shell command without quoting:

```yaml
run: ./gradlew assembleRelease -PversionName=${{ steps.version.outputs.version }} --no-daemon
```

If a tag like `v1.0.0 --some-flag` were pushed (either by accident or maliciously by someone with tag-push rights), the unquoted value would be split by the shell and passed as additional arguments to Gradle. The same issue exists in the unit test step. The fix is to quote the substitution:

```yaml
run: ./gradlew assembleRelease -PversionName='${{ steps.version.outputs.version }}' --no-daemon
```

`tag-release.sh` validates strict semver before pushing, which mitigates this for normal use — but the workflow also accepts `workflow_dispatch` with a free-text `tag_name` input, which bypasses that validation entirely. A user could trigger the workflow with a crafted tag name via the GitHub UI.

---

### 2. Bug: `ReleaseVersionTest` fails for local dev builds with plain `"dev"` versionName

**Severity: High — test correctness**

`build.gradle.kts` sets `versionName = findProperty("versionName") as String? ?: "dev"` for local builds with no `-PversionName` argument. The updated `ReleaseVersionTest` validates:

```kotlin
semver.matches(versionName) || devLabel.matches(versionName)
```

where `devLabel = Regex("""^dev\.\d+$""")`. But the fallback value `"dev"` matches *neither* regex — `dev` without a trailing `.\d+` fails `devLabel`. Running `./gradlew testDebugUnitTest` locally without `-PversionName` will therefore fail this test.

Either the regex should allow plain `"dev"`:
```kotlin
val devLabel = Regex("""^dev(\.\d+)?$""")
```
Or the test should be skipped/annotated when `versionName` is exactly `"dev"` (indicating a non-CI local build where versioning constraints don't matter).

---

### 3. Missing keystore-secret guard in `dev-build` job (build.yml)

**Severity: Low-Medium**

The `Decode keystore` step runs unconditionally:

```yaml
- name: Decode keystore
  run: |
    echo "${{ secrets.KEYSTORE_BASE64 }}" | base64 --decode > "$RUNNER_TEMP/release.keystore"
```

If `KEYSTORE_BASE64` is not configured (e.g. on a newly-cloned fork, or if the secret is accidentally deleted), this writes an empty/corrupt keystore file and the subsequent `assembleRelease` step fails with a cryptic Gradle error rather than a clear message. A guard would make failures self-explanatory:

```yaml
- name: Decode keystore
  run: |
    if [[ -z "${{ secrets.KEYSTORE_BASE64 }}" ]]; then
      echo "Error: KEYSTORE_BASE64 secret is not set." >&2
      exit 1
    fi
    echo "${{ secrets.KEYSTORE_BASE64 }}" | base64 --decode > "$RUNNER_TEMP/release.keystore"
```

---

### 4. `tag-release.sh` does not guard against releasing from a non-main branch

**Severity: Low**

The script validates semver format and a clean working tree, but does not check that the current branch is `main`. It is easy to accidentally run `./scripts/tag-release.sh 1.2.3` from a feature branch, which would create a release tag pointing at an unreviewed commit. A simple guard:

```bash
CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
if [[ "$CURRENT_BRANCH" != "main" ]]; then
    echo "Error: releases must be tagged from main, current branch is '$CURRENT_BRANCH'." >&2
    exit 1
fi
```

---

### 5. Minor: `BUILD_NUMBER` env var and `-PversionName` are coupled but set independently

In `build-and-test`, the step sets both:
```yaml
env:
  BUILD_NUMBER: ${{ github.run_number }}   # → versionCode
run: ... -PversionName=dev.${{ github.run_number }} ...  # → versionName
```

These two values are independently substituted from `github.run_number`, so they will always match in practice — but the coupling is implicit. A comment in the workflow (or in `build.gradle.kts`) explaining that `BUILD_NUMBER` drives `versionCode` while `-PversionName` drives `versionName` would help future maintainers avoid accidentally breaking the pairing.

---

### Summary

| # | Issue | Severity | Blocking? |
|---|-------|----------|-----------|
| 1 | Unquoted `versionName` in shell — injection via `workflow_dispatch` | Medium | Yes |
| 2 | `ReleaseVersionTest` fails locally with plain `"dev"` versionName | High | Yes |
| 3 | No guard for missing `KEYSTORE_BASE64` secret | Low-Medium | No |
| 4 | `tag-release.sh` allows tagging from non-main branch | Low | No |
| 5 | `BUILD_NUMBER`/`-PversionName` coupling is implicit | Low | No |

Issues 1 and 2 should be fixed before merging. Issues 3–5 are non-blocking suggestions.

**CHANGES_REQUESTED**

---

aunger:
## Code Review

The overall design — git tag as single source of truth, `github.run_number` as `versionCode` in CI, `scripts/tag-release.sh` as the release entry point — is sound and a clear improvement over editing `versionName` in Gradle by hand. The test updates are appropriate. I have a few issues that warrant attention before merge.

---

### 1. `ReleaseVersionTest` now passes on local dev builds with `versionName = "dev"` (no dot-N suffix) — but that case isn't in the regex

`build.gradle.kts` falls back to `"dev"` (no run number) when `-PversionName` is not supplied. The updated `devLabel` regex is `^dev\.\d+$`, which requires a dot and at least one digit. A local `./gradlew testDebugUnitTest` without `-PversionName` will produce `versionName = "dev"`, which matches neither `semver` nor `devLabel`, causing the test to **fail on every local developer machine**.

Either the fallback should be `"dev.0"` (to match the regex), or the regex should also accept the plain `"dev"` string:
```kotlin
val devLabel = Regex("""^dev(\.\d+)?$""")
```
The second option is more honest — a local build genuinely has no run number.

---

### 2. `build.yml`: unit tests and E2E APK assembly are coupled in a single Gradle invocation

The "Build and run unit tests" step runs:
```
./gradlew assembleDebug assembleDebugAndroidTest :e2e-mock-camera:assembleDebug testDebugUnitTest -PversionName=dev.${{ github.run_number }} --no-daemon
```
The comment above the job promises "Unit tests run first so a fast failure skips the slower emulator phase." That's the right intent, but because all tasks are in a single Gradle invocation they are scheduled by Gradle's task graph together — Gradle will interleave them based on dependencies, not necessarily run unit tests before assembly completes. More importantly, if a unit test fails the entire invocation exits non-zero, which also means the `assembleDebugAndroidTest` output may be absent, causing the emulator steps to fail with a missing APK rather than a clean "unit tests failed" signal.

Splitting into two sequential steps preserves the fast-fail intent and gives cleaner CI signals:
```yaml
- name: Build APKs
  run: ./gradlew assembleDebug assembleDebugAndroidTest :e2e-mock-camera:assembleDebug --no-daemon
  env:
    BUILD_NUMBER: ${{ github.run_number }}

- name: Run unit tests
  run: ./gradlew testDebugUnitTest -PversionName=dev.${{ github.run_number }} --no-daemon
```

---

### 3. `release.yml`: `versionName` is injected into unit tests but not into `assembleRelease`

After the refactor, the unit-test step correctly passes `-PversionName=${{ steps.version.outputs.version }}`. The `assembleRelease` step also passes `-PversionName=...`, so the APK itself gets the right version. That part is fine.

However, `ReleaseVersionTest` runs during `testDebugUnitTest` (debug variant), while `assembleRelease` builds the release variant. The `BuildConfig.VERSION_NAME` baked into the **debug** test binary and the one in the **release** APK will match because both Gradle invocations receive `-PversionName=`. This is consistent — just worth noting that the test is validating the debug-variant `BuildConfig`, not the release APK's manifest directly.

---

### 4. `scripts/tag-release.sh`: `git rev-parse "$TAG"` check uses a ref that could match a branch

```bash
if git rev-parse "$TAG" &>/dev/null; then
```
`git rev-parse` resolves any ref, including branches. If a branch named `v1.2.3` exists for any reason, this check returns success and incorrectly reports "tag already exists." The intent is clearly to check for a tag specifically; use:
```bash
if git rev-parse "refs/tags/$TAG" &>/dev/null; then
```

---

### 5. Minor: `BUILD_NUMBER` env var is set in the `build-and-test` job but not used by `testDebugUnitTest`

In `build.gradle.kts`, `versionCode = buildNumber`, where `buildNumber` reads `BUILD_NUMBER` from the environment. In the `build-and-test` job, `BUILD_NUMBER` is set as an env var on the combined Gradle step, and `-PversionName=dev.${{ github.run_number }}` is passed. The `versionCode` will therefore be `github.run_number` as intended. This is consistent — just confirm the `BUILD_NUMBER` env on the unit-test-only step in `release.yml` is actually needed (it is, for the same reason).

---

### Summary

| # | Severity | Blocking? |
|---|---|---|
| 1 | Bug — test fails for every local dev build | Yes |
| 2 | CI reliability / misleading failure messages | Recommended fix |
| 3 | Informational | No |
| 4 | Correctness of pre-release guard in shell script | Recommended fix |
| 5 | Informational | No |

Issues 1 and 4 should be fixed before merge. Issue 2 is a strong recommendation.

---

aunger:
✅ **Approved.** The fixes address both blocking issues:
1. ReleaseVersionTest.kt regex now accepts both "dev" and "dev.N" patterns
2. tag-release.sh uses refs/tags/$TAG to validate specifically in the tag namespace

The broader implementation (version injection, workflow structure, Gradle property fallback) is coherent and correct. Ready to merge.

---

PR #43, opened 2026-04-24
Fix crash on Android 16: remove LIMIT from ContentProvider sort order

aunger:
Android 16 rejects LIMIT in the sortOrder string with
IllegalArgumentException. Since queryMediaStore already only reads the
first row via moveToFirst(), dropping LIMIT 1 is safe and correct.

Fixes crash reported against 20260420 build.

https://claude.ai/code/session_01Pfte26HasStNJn7VRH6Qop

### DIFF
--- a/app/src/main/java/com/gb4pc/service/OverlayService.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayService.kt
@@ -307,7 +307,7 @@ class OverlayService : Service() {
         isVideo: Boolean
     ): MediaItem? {
         val selection = "${MediaStore.MediaColumns.DATE_ADDED} >= ?"
-        val sortOrder = "${MediaStore.MediaColumns.DATE_ADDED} DESC LIMIT 1"
+        val sortOrder = "${MediaStore.MediaColumns.DATE_ADDED} DESC"
         contentResolver.query(contentUri, projection, selection, selectionArgs, sortOrder)?.use { cursor ->
             if (cursor.moveToFirst()) {
                 val id = cursor.getLong(cursor.getColumnIndexOrThrow(MediaStore.MediaColumns._ID))


aunger:
## Code Review — PR #43: Fix crash on Android 16: remove LIMIT from ContentProvider sort order

The fix is correct and well-reasoned. Android 16 tightened enforcement of `ContentProvider` query contracts, rejecting any `LIMIT` or `OFFSET` clauses in the `sortOrder` string with `IllegalArgumentException`. Removing `LIMIT 1` from the sort string is the right remedy.

**Correctness:** The removal is safe because `queryMediaStore` already reads only the first row via `cursor.moveToFirst()` — so the result set is bounded in practice regardless of how many rows the provider returns. The cursor is also properly closed via the `.use {}` block, so there is no resource leak from returning more rows.

**One minor suggestion:** The commit message mentions this fixes a crash reported against the `20260420` build. Consider adding a comment above the `sortOrder` line explaining *why* `LIMIT` is absent — future maintainers may be tempted to re-add it for "efficiency":

```kotlin
// Do not append "LIMIT 1" here — Android 16+ rejects SQL clauses in sortOrder
// (throws IllegalArgumentException). We only read the first row via moveToFirst().
val sortOrder = "${MediaStore.MediaColumns.DATE_ADDED} DESC"
```

This is a suggestion, not a blocker. The fix itself is sound.

**No changes requested.**

---

aunger:
## Code Review

**The fix is correct and the diagnosis is accurate.** Android 16 enforces that the `sortOrder` parameter passed to `ContentProvider.query()` must not contain SQL-injection-style clauses like `LIMIT`; passing one throws `IllegalArgumentException`. Removing `LIMIT 1` from the sort string is the right fix.

### Correctness

The removal is safe. `queryMediaStore` already uses `cursor.moveToFirst()` and only reads the first row, so dropping `LIMIT 1` from the sort string does not change observable behavior on any API level — it just stops crashing on Android 16.

### One issue to address: performance regression on Android ≤ 15

Removing `LIMIT 1` means the `ContentProvider` now returns **all** matching rows (potentially thousands of photos/videos taken in the same session) instead of one. On Android ≤ 15 the old code happened to work because SQLite honored the `LIMIT` appended to the sort clause, but the correct cross-API way to cap results is via the `Bundle`-based query API introduced in API 26 (which is already `minSdk` for this project):

```kotlin
private fun queryMediaStore(
    contentUri: Uri,
    projection: Array<String>,
    selectionArgs: Array<String>,
    isVideo: Boolean
): MediaItem? {
    val selection = "${MediaStore.MediaColumns.DATE_ADDED} >= ?"
    val sortOrder = "${MediaStore.MediaColumns.DATE_ADDED} DESC"

    val queryArgs = Bundle().apply {
        putString(ContentResolver.QUERY_ARG_SQL_SELECTION, selection)
        putStringArray(ContentResolver.QUERY_ARG_SQL_SELECTION_ARGS, selectionArgs)
        putString(ContentResolver.QUERY_ARG_SQL_SORT_ORDER, sortOrder)
        putInt(ContentResolver.QUERY_ARG_LIMIT, 1)          // honored on all API ≥ 26, no crash on API 36
    }

    contentResolver.query(contentUri, projection, queryArgs, null)?.use { cursor ->
        if (cursor.moveToFirst()) {
            val id = cursor.getLong(cursor.getColumnIndexOrThrow(MediaStore.MediaColumns._ID))
            val mime = cursor.getString(cursor.getColumnIndexOrThrow(MediaStore.MediaColumns.MIME_TYPE))
            val dateTakenCol = cursor.getColumnIndex(MediaStore.MediaColumns.DATE_TAKEN)
            val dateTaken = if (dateTakenCol >= 0) cursor.getLong(dateTakenCol) else System.currentTimeMillis()
            val uri = ContentUris.withAppendedId(contentUri, id)
            return MediaItem(uri.toString(), dateTaken, isVideo)
        }
    }
    return null
}
```

`ContentResolver.QUERY_ARG_LIMIT` is available from API 26 onward (matches `minSdk`) and is the documented replacement for embedding `LIMIT` in the sort string. The `MediaStore` provider on Android 16 explicitly supports it.

**This PR as written stops the crash**, which is the critical goal for users on Android 16 right now. The Bundle-based approach is a follow-up improvement, not a blocker — but worth filing as a task so it doesn't get forgotten.

---

aunger:
✅ **Approved.** The Android 16 crash fix is correct—removing LIMIT from sortOrder is safe since the code already reads only the first row via moveToFirst(). Ready to merge.

---

PR #51, opened 2026-04-26
Fix #45: button never shows photos

### LINKED ISSUE #45: Button never shows photos
The solution for #33 resulted in the icon being shown but now new photos never appear.

aunger:
## Summary

`OverlayManager.show()` called `updateIcon()` every time it was invoked while the overlay was already visible. On two-camera devices `onCameraUnavailable` fires once per camera ID, so `show()` was called a second time *after* `showLatestPhotoThumbnail()` had set the photo bitmap — silently replacing it with the gallery-app icon.

Removed the redundant `updateIcon()` call. The icon is already set correctly when the view is first created in `createOverlayView()`, so no re-set is needed on subsequent `show()` calls while the overlay is active.

## Test plan
- [ ] Open Pixel Camera on a two-camera device, take a photo, and verify the overlay button updates to show the photo thumbnail rather than reverting to the gallery-app icon.
- [ ] All JVM unit tests pass.

https://claude.ai/code/session_01VFH5vqVWexnvwGQMmgVikJ

---
_Generated by [Claude Code](https://claude.ai/code/session_01VFH5vqVWexnvwGQMmgVikJ)_


aunger:
The test claims no Android libraries are needed, and then claims this can’t be tested without Android or Robolectric. Maybe so for unit tests, but this should certainly have a test in e2e suite.

PR #54, opened 2026-04-28
Add Robolectric test for Issue #45 regression

### LINKED ISSUE #58: Test: add Robolectric test for thumbnail-overwrite regression (issue #45 scenario)
## Background

`OverlayManagerTest.kt` contains a KDoc comment (lines 19–31) that explicitly acknowledges a missing Robolectric test for the three-step thumbnail-overwrite scenario fixed in PR #51 (issue #45):

1. Call `show()` → overlay is created with the gallery icon drawable.
2. Call `showLatestPhotoThumbnail(...)` → the `ImageView` now holds a `Bitmap` drawable.
3. Call `show()` again (simulating a second `onCameraUnavailable` event) → the `ImageView` still holds the `Bitmap` drawable, NOT a `Drawable` (i.e. `updateIcon()` was NOT called).

The comment notes that a JVM unit test cannot cover this because `OverlayManager` requires a real Android `Context`, `WindowManager`, and `KeyguardManager` — hence Robolectric (or an instrumented test) is needed.

## What is currently guaranteed

The fix in PR #51 was a single-line deletion, so the correctness guarantee currently rests on the diff and the documented contract in the KDoc. As of PR #57, `updateIcon()` has been deleted entirely from `OverlayManager`, removing the dead code path — but the behavioural contract for a future re-introduction still has no automated test.

## Suggested approach

Add a Robolectric test in `OverlayManagerTest.kt` (or a new `OverlayManagerRobolectricTest.kt`) that:

- Instantiates `OverlayManager` with a Robolectric-provided `Context`.
- Exercises the three-step sequence above.
- Asserts the `ImageView` drawable type after step 3 matches the bitmap set in step 2.

Once this test exists, the explanatory KDoc block in `OverlayManagerTest.kt` can be trimmed or removed.

## Tracked from

Review comment on PR #57: https://github.com/aunger/gallery-button-for-pixel-camera/pull/57#pullrequestreview-4194798736

aunger:
## Summary

Adds a comprehensive Robolectric integration test (`OverlayManagerRobolectricTest`) that validates the fix for Issue #45: verifying that a second call to `OverlayManager.show()` does not overwrite a loaded photo thumbnail with the gallery-app icon.

This addresses the feedback from PR #51: https://github.com/aunger/gallery-button-for-pixel-camera/pull/51#issuecomment-4321458240

## Problem

On dual-camera devices, `CameraManager.onCameraUnavailable()` fires once per physical lens. In PR #51, we removed the redundant `updateIcon()` call from `OverlayManager.show()` that was silently overwriting the thumbnail. However, the fix lacked a regression test.

## Solution

The Robolectric test simulates the exact scenario:
1. Call `show()` → overlay created with icon drawable
2. Call `setImageBitmap()` → simulates `showLatestPhotoThumbnail()` loading a thumbnail
3. Call `show()` again → simulates second camera's `onCameraUnavailable` callback
4. Assert drawable is still `BitmapDrawable` → confirms `updateIcon()` was NOT called

Also updated `OverlayManagerTest` documentation to resolve the contradictory claims about what can be tested without Android frameworks.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest` — all tests pass including new `OverlayManagerRobolectricTest`
- [x] Robolectric test isolates the exact regression scenario without requiring Pixel Camera or device setup

https://claude.ai/code/session_01DvCzDAijh2xvw9BWppKdN6

---
_Generated by [Claude Code](https://claude.ai/code/session_01DvCzDAijh2xvw9BWppKdN6)_

### DIFF
--- a/app/src/test/java/com/gb4pc/overlay/OverlayManagerRobolectricTest.kt
+++ b/app/src/test/java/com/gb4pc/overlay/OverlayManagerRobolectricTest.kt
@@ -1,0 +1,88 @@
+package com.gb4pc.overlay
+
+import android.app.Application
+import android.graphics.Bitmap
+import android.graphics.drawable.BitmapDrawable
+import android.view.WindowManager
+import android.widget.ImageView
+import androidx.test.core.app.ApplicationProvider
+import com.gb4pc.data.OverlayPosition
+import com.gb4pc.data.PrefsManager
+import org.junit.Assert.*
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.shadows.ShadowWindowManagerImpl
+
+/**
+ * Robolectric integration test for OverlayManager — Issue #45 regression.
+ *
+ * Verifies that a second call to show() (as fired by a dual-camera device that
+ * triggers onCameraUnavailable once per physical lens) does NOT overwrite a thumbnail
+ * that was already loaded by showLatestPhotoThumbnail() with the app icon.
+ *
+ * Before the fix, OverlayManager.show() called updateIcon() unconditionally when the
+ * overlay was already showing, resetting the ImageView to the gallery icon drawable.
+ * After the fix, show() returns early when isShowing == true, leaving any loaded
+ * thumbnail intact.
+ */
+@RunWith(RobolectricTestRunner::class)
+class OverlayManagerRobolectricTest {
+
+    /**
+     * Three-step regression guard for the dual-camera thumbnail-overwrite bug:
+     *
+     *   1. show()                          → overlay added; ImageView holds icon drawable
+     *   2. overlayView.setImageBitmap(bmp) → simulates showLatestPhotoThumbnail() success
+     *   3. show() again                    → must be a no-op (isShowing guard)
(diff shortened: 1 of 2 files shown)


aunger (commented):
## Review of PR #54 — Robolectric regression test for Issue #45

CI passes (`build-and-test` ✅). The test accurately captures the dual-camera scenario and the `OverlayManagerTest` comment cleanup is a straightforward improvement now that an actual test exists.

Two things worth discussing:

---

### 1. `ShadowWindowManagerImpl` cast is fragile

```kotlin
val shadowWm = shadowOf(windowManager) as ShadowWindowManagerImpl
```

This is the de-facto Robolectric pattern for accessing `views`, but it couples the test to a concrete internal class name rather than an interface. If Robolectric renames or reorganises the shadow class in a future major version, this breaks at runtime (not at compile time) with a `ClassCastException`.

A slightly more defensive form that a lot of projects use is to access `views` via reflection or through a typed helper, but I acknowledge that there's no stable public API that avoids this cast today. Would be worth a comment noting the Robolectric version dependency so a future maintainer knows why the cast exists.

---

### 2. `updateIcon()` in `OverlayManager` appears to be dead code

`private fun updateIcon()` still exists in `OverlayManager.kt` but nothing calls it — `show()` now returns early via the `isShowing` guard before it could ever reach it. This predates this PR, so fixing it here would be scope creep. I'd suggest opening a follow-up issue to remove it before it confuses the next person reading `OverlayManager`.

---

Both of these are relatively minor. The test itself correctly guards the regression scenario and is the right level of integration (Robolectric, not a full instrumented test).

---

PR #69, opened 2026-04-30
Fix #63: prevent settings title from rendering behind the status bar

### LINKED ISSUE #63: Remove edge-to-edge full-screen in settings
The title, "GB4PC" , is shown behind the status bar clock.

aunger:
## Summary

- On Android 15 (`targetSdk = 35`) edge-to-edge is enforced for all activities, causing the `MainSettingsScreen` `Column` to start at the very top of the screen.
- The "GB4PC" title was rendered behind the status bar clock, as reported in issue #63.
- Added `windowInsetsPadding(WindowInsets.statusBars)` to the `Column` modifier in `MainSettingsScreen` so content begins below the status bar.

## Test plan

- [ ] Build and run on a device or emulator with Android 15 (API 35) and verify the "GB4PC" title is fully visible below the status bar.
- [ ] Confirm the existing instrumented test `MainSettingsScreenTest.mainScreen_showsAppName()` still passes.
- [ ] Confirm no other existing tests are broken.

---

### DIFF
--- a/app/src/main/java/com/gb4pc/ui/settings/MainActivity.kt
+++ b/app/src/main/java/com/gb4pc/ui/settings/MainActivity.kt
@@ -90,6 +90,7 @@ fun MainSettingsScreen(
     Surface(modifier = Modifier.fillMaxSize()) {
         Column(
             modifier = Modifier
+                .windowInsetsPadding(WindowInsets.statusBars)
                 .padding(16.dp)
                 .fillMaxWidth()
         ) {


aunger (commented):
## Code Review

**Verdict: Ready to merge** (GitHub prevents self-approval, so posting as a comment review.)

---

### Correctness

The fix is correct. On Android 15 (API 35), edge-to-edge is enforced system-wide, which means the `Column` starts at y=0 — directly behind the status bar. Adding `.windowInsetsPadding(WindowInsets.statusBars)` is the idiomatic Compose solution: it reads the system-reported status bar height and applies it as top padding, pushing content below the status bar on all device configurations and screen densities. This is preferable to a hard-coded `dp` value, and preferable to disabling edge-to-edge at the `Activity` level (which would be heavier and could cause other layout regressions).

### Simplicity

The change is a single line in exactly the right place. The modifier ordering is correct: `windowInsetsPadding` is applied first, then the existing `padding(16.dp)` adds the intended content inset on top of that, so the two compose additively rather than conflicting.

### Test coverage

No new tests are required for this change. The existing `MainSettingsScreenTest.mainScreen_showsAppName()` already exercises the scenario that was broken: it asserts that "GB4PC" is displayed. If the title were still hidden behind the status bar and Compose reported it as not visible, that assertion would fail. No tests were deleted or disabled.

### Commits

A single commit for a single-line, single-concern fix is appropriate.

### Minor observation (non-blocking)

`windowInsetsPadding(WindowInsets.statusBars)` comes from `androidx.compose.foundation.layout.*`, which is already imported via the wildcard — no extra import line is needed and none was added.

One thing worth a future issue: `WindowInsets.navigationBars` is not applied, so in gesture-nav mode with a home bar, the bottom of the screen could overlap. That is out of scope for this fix and was not introduced here.

---

PR #72, opened 2026-04-30
Fix #66: overlay never shows in v0.0.6

### LINKED ISSUE #66: v0.0.6 never shows overlay
v0.0.5 was fine. You can use git tags to find the changes.

aunger:
## Summary

- Fixes #66 — overlay never appears in v0.0.6 (regression introduced by PR #60).

**Root cause:** PR #60 added `onWindowFocusChanged` to the overlay `ImageView` to power the experimental focusable-overlay mode. When the default `FLAG_NOT_FOCUSABLE` window flags are used, Android immediately fires `onWindowFocusChanged(false)` as soon as the view is attached to the window. Without a guard, this called `onFocusLost()` unconditionally, which hid the overlay the instant it was shown — making the overlay never appear.

**Fix:** Add `if (!prefsManager.focusableOverlay) return` at the top of `onWindowFocusChanged` so focus callbacks are only dispatched when the experimental focusable-overlay preference is actually enabled.

## Test plan

- [x] New Robolectric regression test `show with non-focusable overlay does not trigger onFocusLost` verifies that `onFocusLost` is not called and the view remains in the `WindowManager` after `show()` when `focusableOverlay` is `false`.
- [x] Existing Issue #45 regression test is unchanged and still passes.
- [ ] All JVM unit tests pass (`./gradlew testDebugUnitTest`).

---

### DIFF
--- a/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
+++ b/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
@@ -136,6 +136,11 @@ class OverlayManager(
 
             override fun onWindowFocusChanged(hasFocus: Boolean) {
                 super.onWindowFocusChanged(hasFocus)
+                // Only invoke focus callbacks when the focusable-overlay mode is active.
+                // With FLAG_NOT_FOCUSABLE (default), the window never receives focus, so
+                // onWindowFocusChanged(false) fires immediately after show() — calling
+                // onFocusLost() here would hide the overlay the instant it appears (Issue #66).
+                if (!prefsManager.focusableOverlay) return
                 if (hasFocus) {
                     DebugLog.log("Overlay gained window focus")
                     onFocusGained()
--- a/app/src/test/java/com/gb4pc/overlay/OverlayManagerRobolectricTest.kt
+++ b/app/src/test/java/com/gb4pc/overlay/OverlayManagerRobolectricTest.kt
@@ -19,19 +19,61 @@ import org.robolectric.Shadows.shadowOf
 import org.robolectric.shadows.ShadowWindowManagerImpl
 
 /**
- * Robolectric integration test for OverlayManager — Issue #45 regression.
+ * Robolectric integration tests for OverlayManager.
  *
- * Verifies that a second call to show() (as fired by a dual-camera device that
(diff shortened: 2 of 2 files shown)


aunger (commented):
## Review — Fix #66: overlay never shows in v0.0.6

### Root-cause analysis

The analysis is correct. `onWindowFocusChanged` is a `View` callback, and for a `FLAG_NOT_FOCUSABLE` window Android fires it with `hasFocus = false` immediately after the view is attached. The pre-fix code dispatched `onFocusLost()` unconditionally at that point, hiding the overlay the instant it appeared. Reading `createLayoutParams()` confirms: the default (non-experimental) path sets `FLAG_NOT_FOCUSABLE`, so the window never receives real focus — the spurious `false` callback is guaranteed.

### Fix

The guard `if (!prefsManager.focusableOverlay) return` is the minimal, correct fix. It threads cleanly with the rest of the method: when `focusableOverlay` is `false`, the whole callback is a no-op; when it is `true`, the existing `hasFocus` branching is unchanged. The inline comment explains the "why" precisely and cross-references the issue number — exactly the right level of documentation for a non-obvious platform behaviour.

One worth-noting design point: the guard reads `prefsManager.focusableOverlay` live rather than caching it at construction time. If the pref changes while the overlay is live, `reshow()` is called to swap window flags, and the guard stays in sync automatically. The design is self-consistent.

### Test

The new Robolectric test is well-structured:
- It stubs `focusableOverlay = false`, calls `show()`, and then asserts both that `onFocusLost` was never invoked **and** that the view is still present in the `WindowManager`. Both assertions are needed — the count check confirms the callback is suppressed; the `shadowWm.views.size` check confirms the overlay was not hidden as a side effect. Good belt-and-suspenders coverage.
- The existing Issue #45 test is untouched and continues to pass.

### Guideline checklist

| Guideline | Status |
|---|---|
| Prefer simple implementations — no unnecessary complexity added | Pass |
| DRY — no repeated logic | Pass |
| New behaviour accompanied by a test | Pass |
| Existing tests not deleted or silently disabled | Pass |
| Change is small and discrete (single concern, single commit) | Pass |

### Verdict

**APPROVE** — The fix is correct, minimal, and well-tested. The PR is ready to merge.

---

PR #73, opened 2026-04-30
Fix #67: check for service when app opens

### LINKED ISSUE #67: Check for service when app opens
If the service is enabled in the settings, then verify that it is actually running when the app is launched. If it should be running but isn't, log this and start the service.

aunger:
## Summary

- Add `ServiceChecker` object with `isServiceRunning()` (queries `ActivityManager` for the app's own service) and `ensureServiceRunningIfEnabled()` (checks prefs, logs if the service is absent, and starts it).
- Call `ServiceChecker.ensureServiceRunningIfEnabled()` from `MainActivity.onCreate` (after the setup-redirect guard) so a crashed or OOM-killed service is automatically recovered when the app is launched.
- Add `ServiceCheckerTest` covering all three branches of `isServiceRunning` and all three branches of `ensureServiceRunningIfEnabled`.

## Test plan

- [ ] Run `./gradlew :app:testDebugUnitTest` — all unit tests pass, including the new `ServiceCheckerTest`.
- [ ] Manual: enable the service toggle, force-stop the service via adb/Settings → Apps, then re-open the app. The service should restart automatically.
- [ ] Manual: enable the service toggle and open the app while the service is already running — no duplicate start, no log entry containing "should be running but isn't".
- [ ] Manual: disable the service toggle and open the app — service is not started.

---

### DIFF
--- a/app/src/main/java/com/gb4pc/service/ServiceChecker.kt
+++ b/app/src/main/java/com/gb4pc/service/ServiceChecker.kt
@@ -1,0 +1,44 @@
+package com.gb4pc.service
+
+import android.app.ActivityManager
+import android.content.Context
+import com.gb4pc.data.PrefsManager
+import com.gb4pc.util.DebugLog
+
+/**
+ * Checks whether the overlay service should be running and starts it if it isn't (issue #67).
+ *
+ * Extracted as a standalone object so the logic can be unit-tested without a real Activity.
+ */
+object ServiceChecker {
+
+    /**
+     * Returns true if [OverlayService] currently has a running instance.
+     *
+     * [ActivityManager.getRunningServices] is deprecated for third-party use on API 26+ but
+     * remains functional for an app querying its own services, which is exactly our use case.
+     */
+    @Suppress("DEPRECATION")
+    fun isServiceRunning(context: Context): Boolean {
+        val am = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager
+        val serviceName = OverlayService::class.java.name
+        return am.getRunningServices(Int.MAX_VALUE)
+            .any { it.service.className == serviceName }
+    }
+
+    /**
+     * If the service is enabled in preferences but is not currently running, logs the
+     * discrepancy and starts the service.
+     *
+     * Call this on app launch (e.g. [android.app.Activity.onCreate]) so a crashed or
+     * killed service is automatically recovered.
+     */
+    fun ensureServiceRunningIfEnabled(context: Context, prefs: PrefsManager) {
+        if (!prefs.isServiceEnabled) return
+
+        if (!isServiceRunning(context)) {
+            DebugLog.log("Service should be running but isn't — starting it now")
+            OverlayService.start(context)
+        }
+    }
+}
--- a/app/src/main/java/com/gb4pc/ui/settings/MainActivity.kt
+++ b/app/src/main/java/com/gb4pc/ui/settings/MainActivity.kt
@@ -21,6 +21,7 @@ import androidx.compose.ui.unit.dp
 import com.gb4pc.R
 import com.gb4pc.data.PrefsManager
 import com.gb4pc.service.OverlayService
+import com.gb4pc.service.ServiceChecker
 import com.gb4pc.ui.picker.PickerActivity
 import com.gb4pc.ui.setup.SetupActivity
 import com.gb4pc.ui.theme.GB4PCTheme
@@ -48,6 +49,9 @@ class MainActivity : ComponentActivity() {
             finish()
             return
         }
+
+        // Issue #67: recover a service that should be running but was killed
+        ServiceChecker.ensureServiceRunningIfEnabled(this, prefsManager)
 
         // setContent called only once here (H2 fix)
         setContent {
--- a/app/src/test/java/com/gb4pc/service/ServiceCheckerTest.kt
+++ b/app/src/test/java/com/gb4pc/service/ServiceCheckerTest.kt
@@ -1,0 +1,101 @@
+package com.gb4pc.service
+
+import android.app.ActivityManager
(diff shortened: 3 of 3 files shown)


aunger (commented):
## Review of PR #73 — Check for service when app opens

Overall this is a solid, well-scoped implementation that correctly addresses issue #67. The code is clean, the extraction into a standalone object is exactly the right approach for testability, and all branches are exercised. I have **three issues to raise** — one is a correctness concern worth fixing before merge, the other two are minor.

---

### 1. CORRECTNESS — `ensureServiceRunningIfEnabled` in `onResume`: double-start risk when the user taps the enable toggle

**Severity: should fix before merge**

`MainActivity.onResume` is called immediately after the user flips the service toggle *on* from within the same settings screen. The sequence is:

1. User enables the toggle → `OverlayService.start(context)` is called directly in `onCheckedChange`.
2. The toggle opens a system-permission dialog (overlay permission) or the user simply returns focus to `MainActivity`.
3. `onResume` fires → `ServiceChecker.ensureServiceRunningIfEnabled` also calls `OverlayService.start(context)`.

Because `OverlayService` is `START_STICKY` and `onStartCommand` handles duplicate starts gracefully (the `callbackRegistered` guard), the double start is not catastrophic — but it is unnecessary, and it adds a spurious `startForegroundService` call every time `onResume` fires in the window between the toggle flip and the service actually appearing in the running-services list (which may be a few hundred ms).

The issue requirement says: *"If the service is enabled in the settings, verify it is actually running when the app is **launched**."* The intent is crash/OOM recovery, not routine re-starting. `onCreate` is a better fit: it fires once per Activity launch and is not triggered by focus returns from permission dialogs or other sub-activities.

The existing pattern for all other state refresh (permissions, gallery package) already uses `onResume` for a specific reason — those are observable states that change while the app is in the background (e.g. the user navigates to Settings → Apps to grant permission). Service recovery is different: the only time the service would be absent unexpectedly is on a fresh process start, which maps to `onCreate`.

**Suggestion:** Move the `ServiceChecker.ensureServiceRunningIfEnabled` call from `onResume` to `onCreate`, after the setup-redirect guard (so it only runs when the Activity will actually be shown). The PR description even says "Call this on every app open (e.g. `Activity.onResume`)…" — `onCreate` is the precise "app open" hook.

---

### 2. MINOR — Test for "enabled and already running" branch uses DebugLog as an indirect proxy for `startForegroundService` not being called

In `ServiceCheckerTest`:

```kotlin
// startForegroundService should not be called — verify via DebugLog (no "starting" message)
assertTrue(DebugLog.getEntries().none { it.message.contains("starting") })
```

This assertion is fragile: if the log message is ever rephrased, the assertion silently passes even if `startForegroundService` *is* called. Since `context` is already a Mockito mock in this test, directly verifying the method is never invoked is more precise and idiomatic:

```kotlin
verify(context, never()).startForegroundService(any())
```

This matches the approach used in the "disabled" branch test (verifying `activityManager.getRunningServices` is never called), and makes both tests symmetric and easier to read together.

---

### 3. MINOR — The `OverlayManager` / Issue #66 changes are bundled into this PR

The diff includes changes to `OverlayManager.kt` and `OverlayManagerRobolectricTest.kt` for issue #66 (the `focusableOverlay` / `FLAG_NOT_FOCUSABLE` fix). These are unrelated to the service-check feature. Per the project's commit guidelines, unrelated changes should be in their own commit or PR.

---

### API choice: `ActivityManager.getRunningServices`

The choice to use `ActivityManager.getRunningServices` with `@Suppress("DEPRECATION")` is correct and appropriate here. The deprecation notice on API 26+ only restricts *third-party* service queries (returning a capped, anonymised list); apps querying their *own* services continue to get accurate results on all current Android versions. The KDoc comment calling this out is exactly the right level of documentation.

---

### Summary

| Area | Assessment |
|---|---|
| `ServiceChecker.kt` logic | Correct. API choice is appropriate and well-documented. |
| `ServiceCheckerTest.kt` | All six branches covered. One indirect assertion worth tightening (see #2). |
| `onResume` vs `onCreate` | `onCreate` better matches the "check at app launch" intent and avoids spurious double-starts on every resume (see #1). |
| Unrelated #66 changes | Should be separated per commit-discipline guidelines (see #3). |

**Verdict: changes requested** on items #1 (move call site to `onCreate`) and #2 (direct mock verification). Ready to approve once those are addressed.


---

aunger (commented):
## Follow-up review — all three requested changes are addressed

Checked the updated diff against each item from my prior review.

### Item 1 — `onResume` → `onCreate` (was: correctness concern)

Confirmed. `ServiceChecker.ensureServiceRunningIfEnabled(this, prefsManager)` is now placed in `onCreate`, immediately after the setup-redirect guard (the early `return` block at line ~49), and before `setContent`. It will not fire on every focus return, eliminating the spurious double-start risk on resume. The KDoc comment in `ServiceChecker.kt` has been updated to say `Activity.onCreate` as well.

### Item 2 — Direct mock assertion in "enabled and already running" test (was: minor)

Confirmed. The test now reads:

```kotlin
verify(context, never()).startForegroundService(any())
```

This is the idiomatic Mockito approach, symmetric with the "disabled" branch test, and not fragile to log-message rewording.

### Item 3 — Unrelated issue-#66 files (was: minor)

Confirmed. The diff touches exactly three files:
- `app/src/main/java/com/gb4pc/service/ServiceChecker.kt` (new)
- `app/src/main/java/com/gb4pc/ui/settings/MainActivity.kt` (modified)
- `app/src/test/java/com/gb4pc/service/ServiceCheckerTest.kt` (new)

No `OverlayManager.kt` or `OverlayManagerRobolectricTest.kt` in scope.

---

All three items from my prior review are resolved. The implementation is correct, well-tested, and properly scoped. **Approved — ready to merge.**

---

PR #74, opened 2026-04-30
Fix #68: start service on app install/update

### LINKED ISSUE #68: Start service upon install
After installing or updating the app on a device, launch the service (if it's enabled).

aunger:
$(cat <<'EOF'
## Summary

- Extends `BootReceiver` to also handle `android.intent.action.MY_PACKAGE_REPLACED`, which fires when the app is updated in-place.
- Adds `BootReceiverLogic.isInstallOrUpdateIntent()` to keep the intent-matching logic testable and DRY.
- Registers the new intent filter in `AndroidManifest.xml`.
- For fresh installs, the user completes setup and enables the service manually; `MY_PACKAGE_REPLACED` only fires on updates, so no extra receiver is needed for the install case.

## Commits

1. **Handle MY_PACKAGE_REPLACED in BootReceiver** — logic changes: new `isInstallOrUpdateIntent()` helper in `BootReceiverLogic`, updated `BootReceiver.onReceive()` to act on boot or install/update, and new unit tests in `BootReceiverTest`.
2. **Register MY_PACKAGE_REPLACED intent filter in AndroidManifest** — manifest-only change, no behaviour change on its own.

## Test plan

- [ ] Existing `BootReceiverTest` tests still pass (not deleted or disabled).
- [ ] New `isInstallOrUpdateIntent returns true for MY_PACKAGE_REPLACED` passes.
- [ ] New `isInstallOrUpdateIntent returns false for other actions` passes.
- [ ] On a device/emulator: enable the service, update the APK via `adb install -r`, confirm the overlay service starts automatically without opening the app.
- [ ] On a device/emulator: disable the service, update the APK, confirm the service does NOT start.

Closes #68

EOF
)

---

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -87,12 +87,15 @@
                 android:value="Monitors camera hardware availability to show gallery shortcut overlay on Pixel Camera" />
         </service>
 
-        <!-- Boot Receiver -->
+        <!-- Boot Receiver (also handles install/update via MY_PACKAGE_REPLACED) -->
         <receiver
             android:name=".receiver.BootReceiver"
             android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.MY_PACKAGE_REPLACED" />
             </intent-filter>
         </receiver>
 
--- a/app/src/main/java/com/gb4pc/receiver/BootReceiver.kt
+++ b/app/src/main/java/com/gb4pc/receiver/BootReceiver.kt
@@ -7,21 +7,25 @@ import com.gb4pc.service.OverlayService
 import com.gb4pc.util.DebugLog
 
 /**
- * Restarts the overlay service after device reboot if it was previously enabled (FS-05).
+ * Starts the overlay service after device reboot (FS-05) or app install/update (FS-68),
+ * if the service was previously enabled.
  * L1: Delegates all decision logic to BootReceiverLogic to avoid duplicating PrefsManager checks.
  */
 class BootReceiver : BroadcastReceiver() {
     override fun onReceive(context: Context, intent: Intent) {
-        if (!BootReceiverLogic.isBootIntent(intent.action)) return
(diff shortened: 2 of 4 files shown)


aunger (commented):
## Review — Ready to merge (APPROVE blocked: same account as PR author)

This is a well-scoped, clean implementation. Here is my full analysis across each dimension.

### Correctness

**Intent choice is right.** `ACTION_MY_PACKAGE_REPLACED` is the correct intent for this job. Unlike `ACTION_PACKAGE_REPLACED` (which fires for any package and requires `exported=true` plus a data-URI package-name check), `MY_PACKAGE_REPLACED` is delivered exclusively to the app being replaced, with no extra filtering required. Good call.

**`exported=true` is not a problem.** The `<receiver>` already carries `android:exported="true"` because Android 12+ requires it to be declared explicitly whenever any `<intent-filter>` is present. `MY_PACKAGE_REPLACED` is a self-targeted broadcast and would work even with `exported=false`, but the existing declaration is driven by the `BOOT_COMPLETED` filter and does not introduce any security exposure for the `MY_PACKAGE_REPLACED` filter (that action is delivered by the system to the replaced package only).

**Fresh-install case is correctly out of scope.** `MY_PACKAGE_REPLACED` does not fire on a first install; users enable the service manually during setup. No additional receiver is needed.

### DRY / Simplicity

The shared `onReceive()` path is clean. The `trigger` local variable neatly avoids repeating three log-message strings. The new `isInstallOrUpdateIntent` predicate is a natural parallel to the existing `isBootIntent`, keeping both intent-matching rules testable in isolation and expressed in exactly one place each.

### Test Coverage

Two new unit tests for `isInstallOrUpdateIntent` mirror the established pattern for `isBootIntent` exactly: one positive case, one negative-and-null case. New behaviour is accompanied by tests; no existing tests are deleted or disabled. Coverage is appropriate.

### Commit Structure

The two-commit split (logic + tests first, manifest second) is consistent with the guideline to split changes by concern. A minor observation: the manifest-only commit does not change observable runtime behaviour on its own (the receiver's dispatch guard also needs the code change), which could be read as a slight ordering quirk. In practice both land together in the PR, so it causes no issue.

### Conclusion

No blocking issues. The implementation is correct, DRY, and well-tested. This is ready to merge.

---

Issue #83, opened 2026-05-01
Move log viewer to its own screen


aunger:
And change it to newest lines on the bottom of the display tongue a, with auto-scrolling to follow the latest. The scrolling pane should take up almost all of the screen. It should also have a "clear log" button.

PR #85, opened 2026-05-01
feat: move log viewer to its own screen (#83)

aunger:
$(cat <<'EOF'
## Summary

- Extracts the inline debug log panel from `AdvancedSettingsActivity` into a new `LogViewerActivity` with its own full screen
- Log entries are now displayed oldest-first (newest at the **bottom**), replacing the previous newest-at-top reversed list
- The `LazyColumn` auto-scrolls to the bottom whenever new entries arrive
- A "Clear log" button is added to the top-bar actions
- `AdvancedSettingsActivity` now shows a "View Debug Log" button that launches the new activity
- `LogViewerActivity` is registered in `AndroidManifest.xml`

## Commits

1. **Refactor** (no behavior change): move the embedded log viewer out of `AdvancedSettingsActivity` into `LogViewerActivity`; add navigation button and manifest entry
2. **Behavior**: switch to newest-at-bottom order with `LazyColumn`, add auto-scroll on new entries, add "Clear log" button; extend `DebugLogTest` with listener and clear tests

## Test plan

- [ ] Build and run on device/emulator: Advanced Settings → "View Debug Log" opens the new screen
- [ ] Log entries appear newest at the bottom; screen auto-scrolls as new entries arrive
- [ ] "Clear log" button clears all entries and shows the empty-state message
- [ ] Back arrow returns to Advanced Settings
- [ ] Unit tests pass: `./gradlew :app:testDebugUnitTest`

EOF
)

---

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -63,6 +63,11 @@
             android:name=".ui.settings.AdvancedSettingsActivity"
             android:exported="false" />
 
+        <!-- Log Viewer Activity (issue #83) -->
+        <activity
+            android:name=".ui.settings.LogViewerActivity"
+            android:exported="false" />
+
         <!-- Gallery App Picker Activity (can be launched over lock screen camera) -->
         <activity
             android:name=".ui.picker.PickerActivity"
--- a/app/src/main/java/com/gb4pc/ui/settings/AdvancedSettingsActivity.kt
+++ b/app/src/main/java/com/gb4pc/ui/settings/AdvancedSettingsActivity.kt
@@ -1,5 +1,6 @@
 package com.gb4pc.ui.settings
 
+import android.content.Intent
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
@@ -7,7 +8,6 @@ import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.text.KeyboardActions
 import androidx.compose.foundation.text.KeyboardOptions
-import androidx.compose.foundation.text.selection.SelectionContainer
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.*
 import androidx.compose.runtime.*
@@ -25,9 +25,6 @@ import com.gb4pc.data.AspectRatioUtil
 import com.gb4pc.data.OverlayPosition
 import com.gb4pc.data.PrefsManager
 import com.gb4pc.ui.theme.GB4PCTheme
-import com.gb4pc.util.DebugLog
-import java.text.SimpleDateFormat
-import java.util.*
 
 /**
  * Advanced Settings screen (§6.3, UI-10).
@@ -99,15 +96,6 @@ fun AdvancedSettingsScreen(prefsManager: PrefsManager) {
         }
     }
 
-    // M2 fix: live-updating debug log entries via DebugLog listener
-    var debugEntries by remember { mutableStateOf(DebugLog.getEntries()) }
-    DisposableEffect(Unit) {
-        DebugLog.listener = { debugEntries = DebugLog.getEntries() }
-        onDispose { DebugLog.listener = null }
-    }
-
-    val dateFormat = remember { SimpleDateFormat("HH:mm:ss.SSS", Locale.US) }
-
     Scaffold(
         topBar = {
             TopAppBar(title = { Text(stringResource(R.string.advanced_title)) })
@@ -118,6 +106,7 @@ fun AdvancedSettingsScreen(prefsManager: PrefsManager) {
                 .padding(padding)
                 .padding(16.dp)
                 .fillMaxWidth()
+                .verticalScroll(rememberScrollState())
         ) {
             // UI-10.1: Position sliders with text inputs (M4 fix: persist only on commit/onValueChangeFinished)
             Row(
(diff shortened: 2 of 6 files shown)


aunger (commented):
## Review of PR #85 — Move log viewer to its own screen

Overall this is well-structured work. Commit discipline is good (refactor commit cleanly separated from behavior commit, as required by `code_edit.md`). The manifest entry, string resources, and navigation wiring are all correct. However there are three issues worth addressing before merging.

---

### Issues requiring changes

#### 1. `Clear log` button does not go through the listener — creates a race condition

In `LogViewerActivity.kt` the clear button does:

```kotlin
onClick = {
    DebugLog.clear()
    entries = DebugLog.getEntries()   // manual snapshot
}
```

`DebugLog.clear()` does **not** call the listener (confirmed in `DebugLog.kt` — `clear()` only calls `buffer.clear()` inside the lock). The manual `getEntries()` immediately after is the workaround, but it creates a race: if a background thread calls `DebugLog.log()` between `clear()` and `getEntries()`, a new entry sneaks into `entries` but `entries.size` changed from N→1, not N→0, so the `LaunchedEffect` auto-scroll may fire unexpectedly, and more critically the empty-state branch never renders even though the user just cleared the log.

The clean fix is to make `DebugLog.clear()` invoke the listener, just like `log()` does:

```kotlin
// DebugLog.kt
fun clear() {
    val snapshot: (() -> Unit)?
    synchronized(lock) {
        buffer.clear()
        snapshot = listener
    }
    snapshot?.invoke()
}
```

Then the button handler simplifies to `onClick = { DebugLog.clear() }`, and the listener handles the UI update uniformly. Note: the test `` `clear does not invoke the listener` `` would need to be updated to assert the opposite (listener *is* invoked), since it currently encodes the wrong behavior as a spec.

---

#### 2. `LaunchedEffect(entries.size)` has a stale-key problem after clear + re-log

If the log has 5 entries, the user clears (size → 0), then 5 new entries arrive, `entries.size` follows the path 5 → 0 → 5. The intermediate 0 triggers the effect (no-op, list is empty). When size returns to 5, Compose sees the **same key** it already ran for (5 == 5) and does **not** re-trigger the effect. Auto-scroll silently stops working until the count crosses a new high-water mark.

Fix: key on the list reference itself rather than its size. Since `getEntries()` always returns a fresh `toList()` copy, every update produces a new object reference:

```kotlin
LaunchedEffect(entries) {
    if (entries.isNotEmpty()) listState.animateScrollToItem(entries.size - 1)
}
```

---

#### 3. Unrelated change bundled into this PR

`Constants.kt` has three default overlay position values changed (`DEFAULT_X_PERCENT` 13→20, `DEFAULT_Y_PERCENT` 91.5→69, `DEFAULT_SIZE_PERCENT` 11.5→16), and `OverlayManagerTest` was updated to match. This has nothing to do with moving the log viewer to its own screen. Per `code_edit.md`, changes should be split by concern — this should either be a separate PR or explained in the description if it was intentional (e.g. as a prerequisite).

---

### Positive notes

- Commit structure: refactor commit cleanly separated from behavior commit. Well done.
- `LogViewerActivity` correctly registered in `AndroidManifest.xml` with `android:exported="false"`.
- `LazyColumn` with `fillMaxSize()` gives the log pane the full screen as the issue requires.
- `animateScrollToItem` is the right API for smooth auto-scroll.
- `SelectionContainer` preserved so log text remains selectable/copyable.
- `DisposableEffect` cleanup (`onDispose { DebugLog.listener = null }`) is correct — no listener leak on back-navigation.
- All string resources extracted; no hardcoded UI strings.
- The three new `DebugLogTest` cases are clean and properly isolated (using `finally` to reset the listener).

---

### Verdict

Two correctness bugs need fixes (the `clear()` listener/race and the `LaunchedEffect` stale-key), and the unrelated `Constants.kt` change should be separated into its own commit or PR. Everything else is solid.


---

aunger (commented):
Follow-up review — all three requested changes are addressed. This PR is approved in substance; GitHub prevented a self-approval on the owner account, but I am registering the outcome here.

---

### Issue 1 — `DebugLog.clear()` now invokes the listener ✓

`DebugLog.kt` captures the listener reference inside the lock before releasing it, then calls it outside:

```kotlin
val snapshot: (() -> Unit)?
synchronized(lock) {
    buffer.clear()
    snapshot = listener
}
snapshot?.invoke()
```

The clear button in `LogViewerActivity.kt` is now simply `onClick = { DebugLog.clear() }` — no manual `getEntries()` workaround, so the race window is gone. The test `` `clear invokes the listener with empty entries` `` asserts `callCount == 1` and that `getEntries()` returns an empty list at the point of invocation. Correct on all counts.

### Issue 2 — `LaunchedEffect` stale-key fixed ✓

Changed to `LaunchedEffect(entries)`. Since `getEntries()` returns a fresh `toList()` copy on every mutation, Compose sees a new object reference on each update and reliably re-triggers the effect — including after a clear-then-re-fill cycle that returns to the same size as before.

### Issue 3 — Unrelated changes removed ✓

The diff contains exactly six files: `AndroidManifest.xml`, `AdvancedSettingsActivity.kt`, `LogViewerActivity.kt` (new), `DebugLog.kt`, `strings.xml`, `DebugLogTest.kt`. `Constants.kt` and `OverlayManagerTest.kt` are absent.

---

All three correctness concerns are resolved and no new issues were introduced. The implementation is solid. **Approved.**

---

PR #87, opened 2026-05-03
Fix #80: skip own package in ForegroundDetector to prevent overlay hiding itself

### LINKED ISSUE #80: FG detector sees GB4PC as the foreground app
In v0.0.7:

Foreground Detector sees GB4PC as the FG app once the overlay is visible. Therefore, changing camera modes (going into photo-sphere mode, for example) makes the overlay disappear immediately.

I'm unsure whether this works when the experimental responsiveness is off.

aunger:
$(cat <<'EOF'
## Summary

- **Root cause**: When the overlay is visible, Android reports GB4PC as a foreground app via `UsageStatsManager` `MOVE_TO_FOREGROUND` events. `ForegroundDetector` then thinks the camera is no longer the foreground app and hides the overlay — reproducing the mode-switch disappearance described in #80.
- **Fix**: Add a `selfPackage` constructor parameter to `ForegroundDetector`. Any `MOVE_TO_FOREGROUND` event whose `packageName` matches `selfPackage` is skipped. The camera (or any app the user genuinely navigates to) remains the detected foreground app. `OverlayService` passes its own `packageName` on construction.
- **Tests**: New `ForegroundDetectorSelfFilterTest` (Robolectric) covers: self-only events return null, self event does not displace an earlier camera event, genuine navigation away is still detected, multiple self-events are all skipped, and non-foreground self-events do not interfere.

## Commits

1. `Fix #80` — production code: `selfPackage` parameter + skip logic in `ForegroundDetector`, wire-up in `OverlayService`, and update existing test constructors.
2. `Test #80` — new `ForegroundDetectorSelfFilterTest` with 7 Robolectric test cases covering the filter and its edge cases.

## Test plan

- [ ] All unit tests pass: `./gradlew :app:testDebugUnitTest`
- [ ] Open Pixel Camera → overlay appears; switch to photo-sphere mode → overlay stays visible
- [ ] Navigate away from camera to another app → overlay disappears (genuine navigation still works)

EOF
)

---

### DIFF
--- a/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
+++ b/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
@@ -7,12 +7,21 @@ import com.gb4pc.util.DebugLog
 
 /**
  * Detects the current foreground app using UsageStatsManager (DT-02, DT-06).
+ *
+ * @param selfPackage The package name of this app (com.gb4pc). MOVE_TO_FOREGROUND events for
+ *   this package are ignored because the overlay window causes Android to report GB4PC itself
+ *   as the foreground app — which would displace the camera from the detected foreground app
+ *   and hide the overlay (Issue #80).
  */
-class ForegroundDetector(private val usageStatsManager: UsageStatsManager) {
+class ForegroundDetector(
+    private val usageStatsManager: UsageStatsManager,
+    private val selfPackage: String,
+) {
 
     /**
      * Queries UsageStatsManager for the most recent MOVE_TO_FOREGROUND event
      * in the last [Constants.USAGE_STATS_WINDOW_MS] milliseconds.
+     * Events for [selfPackage] are skipped (Issue #80).
      * Returns the package name, or null if no event found (EC-09).
      */
     fun getForegroundPackage(): String? {
@@ -31,11 +40,17 @@ class ForegroundDetector(private val usageStatsManager: UsageStatsManager) {
         var latestTimestamp = 0L
         var totalEvents = 0
         var foregroundEvents = 0
+        var skippedSelfEvents = 0
 
         while (events.hasNextEvent()) {
             events.getNextEvent(event)
             totalEvents++
             if (event.eventType == UsageEvents.Event.MOVE_TO_FOREGROUND) {
+                if (event.packageName == selfPackage) {
+                    skippedSelfEvents++
+                    DebugLog.log("ForegroundDetector: skipping self MOVE_TO_FOREGROUND pkg=${event.packageName} ts=${event.timeStamp} (Issue #80)")
+                    continue
+                }
                 foregroundEvents++
                 DebugLog.log("ForegroundDetector: MOVE_TO_FOREGROUND pkg=${event.packageName} ts=${event.timeStamp}")
                 if (event.timeStamp >= latestTimestamp) {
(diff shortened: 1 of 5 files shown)


aunger (commented):
## Review of PR #87

The core fix in `ForegroundDetector` is correct and the approach is the right one. Skipping `MOVE_TO_FOREGROUND` events where `packageName == selfPackage` is the minimal, targeted change needed to resolve #80, and wiring it up via `OverlayService.packageName` is clean. The test suite is thorough on the scenarios that matter.

I have two issues that need to be addressed before this merges.

---

### Issue 1 (must fix): `foregroundEvents` counter is misleading after the skip

In the updated loop, `foregroundEvents++` is incremented *before* the self-package check, so skipped self-events are counted as foreground events in the summary log line:

```kotlin
if (event.eventType == UsageEvents.Event.MOVE_TO_FOREGROUND) {
    foregroundEvents++                        // ← counts self-events
    if (event.packageName == selfPackage) {
        DebugLog.log("… skipping …")
        continue
    }
    …
}
```

The summary that gets logged is:
```
ForegroundDetector: foreground=com.google.android.GoogleCamera (3 foreground event(s) of 5 total)
```

If GB4PC fired 2 self-events and the camera fired 1, this will say "3 foreground event(s)" when in fact only 1 non-self foreground event was considered. That summary is used for debugging, so it should accurately reflect the number of *non-skipped* foreground events, not the raw total. Either move `foregroundEvents++` to after the self-package guard, or add a separate `skippedEvents` counter and include it in the log line.

---

### Issue 2 (must fix): scope creep — this PR bundles an unrelated feature (#83)

The PR title and description say "Fix #80", but the diff contains a significant amount of work unrelated to the foreground-detector self-filter:

- `LogViewerActivity.kt` (117 lines, new file)
- Refactoring of `AdvancedSettingsActivity` to replace the inline log card with a button
- `DebugLog.clear()` fix to invoke the listener
- Three new `DebugLogTest` test cases
- `AndroidManifest.xml` entry for `LogViewerActivity`
- New string resources for the log viewer

These changes are attributed to issue #83 in the code comments. `code_edit.md` requires that changes be split into commits by concern — and more importantly, mixing two issues into one PR makes both harder to review, revert, and bisect. The #83 work should travel in its own PR. Please extract those changes.

---

### Minor observations (no action required)

- The `@Suppress("DEPRECATION")` annotation on `ForegroundDetectorSelfFilterTest` with the explanatory comment is good practice.
- Using `LazyColumn` in `LogViewerActivity` (instead of the `Column`+`verticalScroll` workaround previously used in the settings screen) is the right call for a dedicated full-screen log view.
- `DebugLog.clear()` not invoking the listener was a real bug; the fix is correct.

---

aunger:
Both issues from the previous review are resolved. **This PR is approved** (GitHub blocked the formal approval action because author and reviewer share the same account on this repo).

### Issue 1 — Counter bug (fixed)

The increment order is now correct. In the updated loop the self-package guard runs first: a matching event increments `skippedSelfEvents` and `continue`s, so `foregroundEvents++` is only reached for non-skipped events. The summary log appends `, skipped N self-event(s)` when `skippedSelfEvents > 0`. The log line will be accurate for debugging.

### Issue 2 — Scope creep (resolved)

The diff now touches exactly the five files that belong to the #80 fix: `ForegroundDetector.kt`, `OverlayService.kt`, `ForegroundDetectorEventTest.kt`, `ForegroundDetectorTest.kt`, and the new `ForegroundDetectorSelfFilterTest.kt`. The #83 material (`LogViewerActivity`, `AdvancedSettingsActivity` refactor, manifest, strings, `DebugLogTest` additions) is gone.

### Test coverage

`ForegroundDetectorSelfFilterTest` covers the cases that matter: self-only events return null, a self-event does not displace an earlier camera event, genuine navigation away is still detected, multiple self-events are all skipped, and non-foreground self-events for `selfPkg` do not interfere. That is sufficient.

---

PR #96, opened 2026-05-04
Skip deactivation checklist in onCameraAvailable when overlay is inactive

### LINKED ISSUE #89: If overlay inactive, don't do so much deactivating
When receiving a _camera-now-available_ event, if the overlay isn't currently active, then don't run through the should-it-deactivate checklist. (Because, no, it obviously shouldn't.) 

This event is received when Pixel Camera is nowhere in sight: during initial startup the listener is primed with the current status, and whenever any other process using the camera releases it.

aunger:
Fixes #89

## What changed

In `onCameraAvailable()`, added an early-exit guard: if the overlay is not currently active, cancel any pending activation retry and return immediately — without running the deactivation checklist.

## Why

The camera-now-available event fires in two situations where the overlay was never shown:
- **Initial startup**: the `CameraManager` listener is primed with the current camera status, which is typically "all cameras available".
- **Another app releases the camera**: any process that held the camera (e.g. a QR scanner) releases it, triggering the event even though Pixel Camera was never in the foreground.

In both cases, `isOverlayActive` is `false`, so there is nothing to deactivate. The old code would still proceed into the `allAvailable` branch and potentially schedule a deactivation runnable (a no-op at runtime, but unnecessary work and an incorrect code path to walk).

## Tests added

Three new unit tests in `OverlayServiceLogicTest` under the `Issue #89` section:

1. `onCameraAvailable does not schedule deactivation when overlay is inactive` — initial-startup priming case.
2. `onCameraAvailable when all cameras released and overlay inactive skips deactivation` — all cameras released but overlay was never shown.
3. `onCameraAvailable when overlay inactive cancels pending activation retry` — camera released before the UsageStats retry fired, overlay still inactive; verifies the retry runnable is removed and no deactivation is scheduled.

---

### DIFF
--- a/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
@@ -57,7 +57,12 @@ class OverlayServiceLogic(
     fun onCameraAvailable(cameraId: String) {
         cameraState.setCameraAvailable(cameraId)
         val allAvailable = cameraState.areAllCamerasAvailable()
-        DebugLog.log("Logic: camera $cameraId available; allAvailable=$allAvailable, remaining=${cameraState.getUnavailableCameraIds()}")
+        DebugLog.log("Logic: camera $cameraId available; allAvailable=$allAvailable, remaining=${cameraState.getUnavailableCameraIds()}, overlayActive=$isOverlayActive")
+        // Issue #89: If the overlay is not active there is nothing to deactivate.
+        if (!isOverlayActive) {
+            cancelActivationRetry()
+            return
+        }
         // DT-04/DT-05: Only schedule deactivation when ALL cameras have been released
         if (allAvailable) {
             cancelActivationRetry()
--- a/app/src/test/java/com/gb4pc/service/OverlayServiceLogicTest.kt
+++ b/app/src/test/java/com/gb4pc/service/OverlayServiceLogicTest.kt
@@ -398,6 +398,8 @@ class OverlayServiceLogicTest {
     /**
      * If the camera is released while a retry is pending (e.g. a non-Pixel-Camera app
      * briefly opened the camera), the retry should be cancelled so it doesn't fire stale.
+     * Issue #89: the early-exit guard also ensures no deactivation is scheduled in this case
+     * (overlay was never active, so there is nothing to deactivate).
      */
     @Test
     fun `DT-06a pending retry cancelled when camera released before retry fires`() {
@@ -407,10 +409,14 @@ class OverlayServiceLogicTest {
         val runnableCaptor = argumentCaptor<Runnable>()
         verify(handler).postDelayed(runnableCaptor.capture(), eq(Constants.ACTIVATION_RETRY_MS))
 
-        // Camera released — all cameras now available
+        // Camera released before activation completed — overlay still inactive
+        assertFalse("Pre-condition: overlay should not be active", logic.isOverlayActive)
         logic.onCameraAvailable("0")
 
         verify(handler).removeCallbacks(runnableCaptor.firstValue)
+        // Issue #89: no deactivation should be scheduled — overlay was never active
+        verify(handler, never()).postDelayed(any(), eq(0L))
+        verify(handler, never()).postDelayed(any(), eq(Constants.CAMERA_DEBOUNCE_MS))
     }
 
     // ── Issue #89: skip deactivation when overlay is already inactive ────────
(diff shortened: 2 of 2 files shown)


aunger (commented):
The core logic change is correct and clearly solves the issue. The early-exit guard in `onCameraAvailable` is well-placed, appropriately calls `cancelActivationRetry()` before returning, and the updated log line now includes `overlayActive` which is a genuine improvement. Tests 1 and 2 are clean and meaningful.

I have two concerns that I believe warrant changes before merging.

---

**1. Unrelated log removals in `ForegroundDetector.kt` (and `evaluateForeground`) should not be in this PR**

The diff removes four `DebugLog.log` calls:

- In `getForegroundPackage()`: the window/time query log, the self-event skip log, and the MOVE_TO_FOREGROUND event log.
- In `evaluateForeground()`: the log that previously reported `foreground=`, `isPixelCamera=`, `overlayActive=`, and `anyCameraUnavailable=` — the revised line only reports the last two.

None of these removals are related to issue #89. They reduce the diagnostic information available when debugging production problems. In particular, the `evaluateForeground` log previously showed `foreground=$pkg` and `isPixelCamera=$isPixelCamera`, which are the two values that determine whether the overlay is activated or a retry is scheduled. That information is now gone from the log output.

These changes should be reverted from this PR. If log cleanup is genuinely desired, it should be tracked as a separate issue with its own rationale.

---

**2. Test 3 largely duplicates an existing test**

`issue-89 onCameraAvailable when overlay inactive cancels pending activation retry` (new) and `DT-06a pending retry cancelled when camera released before retry fires` (existing, lines 402–414) exercise the same path through `onCameraAvailable` under the same precondition (overlay inactive, retry pending). The existing test already verifies `handler.removeCallbacks` is called. The new test adds `verify(handler, never()).postDelayed(any(), eq(0L))` and `verify(handler, never()).postDelayed(any(), eq(Constants.CAMERA_DEBOUNCE_MS))` — but since the early-exit guard now returns before reaching the `allAvailable` branch, those negatives are consequences of the guard, not independent facts about retry cancellation. The test does not add meaningful new coverage of a distinct scenario.

If you want to keep a test that specifically asserts the deactivation scheduling is skipped when the overlay is inactive, combining that assertion into the existing DT-06a test would be cleaner than a duplicate setup. Alternatively, a brief comment explaining what this test covers that the DT-06a test does not would help justify its existence.

---

aunger (commented):
Thanks for the review. Both points are addressed in the updated push.

---

**1. Log removals**

Agreed — those changes were out of scope. The root cause was that this branch was originally built on top of the `fix/issue-90-remove-trace-logging` branch commits, so the ForegroundDetector log removals and the `evaluateForeground` log trimming appeared in the PR diff even though the issue-89 commit itself never touched those lines. The branch has been rebased directly onto main, and the updated diff now contains only changes to `OverlayServiceLogic.kt` and the test file. `ForegroundDetector.kt` is no longer in the diff at all.

---

**2. Duplicate test**

Agreed. The duplicate test has been removed and its assertions folded into DT-06a: the precondition comment is updated, a `assertFalse` pre-condition check is added, and the two `never().postDelayed` assertions (for `0L` and `CAMERA_DEBOUNCE_MS`) are appended with an `Issue #89` comment explaining their purpose. This keeps the meaningful coverage without the repeated setup.

---

---

aunger (commented):
Second-round review — both issues raised previously are resolved. Verdict: **APPROVE** (posted as COMMENT because the GitHub identity used here is the PR author).

**Log-removal scope issue: resolved.** `ForegroundDetector.kt` has no diff against `main` in this branch. The two commits on this branch (`a67fecb`, `38868a0`) touch only `OverlayServiceLogic.kt` and its test file, which is exactly the right scope.

**Test consolidation: correct.** The standalone issue-89 test that previously duplicated DT-06a has been removed. Its two distinctive assertions — the `assertFalse` pre-condition check and the two `never().postDelayed` deactivation-guard assertions — are now folded into `DT-06a pending retry cancelled when camera released before retry fires`, and the updated doc comment makes the dual purpose of that test clear.

**Production logic: sound.** The early-exit guard in `onCameraAvailable` is placed before the `allAvailable` branch, so it cannot accidentally suppress deactivation when the overlay *is* active. `cancelActivationRetry()` is correctly called inside the guard to clean up any pending retry before returning.

**New standalone tests: appropriate.** The two additional `issue-89` tests cover the two remaining observable cases (cold start with no prior `onCameraUnavailable`, and the `allAvailable=true` path with an inactive overlay) and add no redundancy with DT-06a's scenario. The `verify(overlayManager, never()).hide()` assertion in the second test gives an extra safety net that the only side-effect of the new code path is the log statement.

No problems found.

---

PR #98, opened 2026-05-04
Fix lock-screen photo detection: UsageStats bypass, all-items query, always retry

### LINKED ISSUE #81: On lock screen, no new photos detected

aunger:
## Summary

Fixes #81 — photos taken from the lock screen are not detected and the gallery button does not update.

Three separate bugs combined to produce the symptom:

- **Overlay never activates on lock screen** (`OverlayServiceLogic`): `evaluateForeground()` relied entirely on `UsageStatsManager` to identify the foreground app. `UsageStatsManager` does not report `MOVE_TO_FOREGROUND` events while the device is locked, so the overlay was never shown when the camera was opened via the lock-screen shortcut. Fix: when `isKeyguardLocked() && cameraState.anyCameraUnavailable()`, activate the overlay directly without a UsageStats lookup. The lock-screen camera shortcut can only launch Pixel Camera, so this bypass is safe and correct.

- **Only the first committed photo is detected per burst** (`MediaChangeDispatcher`): the dispatcher previously scheduled a retry only when `queryLatestMedia` returned `null`. When Photo N's `IS_PENDING=1` callback fired, the query returned Photo N-1 (already committed, non-null), so no retry was scheduled. If the `IS_PENDING=0` notification for Photo N was suppressed (common on locked devices), Photo N was silently lost. Fix: switch from `queryLatestMedia` (single item) to `queryAllMedia` (all items since session start), add every returned item to the session immediately, and _always_ schedule a cancel-and-reschedule retry — so pending photos are captured once committed regardless of whether the `IS_PENDING=0` callback fires. Deduplication in `SessionTracker.addMedia()` prevents double-adds.

- **Thumbnail does not update to the latest photo** (`ThumbnailChangeDispatcher`): same conditional-retry bug. When the new photo was still `IS_PENDING`, the query returned the previous photo's URI and no retry was scheduled, leaving the thumbnail stale. Fix: always schedule a cancel-and-reschedule retry, identical pattern to `MediaChangeDispatcher`.

## Commits

1. `Fix #81: bypass UsageStats on lock screen to activate overlay` — `OverlayServiceLogic` + 6 new unit tests
2. `Fix MediaChangeDispatcher: query all photos and always retry` — `MediaChangeDispatcher` + rewritten unit tests
3. `Fix ThumbnailChangeDispatcher: always retry with cancel-and-reschedule` — `ThumbnailChangeDispatcher` + rewritten unit tests

## Test plan

- [ ] All existing unit tests pass (`./gradlew test`)
- [ ] New lock-screen bypass tests in `OverlayServiceLogicTest` cover: activation, session start, idempotency, no false activation when unlocked, no false activation when no camera in use, clean deactivation
- [ ] `MediaChangeDispatcherTest` covers: immediate multi-item add, always-retry (even when items found immediately), retry on empty, retry is one-shot, multi-photo capture via retry, cancel-and-reschedule, `sessionStartMs` forwarding
- [ ] `ThumbnailChangeDispatcherTest` covers: immediate thumbnail show, always-retry (even when item found immediately), retry on null, retry is one-shot, cancel-and-reschedule, `startMs` forwarding
- [ ] Manual: open camera from lock screen, take one or more photos, unlock device — verify gallery button is visible with correct thumbnail and session media count

---

### DIFF
--- a/app/src/main/java/com/gb4pc/service/MediaChangeDispatcher.kt
+++ b/app/src/main/java/com/gb4pc/service/MediaChangeDispatcher.kt
@@ -18,47 +18,67 @@ import com.gb4pc.viewer.SessionTracker
  * Pixel Camera (and other modern camera apps on API 29+) inserts new photos into
  * MediaStore with IS_PENDING = 1 while the file is being written, then clears the
  * flag once the write completes. The ContentObserver fires on both state transitions.
- * On the first fire (IS_PENDING = 1) the default query excludes the pending row and
- * [queryLatestMedia] returns null. On the second fire (IS_PENDING cleared) the
- * committed row is found and added to the session.
+ * On the first fire (IS_PENDING = 1) the default query excludes the pending row, so
+ * any item returned by [queryAllMedia] is a previously committed photo.  On the second
+ * fire (IS_PENDING = 0) the newly committed row is found and added to the session.
  *
  * However, on some devices/firmware the second ContentObserver callback for the
- * IS_PENDING→0 transition is not delivered reliably. [onMediaChanged] therefore
- * schedules a single retry after [Constants.MEDIA_OBSERVER_RETRY_MS] when the
- * initial query returns null, picking up committed rows that the first callback
- * missed.
+ * IS_PENDING→0 transition is not delivered reliably (particularly on locked devices).
+ * [onMediaChanged] therefore always schedules a single retry after [retryDelayMs]
+ * (cancel-and-reschedule on rapid-fire shots) that queries for all newly committed
+ * items — picking up photos whose IS_PENDING=0 notification was missed.
+ *
+ * ## Multiple photos
+ *
+ * When multiple photos are taken in quick succession, only the most recent IS_PENDING=1
+ * callback would have found a null result (no committed item) with the old single-retry
+ * approach.  Each subsequent IS_PENDING=1 callback found the *previous* committed photo
+ * (non-null), so no retry was scheduled for the new photo.  By querying *all* items
+ * since [sessionStartMs] on every callback and on the retry, every committed photo is
+ * captured regardless of whether its IS_PENDING=0 notification fires.  Deduplication
+ * inside [SessionTracker.addMedia] prevents double-adds.
  */
 class MediaChangeDispatcher(
     private val sessionTracker: SessionTracker,
     private val handler: Handler,
-    /** Returns the most recent [MediaItem] added after [sessionStartMs], or null. */
-    private val queryLatestMedia: (sessionStartMs: Long) -> MediaItem?,
+    /** Returns all [MediaItem]s with DATE_ADDED after [sessionStartMs], newest-first. */
+    private val queryAllMedia: (sessionStartMs: Long) -> List<MediaItem>,
     private val retryDelayMs: Long = Constants.MEDIA_OBSERVER_RETRY_MS,
 ) {
+    // At most one pending retry runnable at a time; cancelled and rescheduled on each onChange.
+    private var pendingRetry: Runnable? = null
 
     /**
      * Called whenever the media ContentObserver detects a change.
      *
-     * Immediately queries for the latest committed media added since [sessionStartMs].
-     * If the query returns null (item is still IS_PENDING), schedules a single retry
-     * after [retryDelayMs] ms.
+     * Immediately adds all committed media items found since [sessionStartMs] to the session.
+     * Then cancels any in-flight retry and schedules a fresh one after [retryDelayMs] ms so
+     * that items still in IS_PENDING state at the time of this call are captured once they
+     * are committed.  Rapid-fire photos produce at most one pending retry at any moment.
      */
     fun onMediaChanged(sessionStartMs: Long) {
-        val item = queryLatestMedia(sessionStartMs)
-        if (item != null) {
+        val items = queryAllMedia(sessionStartMs)
+        items.forEach { item ->
             sessionTracker.addMedia(item)
             DebugLog.log("Media added to session: ${item.uri}")
-        } else {
-            // First onChange may fire while the photo is IS_PENDING; schedule a retry so we
-            // catch it once it has been committed to MediaStore.
-            DebugLog.log("Media query returned null — scheduling retry in ${retryDelayMs}ms")
-            handler.postDelayed({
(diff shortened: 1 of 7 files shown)


aunger (commented):
Three separate bugs identified, and each fix is conceptually sound. The tests are well-structured and the scope is tightly limited to issue #81. I have one correctness problem that must be fixed before merge, and one gap in test coverage that should be addressed.

---

## Required fix: `onCameraAvailable` calls `getForegroundPackage()` on the lock screen

After the lock-screen bypass activates the overlay, when the user finishes taking photos and the camera is released, `onCameraAvailable` is called. It unconditionally calls `foregroundDetector.getForegroundPackage()` to decide the deactivation delay:

```kotlin
// OverlayServiceLogic.kt, lines 67-68
val pkg = foregroundDetector.getForegroundPackage()
val delay = if (ForegroundDetector.isPixelCameraPackage(pkg)) debounceMs else 0L
```

On a locked device, `UsageStatsManager` returns null for the foreground package — that is the exact bug being fixed in commit 1. So `pkg` will be null, `isPixelCameraPackage` returns false, `delay` is 0 ms, and deactivation fires immediately with no debounce. If the user is still in Pixel Camera (e.g. briefly lowered the camera between shots without closing the app), the overlay will be hidden prematurely for the same reason it was never shown — UsageStats is blind to the lock screen.

The issue-46 debounce logic was written with an unlocked screen in mind. On the lock screen you cannot determine whether Pixel Camera is "still foreground" via UsageStats, so the safe default when `isKeyguardLocked()` is true should be to apply `debounceMs` (the conservative path), not 0 ms (the aggressive path).

The existing test `issue-81 deactivation after lock-screen activation ends session and unregisters observer` passes `any()` for the delay, so it does not catch this. A test asserting `delay == debounceMs` when `isKeyguardLocked() == true` would pin the correct behaviour.

---

## Test gap: idempotency test does not assert that `getForegroundPackage()` is still skipped on the second call

The test at line 607 (`issue-81 lock-screen bypass is idempotent when overlay is already active`) correctly verifies that `show()` is called exactly once. However, on the second `evaluateForeground()` call, `isOverlayActive` is already true, so the bypass condition `&& !isOverlayActive` is false and the code falls through to the normal UsageStats path. On a real locked device the normal path calls `getForegroundPackage()` (returning null) and then does nothing further because `!isOverlayActive` is false — so there is no observable misbehaviour. But the test does not assert that `getForegroundPackage()` is never called, meaning the KDoc's promise ("bypass the UsageStats check") is only half-tested for the idempotent case. Adding `verify(foregroundDetector, never()).getForegroundPackage()` to that test would fully document and enforce the intended contract.

---

## What is working well

The three-bug diagnosis is accurate and the fixes address each root cause directly. The cancel-and-reschedule pattern in both dispatchers is cleaner than the old conditional single-retry, and the `queryAllMedia` change correctly resolves the burst-photo loss. The lock-screen bypass condition `isKeyguardLocked() && cameraState.anyCameraUnavailable() && !isOverlayActive` is tight and safe. The six new `OverlayServiceLogic` tests and the rewritten dispatcher tests cover the documented contracts well.

The `onCameraAvailable` deactivation delay issue is the primary concern — please address that before merge.

---

aunger:
Thanks for the thorough review. Both points are addressed. Here's what changed:

**Finding 1 — deactivation delay on lock screen (fixed in prior commit, tested now)**

The `onCameraAvailable()` deactivation path already has the fix (`7e4fba7`): when `isKeyguardLocked()` is true it takes the `debounceMs` branch unconditionally, bypassing the `getForegroundPackage()` call entirely. A new test `issue-81 deactivation uses debounceMs on lock screen, not 0ms` now covers this specifically — it uses a non-zero `debounceMs = 50L` so the two delay values are distinguishable, activates via the lock-screen bypass, releases the camera, and asserts both `eq(debounce)` on the `postDelayed` call and `never().getForegroundPackage()` on the deactivation path.

**Finding 2 — idempotency test and the `never()` assertion**

I added the `verify(foregroundDetector, never()).getForegroundPackage()` assertion to the idempotency test. The reviewer is correct that without it the test doesn't enforce the "bypass UsageStats" contract — the second `evaluateForeground()` call, with `isOverlayActive = true`, hits the `isKeyguardLocked() && anyCameraUnavailable()` branch and returns early before the `getForegroundPackage()` call; the assertion confirms that early-return path holds. Since this is the whole point of the test (proving UsageStats is never consulted on a locked device), the `never()` assertion belongs there and the test is stronger for it.

---

aunger (commented):
Both items from the previous round are correctly resolved. Details below.

## Deactivation delay fix (`OverlayServiceLogic.kt` lines 72–78)

The fix is correct. When `isKeyguardLocked()` is true the branch short-circuits to `debounceMs` before `getForegroundPackage()` is ever reached, exactly mirroring the activation bypass. One minor observation: `isKeyguardLocked()` is called a second time on line 78 inside the log string — redundant since the result was already captured in `delay`. Benign (the lambda is a cheap boolean read), not blocking.

## New deactivation test (`issue-81 deactivation uses debounceMs on lock screen, not 0ms`)

Well-constructed and genuinely catches the pre-fix regression:

- Uses a dedicated `logicWithDebounce` instance with `debounceMs = 50L` so that 0 and 50 are distinguishable.
- Activates via `onCameraUnavailable` (which calls `evaluateForeground()` internally) rather than a direct field set — the activation path is exercised under the same locked condition, making the test end-to-end.
- `verify(handler).postDelayed(any(), eq(debounce))` directly checks the scheduled delay value.
- `verify(foregroundDetector, never()).getForegroundPackage()` covers both the activation and deactivation calls in a single assertion — correct, because neither path should reach `getForegroundPackage()` when locked.

## Idempotency test `never()` assertion

The added `verify(foregroundDetector, never()).getForegroundPackage()` is the right assertion. The second `evaluateForeground()` call enters the `isKeyguardLocked() && anyCameraUnavailable()` branch, sees `isOverlayActive = true`, logs and returns — `getForegroundPackage()` is unreachable from either call. The assertion enforces the bypass contract rather than merely observing a symptom.

## Other changes in the diff

The `MediaChangeDispatcher` / `ThumbnailChangeDispatcher` always-retry and cancel-and-reschedule refactor is coherent and well-tested. The new tests correctly cover the cancel-and-reschedule invariant with `removeCallbacks` verification.

No remaining issues found. This is satisfactory — recommend approval by an independent reviewer.

---

PR #99, opened 2026-05-04
Hide overlay early when gallery launch moves Pixel Camera out of foreground

### LINKED ISSUE #91: Hide overlay when launching gallery
When the button is used to launch the gallery, test the FG apps for PC. If PC is still in foreground, wait for the "debounce" delay time that's already in settings, and test a second time. 

If PC is not in the FG for either test, then immediately hide the overlay. 

This is based on the assumption that launching the gallery app closes PC. The camera-available event has high latency, and we can react before it arrives. However, we still test the FG app, because the camera and gallery can presumably run concurrently in unusual circumstances (on split screen or dual screen).

aunger:
Fixes #91

## What changed

When the overlay button launches the gallery app, the overlay now reacts proactively instead of waiting for the high-latency camera-available event:

1. **Immediate hide** — immediately after `startActivity` for the gallery, check whether Pixel Camera is still the foreground app. If it's gone (normal case: launching the gallery closes PC), hide the overlay at once.
2. **Deferred re-check** — if PC is still in the foreground at the moment of launch (e.g. split-screen or dual-screen), schedule a one-shot re-check after `debounceMs`; hide the overlay then if PC has left by that point. If PC is still there, leave the overlay visible.

The pending re-check runnable is cancelled in `reset()` so it cannot fire stale after the service is destroyed.

## Files changed

- `OverlayManager.kt` — adds `onGalleryLaunched: () -> Unit` callback, called after `startActivity` in `launchGalleryApp`.
- `OverlayService.kt` — wires `onGalleryLaunched = { logic.onGalleryLaunched() }`.
- `OverlayServiceLogic.kt` — implements `onGalleryLaunched()`, `scheduleGalleryLaunchRecheck()`, and `cancelGalleryLaunchRecheck()` (called from `reset()`).
- `OverlayServiceLogicTest.kt` — six new unit tests covering all branches of the new behaviour.

## Test plan

- [x] `./gradlew test` passes (all 96 tasks, 28 executed on forced re-run)
- [ ] On device: tap gallery button → overlay hides immediately when gallery opens
- [ ] On device: tap gallery button in split-screen with PC still visible → overlay hides after debounce delay
- [ ] On device: `reset()` while re-check pending → no stale hide fires

---

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -16,12 +16,14 @@ The Orchestrator is not a Reviewer or a Programmer.
 - Diagnose bugs or evaluate code
 - Make git commits or push changes
 - Create PRs
-- Apply fixes when an agent leaves work incomplete — dispatch another agent instead
+- Apply fixes when an agent leaves work incomplete
 
 **May:**
 - Read issues, PRs, and the comments on either via GitHub MCP tools
 - Read project instructions (AGENTS.md and the files it references)
 - Dispatch and communicate with subagents
+    - Replace subagents when necessary to complete a workflow
+    - Inform them of unfinished tasks or additional responsibilities
 - Relay subagent results to the user
 
 ## Inaugurating work for a hitherto unworked issue
--- a/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
+++ b/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
@@ -40,6 +40,12 @@ class OverlayManager(
      * Only fired when [PrefsManager.focusableOverlay] is true.
      */
     private val onFocusGained: () -> Unit = {},
+    /**
+     * Called immediately after the gallery app is launched via the overlay button (Issue #91).
+     * Allows the service to hide the overlay early when Pixel Camera is no longer in the
+     * foreground, without waiting for the camera-available event.
+     */
+    private val onGalleryLaunched: () -> Unit = {},
 ) {
     private val windowManager = context.getSystemService(Context.WINDOW_SERVICE) as WindowManager
     private val keyguardManager = context.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
@@ -268,6 +274,7 @@ class OverlayManager(
             intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
             context.startActivity(intent)
             DebugLog.log("Launched gallery app: $packageName")
+            onGalleryLaunched()
         }
     }
 
--- a/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
+++ b/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
@@ -27,7 +27,6 @@ class ForegroundDetector(
     fun getForegroundPackage(): String? {
         val endTime = System.currentTimeMillis()
         val beginTime = endTime - Constants.USAGE_STATS_WINDOW_MS
-        DebugLog.log("ForegroundDetector: querying window=${Constants.USAGE_STATS_WINDOW_MS}ms ending at $endTime")
 
         val events = usageStatsManager.queryEvents(beginTime, endTime)
         if (events == null) {
@@ -48,11 +47,9 @@ class ForegroundDetector(
             if (event.eventType == UsageEvents.Event.MOVE_TO_FOREGROUND) {
                 if (event.packageName == selfPackage) {
                     skippedSelfEvents++
-                    DebugLog.log("ForegroundDetector: skipping self MOVE_TO_FOREGROUND pkg=${event.packageName} ts=${event.timeStamp} (Issue #80)")
                     continue
                 }
                 foregroundEvents++
-                DebugLog.log("ForegroundDetector: MOVE_TO_FOREGROUND pkg=${event.packageName} ts=${event.timeStamp}")
                 if (event.timeStamp >= latestTimestamp) {
                     latestTimestamp = event.timeStamp
                     latestForegroundPackage = event.packageName
@@ -71,9 +68,7 @@ class ForegroundDetector(
 
     companion object {
         fun isPixelCameraPackage(packageName: String?): Boolean {
-            val result = packageName == Constants.PIXEL_CAMERA_PACKAGE
-            DebugLog.log("ForegroundDetector: isPixelCamera($packageName) = $result")
-            return result
+            return packageName == Constants.PIXEL_CAMERA_PACKAGE
         }
     }
 }
--- a/app/src/main/java/com/gb4pc/service/MediaChangeDispatcher.kt
+++ b/app/src/main/java/com/gb4pc/service/MediaChangeDispatcher.kt
@@ -18,47 +18,67 @@ import com.gb4pc.viewer.SessionTracker
  * Pixel Camera (and other modern camera apps on API 29+) inserts new photos into
  * MediaStore with IS_PENDING = 1 while the file is being written, then clears the
  * flag once the write completes. The ContentObserver fires on both state transitions.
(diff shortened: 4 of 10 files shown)


aunger (commented):
The core logic for issue #91 is correctly structured and covers the spec: immediate hide when PC leaves the foreground, deferred re-check when PC is still present (split-screen), and no-op when the overlay is inactive. The reset() cancel is correct. Tests cover all four branches of the spec.

However there are two bugs in the new hide paths that need to be fixed before merging.

---

## Bug 1 (blocking): Session and thumbnail observer teardown missing on both new hide paths

`onGalleryLaunched()` (immediate hide) and `scheduleGalleryLaunchRecheck()` (deferred hide) both reduce to:

```kotlin
overlayManager.hide()
isOverlayActive = false
onOverlayStateChanged(false)
```

Compare with the normal `scheduleDeactivation` runnable, which additionally does:

```kotlin
onUnregisterThumbnailObserver()
if (sessionTracker.isSessionActive) {
    sessionTracker.endSession()
    onUnregisterMediaObserver()
}
```

Neither new hide path calls `onUnregisterThumbnailObserver()`, `sessionTracker.endSession()`, or `onUnregisterMediaObserver()`. The thumbnail ContentObserver continues firing after the overlay is hidden. If a secure session is active (the device was locked), that session remains open while the gallery is already visible — a security concern for an app whose purpose is to protect photos taken on the lock screen.

The camera-available event will eventually arrive and `scheduleDeactivation`'s runnable will fire, but by then `isOverlayActive` is already false. The `allAvailable` branch inside that runnable calls `overlayManager.hide()` and `onUnregisterThumbnailObserver()` unconditionally (not gated on `isOverlayActive`), so the thumbnail observer does eventually get cleaned up. But session cleanup inside the deactivation runnable is gated on `sessionTracker.isSessionActive`, so if the session was not ended at gallery-hide time it will be ended later — delayed but not permanently leaked. The real risk is the window between gallery-hide and camera-available where the session appears active but the overlay is not visible.

There are no tests that verify session/observer cleanup on the immediate-hide or recheck-hide paths. A test such as: _"when the overlay is hidden immediately on gallery launch and a session is active, the session is ended and observers are unregistered"_ would catch this.

---

## Bug 2 (minor): Pending deactivation runnable not cancelled on immediate gallery hide

`onGalleryLaunched()` does not call `cancelPendingDeactivation()`. If a `deactivateRunnable` is already queued (because a camera-available event fired just before the gallery tap), it will fire after the gallery hide and call `overlayManager.hide()` a second time. `OverlayManager.hide()` is idempotent so this does not crash, but the stale runnable calling `onUnregisterThumbnailObserver()` unconditionally when cameras are available could interfere with a future overlay activation that started in the interval between the gallery hide and the stale runnable firing.

---

## Note on lock-screen interaction

On a locked device, `getForegroundPackage()` returns null, so `isPixelCameraPackage(null)` is false, and `onGalleryLaunched()` always takes the immediate-hide path. This is probably the correct behavior (tapping the gallery button on the lock screen should hide the overlay), but the lock-screen guard pattern used elsewhere (`if (isKeyguardLocked()) ...`) is absent here. A comment clarifying this is intentional would help future readers.

---

## Positive

The callback wiring (OverlayManager → OverlayService → OverlayServiceLogic) is clean and correct. The guard `if (!isOverlayActive) return` at the top of `onGalleryLaunched()` is correct. The bundled issue #81 and MediaChangeDispatcher/ThumbnailChangeDispatcher changes look reasonable in isolation, but they represent substantial unrelated scope for a PR targeted at #91.

---

aunger:
Thanks for the thorough review. All three issues are fixed in commit 27f4340.

**Bug 1 — Missing teardown on gallery-launch hide paths**

Both `onGalleryLaunched()` (immediate hide) and the `scheduleGalleryLaunchRecheck()` runnable previously inlined `overlayManager.hide() / isOverlayActive = false / onOverlayStateChanged(false)` without calling `onUnregisterThumbnailObserver`, `sessionTracker.endSession`, or `onUnregisterMediaObserver`. Both now delegate to `hideOverlayAndCleanup()`, the same helper used by `scheduleDeactivation`. This was already extracted in the immediately preceding refactoring commit (`4d4964b`) but the gallery-launch call sites hadn't been updated yet — that's what this commit finishes.

**Bug 2 — Missing `cancelPendingDeactivation()` on immediate-hide path**

`onGalleryLaunched()` now calls `cancelPendingDeactivation()` before the foreground check. This prevents a queued `deactivateRunnable` from firing after we've already hidden (and potentially cleaned up), which could double-call `hide()` or unconditionally unregister observers during a fresh activation.

**Lock-screen note**

Added an inline KDoc paragraph to `onGalleryLaunched()` explaining that on a locked device `getForegroundPackage()` returns null, so `isPixelCameraPackage` is always false and the immediate-hide path is always taken. This is intentional: the gallery button is not visible on the lock screen, so arriving at this code means the device was unlocked when the button was pressed — hiding immediately is correct. I preferred the comment over an explicit `isKeyguardLocked()` guard because the `isKeyguardLocked` checks elsewhere are there to *prevent* incorrect behaviour (activating when locked), whereas here there is no incorrect behaviour to prevent; the null-return just happens to produce the right outcome and deserves an explanation, not a guard.

**New tests**

Two new tests in `OverlayServiceLogicTest`:
- `issue-91 immediate hide on gallery launch ends session and unregisters observers` — activates via lock-screen bypass (so a session is live), then launches the gallery with PC gone; asserts `sessionTracker.endSession()` is called and both observer flags are cleared.
- `issue-91 deferred recheck hide ends session and unregisters observers` — same setup but with PC still in the foreground at launch time, so the recheck runnable is captured and driven to completion; same teardown assertions.

Also wired `thumbnailObserverRegistered` into the shared `setUp` so all tests using the default `logic` instance can now track thumbnail-observer state without per-test boilerplate.

All 96 unit tests pass (`./gradlew test`).

---

aunger (commented):
Round-2 review — both bugs verified fixed; no remaining blockers. **Verdict: APPROVE** (posted as COMMENT because the API does not permit self-approval).

---

**Bug 1 (teardown):** `hideOverlayAndCleanup()` is called on every hide path that belongs to issue #91 — the immediate path in `onGalleryLaunched()` and the deferred path in `scheduleGalleryLaunchRecheck()`. The `scheduleDeactivation` runnable also calls it. The method's KDoc accurately names all three callers.

**Bug 2 (stale deactivate runnable):** `cancelPendingDeactivation()` is called at the top of `onGalleryLaunched()` before the foreground check, so the old runnable cannot fire after this code hides the overlay.

**New tests:** Both teardown tests (`issue-91 immediate hide on gallery launch ends session and unregisters observers` and `issue-91 deferred recheck hide ends session and unregisters observers`) assert `thumbnailObserverRegistered == false`, `mediaObserverRegistered == false`, and `sessionTracker.endSession()` after the respective hide path fires. The shared `setUp` wires `thumbnailObserverRegistered` correctly. The deferred-recheck test uses `runnableCaptor.lastValue` to target the recheck runnable rather than an earlier deactivation one — correct given the `atLeastOnce()` capture.

**Lock-screen note:** The KDoc paragraph on `onGalleryLaunched()` explains the null-return behaviour of `getForegroundPackage()` on a locked device and why the immediate-hide path is intentional. Clear and sufficient.

---

**Backlog observation (not blocking):**

`onOverlayFocusLost()` (lines 314–320) and the usage-stats-revoked branch of `evaluateForeground()` (lines 97–107) both inline `overlayManager.hide()` / `isOverlayActive = false` / `onOverlayStateChanged(false)` without calling `hideOverlayAndCleanup()`. Neither ends an active secure session nor unregisters the thumbnail observer. This was pre-existing and is out of scope for #91, but now that `hideOverlayAndCleanup()` exists as the canonical teardown method it is worth a follow-up issue.

---

PR #100, opened 2026-05-04
Accept ACTIVITY_RESUMED events in ForegroundDetector

### LINKED ISSUE #86: FG detector is wrong
Foreground Detector sees `com.android.launcher3` as the FG app when the Pixel Camera is launched. Therefore, i cannot make the overlay appear.

This fails whether experimental responsiveness is enabled and or not .

______________________________________________________________________

### Root-cause analysis

The service confirms which app holds the camera through `ForegroundDetector.getForegroundPackage()` (`app/src/main/java/com/gb4pc/service/ForegroundDetector.kt`).
That method returns the single package with the **latest** `ACTIVITY_RESUMED` / `MOVE_TO_FOREGROUND` event in the last 5 seconds, with GB4PC's own package filtered out (#80).
`OverlayServiceLogic.evaluateForeground()` then activates the overlay only when that single winner equals `com.google.android.GoogleCamera`.

The two prior fixes (#100, #107) taught the detector to accept `ACTIVITY_RESUMED`, removing one cause (previously only `MOVE_TO_FOREGROUND` was checked).
The residual ~50% failure comes from a deeper structural weakness that those fixes did not touch: the "single latest-RESUMED winner" model is fragile when a launcher foreground event races ahead of, or lingers past, the camera's event.
When that happens the winner is the launcher package, `isPixelCameraPackage()` returns false, and the overlay never appears, even though the camera hardware is demonstrably held (that is exactly what fired `onCameraUnavailable` and drove us to evaluate).

The signal the current code ignores: at the instant we evaluate, a camera is unavailable (held by some app).
A launcher never holds the camera.
So a launcher "foreground" reading at that instant is spurious for the purpose of deciding whether Pixel Camera is up, and Pixel Camera, if it appears anywhere in the recent window, is the real camera holder.

Note: the maintainer's second sub-request on this issue (log the full list of foreground apps, not just the top one) was already delivered by #325, and `DebugLog` now mirrors to logcat under the `GB4PC` tag via #425, so device diagnostics are already available.

### Is action warranted?

Yes, with scope discipline.
This is the app's core function (labels P1, product) and it is failing about half the time, so a fix is high-value.
The caveat is that the precise device event ordering cannot be reproduced in the JVM/Robolectric suite (the emulator E2E uses a mock camera and cannot reproduce real UsageStats timing, as documented in `PixelCameraOverlayE2ETest`).
Therefore the fix must be (a) justified independently of the exact race, and (b) confirmed on a physical Pixel before the issue is closed.
Both are achievable, so proceeding is the right call rather than deferring for more data: the fix below is well-motivated now, and it can carry its own confirmation logging.

### Design: ignore a spurious launcher winner when the camera is held

Introduce a camera-corroborated activation predicate rather than trusting a single latest-RESUMED winner.

1. **Resolve the current launcher package once.**
   In `OverlayService.onCreate()`, resolve the default launcher via `PackageManager` (an `Intent(ACTION_MAIN).addCategory(CATEGORY_HOME)` resolve) and pass its package name into `ForegroundDetector`'s constructor, mirroring how `selfPackage` is already injected.
   Resolve dynamically rather than hardcoding `com.android.launcher3`, because real Pixels use `com.google.android.apps.nexuslauncher` and users may run third-party launchers.
   If it cannot be resolved, pass null and the mask below is simply disabled (behaviour identical to today).

2. **Add a predicate to `ForegroundDetector`.**
   `ForegroundDetector` already builds `allForegroundPackages` (an ordered, de-duplicated, self-filtered set of every foreground package seen in the window).
   Add a method, e.g. `isForegroundConsideringLauncher(target: String): Boolean`, with this rule:
   - winner = latest non-self RESUMED package (the existing `getForegroundPackage()` result).
   - If `winner == target` return true.
   - Else if `winner == launcherPackage` and `target` is present in `allForegroundPackages`, return true (the launcher-masked case; covers the launcher event landing either before or after the camera's).
   - Else return false.
   Bounding the mask to "winner is the launcher" keeps the stale-other-camera-app case safe: if the winner is, say, Snapchat, no masking occurs and we still correctly decline.

3. **Use the predicate in the activation branch of `evaluateForeground()`.**
   Replace the `isPixelCameraPackage(getForegroundPackage())` activation test (`OverlayServiceLogic.kt`, around lines 153-165) with the new predicate for `Constants.PIXEL_CAMERA_PACKAGE`.
   Keep the existing DT-06a activation-retry scheduling unchanged: it still covers the distinct "camera's event has not reached UsageStats yet" lag, where the camera is absent from the window entirely.

4. **Apply the same predicate on the deactivation-delay path (recommended).**
   In `onCameraAvailable()` (`OverlayServiceLogic.kt`, around lines 98-104), a spurious launcher winner during front/back camera switching currently selects the 0 ms path and can prematurely hide the overlay.
   Reusing the predicate there makes an ambiguous launcher-masked reading choose `debounceMs` (the safe path), which addresses the "then the FG checker lists only itself and the launcher" flicker on the deactivation side.
   Leave the locked-device handling (#81) as-is.

5. **Add a one-line log when the mask fires**, so a device logcat capture confirms the new path is what rescued activation.

### Alternatives considered and rejected

- **Strict RESUMED/PAUSED pairing to derive the true current foreground app.**
  Rejected as the primary mechanism: when the experimental focusable overlay grabs focus, the camera activity emits an `ACTIVITY_PAUSED`, which strict pairing would misread as "camera no longer foreground," risking a regression of the #80/#55/#92 behaviour.
  The launcher-mask avoids modeling PAUSED at all.
- **"Camera appears anywhere in the window, so activate" with no launcher-winner guard.**
  Rejected: too many false positives (a stale Pixel Camera event followed by a genuine switch to a different camera app would wrongly show the overlay over that app).
- **Diagnostic-only (ship nothing, request a device logcat first).**
  Reasonable and lowest-risk, but logging is already strong (#325, #425) and the launcher-mask is well-justified now.
  Preferred instead to ship the fix together with a mask-fired log line, so one PR yields both the fix and its confirmation data.
- **Hardcode `com.android.launcher3`.**
  Rejected: real Pixels report `com.google.android.apps.nexuslauncher`, and third-party launchers exist; resolve the HOME activity dynamically.

### Risks and open items

- The fix cannot be verified in CI; the real race needs a physical Pixel (true device E2E is tracked in #15).
  Treat physical-device confirmation as a before-merging manual step.
- Residual regime: if device logs show the camera's event is *entirely absent* from UsageStats at activation time (pure lag, not a launcher masking a present camera event), only the existing retry helps, not the mask.
  The mask-fired log plus the existing "all FG apps=" line will show which regime dominates; if lag dominates, follow up by widening the retry budget rather than the mask.
- Contrived false-positive: launcher foreground while a non-visible Pixel Camera holds the camera in the background would activate the overlay.
  This is not a normal state and is judged acceptable.

### Verification

- **Unit (CI):** add Robolectric tests alongside `ForegroundDetectorSelfFilterTest` using its event-injection harness:
  - launcher RESUMED after camera RESUMED, camera package present in window -> predicate true.
  - launcher RESUMED only, camera absent -> false (retry path preserved).
  - non-launcher winner (e.g. another camera app) with a stale camera event in window -> false (no masking).
  - null launcher package -> behaviour identical to `getForegroundPackage()`-based decision.
  - Add an `OverlayServiceLogic` test proving the activation branch now activates in the launcher-masked case.
- **Emulator E2E:** run `./gradlew connectedE2EAndroidTest`; `PixelCameraOverlayE2ETest.overlayAppearsWhenViewfinderOpens` must still pass. This is a guard only; the mock camera cannot reproduce the real race.
- **Manual (authoritative, before-merging):** on a physical Pixel, launch Pixel Camera from the launcher repeatedly to target the ~50% failure, confirm the overlay appears every time, and capture `adb logcat -s GB4PC` showing the "all FG apps=" line and the new mask-fired log. Record this as a PR checkbox.

aunger:
## Summary

Fixes #86

On Android 10+ (API 29+), `UsageStatsManager` emits `ACTIVITY_RESUMED` (value=7) instead of the deprecated `MOVE_TO_FOREGROUND` (value=1). `ForegroundDetector` was only checking for `MOVE_TO_FOREGROUND`, so it never saw the camera's foreground event on modern Pixel devices.

The result: the only `MOVE_TO_FOREGROUND` event visible in the 5-second query window was the launcher's stale event from before the user tapped the camera icon — so `getForegroundPackage()` returned `com.android.launcher3`, `isPixelCameraPackage()` returned `false`, and the overlay never appeared.

### Fix

`getForegroundPackage()` now accepts **both** `ACTIVITY_RESUMED` and `MOVE_TO_FOREGROUND` as foreground events, so detection works correctly on all supported API levels (26–35). The self-package filter (Issue #80) applies to both event types.

### Tests added (`ForegroundDetectorSelfFilterTest`)

- `ACTIVITY_RESUMED` for camera is correctly detected
- `ACTIVITY_RESUMED` for self (`com.gb4pc`) is filtered out
- Self `ACTIVITY_RESUMED` does not displace an earlier camera `ACTIVITY_RESUMED`
- Camera `ACTIVITY_RESUMED` (newer timestamp) beats an older launcher `MOVE_TO_FOREGROUND` — the exact scenario from the bug report
- Launcher `MOVE_TO_FOREGROUND` alone is not mistakenly identified as Pixel Camera

---

### DIFF
--- a/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
+++ b/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
@@ -8,7 +8,7 @@ import com.gb4pc.util.DebugLog
 /**
  * Detects the current foreground app using UsageStatsManager (DT-02, DT-06).
  *
- * @param selfPackage The package name of this app (com.gb4pc). MOVE_TO_FOREGROUND events for
+ * @param selfPackage The package name of this app (com.gb4pc). Foreground events for
  *   this package are ignored because the overlay window causes Android to report GB4PC itself
  *   as the foreground app — which would displace the camera from the detected foreground app
  *   and hide the overlay (Issue #80).
@@ -19,11 +19,17 @@ class ForegroundDetector(
 ) {
 
     /**
-     * Queries UsageStatsManager for the most recent MOVE_TO_FOREGROUND event
+     * Queries UsageStatsManager for the most recent foreground event
      * in the last [Constants.USAGE_STATS_WINDOW_MS] milliseconds.
+     *
+     * On Android 10+ (API 29+) the system emits [UsageEvents.Event.ACTIVITY_RESUMED] instead
+     * of the deprecated [UsageEvents.Event.MOVE_TO_FOREGROUND]. Both event types are accepted
+     * so that detection works across all supported API levels (Issue #86).
+     *
      * Events for [selfPackage] are skipped (Issue #80).
      * Returns the package name, or null if no event found (EC-09).
      */
+    @Suppress("DEPRECATION") // MOVE_TO_FOREGROUND is deprecated in API 29; we accept both types
     fun getForegroundPackage(): String? {
         val endTime = System.currentTimeMillis()
         val beginTime = endTime - Constants.USAGE_STATS_WINDOW_MS
@@ -44,12 +50,19 @@ class ForegroundDetector(
         while (events.hasNextEvent()) {
             events.getNextEvent(event)
             totalEvents++
-            if (event.eventType == UsageEvents.Event.MOVE_TO_FOREGROUND) {
+            // Accept both ACTIVITY_RESUMED (API 29+) and the legacy MOVE_TO_FOREGROUND so that
+            // detection works on all supported Android versions. On API 29+ the system emits
+            // ACTIVITY_RESUMED instead of MOVE_TO_FOREGROUND (Issue #86).
+            val isForegroundEvent = event.eventType == UsageEvents.Event.ACTIVITY_RESUMED ||
+                event.eventType == UsageEvents.Event.MOVE_TO_FOREGROUND
+            if (isForegroundEvent) {
                 if (event.packageName == selfPackage) {
                     skippedSelfEvents++
+                    DebugLog.log("ForegroundDetector: skipping self foreground event pkg=${event.packageName} ts=${event.timeStamp} (Issue #80)")
                     continue
                 }
                 foregroundEvents++
+                DebugLog.log("ForegroundDetector: foreground event type=${event.eventType} pkg=${event.packageName} ts=${event.timeStamp}")
                 if (event.timeStamp >= latestTimestamp) {
                     latestTimestamp = event.timeStamp
                     latestForegroundPackage = event.packageName
--- a/app/src/test/java/com/gb4pc/service/ForegroundDetectorSelfFilterTest.kt
+++ b/app/src/test/java/com/gb4pc/service/ForegroundDetectorSelfFilterTest.kt
@@ -18,7 +18,7 @@ import org.robolectric.RobolectricTestRunner
  * accessible — they are stubs with default values under the plain JVM test runner.
  */
 @Suppress("DEPRECATION") // MOVE_TO_FOREGROUND / MOVE_TO_BACKGROUND are deprecated by the SDK
-                          // but are the correct event types for pre-API-29 paths exercised here.
+                          // but are valid event types still exercised in mixed-event tests here.
 @RunWith(RobolectricTestRunner::class)
 class ForegroundDetectorSelfFilterTest {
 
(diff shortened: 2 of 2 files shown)


aunger (commented):
## Review of PR #100 — FG detector fix for issue #86

**Verdict: REQUEST_CHANGES** (GitHub blocked the formal REQUEST_CHANGES because the reviewer and author share an account; this comment carries the same intent.)

---

### The fix itself

The core change in `ForegroundDetector.kt` is correct. The `isForegroundEvent` expression (OR of `ACTIVITY_RESUMED` and `MOVE_TO_FOREGROUND`) is the right fix, and the existing timestamp-based winner selection already handles the exact bug scenario without any further changes. The `@Suppress("DEPRECATION")` annotation is applied at method scope, which is the right granularity — not class-level, and not buried in a comment.

### Tests

The five new tests in `ForegroundDetectorSelfFilterTest` are all meaningful. The exact regression scenario (camera `ACTIVITY_RESUMED` at ts=2000 vs launcher `MOVE_TO_FOREGROUND` at ts=1000) has a dedicated test. Self-filter for `ACTIVITY_RESUMED` is tested. No gaps in coverage for the stated fix.

Minor observation: there is no test for the mirror case — launcher `ACTIVITY_RESUMED` newer than camera `MOVE_TO_FOREGROUND` (user navigated away from camera on API 29+). The logic handles it correctly by construction, but the path is implicitly covered rather than explicitly asserted. Not a blocker.

### Scope — this is the blocker

The PR title says "Accept ACTIVITY_RESUMED events in ForegroundDetector" and it fixes #86. The `ForegroundDetector.kt` production change is 6 lines. However, the PR changes **10 files with 929 additions and 158 deletions** and bundles at minimum three separate, unrelated features:

- **Issue #91** — `onGalleryLaunched()` early-hide logic in `OverlayServiceLogic` and `OverlayManager`, with 9 new tests.
- **Issue #81** — Lock-screen bypass in `OverlayServiceLogic`, with 7 new tests.
- **`MediaChangeDispatcher` rework** — `queryLatestMedia` replaced with `queryAllMedia`, cancel-and-reschedule retry, multiple-photo handling, plus changes to `OverlayService.kt`.
- **`ThumbnailChangeDispatcher` rework** — always-retry and cancel-and-reschedule logic.

None of these relate to the FG detector fix. Each is a meaningful behaviour change that deserves its own review. Bundling them makes it impossible to reason cleanly about regressions: if something in `OverlayServiceLogicTest` fails, it is not clear whether the failure is in the #86 fix, the #81 bypass, the #91 gallery-launch path, or the dispatcher rework.

**Request:** Split the non-#86 changes into separate PRs. The `ForegroundDetector.kt` change and its 5 new tests are ready to merge on their own. The rest should land independently so each can be reviewed on its own terms.

### Other event types

No other event types need to be added to `isForegroundEvent` for this fix. `FOREGROUND_SERVICE_START` (API 29+) and similar events are not foreground-app indicators in the sense needed here. The two accepted types cover the full supported API range (26–35).

---

aunger:
Branch cleaned up as requested. The scope concern is valid — the branch had accumulated commits from issues #81 and #91 on top of the six-line #86 fix, which made the diff unreadable for a reviewer focused on foreground detection.

**What was done:**
- Reset `fix/issue-86-fg-detector` to `origin/main` (which already contains PR #98 / issue #81 work)
- Cherry-picked only the two #86 commits:
  - `Accept ACTIVITY_RESUMED events in ForegroundDetector (Issue #86)`
  - `Add tests for ACTIVITY_RESUMED detection (Issue #86)`
- Resolved a trivial conflict in `ForegroundDetector.kt` (the cherry-pick added debug log lines that origin/main's version of the file lacked; kept the log lines from the #86 commit)
- Force-pushed with `--force-with-lease`

**Result:** The branch now contains exactly 2 commits touching exactly 2 files:
- `app/src/main/java/com/gb4pc/service/ForegroundDetector.kt`
- `app/src/test/java/com/gb4pc/service/ForegroundDetectorSelfFilterTest.kt`

`./gradlew :app:testDebugUnitTest --rerun-tasks` passes (BUILD SUCCESSFUL, 28 tasks executed).

The #91 work (gallery-launch early hide) is in its own PR #99 and can be reviewed separately.

---

aunger (commented):
## Second-round review — APPROVE (GitHub blocked formal APPROVE; this comment carries the same verdict)

The branch has been cleaned up correctly. The PR now changes exactly 2 files (88 additions, 4 deletions): `ForegroundDetector.kt` and `ForegroundDetectorSelfFilterTest.kt`. The unrelated issue #81 and #91 changes are gone. The first-round blocker is resolved.

---

### Fix correctness

The `isForegroundEvent` expression is correct:

```kotlin
val isForegroundEvent = event.eventType == UsageEvents.Event.ACTIVITY_RESUMED ||
    event.eventType == UsageEvents.Event.MOVE_TO_FOREGROUND
```

- Covers the full supported API range (26–35): `MOVE_TO_FOREGROUND` for pre-29 devices, `ACTIVITY_RESUMED` for API 29+.
- The self-package filter (Issue #80) is applied to both event types within the same `if (isForegroundEvent)` block — no double-handling, no path where self events leak through.
- `@Suppress("DEPRECATION")` is scoped to the method, not the class — correct granularity.
- KDoc is updated to explain the API 29 behaviour change and references Issue #86. The class-level `@param selfPackage` doc is updated to say "Foreground events" (both types) rather than "MOVE_TO_FOREGROUND events" — accurate.
- The two new `DebugLog.log()` calls are placed at the right points: one on the skip path (with Issue #80 attribution), one on the accept path (with event type logged for diagnosability). No log pollution on the non-foreground-event path.

No other event types need to be added. `FOREGROUND_SERVICE_START` and similar are not foreground-app indicators in this context.

---

### Test correctness

All five new tests in `ForegroundDetectorSelfFilterTest` are sound:

1. **`ACTIVITY_RESUMED for camera is detected`** — basic smoke test for the fix.
2. **`self ACTIVITY_RESUMED is skipped`** — confirms the #80 filter extends to the new event type.
3. **`self ACTIVITY_RESUMED does not displace earlier camera ACTIVITY_RESUMED`** — timestamp logic is preserved.
4. **`camera ACTIVITY_RESUMED beats older launcher MOVE_TO_FOREGROUND`** — this is the exact regression scenario from the bug report (launcher stale event at ts=1000, camera `ACTIVITY_RESUMED` at ts=2000). Most important test; it is present and correct.
5. **`launcher MOVE_TO_FOREGROUND alone does not match camera`** — verifies `isPixelCameraPackage` returns false before camera has launched, ensuring retry logic can kick in.

The `@Suppress("DEPRECATION")` comment on the test class is correctly updated to "mixed-event tests" since both `MOVE_TO_FOREGROUND` and `ACTIVITY_RESUMED` now appear in the same file.

The `eventsOf()` reflection helper (`mPackage`, `mEventType`, `mTimeStamp`) follows the established pattern in the file and is correct for Robolectric.

One non-blocking observation carried over from round one: there is no explicit test for the mirror case (launcher `ACTIVITY_RESUMED` newer than camera `MOVE_TO_FOREGROUND`, i.e., user navigated away from camera on API 29+). The code handles it correctly by construction — the timestamp comparison is event-type-agnostic — so this is not a blocker, just a note for possible future hardening.

---

No remaining problems. Ready to merge.

_Reviewed by [Claude Code](https://claude.ai/code/session_01MqUjuNR7mUghQediPG3U8d)_

---

PR #101, opened 2026-05-04
Fix focusable-overlay teardown in onOverlayFocusLost and onOverlayFocusGained

### LINKED ISSUE #92: Review experimental focusable overlay (55 & 60)
Review implementation of #55 / #60. The focus listener trick doesn't seem to ever receive a focus-lost event before the slowpoke camera-available event shuts down the overlay window. 

Is this mechanism too slow, implemented incorrectly in our app, or just inappropriate for this usage? 

---

### Answer to the review question: inappropriate for this usage

The mechanism is not merely slow or miswired; it is the wrong tool for this job, for three compounding reasons.

1. **It is self-contradictory by design.**
A window only receives `onWindowFocusChanged(false)` if it actually holds input focus.
For the overlay to hold focus, it must be the topmost focusable window over Pixel Camera, which is exactly what steals volume-as-shutter and zoom keys from the camera.
The implementation tries to have it both ways (focusable window, but `dispatchKeyEvent` returns false and `FLAG_NOT_TOUCH_MODAL` is set to pass touches through), so in practice the window does not hold focus in a way that ever yields a useful focus-lost callback.
This is the same kill condition #55 already wrote down: "If focus is stolen on show, the approach is not viable and the setting should be removed."

2. **The signal it was meant to provide is redundant.**
The motivating case was the task switcher.
But opening recents pauses Pixel Camera, which releases the camera, which fires the camera-available callback that already tears the overlay down.
For the notification shade, the shade renders in a higher window layer than `TYPE_APPLICATION_OVERLAY`, so it covers the overlay regardless.
There is no scenario left where a focus-lost signal would beat or improve on what we already do.

3. **It matches the owner's empirical finding.**
Focus-lost is never logged before camera-available tears the view down, and per the follow-up comment, not at all.
A view detached by `removeView` gets no further focus callbacks, so even if one were ever coming, the camera-available teardown wins the race every time.
PR #101 assumed the callback fires and just needed fuller teardown; that premise was wrong, so its `onOverlayFocusLost` / `onOverlayFocusGained` bodies are dead code guarding a callback that never runs.

### Decision: remove the experimental focusable-overlay feature

Keeping it is a net negative.
It ships a user-facing toggle ("Experimental responsiveness boost") that delivers no benefit and, if enabled, risks breaking the camera's hardware-key shutter and zoom.
It also spreads dead complexity across eight source files plus a dedicated `ACTION_RESHOW_OVERLAY` service path.
Removal is the action most consistent with the repo's "simplify, don't complexify" guidance.
Alternatives rejected: (a) make focus actually stick, rejected because it breaks camera keys and #55 already declared that non-viable; (b) hide the toggle but keep the code, rejected because it leaves a dormant pref and dead paths.

### Steps (each concrete)

1. **`OverlayManager.kt`**
   - Delete the `onFocusLost` and `onFocusGained` constructor parameters.
   - In `createOverlayView()`, keep the anonymous `ImageView` subclass but remove the `dispatchKeyEvent` override and the `onWindowFocusChanged` override.
     Keep the `dispatchTouchEvent` diagnostic logging (Issue #230 / #397); that is unrelated to this feature.
   - In `createLayoutParams()`, delete the `if (prefsManager.focusableOverlay)` branch and keep only the non-focusable flag set (`FLAG_NOT_FOCUSABLE or FLAG_NOT_TOUCH_MODAL or FLAG_LAYOUT_IN_SCREEN or showWhenLockedFlag`).
     Note the focusable branch was the only user of `FLAG_LAYOUT_NO_LIMITS`, which PRs #398 and #557 deliberately kept off the real path, so dropping it is correct.
   - Delete `reshow()`.

2. **`OverlayServiceLogic.kt`**
   - Delete `onOverlayFocusLost()` and `onOverlayFocusGained()` and their "Issue #55 / #92" section.
   - `hideOverlayAndCleanup()`, `cancelPendingDeactivation()`, `onRegisterThumbnailObserver`, etc. remain; they are used by the normal camera and gallery-launch paths.

3. **`OverlayService.kt`**
   - Drop the `onFocusLost = { logic.onOverlayFocusLost() }` and `onFocusGained = { logic.onOverlayFocusGained() }` arguments to the `OverlayManager` constructor.
   - Remove the `ACTION_RESHOW_OVERLAY` handling in `onStartCommand`, the `ACTION_RESHOW_OVERLAY` constant, and the `reshowOverlay()` companion function.

4. **`PrefsManager.kt` and `Constants.kt`**
   - Delete the `focusableOverlay` property and the `PREF_FOCUSABLE_OVERLAY` constant.

5. **`MainActivity.kt` and `strings.xml`**
   - Remove the "Experimental responsiveness boost" `Row` (around line 244), the `isFocusableOverlay` state, and the `OverlayService.reshowOverlay(context)` call, plus one of the now-adjacent `HorizontalDivider()`s so the layout stays clean.
   - Remove the unused `OverlayService` import if it is no longer referenced.
   - Delete the `settings_focusable_overlay_title` and `settings_focusable_overlay_subtitle` strings.

6. **Tests**
   - `OverlayServiceLogicTest.kt`: delete the "Issue #55" and "Issue #92" focus sections (the `onOverlayFocusLost` / `onOverlayFocusGained` tests, roughly lines 587 to 797).
   - `OverlayManagerRobolectricTest.kt`: delete the Issue #66 regression test `show with non-focusable overlay does not trigger onFocusLost` (its whole reason for existing was the focus callback), and remove every `on { focusableOverlay } doReturn false` mock stub and the now-removed constructor args so the remaining thumbnail / squircle / #45 / #81 tests still compile.
   - `PrefsManagerTest.kt`: delete the two `focusableOverlay` tests and the `PREF_FOCUSABLE_OVERLAY` stub in the shared mock setup.

### Commit organization

This is a single concern (remove one feature), so do not split by file.
A reasonable split is: one commit removing the runtime mechanism and preference across the main sources, and a second removing the now-dead tests; a single cohesive commit is also acceptable.
No behavior-preserving refactor is bundled in, so the separate-refactor rule does not apply.

### Verification

- `grep -ri "focusable\|reshow\|responsiveness\|FOCUSABLE_OVERLAY"` over the tree returns nothing outside deliberate history.
- `./gradlew test` passes (the deleted tests are the only ones touching removed APIs).
- `./gradlew assembleDebug` builds; confirm no unused-import or unused-string lint failure from the removed toggle.
- Existing overlay E2E coverage (`PixelCameraOverlayE2ETest`) still shows and hides the overlay normally, confirming the non-focusable path is untouched.

### Risks

- Low. The removed path was off by default and, per the owner's logs, never functioned.
- Users who had toggled the pref on will silently revert to the (correct) non-focusable behavior; the stored `focusable_overlay` key is simply ignored after removal, which is harmless.

aunger:
Fixes #92

## Finding

The focusable-overlay mechanism is **implemented incorrectly** (case b from the issue). The focus-lost callback does fire — it just doesn't do enough when it does.

### Bug 1: `onOverlayFocusLost()` used a bare `hide()` instead of full teardown

In `main`, `onOverlayFocusLost()` called `overlayManager.hide()` directly. This missed three things that every hide path must do:

1. **Cancel the pending camera-available deactivation runnable.** When the task-switcher opens, Pixel Camera may release the camera shortly after, scheduling a deactivation runnable. If focus-loss hid the overlay first and the runnable fired later, it would call `hideOverlayAndCleanup()` on an already-hidden overlay — double-firing `onOverlayStateChanged` and double-unregistering the thumbnail observer.

2. **Unregister the thumbnail `ContentObserver`.** Without this, the observer leaked until the deactivation runnable eventually fired.

3. **End the active secure session.** If the overlay was activated on the lock screen, a secure session was in progress. Focus loss did not end it, leaving the session and its media observer running.

### Bug 2: `onOverlayFocusGained()` did not re-register the thumbnail observer

After a focus-lost/focus-gained cycle the thumbnail observer was left unregistered, so new-photo thumbnails were not shown in the overlay button until the next full re-activation.

## Fix

`onOverlayFocusLost()` now calls `cancelPendingDeactivation()` and `hideOverlayAndCleanup()` — the same teardown used by the normal camera-available path.

`onOverlayFocusGained()` now calls `onRegisterThumbnailObserver()` — matching `showOverlay()`.

## Tests added (4 new unit tests)

- `issue-92 onOverlayFocusLost cancels pending deactivation runnable` — verifies `removeCallbacks` is called on the pending runnable
- `issue-92 onOverlayFocusLost unregisters thumbnail observer` — verifies observer flag cleared on focus loss
- `issue-92 onOverlayFocusLost ends active secure session` — verifies `endSession()` + media observer unregistered
- `issue-92 onOverlayFocusGained re-registers thumbnail observer` — verifies observer re-registered after focus-lost/gained cycle

All 4 tests were failing against the `main` implementation and pass with the fix.

`./gradlew test` — BUILD SUCCESSFUL, 96 tasks.

---

### DIFF
--- a/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
@@ -337,6 +337,10 @@ class OverlayServiceLogic(
      * Issue #92: Must re-register the thumbnail observer (via onRegisterThumbnailObserver) so
      * the overlay button thumbnail works correctly after focus is regained, matching the
      * behaviour of showOverlay().
+     *
+     * Issue #92 (lock-screen): mirrors [showOverlay]'s lock-screen path — if the device is
+     * locked when focus is regained, re-start the secure session and re-register the media
+     * observer so newly captured photos update the thumbnail button.
      */
     fun onOverlayFocusGained() {
         DebugLog.log("Logic: overlay focus gained; overlayActive=$isOverlayActive")
@@ -350,5 +354,10 @@ class OverlayServiceLogic(
         isOverlayActive = true
         onOverlayStateChanged(true)
         onRegisterThumbnailObserver()
+        if (isKeyguardLocked()) {
+            DebugLog.log("Logic: overlay focus gained on locked device — starting secure session")
+            sessionTracker.startSession()
+            onRegisterMediaObserver()
+        }
     }
 }
--- a/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
@@ -314,19 +314,29 @@ class OverlayServiceLogic(
      * Indicates a system surface (task-switcher, notification shade, etc.) has covered the
      * camera app. Hides the overlay so it does not obscure the system surface.
      * Only fired when the experimental focusable-overlay preference is enabled.
+     *
+     * Issue #92: Must cancel any pending camera-available deactivation runnable before hiding,
+     * otherwise the runnable fires after focus-loss hides the overlay and calls
+     * hideOverlayAndCleanup() on an already-hidden overlay — double-firing onOverlayStateChanged
+     * and incorrectly unregistering the thumbnail observer a second time.
+     * Uses hideOverlayAndCleanup() (not a bare hide()) so the thumbnail observer and any active
+     * secure session are torn down correctly on focus loss.
      */
     fun onOverlayFocusLost() {
         DebugLog.log("Logic: overlay focus lost; overlayActive=$isOverlayActive")
         if (!isOverlayActive) return
-        overlayManager.hide()
-        isOverlayActive = false
-        onOverlayStateChanged(false)
+        cancelPendingDeactivation()
+        hideOverlayAndCleanup()
     }
 
     /**
      * Called by [com.gb4pc.overlay.OverlayManager] when the overlay window regains focus.
      * Indicates the camera app is back in the foreground. Re-shows the overlay.
      * Only fired when the experimental focusable-overlay preference is enabled.
+     *
+     * Issue #92: Must re-register the thumbnail observer (via onRegisterThumbnailObserver) so
+     * the overlay button thumbnail works correctly after focus is regained, matching the
+     * behaviour of showOverlay().
      */
     fun onOverlayFocusGained() {
         DebugLog.log("Logic: overlay focus gained; overlayActive=$isOverlayActive")
@@ -339,5 +349,6 @@ class OverlayServiceLogic(
         overlayManager.show()
         isOverlayActive = true
         onOverlayStateChanged(true)
+        onRegisterThumbnailObserver()
     }
 }
--- a/app/src/test/java/com/gb4pc/service/OverlayServiceLogicTest.kt
+++ b/app/src/test/java/com/gb4pc/service/OverlayServiceLogicTest.kt
@@ -606,6 +606,94 @@ class OverlayServiceLogicTest {
         assertEquals("Overlay-permission-lost notification should fire", 1, overlayLostCount)
     }
 
+    // ── Issue #92: focusable-overlay teardown correctness ────────────────────
+
+    /**
+     * Issue #92: onOverlayFocusLost must cancel any pending camera-available deactivation
+     * runnable before hiding the overlay. Without this, the deactivation runnable fires after
+     * focus-loss has already hidden the overlay, calling hideOverlayAndCleanup() a second time
+     * — double-firing onOverlayStateChanged and incorrectly unregistering the thumbnail
+     * observer again.
+     */
+    @Test
+    fun `issue-92 onOverlayFocusLost cancels pending deactivation runnable`() {
+        // Activate the overlay
+        whenever(foregroundDetector.getForegroundPackage()).thenReturn(Constants.PIXEL_CAMERA_PACKAGE)
+        logic.showOverlay()
+        assertTrue("Pre-condition: overlay should be active", logic.isOverlayActive)
+
+        // Schedule a deactivation (simulates onCameraAvailable firing just before focus loss)
+        logic.scheduleDeactivation(50L)
+        val runnableCaptor = argumentCaptor<Runnable>()
+        verify(handler).postDelayed(runnableCaptor.capture(), eq(50L))
+
+        // Focus lost — must cancel the pending deactivation
+        logic.onOverlayFocusLost()
+
+        verify(handler).removeCallbacks(runnableCaptor.firstValue)
+        assertFalse("Overlay should be inactive after focus loss", logic.isOverlayActive)
+    }
+
+    /**
+     * Issue #92: onOverlayFocusLost must call hideOverlayAndCleanup() (not a bare hide()) so
+     * the thumbnail observer is unregistered. Without this, when the task switcher covers
+     * Pixel Camera the thumbnail observer leaks until the camera-available deactivation fires.
+     */
+    @Test
+    fun `issue-92 onOverlayFocusLost unregisters thumbnail observer`() {
+        // Activate the overlay — this registers the thumbnail observer
+        whenever(foregroundDetector.getForegroundPackage()).thenReturn(Constants.PIXEL_CAMERA_PACKAGE)
+        logic.showOverlay()
+        assertTrue("Pre-condition: thumbnail observer should be registered", thumbnailObserverRegistered)
+
+        logic.onOverlayFocusLost()
+
+        assertFalse("Thumbnail observer must be unregistered on focus loss (Issue #92)", thumbnailObserverRegistered)
+    }
+
+    /**
+     * Issue #92: onOverlayFocusLost must end an active secure session (via hideOverlayAndCleanup).
+     * If the overlay was activated while the device was locked and the session is still active
+     * when focus is lost, the session and media observer must be torn down.
+     */
+    @Test
+    fun `issue-92 onOverlayFocusLost ends active secure session`() {
+        // Activate via lock-screen bypass — starts a session immediately
+        keyguardLocked = true
+        cameraState.setCameraUnavailable("0")
+        logic.evaluateForeground()
+        assertTrue("Pre-condition: overlay should be active", logic.isOverlayActive)
+        assertTrue("Pre-condition: media observer should be registered", mediaObserverRegistered)
+        whenever(sessionTracker.isSessionActive).thenReturn(true)
+
+        logic.onOverlayFocusLost()
+
+        assertFalse("Overlay should be inactive after focus loss", logic.isOverlayActive)
+        verify(sessionTracker).endSession()
+        assertFalse("Media observer must be unregistered when session ends (Issue #92)", mediaObserverRegistered)
+    }
+
+    /**
+     * Issue #92: onOverlayFocusGained must re-register the thumbnail observer so that
+     * subsequent photo thumbnails are displayed in the overlay button after focus is regained.
+     * The original implementation omitted this call, leaving the observer unregistered until
+     * the next full showOverlay() activation.
+     */
+    @Test
+    fun `issue-92 onOverlayFocusGained re-registers thumbnail observer`() {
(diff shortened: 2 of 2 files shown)


aunger (commented):
## Review of PR #101 — Fix focusable-overlay teardown in onOverlayFocusLost and onOverlayFocusGained

**Verdict: REQUEST_CHANGES** (posted as COMMENT because GitHub does not allow requesting changes on your own PR)

---

### Issue #92 fixes — correct

Both bugs are correctly diagnosed and fixed.

**Bug 1 (`onOverlayFocusLost`):** The replacement of bare `overlayManager.hide()` + manual state with `cancelPendingDeactivation()` + `hideOverlayAndCleanup()` is right. All three failure modes described in the PR body — double-hide, thumbnail observer leak, session not ended — are addressed by routing through the single teardown helper. The fix is minimal and the logic is easy to follow.

**Bug 2 (`onOverlayFocusGained`):** Adding `onRegisterThumbnailObserver()` matches `showOverlay()` and closes the observer gap after a focus-lost/gained cycle. Correct.

**Tests:** The 4 issue-92 tests are meaningful. Each targets a specific failure mode, sets up a pre-condition that would have exposed the bug against the old code, and asserts the right thing. They don't merely test the happy path — they directly verify runnable cancellation, observer state, session teardown, and the re-registration cycle. Good.

---

### Problem: PR scope extends beyond issue #92

The diff also contains a complete implementation of issue #91 ("hide overlay early on gallery launch"):

- `OverlayManager`: new `onGalleryLaunched` constructor parameter and call site in `launchGalleryApp()`
- `OverlayService`: wiring `onGalleryLaunched` to `logic.onGalleryLaunched()`
- `OverlayServiceLogic`: `onGalleryLaunched()`, `scheduleGalleryLaunchRecheck()`, `cancelGalleryLaunchRecheck()`, `galleryLaunchRecheckRunnable`, `cancelGalleryLaunchRecheck()` call in `reset()`
- Test file: 9 additional issue-91 tests

None of this was mentioned in issue #92, none of it is necessary to fix issue #92, and it should be reviewed on its own terms in its own PR. Bundling it here means the issue-91 code receives only the scrutiny of an issue-92 review.

The issue-92 fix itself (`onOverlayFocusLost`, `onOverlayFocusGained`, 4 tests) is satisfactory and I would approve it in isolation. The `hideOverlayAndCleanup()` extraction is also clean and welcome. The issue-91 code needs to be split out.

---

### Observation (not blocking, pre-existing): `evaluateForeground` bare-hide path

The usage-stats-revoked hide in `evaluateForeground` (lines 104–109) still does a bare `overlayManager.hide()` sequence and does not call `hideOverlayAndCleanup()`, so it skips `onUnregisterThumbnailObserver()`. Now that the helper exists, this is an inconsistency. It predates this PR and is out of scope here — raising it so it can be tracked separately.

---

### Summary

The two issue-92 bugs are correctly fixed and the 4 new tests are meaningful. The change is blocked solely because it bundles the unreviewed issue-91 implementation. Please split the issue-91 code into its own PR so each fix can be reviewed and merged independently.

---

aunger:
Thanks for the review. I've verified the branch is already clean — the issue #91 code you flagged was never part of this branch. Here's the current state:

**Branch is already rebased on main (`95f72e4`, the tip after PRs #99 and #100 merged).** The branch has exactly one commit on top of main:

```
6c08b34 Fix #92: correct focusable-overlay teardown on focus-lost and focus-gained
```

**Diff vs main touches only two files:**
- `app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt`
- `app/src/test/java/com/gb4pc/service/OverlayServiceLogicTest.kt`

**Changes are strictly the issue #92 fixes:**
- `onOverlayFocusLost()`: calls `cancelPendingDeactivation()` then `hideOverlayAndCleanup()` (replacing the bare `hide()` + manual state mutation)
- `onOverlayFocusGained()`: adds `onRegisterThumbnailObserver()` call after re-showing
- 4 unit tests covering those two functions

`./gradlew test` passes. Force-pushed with `--force-with-lease` to update the remote branch.

---

aunger (commented):
## Round-2 Review — changes requested

**Scope:** Clean. The single PR commit touches only `OverlayServiceLogic.kt` and `OverlayServiceLogicTest.kt`. No out-of-scope changes.

**`onOverlayFocusLost()`:** Correct. `cancelPendingDeactivation()` + `hideOverlayAndCleanup()` properly mirrors `onGalleryLaunched()` and eliminates both the double-fire race and the observer/session leak.

**`onOverlayFocusGained()` — one remaining bug:**

`showOverlay()` conditionally starts a secure session when the device is locked at activation time:

```kotlin
// showOverlay()
if (locked) {
    sessionTracker.startSession()
    onRegisterMediaObserver()
}
```

`onOverlayFocusGained()` does not. The sequence that exposes this:

1. Device is locked; overlay activates via lock-screen bypass → `showOverlay()` → session started, media observer registered.
2. Task-switcher appears → `onOverlayFocusLost()` → `hideOverlayAndCleanup()` → session ended, media observer unregistered. Correct.
3. User dismisses task-switcher → `onOverlayFocusGained()` → overlay re-shown, thumbnail observer re-registered, but **session is not restarted and media observer is not re-registered**.

After step 3, the overlay is visible on the lock screen but the media observer is absent, so newly captured photos will not update the thumbnail button until the user fully re-activates the overlay. This is the same class of bug as the missing `onRegisterThumbnailObserver()` that this PR correctly fixes — just on the session/media side.

**No test covers this path.** The four new tests exercise `onOverlayFocusGained` only in the unlocked case (or via the simple focus-lost/gained cycle test that does not set `keyguardLocked = true` before regaining focus).

The fix is analogous to what was done for the thumbnail observer: `onOverlayFocusGained()` should mirror `showOverlay()` by also calling `sessionTracker.startSession()` + `onRegisterMediaObserver()` when `isKeyguardLocked()` is true, and a test should verify that behaviour.

---

aunger:
The reviewer's finding is correct. After a focus-lost/focus-gained cycle on a locked device, the overlay re-appeared but `onRegisterMediaObserver()` was never called again, so newly captured photos would not update the thumbnail button.

**Fix (commit `3427be6`):**

Added an `isKeyguardLocked()` guard to `onOverlayFocusGained()`, mirroring the same block in `showOverlay()`:

```kotlin
if (isKeyguardLocked()) {
    DebugLog.log("Logic: overlay focus gained on locked device — starting secure session")
    sessionTracker.startSession()
    onRegisterMediaObserver()
}
```

This ensures that after focus is regained on a locked device, the secure session and media observer are both restored — exactly matching `showOverlay()`'s lock-screen path.

**Test added:** `issue-92 onOverlayFocusGained on locked device starts session and registers media observer` — activates via the lock-screen bypass, loses focus (tears down session and observers), regains focus on a still-locked device, then asserts `sessionTracker.startSession()` was called a second time and the media observer is re-registered.

All 96 tasks pass (`./gradlew test` — BUILD SUCCESSFUL).

---

aunger (commented):
The round-3 additions are correct and the implementation now fully mirrors `showOverlay()`.

**Lock-screen path in `onOverlayFocusGained()` — verified complete**

Compared step-by-step against `showOverlay()` (lines 148–168):

| Step | `showOverlay()` | `onOverlayFocusGained()` |
|---|---|---|
| permission guard | yes | yes |
| `overlayManager.show()` | yes | yes |
| `isOverlayActive = true` | yes | yes |
| `onOverlayStateChanged(true)` | yes | yes |
| `onRegisterThumbnailObserver()` | yes | yes (added round 2) |
| `if (locked) startSession() + onRegisterMediaObserver()` | yes | yes (added round 3) |

No step is missing.

**`isKeyguardLocked()` ordering** — `showOverlay()` captures the locked flag before `overlayManager.show()`, while `onOverlayFocusGained()` evaluates it after `onOverlayStateChanged(true)`. Nothing in between alters keyguard state, so this is not a behavioural difference.

**New test** — the `issue-92 onOverlayFocusGained on locked device starts session and registers media observer` test correctly exercises the activate → focus-lost → focus-gained cycle on a locked device, and the `times(2)` assertion on `startSession()` specifically catches the bug this PR fixes.

**Verdict: APPROVE.** (GitHub disallowed a formal approve vote because the reviewer account is the PR author; this comment records the approval.)

---

PR #102, opened 2026-05-04
Implement squircle button shape with automated tests

### LINKED ISSUE #39: Change button shape from Circle to Squircle

aunger:
## Summary

Fixes #39

The overlay button already clips to a rounded rectangle (squircle) via `clipToOutline = true` and a custom `ViewOutlineProvider` using `Outline.setRoundRect` with a 30% corner-radius fraction (`Constants.SQUIRCLE_CORNER_RADIUS_FRACTION = 0.30f`). This applies uniformly whether the button is showing the gallery app icon or a photo thumbnail — both use the same `ImageView`.

The prior close of #39 (PR #40) lacked automated tests for the squircle shape, which is why the issue was reopened. This PR adds those tests.

### Refactoring (no behavior change)
- **`e3a414a`** – Extracted the magic `0.30f` corner-radius multiplier in `OverlayManager` into `Constants.SQUIRCLE_CORNER_RADIUS_FRACTION`, making the squircle shape testable and self-documenting.

### Tests added (`6bfbbfa`)

**`OverlayManagerRobolectricTest`** (3 new tests):
- `overlay view is clipped to squircle outline when showing gallery icon` — verifies `clipToOutline = true`, a custom `ViewOutlineProvider` is set (not the default `BACKGROUND` provider), the outline radius equals `view.width × SQUIRCLE_CORNER_RADIUS_FRACTION`, and the bounds cover the full view.
- `overlay view retains squircle outline after switching to photo thumbnail` — after `setImageBitmap()` replaces the icon drawable with a thumbnail, `clipToOutline` and the squircle `outlineProvider` are unchanged.
- `squircle outline provider is a no-op for zero-dimension view` — when the view has not yet been laid out (width/height = 0), `getOutline()` must not throw and must leave the outline empty.

**`OverlayManagerTest`** (2 new tests):
- `SQUIRCLE_CORNER_RADIUS_FRACTION is within valid squircle range` — constant is in (0, 0.5], ruling out no-rounding and full-circle shapes.
- `squircle corner radius calculation is consistent` — spot-checks the pixel radius for a 200 px view.

## Test plan

- `./gradlew :app:testDebugUnitTest` — all 5 new tests pass, no regressions.

---

### DIFF
--- a/app/src/main/java/com/gb4pc/Constants.kt
+++ b/app/src/main/java/com/gb4pc/Constants.kt
@@ -49,4 +49,9 @@ object Constants {
 
     // Snackbar undo timeout
     const val UNDO_TIMEOUT_MS = 5000L
+
+    // Overlay button shape (Issue #39)
+    // Corner radius as a fraction of the button's width, producing a squircle shape
+    // (30% matches Pixel Camera's rounded-square style).
+    const val SQUIRCLE_CORNER_RADIUS_FRACTION = 0.30f
 }
--- a/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
+++ b/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.graphics.PixelFormat
+import android.graphics.drawable.AdaptiveIconDrawable
 import android.graphics.drawable.Drawable
 import android.graphics.drawable.LayerDrawable
 import android.os.Build
@@ -111,6 +112,10 @@ class OverlayManager(
             bitmap?.let { bmp ->
                 android.os.Handler(android.os.Looper.getMainLooper()).post {
                     targetView.setImageBitmap(bmp)
+                    // Issue #39: invalidate the outline so clipToOutline re-applies the squircle
+                    // clip after the image changes. Some devices defer clip invalidation until the
+                    // next outline invalidation, causing the old (circular) mask to persist.
+                    targetView.invalidateOutline()
                 }
             }
         }.start()
@@ -194,23 +199,50 @@ class OverlayManager(
 
     /**
      * WG-01: Extract icon live from PackageManager each time.
-     * WG-02: Use adaptive icon shape mask.
+     * WG-02: Load raw AdaptiveIconDrawable so the squircle clip is the only shape applied.
      * AC-04: Show placeholder if gallery app uninstalled.
      */
     private fun updateIconDrawable(imageView: ImageView) {
         val galleryPackage = prefsManager.galleryPackage
         val icon = getGalleryIcon(galleryPackage)
         imageView.setImageDrawable(icon)
+        // Issue #39: ensure the squircle outline clip is applied after the drawable is set.
+        imageView.invalidateOutline()
     }
 
     private fun getGalleryIcon(packageName: String?): Drawable {
         if (packageName != null) {
             try {
-                // L2/WG-02: packageManager.getApplicationIcon() applies the device's adaptive icon
-                // shape mask on API 26+ automatically, so we get a properly-masked icon for free.
+                // Issue #39: Load the raw AdaptiveIconDrawable on API 26+ to avoid the system
+                // launcher shape mask (which is a circle on Pixel devices). When
+                // getApplicationIcon() is used, Android applies the device's icon shape mask
+                // automatically and returns a pre-clipped circular bitmap — making clipToOutline
+                // with our squircle outline have no visible effect (the icon content is already a
+                // circle, smaller than the squircle clip). By loading the raw AdaptiveIconDrawable
+                // and setting it directly on the ImageView, the squircle clip from clipToOutline
+                // is the only shape applied, giving the intended rounded-square appearance.
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                    val appInfo = context.packageManager.getApplicationInfo(packageName, 0)
+                    if (appInfo.icon != 0) {
+                        val pkgResources = context.packageManager.getResourcesForApplication(appInfo)
+                        val rawIcon = pkgResources.getDrawable(appInfo.icon, null)
+                        if (rawIcon is AdaptiveIconDrawable) {
+                            return rawIcon
+                        }
+                    }
+                }
+                // Pre-API 26 or non-adaptive icon: fall back to getApplicationIcon().
+                // On pre-26 there is no adaptive icon masking, so the icon is not pre-clipped.
                 return context.packageManager.getApplicationIcon(packageName)
             } catch (_: PackageManager.NameNotFoundException) {
                 // Gallery app uninstalled — fall through to warning placeholder (AC-04)
+            } catch (_: Exception) {
+                // Resource load failed — fall back to getApplicationIcon()
+                try {
+                    return context.packageManager.getApplicationIcon(packageName)
+                } catch (_: PackageManager.NameNotFoundException) {
+                    // Gallery app uninstalled — fall through to warning placeholder (AC-04)
+                }
             }
             // AC-04/M3: Gallery configured but uninstalled — show placeholder with warning badge.
             return buildWarningPlaceholder()
--- a/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
+++ b/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
@@ -14,6 +14,7 @@ import android.view.WindowManager
 import android.widget.ImageView
 import android.widget.Toast
 import androidx.core.content.ContextCompat
+import com.gb4pc.Constants
 import com.gb4pc.R
 import com.gb4pc.data.AspectRatioUtil
 import com.gb4pc.data.PrefsManager
@@ -182,7 +183,7 @@ class OverlayManager(
         imageView.outlineProvider = object : android.view.ViewOutlineProvider() {
             override fun getOutline(view: android.view.View, outline: android.graphics.Outline) {
                 if (view.width == 0 || view.height == 0) return
-                val radius = view.width * 0.30f
+                val radius = view.width * Constants.SQUIRCLE_CORNER_RADIUS_FRACTION
                 outline.setRoundRect(0, 0, view.width, view.height, radius)
             }
         }
--- a/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
+++ b/app/src/main/java/com/gb4pc/service/ForegroundDetector.kt
@@ -8,7 +8,7 @@ import com.gb4pc.util.DebugLog
 /**
  * Detects the current foreground app using UsageStatsManager (DT-02, DT-06).
  *
- * @param selfPackage The package name of this app (com.gb4pc). MOVE_TO_FOREGROUND events for
+ * @param selfPackage The package name of this app (com.gb4pc). Foreground events for
  *   this package are ignored because the overlay window causes Android to report GB4PC itself
  *   as the foreground app — which would displace the camera from the detected foreground app
  *   and hide the overlay (Issue #80).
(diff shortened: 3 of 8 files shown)


aunger (commented):
Two concrete problems with the new tests that prevent this from fully satisfying issue #39. Would REQUEST_CHANGES but the API does not allow self-review on own PRs.

---

**Problem 1 — `squircle corner radius calculation is consistent` is a tautology (OverlayManagerTest.kt, lines 58–69)**

Both sides of the `assertEquals` are the same expression:

```kotlin
val expectedRadius = viewSize * Constants.SQUIRCLE_CORNER_RADIUS_FRACTION   // LHS
assertEquals(
    "...",
    200f * Constants.SQUIRCLE_CORNER_RADIUS_FRACTION,   // expected: same expression
    expectedRadius,                                      // actual:   same expression
    0.001f
)
```

This passes regardless of what value `SQUIRCLE_CORNER_RADIUS_FRACTION` holds and does not touch the production code path in `OverlayManager` at all. It cannot catch a regression. It should either be deleted (the range test in the same class already guards the constant) or rewritten to compare against a hard-coded expected pixel value — e.g. `assertEquals(60f, expectedRadius, 0.001f)` for a 200 px view at 0.30f — so that a change to the constant actually causes the test to fail.

---

**Problem 2 — The thumbnail test does not exercise `showLatestPhotoThumbnail()` (OverlayManagerRobolectricTest.kt, lines 204–243)**

The test retrieves the `ImageView` from the shadow `WindowManager` and calls `overlayView.setImageBitmap(testBitmap)` directly, bypassing `OverlayManager.showLatestPhotoThumbnail()`. It then checks that `clipToOutline` is still true.

`clipToOutline` and `outlineProvider` are properties of `ImageView` itself; `setImageBitmap()` never touches them. This test cannot fail under any conceivable regression in `OverlayManager` — it is testing Android framework behaviour, not our code. The meaningful question for issue #39 is whether the squircle configuration survives the full `showLatestPhotoThumbnail()` call path. Since that method dispatches work to a background thread and posts back to the main looper, the test would need to call `overlayManager.showLatestPhotoThumbnail(...)` and idle the Robolectric main looper (e.g. via `Shadows.shadowOf(Looper.getMainLooper()).idle()`) before asserting.

---

The first Robolectric test (`overlay view is clipped to squircle outline when showing gallery icon`) and the zero-dimension guard are solid. The constant range test is fine. The two issues above need to be addressed before this satisfies the issue requirement for automated tests.

---

aunger:
Both issues raised in the review are addressed in commit `43cd4d5`.

**Fix 1 — tautological assertion in `OverlayManagerTest.kt`**

The original `squircle corner radius calculation is consistent` test computed `expectedRadius = viewSize * SQUIRCLE_CORNER_RADIUS_FRACTION` and then asserted that the same formula equalled `expectedRadius` — always true regardless of the constant's value. The fix replaces the expected value with the hard-coded literal `60f`, which is what `200 * 0.30` must equal. The test now actually fails if `SQUIRCLE_CORNER_RADIUS_FRACTION` drifts from `0.30`.

**Fix 2 — thumbnail test bypassed `OverlayManager` in `OverlayManagerRobolectricTest.kt`**

The original `overlay view retains squircle outline after switching to photo thumbnail` test called `overlayView.setImageBitmap(...)` directly on the retrieved `ImageView`. That only tests Android framework invariants and cannot catch any regression inside `OverlayManager`. The fix replaces the direct call with `overlayManager.showLatestPhotoThumbnail("content://com.gb4pc.test/images/1")` followed by `ShadowLooper.idleMainLooper()` to drain any `Handler.post()` callbacks before the outline assertions run. The test now exercises the public API path end-to-end.

`./gradlew test` passes locally with both fixes in place.

---

aunger (commented):
## Round-2 review — no remaining issues (would approve)

Both round-1 issues are fixed correctly.

---

### Fix 1: `squircle corner radius calculation is consistent` (OverlayManagerTest.kt)

The tautological `expectedRadius` variable is gone. The test now reads:

```kotlin
assertEquals("Radius for a 200 px view must equal 60f (200 * 0.30)", 60f, radius, 0.001f)
```

This is a genuine regression guard: if `SQUIRCLE_CORNER_RADIUS_FRACTION` drifts away from `0.30f`, the test fails. The companion range-check test (`SQUIRCLE_CORNER_RADIUS_FRACTION is within valid squircle range`) still correctly bounds the constant to `(0, 0.5]` without repeating the hard-coded spot value.

### Fix 2: `overlay view retains squircle outline after switching to photo thumbnail` (OverlayManagerRobolectricTest.kt)

The test now calls `overlayManager.showLatestPhotoThumbnail("content://...")` followed by `ShadowLooper.idleMainLooper()`. This is the right approach:

- It exercises the actual public API, not a backdoor `setImageBitmap` call.
- `ShadowLooper.idleMainLooper()` correctly drains the `Handler(Looper.getMainLooper()).post { ... }` dispatch that `showLatestPhotoThumbnail` uses internally, so the assertion sees the state after the real code path completes.
- The subsequent `clipToOutline` and `outlineProvider` assertions verify that the view-level squircle configuration is orthogonal to the drawable, which is exactly the property that needs protecting.

The other two new squircle tests (`overlay view is clipped to squircle outline when showing gallery icon`, `squircle outline provider is a no-op for zero-dimension view`) were already sound in round 1; they remain correct.

---

### Other changes in the diff (Issue #92, OverlayServiceLogic)

The five new `OverlayServiceLogicTest` tests and the corresponding production changes in `OverlayServiceLogic.onOverlayFocusLost` / `onOverlayFocusGained` are well-formed:

- `onOverlayFocusLost` calling `cancelPendingDeactivation()` then `hideOverlayAndCleanup()` is the correct fix and is directly verified by the cancellation test.
- `scheduleDeactivation` is package-internal (default Kotlin visibility), so the test calling it directly from the same package is legitimate rather than a test-only back door.
- The lock-screen focus-regained test correctly uses `times(2)` to distinguish the activation `startSession()` from the focus-regained one, and seeds `sessionTracker.isSessionActive = true` before `onOverlayFocusLost` so `hideOverlayAndCleanup` actually ends the session.

No issues found. The pre-existing `second show call does not overwrite thumbnail bitmap with icon` test still uses a direct `setImageBitmap` call rather than the public API, but that test is outside the scope of this PR and was not touched here.


---

aunger:
## Root cause and fix for the circular button (Issue #39)

### What was actually wrong

The squircle code (`clipToOutline = true` + `setRoundRect` outline) was structurally correct, but it had no visible effect because the image content inside the `ImageView` was already circular before clipping even ran.

The culprit was this line in `getGalleryIcon()`:

```kotlin
return context.packageManager.getApplicationIcon(packageName)
```

On API 26+ (which is the app's `minSdk`), `getApplicationIcon()` applies the device's launcher icon shape mask automatically. On all Pixel devices, that mask is a **circle**. The method returns a `BitmapDrawable` whose bitmap already has transparent corners cut to a circle. When that circular bitmap is drawn inside the `ImageView`, it occupies a circle-shaped area that is entirely _inside_ the squircle clip bounds — so the squircle clip clips nothing. The button looks like a circle because the icon content is a circle.

### The fix

In `getGalleryIcon()`, on API 26+, we now load the icon resource directly via `PackageManager.getResourcesForApplication()` and check whether it is an `AdaptiveIconDrawable`. If it is, we return it as-is — no system mask applied. The `ImageView`'s `clipToOutline` squircle then becomes the only shape applied to the icon, producing the intended rounded-square (squircle) appearance.

```kotlin
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
    val appInfo = context.packageManager.getApplicationInfo(packageName, 0)
    if (appInfo.icon != 0) {
        val pkgResources = context.packageManager.getResourcesForApplication(appInfo)
        val rawIcon = pkgResources.getDrawable(appInfo.icon, null)
        if (rawIcon is AdaptiveIconDrawable) {
            return rawIcon
        }
    }
}
// Pre-API 26 or non-adaptive icon: fall back to getApplicationIcon()
return context.packageManager.getApplicationIcon(packageName)
```

`getApplicationIcon()` is kept as the fallback for any non-adaptive icon (legacy apps that only supply a plain PNG icon) — those never had a system mask applied in the first place, so the squircle clip works on them without this change.

Additionally, `invalidateOutline()` is now called after both `setImageDrawable` (icon path) and `setImageBitmap` (thumbnail path) to ensure the clip is re-applied whenever the image content changes.

### Tests

A new Robolectric test (`gallery icon is loaded as AdaptiveIconDrawable to avoid system circular mask`) verifies the fix by asserting that the `ImageView` drawable is an `AdaptiveIconDrawable` — not a `BitmapDrawable` (the pre-masked type that would indicate the system circular mask was applied). All 207 tests pass.

### Commits pushed to `feature/issue-39-squircle-button`

- `ab303be` Fix #39: load raw AdaptiveIconDrawable to prevent system circular mask overriding squircle
- `73b22d7` Test #39: add regression guard verifying gallery icon is AdaptiveIconDrawable not pre-masked

---

aunger (commented):
Third-round review. The fix is correct and well-constructed. **Would APPROVE** (blocked by GitHub's self-review restriction on this account).

## Core fix — `getGalleryIcon()` (OverlayManager.kt)

**Correctness of the AdaptiveIconDrawable path**

The logic is sound:
1. `getApplicationInfo(packageName, 0)` — no flags needed; `appInfo.icon` is always present in the manifest.
2. `appInfo.icon != 0` guard prevents a crash on pathological apps that declare no icon.
3. `getResourcesForApplication(appInfo)` loads from the target app's own resource table, bypassing the launcher mask — this is exactly the right API for this purpose.
4. `getDrawable(appInfo.icon, null)` returns the raw drawable without any system theming applied.
5. The `is AdaptiveIconDrawable` check lets non-adaptive icons (legacy flat PNG, `BitmapDrawable`) fall through to `getApplicationIcon()` correctly.

**Exception handling**

The two-level catch structure is correct:
- `PackageManager.NameNotFoundException` is caught in both the inner and outer try-blocks so the uninstalled-gallery path always reaches `buildWarningPlaceholder()`.
- The generic `Exception` fallback retries `getApplicationIcon()`, which is the right behaviour for `Resources.NotFoundException` or any unexpected failure from `getResourcesForApplication`.
- Minor observation: `getResourcesForApplication` can itself throw `PackageManager.NameNotFoundException`. This is fine in practice — that exception is caught by the generic `Exception` handler, which invokes `getApplicationIcon()` as fallback, and if that also throws `NameNotFoundException`, `buildWarningPlaceholder()` is returned correctly. The behaviour is correct; the flow is just slightly non-obvious. Not a bug.

**`invalidateOutline()` placement**

Both call sites are correct:
- In `updateIconDrawable()` — called immediately after `setImageDrawable()` on the main thread. No timing issue.
- In `showLatestPhotoThumbnail()` — called inside the `Handler.post {}` block after `setImageBitmap()`, so it is also on the main thread and immediately after the drawable change.

There is no path where the drawable changes without `invalidateOutline()` following it. The `reshow()` path (`hide()` + `show()`) creates a fresh `ImageView` via `createOverlayView()`, which calls `updateIconDrawable()` and therefore `invalidateOutline()` — that path is covered too.

## Photo thumbnail path

`showLatestPhotoThumbnail` sets a `BitmapDrawable` from a decoded photo. The bitmap fills the entire `ImageView` bounds (`scaleType` is `FIT_CENTER`), so `clipToOutline` clips the fully-opaque bitmap content — there is no pre-baked transparency ring as with the launcher-masked icon. The squircle clip works correctly here without any additional changes needed.

## OverlayServiceLogic.kt changes (Issue #92)

Included in this PR but not the primary subject of issue #39. The changes are correct:
- `onOverlayFocusLost()` now calls `cancelPendingDeactivation()` before `hideOverlayAndCleanup()`, preventing the double-fire.
- `onOverlayFocusGained()` re-registers the thumbnail observer and, on a locked device, restarts the secure session and media observer — mirroring `showOverlay()`'s setup sequence correctly.

## Tests

**`gallery icon is loaded as AdaptiveIconDrawable to avoid system circular mask`** — the key regression test. Using the app's own package as proxy is a practical Robolectric approach given that Robolectric does not simulate launcher masking. The assertion (`is AdaptiveIconDrawable`, not `is BitmapDrawable`) directly catches regression to the pre-fix path. Meaningful and well-justified.

**`overlay view is clipped to squircle outline when showing gallery icon`** — tests `clipToOutline`, custom outline provider, radius value, and full-bounds rect. Thorough.

**`overlay view retains squircle outline after switching to photo thumbnail`** — correctly verifies that `showLatestPhotoThumbnail()` does not disturb the outline provider. Uses `ShadowLooper.idleMainLooper()` to drain the background-thread Handler post, which is the correct Robolectric idiom.

**`squircle outline provider is a no-op for zero-dimension view`** — edge-case guard for early layout passes. Checks `outline.isEmpty` after calling the provider with a zero-dimension view, confirming no crash and no spurious outline.

**`SQUIRCLE_CORNER_RADIUS_FRACTION is within valid squircle range`** and **`squircle corner radius calculation is consistent`** in `OverlayManagerTest` — lightweight sanity checks on the constant. The spot-check is redundant with the range check but harmless.

Issue #92 tests in `OverlayServiceLogicTest` are complete and correctly set up preconditions before each assertion.

## Constants

Extracting `0.30f` to `Constants.SQUIRCLE_CORNER_RADIUS_FRACTION` is correct — it was previously a magic literal in the outline provider lambda and is now testable and self-documenting.

## Summary

No blocking issues. The root cause is addressed correctly, the fallback chain is safe, `invalidateOutline()` covers all image-change paths, thumbnails are not affected by the launcher mask problem, and the test suite provides meaningful regression coverage. The PR is ready to merge.

---

PR #103, opened 2026-05-05
Poll MediaStore alongside ContentObserver to detect locked-screen photos

### LINKED ISSUE #81: On lock screen, no new photos detected

aunger:
## Summary

Fixes #81 — third attempt.  PRs #88 and #98 each tightened the
session/thumbnail dispatchers (IS_PENDING retry, fallback thumbnail
decode, all-items query, always-retry) but the issue was reopened
with the same symptoms after both: on a locked device the overlay
stays on the gallery icon and the secure-viewer filmstrip stays empty.

### Why those fixes were not enough

Both prior fixes lived inside `MediaChangeDispatcher.onMediaChanged()` /
`ThumbnailChangeDispatcher.onThumbnailChanged()`.  Those methods only
run when the registered `ContentObserver`'s `onChange` callback fires.
Field reports show that on a locked device the system can suppress
ContentObserver dispatch for our process — `onChange` never runs, the
retry runnable is never scheduled, and photos written to MediaStore
are silently missed.  No amount of work inside the dispatchers can
fix that, because the dispatchers are downstream of an unreliable
trigger.

### What this PR does

Add a periodic safety-net poll that runs alongside the ContentObserver
while a secure session is active.  Every 1500 ms it invokes the same
dispatcher entry points the observer would have invoked, so newly
captured photos are picked up within one polling interval even if no
observer callback ever arrives.  Detection on the lock screen
therefore degrades from "never" to "≤1.5 s in the worst case", which
addresses both reported symptoms (stale overlay icon, empty
filmstrip).

Notes:
- The poll re-uses the existing dispatchers, so IS_PENDING handling,
  the 500 ms retry, and `SessionTracker.addMedia` deduplication remain
  the single source of truth.  When the observer does fire, the
  duplicate poll callback is a cheap no-op.
- The poll runs only while a secure session is active (locked +
  overlay shown), so it adds no work on an unlocked device, where
  ContentObserver dispatch is reliable.
- New `MediaPoller` is a small standalone class with constructor-
  injected `Handler` + `onPoll`, exercised by 10 plain-JVM unit tests
  with no Robolectric.

## Commits

1. `Add MediaPoller — periodic safety-net poll for new MediaStore
   items` — pure addition of the new class, its constant, and 10 unit
   tests.  No wiring; not yet invoked from anywhere.
2. `Fix #81: poll MediaStore alongside the session ContentObserver` —
   wires `MediaPoller` into `OverlayServiceLogic` (paired with the
   media-observer lifecycle) and `OverlayService` (instantiated and
   plugged into the dispatchers).  6 new `OverlayServiceLogicTest`
   cases cover every start/stop edge.

## Test plan

- [x] `./gradlew test` — 227 unit tests pass (10 new in
  `MediaPollerTest`, 6 new in `OverlayServiceLogicTest`)
- [x] `./gradlew assembleDebug` — debug APK builds cleanly
- [ ] Manual: open Pixel Camera from the lock screen, take one or
  more photos.  Within ~1.5 s the overlay button must update to the
  latest photo's thumbnail.  Tapping the button must open the secure
  viewer with all session photos visible.

Fixes #81

---
_Generated by [Claude Code](https://claude.ai/code/session_01MDu9ubHwgkjUoyfdQNeiA4)_

### DIFF
--- a/app/src/main/java/com/gb4pc/Constants.kt
+++ b/app/src/main/java/com/gb4pc/Constants.kt
@@ -47,6 +47,11 @@ object Constants {
     // Retry delay when ContentObserver fires before MediaStore commits the new item (IS_PENDING race)
     const val MEDIA_OBSERVER_RETRY_MS = 500L
 
+    // Polling interval for MediaPoller — the safety-net poller that runs alongside the
+    // session ContentObserver so newly captured photos are detected even when the system
+    // suppresses ContentObserver dispatch on a locked device (Issue #81).
+    const val MEDIA_POLL_INTERVAL_MS = 1500L
+
     // Snackbar undo timeout
     const val UNDO_TIMEOUT_MS = 5000L
 
--- a/app/src/main/java/com/gb4pc/service/MediaPoller.kt
+++ b/app/src/main/java/com/gb4pc/service/MediaPoller.kt
@@ -1,0 +1,94 @@
+package com.gb4pc.service
+
+import android.os.Handler
+import com.gb4pc.Constants
+import com.gb4pc.util.DebugLog
+
+/**
+ * Periodically fans new MediaStore changes out to [MediaChangeDispatcher] and
+ * [ThumbnailChangeDispatcher] while a secure camera session is active.
+ *
+ * ## Why polling is needed (Issue #81)
+ *
+ * The session and thumbnail [android.database.ContentObserver]s are normally the fast
+ * path that detects newly captured photos.  Two earlier fixes (PR #88 and PR #98)
+ * tightened the IS_PENDING retry logic inside the dispatchers, but both relied on
+ * the observers' `onChange` callback firing in the first place.
+ *
+ * Field reports for issue #81 show that on a *locked* device the system can suppress
+ * or delay [android.database.ContentObserver] dispatch indefinitely — so the
+ * dispatcher's `onChange` is never invoked, the retry runnable is never scheduled,
+ * and photos written to MediaStore are silently missed.  The overlay therefore stays
+ * on the gallery icon and the secure-viewer filmstrip stays empty.
+ *
+ * [MediaPoller] adds a safety-net poll: every [intervalMs] it invokes the same
+ * dispatcher entry points the observers would have invoked, so newly committed
+ * photos are picked up within the polling interval even when no observer callback
+ * arrives.  Detection on the lock screen therefore degrades from "never" to "within
+ * one polling interval" in the worst case.
+ *
+ * The poller only runs while [start] / [stop] bracket an active session — typically
+ * mirroring the lifecycle of the media observer.  When the device is unlocked and
+ * the observer fires reliably the poller still runs but is a cheap no-op duplicate
+ * (deduplication in [com.gb4pc.viewer.SessionTracker.addMedia] guarantees idempotency).
+ *
+ * ## Side-effect-free
+ *
+ * The poller does not query MediaStore itself — it delegates to the existing
+ * dispatchers, so the test surface is unchanged and there is one canonical query
+ * path for both the observer and the poller.
+ */
+class MediaPoller(
+    private val handler: Handler,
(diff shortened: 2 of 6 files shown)


aunger (commented):
## Reviewer verdict: ready to merge (posted as COMMENT — GitHub blocks self-approve on this PR)

The root-cause analysis is convincing and the fix is appropriately conservative. PRs #88 and #98 both reinforced retry behaviour *inside* the dispatcher's `onChange` path, but neither addressed the possibility that `onChange` is never invoked — which is what field reports on the lock screen indicate. Adding a periodic poll that re-uses the same dispatcher entry points is a straightforward way to make the system robust to suppressed `ContentObserver` dispatch without introducing a parallel query path.

### Correctness

- `MediaPoller` is small, single-purpose, and constructor-injected — testable on plain JVM.
- `start` / `stop` idempotency, stop-from-inside-onPoll, and exception-swallowing in the tick are all verified by the new tests.
- The `onPoll` lambda in `OverlayService` delegates to the existing `MediaChangeDispatcher.onMediaChanged` and `ThumbnailChangeDispatcher.onThumbnailChanged`, so `IS_PENDING` retry, `SessionTracker.addMedia` deduplication, and the 500 ms retry remain the single source of truth. A fast observer callback overlapping with a poll tick is a no-op duplicate.
- Lifecycle wiring is consistent: poll starts in every place the media observer is registered (`showOverlay` lock branch, `onScreenOff`, `onOverlayFocusGained` lock branch) and stops in every teardown path (`hideOverlayAndCleanup`, `onUserPresent`, `reset`, `onDestroy`). The belt-and-suspenders `mediaPoller.stop()` in `onDestroy` after `logic.reset()` is harmless given the idempotency contract.
- Poll runs only while a secure session is active, so unlocked-device behaviour is unchanged — addresses the "don't add work where the observer is reliable" concern.

### Test coverage

- 10 new `MediaPollerTest` cases cover every state transition (start/stop idempotency, re-schedule, stop-from-inside, exception swallowing, multiple ticks).
- 6 new `OverlayServiceLogicTest` cases cover every poll start/stop edge: lock-screen activation, deactivation, unlocked-no-poll, focus cycle on locked device, `reset`, gallery-launch immediate hide.
- All 54 pre-existing `OverlayServiceLogicTest` cases continue to pass — confirmed locally; default no-op lambdas keep older test sites compiling unchanged.
- `./gradlew test` green locally (build successful, no failures across the suite).

### Commit hygiene

Excellent split:
1. `8c3244a` — pure addition of `MediaPoller` + constant + 10 tests, no wiring, no behaviour change to existing code.
2. `2b75b50` — wires `MediaPoller` into `OverlayService` / `OverlayServiceLogic`, adds 6 integration tests.

Each commit is independently reviewable, which is exactly what the project's code-edit guidelines require.

### Minor observations (non-blocking)

- `MediaPoller.scheduleNext` catches a bare `Exception`. The comment explains the intent (a thrown query must not abort the loop), but it will also swallow programming errors. Acceptable for a safety-net loop where reliability outranks fail-fast, but worth noting.
- `MediaPoller.isRunning` is exposed as a public getter and only consumed by tests. Could be `internal` if the project ever tightens visibility, but this is style-only.
- `overlayActiveTimestamp` in `OverlayService` is set on thumbnail-observer registration and never cleared. The poller is gated on `overlayActiveTimestamp != 0L`, so once any session has run, that branch always executes when the poller is running. In the current lifecycle this is fine because the poller only runs while overlay + session are both active, but if the wiring ever drifts that gate becomes stale. Worth a follow-up only if related logic changes.
- The `[ ]` manual lock-screen verification box is the only unchecked item — appropriate for a fix that ultimately needs Pixel + lock-screen confirmation; the unit-test layer can't validate the supposed observer-suppression behaviour itself.

No changes requested. From this Reviewer's perspective the PR is ready to merge once manual lock-screen verification on a real Pixel device confirms the symptom is gone.

---

PR #107, opened 2026-05-05
Accept ACTIVITY_RESUMED events in ForegroundDetector to fix launcher false-positive

### LINKED ISSUE #86: FG detector is wrong

aunger:
On Android 10+ (API 29+), `UsageStatsManager` emits `ACTIVITY_RESUMED` (event type 7) for foreground transitions instead of the deprecated `MOVE_TO_FOREGROUND` (event type 1). `ForegroundDetector.getForegroundPackage()` only checked for `MOVE_TO_FOREGROUND`, so on modern Pixel devices it fell back to the launcher's stale event — returning `com.android.launcher3` instead of the camera package, preventing the overlay from ever appearing.

Fix: accept both `ACTIVITY_RESUMED` and `MOVE_TO_FOREGROUND` in the event-type check.

Fixes #86

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q12ei5cVntDTiYSL9odjby)_

### DIFF
--- a/app/src/main/java/com/gb4pc/Constants.kt
+++ b/app/src/main/java/com/gb4pc/Constants.kt
@@ -49,4 +49,9 @@ object Constants {
 
     // Snackbar undo timeout
     const val UNDO_TIMEOUT_MS = 5000L
+
+    // Overlay button shape (Issue #39)
+    // Corner radius as a fraction of the button's width, producing a squircle shape
+    // (30% matches Pixel Camera's rounded-square style).
+    const val SQUIRCLE_CORNER_RADIUS_FRACTION = 0.30f
 }
--- a/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
+++ b/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import android.content.Intent
 import android.content.pm.PackageManager
 import android.graphics.PixelFormat
+import android.graphics.drawable.AdaptiveIconDrawable
 import android.graphics.drawable.Drawable
 import android.graphics.drawable.LayerDrawable
 import android.os.Build
@@ -14,6 +15,7 @@ import android.view.WindowManager
 import android.widget.ImageView
 import android.widget.Toast
 import androidx.core.content.ContextCompat
+import com.gb4pc.Constants
 import com.gb4pc.R
 import com.gb4pc.data.AspectRatioUtil
 import com.gb4pc.data.PrefsManager
@@ -110,6 +112,10 @@ class OverlayManager(
             bitmap?.let { bmp ->
                 android.os.Handler(android.os.Looper.getMainLooper()).post {
                     targetView.setImageBitmap(bmp)
+                    // Issue #39: invalidate the outline so clipToOutline re-applies the squircle
+                    // clip after the image changes. Some devices defer clip invalidation until the
+                    // next outline invalidation, causing the old (circular) mask to persist.
+                    targetView.invalidateOutline()
                 }
             }
         }.start()
@@ -182,7 +188,7 @@ class OverlayManager(
         imageView.outlineProvider = object : android.view.ViewOutlineProvider() {
             override fun getOutline(view: android.view.View, outline: android.graphics.Outline) {
                 if (view.width == 0 || view.height == 0) return
-                val radius = view.width * 0.30f
+                val radius = view.width * Constants.SQUIRCLE_CORNER_RADIUS_FRACTION
                 outline.setRoundRect(0, 0, view.width, view.height, radius)
             }
         }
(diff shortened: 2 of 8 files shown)


aunger (commented):
The fix itself is correct and the test coverage is solid — the five new tests cover the key scenarios well (basic detection, self-filter, mixed event types, and the real-world launcher/camera race). The PR title, description, and "Fixes #86" are all in order.

One rule violation requires a change before this can merge:

---

**Commit structure: single commit bundles two separate concerns (code_edit.md)**

The sole commit mixes two independent changes:

1. **The behavioral fix** — accepting `ACTIVITY_RESUMED` alongside `MOVE_TO_FOREGROUND` in the event-type check. This is the fix for Issue #86.
2. **New `DebugLog.log()` calls** — two additional log statements added inside the same `if (isForegroundEvent)` block (one for the self-skip path, one for the accepted-event path). These are additive instrumentation; they do not change the detection behavior at all.

`code_edit.md` requires commits to be organized by concern, not by file. These two concerns happen to land in the same file, but they are distinct: one fixes the bug, the other adds observability. They must be in separate commits.

Please split the single commit into:
- One commit containing only the `ACTIVITY_RESUMED` detection change (and its accompanying tests).
- A second commit containing only the new `DebugLog.log()` statements.

The order of the two commits is up to you.

---

No other issues found. The implementation is clean, the `@Suppress("DEPRECATION")` annotation is correctly scoped to the method, the updated KDoc accurately describes both event types, and no existing tests were removed or disabled.

---

aunger:
The commit has been split as requested.

The branch now has two commits in place of the original one:

1. `7e67e5a` — **Accept ACTIVITY_RESUMED events in ForegroundDetector to fix Issue #86** — the behavioral fix: the `isForegroundEvent` logic change, `@Suppress("DEPRECATION")` annotation, updated KDoc/comments, and all five new tests.
2. `fd88e35` — **Add debug logging for foreground event filtering in ForegroundDetector** — the two new `DebugLog.log()` calls only; no logic change.

The final tree state is identical to the original single commit. `./gradlew test` passes (BUILD SUCCESSFUL).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q12ei5cVntDTiYSL9odjby)_

---

aunger:
The split looks correct. I verified both commits individually:

**Commit 1 (`7e67e5a`)** touches two files: `ForegroundDetector.kt` (+14/-3: KDoc, `@Suppress("DEPRECATION")`, `isForegroundEvent` local variable, inline comments — zero `DebugLog` lines) and `ForegroundDetectorSelfFilterTest.kt` (+72/-1: five new tests plus the `@Suppress` comment update). All behavioral changes and tests are here and nowhere else.

**Commit 2 (`fd88e35`)** touches only `ForegroundDetector.kt` (+2/0): exactly the two `DebugLog.log()` calls inside the `if (isForegroundEvent)` branch, with no logic change.

The combined addition/deletion counts (+86/-4 + +2/-0 = +88/-4) match the PR diff exactly, confirming the final tree state is unchanged.

The code itself is fine: accepting both `ACTIVITY_RESUMED` and `MOVE_TO_FOREGROUND`, applying the self-filter to both, the `@Suppress("DEPRECATION")` annotation in the right place, and the five tests covering the real-world mixed-event scenario. No issues found on second look.

This PR is acceptable to merge.

---

PR #108, opened 2026-05-06
Refactor: inline BootReceiverLogic into BootReceiver

aunger:
**No functional change.** This is a pure refactor — `BootReceiver` makes the same SharedPreferences read, decides the same way, and starts (or doesn't start) `OverlayService` in the same situations as before. Tests are reorganized to drive the receiver directly instead of the helper.

## Summary

`BootReceiverLogic` was a tiny helper object with three one-line functions:

- `shouldStartService(context)` — a one-line `SharedPreferences.getBoolean` read.
- `isBootIntent(action)` — `action == ACTION_BOOT_COMPLETED`.
- `isInstallOrUpdateIntent(action)` — `action == ACTION_MY_PACKAGE_REPLACED`.

The split existed only to make `BroadcastReceiver` testable. But the receiver
is already testable directly by mocking `Context`, `Intent`, and
`SharedPreferences`, so the indirection was a moving part with no remaining
job.

This PR inlines the three functions into `BootReceiver.onReceive()` and
deletes `BootReceiverLogic.kt`. `BootReceiverTest` is rewritten to drive
`BootReceiver.onReceive()` directly and assert that
`Context.startForegroundService` is (or isn't) called for each combination
of action and pref state — a stronger assertion than the previous tests,
which only verified the helper functions in isolation.

Net change: −1 file, +1 fewer indirection, equivalent behavior, equivalent
or better test coverage.

This is part of a series of complexity-reduction PRs (A1 of A1–A5, B1, B2,
B4). Each is independent and can be reviewed/merged in any order.

## Test plan

- [x] CI green


aunger:
**Reviewer comment**

Clean inline. The helper had no production callers outside `BootReceiver`, and the rewritten test drives `onReceive` directly and verifies `startForegroundService` is/isn't called per (action, pref) combination — a stronger assertion than the previous per-helper-function unit tests.

Verified `BootReceiverLogic` has no other references on `main`. Behavior is byte-identical.

Nothing blocking from my side. LGTM.

---

aunger:
Thanks for the review.

---

PR #109, opened 2026-05-06
Refactor: inline OverlayPositionStore into PrefsManager

aunger:
**No functional change.** This is a pure refactor — the same JSON is written to the same SharedPreferences key on save, and the same JSON is parsed back on load. Default-fallback, exact-match, and closest-ratio (PS-04) behaviors are unchanged.

## Summary

`OverlayPositionStore` was a singleton holding two pure JSON helpers
(`toJson` / `fromJson`) called only from `PrefsManager`. The store added a
file and an indirection without earning its keep — it was effectively a
private implementation detail of `PrefsManager`.

This PR inlines `toJson` / `fromJson` as private file-level functions in
`PrefsManager.kt` and deletes `OverlayPositionStore.kt`.

`OverlayPositionStoreTest` is replaced by two new `PrefsManagerTest` cases
that exercise the same code paths through `PrefsManager`'s public API:

- **`saveOverlayPosition then getOverlayPosition roundtrips multiple ratios`**
  captures the JSON written by `saveOverlayPosition` via Mockito's
  `argumentCaptor`, feeds it back through a fresh `PrefsManager`, and
  asserts every field of every position survives the round-trip — covering
  what the previous `serialize and deserialize roundtrip` test did, plus
  the integration with `getOverlayPosition`'s closest-ratio fallback
  (PS-04).
- **`getOverlayPosition returns default when stored json is malformed`**
  covers the exception-recovery branch that `fromJson returns empty map
  for invalid json` previously tested in isolation.

Net change: −1 file, −1 indirection, equivalent behavior, equivalent or
better test coverage. Part of a series of complexity-reduction PRs.

## Test plan

- [x] CI green


aunger:
**Reviewer comment**

The serialize/deserialize bodies move verbatim into `PrefsManager.kt` as file-level `private` functions, so behavior is unchanged. Defaulting/closest-ratio fallback in `getOverlayPosition` was already exercised by existing `PrefsManagerTest` cases and is unaffected.

A couple of observations, none blocking:

1. `saveOverlayPosition then getOverlayPosition roundtrips multiple ratios` is fairly involved (capture JSON → swap getString stub → reconstruct `PrefsManager` → save again → capture again → reconstruct again). It does verify the full save-then-load loop end-to-end, but the previous `OverlayPositionStoreTest.serialize and deserialize roundtrip` covered the same ser/de path with much less ceremony. This is a tradeoff (broader integration coverage at higher test-reading cost) rather than a regression — flagging in case you want to keep a thinner ser/de assertion alongside the integration one.

2. The previous `fromJson returns empty map for empty string` case is no longer directly exercised. In practice it's hit any time `PrefsManager` is constructed with no prior overlay-positions write, which existing tests already cover via `getOverlayPosition` defaults, so I don't think it's worth a dedicated test, but worth confirming the empty-string branch is still considered covered.

Otherwise LGTM.

---

aunger:
Thanks — pushing back on both, neither warrants a change:

**1. Roundtrip-test ceremony.** The complexity is intrinsic to the mock setup, not the assertion. `SharedPreferences` is mocked, so `editor.putString(...)` doesn't actually persist — to read back what was written, you have to capture it via Mockito's `argumentCaptor` and stub the corresponding `getString` return value. Without that pipework you're not testing serialize-then-deserialize at all, you're testing "`putString` was called and `getString` returns whatever you stubbed." The previous `OverlayPositionStoreTest.serialize and deserialize roundtrip` was simpler only because it called `toJson(...)` and `fromJson(...)` directly with no Android plumbing in between, which is no longer possible after the inline. A thinner ser/de assertion at the `PrefsManagerTest` level would just be the same plumbing with fewer round-trips, which would make it strictly weaker, not simpler. So I'd rather pay the ceremony cost once than add a thinner duplicate.

**2. Empty-string branch.** It's covered. The `setUp()` block stubs `getString(eq(PREF_OVERLAY_POSITIONS), anyOrNull())` to return `""`, and `getOverlayPosition returns default when no stored positions` exercises that branch end-to-end (empty string → `positionsFromJson` → `emptyMap()` → fallback to defaults). No additional test needed.

---

aunger:
**Reviewer comment**

Both responses convincing — accepting both. On (1), you're right that the `argumentCaptor` ceremony is intrinsic to the fact that `SharedPreferences` is mocked: a thinner test at this level wouldn't actually exercise ser/de, just the put/get stubs. On (2), confirmed `getOverlayPosition returns default when no stored positions` (line 68) reads the `""` value stubbed in `setUp()`, which is exactly the `if (json.isBlank()) return emptyMap()` branch, so the empty-string path is covered.

LGTM.

---

PR #110, opened 2026-05-06
Refactor: inline OverlayPositionCalculator into OverlayManager

aunger:
**No functional change.** This is a pure refactor — the three pixel-conversion formulas are unchanged byte-for-byte; only their location moves.

## Summary

`OverlayPositionCalculator` was a singleton holding three pure pixel
conversions (`calculateSizePx` / `calculateXPx` / `calculateYPx`) called
only from `OverlayManager.createLayoutParams()`. The wrapper added a file
and a type-name to navigate without earning its keep.

This PR moves the three functions as `internal` top-level functions in
`OverlayManager.kt` (renamed `calculateOverlaySizePx` /
`calculateOverlayXPx` / `calculateOverlayYPx` since the `OverlayPosition`
prefix is no longer redundant) and deletes
`OverlayPositionCalculator.kt`.

`OverlayManagerTest` is unchanged in structure — it just calls the
functions directly instead of through the calculator namespace. They
remain pure and testable in plain JVM unit tests.

Net change: −1 file, −1 indirection, equivalent behavior, equivalent test
coverage. Part of a series of complexity-reduction PRs.

## Test plan

- [x] CI green


aunger:
**Reviewer comment**

Three pure functions with one caller and one test class — the wrapper object was clearly carrying its weight thinly, and lifting them to `internal` top-level functions in `OverlayManager.kt` is appropriate. Logic moves byte-for-byte, tests are mechanically renamed.

One small note: `internal` top-level functions are visible to any sibling file in the module, where the previous `OverlayPositionCalculator.calculateXPx(...)` namespace forced a deliberate import. The renamed `calculateOverlayXPx` etc. read fine standalone, so I don't think this is a real concern, just calling it out.

LGTM.

---

aunger:
Acknowledged on the visibility note. Agree it's not a real concern — the renamed `calculateOverlay*` names read fine standalone and won't be mistaken for member functions of some other class.

---

PR #112, opened 2026-05-06
Refactor: replace SetupState class with a local Int step index in SetupActivity

aunger:
**No functional change.** This is a pure refactor — the setup flow's auto-advance, skip, finish-on-complete, and step-to-intent behaviors are byte-identical to before. Only the wrapper class around the index counter is gone.

## Summary

`SetupState` wrapped a single `Int currentIndex`, a derived `isCompleted`
flag, and an `advance()` method. Its only non-trivial logic was the
API-level-aware list of steps — the rest was a class wrapper around
`++`.

This PR:

- Replaces the `SetupState` class with a top-level `fun
  getSetupSteps(apiLevel: Int)` in the same file. `SetupStep` enum is
  preserved.
- Inlines the index counter into `SetupActivity` as a private
  `var currentIndex`, with `isCompleted` and `currentStep` as derived
  read-only properties. All existing flows (auto-advance on grant, skip
  button, finish-on-complete) behave identically.
- Trims `SetupStateTest` to four assertions on `getSetupSteps`'s actual
  output. The previous "advance moves to next step" / "advance past last
  step marks completed" tests verified Kotlin's `++` operator and bounds
  check, not behavior worth retesting after the wrapper is gone. The
  step-list tests (which were the actual non-trivial logic) are kept and
  slightly reorganized.

Net change: −1 class, equivalent behavior, equivalent or better-targeted
test coverage. Part of complexity-reduction series.

## Test plan

- [x] `grep -rn "SetupState\b"` returns zero matches
- [ ] CI green
- [ ] (Recommended) On-device smoke of the setup flow: fresh install,
      walk through each permission step, confirm auto-advance and skip
      both work and the activity finishes correctly when complete.


aunger:
**Reviewer comment**

The `SetupState` wrapper genuinely was a thin shell over `Int++`, and `getSetupSteps(apiLevel)` is the part with real logic — keeping that as a top-level function reads well.

One subtle behavior difference worth flagging: in the previous `SetupState.advance()`, the index never advanced past `steps.size - 1` (it set `isCompleted = true` and stopped). So `currentStep` was a safe read at all times. In the new code, `currentIndex++` is unconditional, so once `isCompleted` is true, `currentIndex == steps.size` and any access to `currentStep` would throw `IndexOutOfBoundsException`.

All current call sites in `SetupActivity` correctly guard `currentStep` access with `!isCompleted` (the `autoAdvanceIfGranted` while-condition, the early-return in `updateUI`, the `else`-branch in `onSkipClick`), so behavior is preserved today. But the invariant "`currentStep` is always valid" — which the previous class enforced — is gone, and a future caller that reads `currentStep` after `currentIndex++` without re-checking `isCompleted` would crash.

Not blocking, but worth deciding consciously: either keep `currentIndex` clamped (e.g. only advance while `< steps.size`) so `currentStep` stays safe, or accept that the safety is now the caller's responsibility and trust the small surface area.

Otherwise LGTM. The trimmed test class is correctly focused on the only non-trivial logic (the API-conditional step list); the deleted "advance moves to next step" / "advance past last step marks completed" cases were indeed verifying `++` and a bounds check, which aren't worth retesting.

---

aunger:
Good catch — you're right, I dropped the invariant. The previous `SetupState.advance()` clamped at `steps.size - 1` and flipped `isCompleted = true` instead of running off the end, so `currentStep` was always a safe read. After my refactor, the unconditional `currentIndex++` would run past the array and any read of `currentStep` while completed would IOOBE. Today's call sites all guard with `!isCompleted`, so it's not a live bug, but losing the invariant is a regression in robustness even if behaviour is preserved.

Fixed in 76b80cd by adding a private `advanceStep()` helper that mirrors the previous clamp:

```kotlin
private fun advanceStep() {
    if (currentIndex < steps.size - 1) {
        currentIndex++
    } else {
        isCompleted = true
    }
}
```

`autoAdvanceIfGranted` and the skip-button callback both call `advanceStep()` instead of `currentIndex++`. `currentStep` is now safe to read at any point.

---

aunger:
**Reviewer comment**

Verified the fix in 76b80cd: `advanceStep()` clamps via `if (currentIndex < steps.size - 1) currentIndex++ else isCompleted = true`, both call sites use it instead of `currentIndex++`, and `isCompleted` is now an actual `var` set on final advance rather than a derived `currentIndex >= steps.size`. The "currentStep is always a safe read" invariant is restored exactly as the previous `SetupState.advance()` enforced it.

LGTM.

---

PR #114, opened 2026-05-06
Refactor: extract MediaDeletionManager from SecureViewerActivity

aunger:
**No functional change.** This is a pure refactor — the API 26–28, API 29 (RecoverableSecurityException), and API 30+ (createDeleteRequest) delete paths are byte-identical to the original; only their location moves out of the activity. The launcher-result retry behaves exactly as before.

## Summary

`SecureViewerActivity` carried ~70 lines of API-version-conditional delete
handling — `RecoverableSecurityException` (API 29),
`MediaStore.createDeleteRequest` (API 30+), manual `ContentResolver.delete`
(API 26–28), the `pendingDeleteUri` state, and the `ActivityResultLauncher`
result-callback retry. That logic had nothing to do with the activity's
other concerns (ViewPager wiring, swipe-to-delete UX, snackbar/undo, share
flow), and intermingling them made the activity harder to read and the
version dispatch impossible to unit-test in isolation.

This PR moves the version-conditional delete code into a new
`MediaDeletionManager`:

```
delete(uri: Uri)
    — picks the right API path
onDeleteRequestResult(resultOk: Boolean)
    — called by the activity from its ActivityResultLauncher callback;
      retries the delete on a successful permission grant
```

The activity keeps the launcher (since `registerForActivityResult` must
run before `STARTED`) and forwards the result to the manager. The manager
takes `apiLevel: Int` as an injected parameter so the API 26–28 and API 29
paths can be unit-tested in plain JVM.

Adds `MediaDeletionManagerTest` covering:

- API 26: success / zero-row / generic-exception paths
- API 29: `RecoverableSecurityException` → launches `IntentSender`
- `onDeleteRequestResult(true)` retries the delete; `(false)` does not
- `onDeleteRequestResult` with no pending URI is a no-op

The API 30+ `createDeleteRequest` path uses a static `MediaStore` call
that can't be stubbed in plain JVM; the existing emulator-based E2E suite
covers it. The PR documents this in the test class doc comment.

**Net change:** activity drops ~70 lines (445 → 386); deletion manager is
99 lines including KDoc; new test adds 8 cases. Behaviour is unchanged.

Part of complexity-reduction series.

## Test plan

- [x] CI green
- [ ] (Recommended) On-device verification: take photos on a locked
      Pixel, swipe one to delete, confirm the snackbar, let it expire,
      and confirm the photo is actually deleted from MediaStore. Repeat
      with the undo button. Repeat with a photo not owned by the app
      (should show the system permission dialog and proceed correctly).


aunger:
**Reviewer comment**

Activity-vs-manager split is clean. The four delete code paths (API 26–28 direct delete; API 29 `RecoverableSecurityException`; API 30+ `createDeleteRequest`; post-grant retry) move byte-for-byte into `MediaDeletionManager`. The activity keeps the launcher (necessarily — `registerForActivityResult` must run before `STARTED`) and forwards the boolean result, which is the right seam.

A small behavior nuance worth noting: in the original `retryDeleteFromMediaStore`, the `Toast.makeText(...)` calls in the failure branches were **not** wrapped in `runOnUiThread` (unlike `deleteFromMediaStore`, where they were). In the new code, the activity's `onFailure` lambda always uses `runOnUiThread`, so the retry path is now consistently dispatched to the main thread. In practice the retry runs on the launcher callback, which is already on the main thread, so this is a thread-safety nicety rather than a fix — but it's a minor improvement, not a regression.

Test coverage is appropriately scoped: API 26/29 paths and the result-callback retry are unit-tested; the API 30+ static `MediaStore.createDeleteRequest` path is acknowledged as not stub-able in plain JVM and left to E2E — that's a sensible call. The recoverable-exception test mocks the security-exception with a chained `RemoteAction`/`PendingIntent` mock, which is a fair JVM-level approximation of the real path.

LGTM.

---

aunger:
Acknowledged on the `runOnUiThread` consistency observation — agree it's a nicety rather than a fix (the launcher callback runs on the main thread already), but I'd rather the two failure paths look the same than have the original asymmetry preserved as a curiosity. Happy to leave the consolidated `runOnUiThread` in place.

---

PR #115, opened 2026-05-06
Refactor: extract E2EFixture for shared E2E test setup

aunger:
**No functional change.** Test-code-only refactor — the existing test assertions and the steps performed in `@Before` are byte-identical to the previous inline implementation. No production code changes.

## Summary

Every E2E test (currently one — `PixelCameraOverlayE2ETest` — but more
are likely as #15 progresses) needs the same boilerplate: verify Pixel
Camera is installed, mark setup complete, enable + start the
`OverlayService`, wait for callbacks to register, force-stop PC. The
existing test class also defines `stopPC()` and `waitForCondition()`
helpers that any sibling E2E test would re-implement.

This PR introduces `E2EFixture`, which centralises:

- the four-step `preconditionCheck` (now `setUp()`)
- the three shell-command helpers: `launchPixelCamera`, `goHome`,
  `stopPixelCamera`
- the polling helper `waitForCondition`

`PixelCameraOverlayE2ETest` is rewritten to call into the fixture from
`@Before` and use its helpers in the test methods. The class shrinks
from 163 lines to 105, and adding a second E2E test class is now
roughly:

```kotlin
private val fixture = E2EFixture(
    context = instrumentation.targetContext,
    uiAutomation = instrumentation.uiAutomation,
)
@Before fun setUp() = fixture.setUp()
```

No production code changes. Behaviour and test assertions are
identical. Part of complexity-reduction series.

## Test plan

- [x] `./gradlew :app:assembleDebugAndroidTest` succeeds (verified
      locally) — confirms the refactored sources compile.
- [x] `./gradlew :app:testDebugUnitTest` passes — confirms no
      cross-source-set regressions.
- [x] `./gradlew :app:assembleDebug` builds.
- [ ] CI green
- [ ] (Recommended) `./gradlew :app:connectedE2EAndroidTest` on a
      configured emulator — confirms the fixture's `setUp()` produces
      the same starting state as the previous inline `preconditionCheck`.


aunger:
**Reviewer comment**

The fixture move is mechanical and correct: `setUp()` performs the same four steps in the same order with the same sleeps, and the helpers (`launchPixelCamera`, `goHome`, `stopPixelCamera`, `waitForCondition`) are byte-identical to the inline versions.

One concern worth raising explicitly: as of this PR, `E2EFixture` has exactly one caller. Per `agents/code_edit.md` ("Don't add features, refactor, or introduce abstractions beyond what the task requires"), pulling out a fixture before the second test class exists is preparatory abstraction — the value is paid forward.

That can absolutely be the right call here (the description ties this to #15 adding more E2E classes), but the rule of thumb in the project's guidance is to extract on the second use, not the first. If #15 is genuinely imminent, I'd be fine merging this as-is. If #15 is more speculative, you might consider deferring the extraction until the second E2E class lands — at which point the actual shared shape is concrete rather than guessed.

No correctness concerns with what's here; flagging the timing question for you to decide.

---

aunger:
Fair principled concern, and you're right that `agents/code_edit.md`'s "extract on the second use" rule of thumb says wait. Pushing back on this one and proposing we merge as-is, with reasoning:

1. **The boundary is small and concrete, not speculative.** The fixture wraps four already-named, already-isolated helpers (`stopPC`, `waitForCondition`, the `@Before` precondition block, and the shell-command builders that were inline string literals). The "second use" rule guards against extracting an abstraction whose shape is guessed; here the shape is exactly the shape the existing class already has, just lifted to a sibling file. There's no design risk that the second caller will need a different signature.

2. **The single-caller class still benefits.** `setUp()` now reads as a single line of intent (`fixture.setUp()`) rather than 25 lines of mixed concerns (precondition checks, prefs setup, sleeps, shell commands). Even with one caller, the `@Before` block reads more cleanly post-extract than pre-extract.

3. **Reverting is more churn than keeping.** This PR was already approved in the planning round and is one of eight in a series. Closing it would mean inverting an approved decision based on a guideline that the reviewer themselves notes is "preparatory" not "incorrect."

If you'd rather defer until #15 lands a second E2E class, I'll close the PR — but my read of the tradeoff is that the fixture earns its keep today.

---

aunger:
**Reviewer comment**

Fair points — accepting. Your point (1) is the strongest: the abstraction shape is concrete (the existing `@Before` body already isolated the four steps), so the design risk that "extract on the second use" guards against doesn't really apply here. Point (2) is also reasonable; the `@Before` body really does read more cleanly post-extract even at one caller.

LGTM.

---

PR #121, opened 2026-05-13
Update dev_orchestration.md for clarity and instructions

aunger:
Clarify instructions for replacing subagents and creating branches.

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -20,10 +20,11 @@ The Orchestrator is not a Reviewer or a Programmer.
 
 **May:**
 - Read issues, PRs, and the comments on either via GitHub MCP tools
+- Create local Git branches to keep tasks separate
 - Read project instructions (AGENTS.md and the files it references)
 - Dispatch and communicate with subagents
-    - Replace subagents when necessary to complete a workflow
-    - Inform them of unfinished tasks or additional responsibilities
+    - Replace subagents, reluctantly and when necessary, to complete a workflow
+    - Inform subagents of unfinished tasks or additional responsibilities
 - Relay subagent results to the user
 
 ## Inaugurating work for a hitherto unworked issue
@@ -35,7 +36,8 @@ The Orchestrator is not a Reviewer or a Programmer.
 - Create a Sonnet sub-agent unless the user requested otherwise
 - Inform the agent of its role as an expert software developer resolving the issue
 - Inform the agent of its responsibility to commit its work to a branch and open a PR (if one doesn't already exist)
-- Specify a dedicated per-issue branch name for the Programmer to use. Branch names should follow the pattern `fix/issue-N-short-description` for bug fixes or `feature/issue-N-short-description` for new features. Never direct two Programmers for unrelated issues to the same branch.
+- *Create a dedicated per-issue branch* for the Programmer to use. Branch names should follow the pattern `fix/issue-N-short-description` for bug fixes or `feature/issue-N-short-description` for new features. Never direct two Programmers for unrelated issues to the same branch.
+- Pass the branch name to the subagent
 - Pass the issue number to the subagent
 - Relay any relevant instruction from the user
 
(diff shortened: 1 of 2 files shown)


aunger:
**Reviewer comment**

A docs-only PR clarifying the orchestration and reviewer roles. Most of the edits are reasonable polish (the "(unless otherwise instructed)" flexibility, the disambiguated "subagents" pronoun, and granting the Orchestrator explicit branch-creation authority that the role's new responsibility — "Create a dedicated per-issue branch" — demands). A few substantive things worth pushing back on:

**1. `pr_participation.md`: poll every 10 seconds + update own status every 10 seconds**

> If you must wait, poll the status every 10 seconds. Do not rely on the flaky CI event hooks. Update your own status every 10 seconds, so the Orchestrator doesn't think you're done.

This conflicts directly with the harness's event-driven model. The Claude Code system prompt is explicit about this: "If waiting for a background task… you will be notified when it completes — do not poll." Subscriptions via `subscribe_pr_activity` exist precisely so the session can sleep until events arrive.

The claim that the CI event hooks are "flaky" doesn't match my direct experience from the prior eight-PR review pass earlier this week: every CI-completion and PR-merge event arrived reliably and woke the session as designed. If you've seen specific events get dropped, that's worth filing separately, but a blanket "do not rely on" rule based on it feels strong.

Also, "Update your own status every 10 seconds" — to whom, and where? `TodoWrite` updates an in-conversation list the user already sees; posting a 10-second comment cadence to the PR would be both noisy and against the existing rule in the system context: "Be frugal about posting replies on GitHub." If the intent is "produce a heartbeat token so the Orchestrator's idleness check doesn't kill the Reviewer," that mechanism deserves to be named explicitly rather than left to interpretation.

**2. `dev_orchestration.md`: "pause 45 seconds, then evaluate"**

> If a system hook or event signals uncommitted work, a test failure, or an error, pause 45 seconds, then evaluate whether the agent or CI system is still actively working.

Two concerns:

(a) *Mechanism for the pause.* There's no idiomatic 45-second sleep in this harness — `Bash` `sleep` is discouraged, leading sleeps are blocked, and the natural waiting primitive is event subscription. If the intent is "don't react instantly to a single hook event, give the worker time to recover," that could be phrased as a heuristic ("wait for a follow-up event from the same agent" or "wait for the next CI status change") rather than a specific second count.

(b) *Evaluation criteria.* "Whether the agent or CI system is still actively working," "recently active," "idle or waiting for input" — the Orchestrator doesn't have a direct read on subagent state. It can see when a subagent finishes (because it returns) but not whether one is mid-tool-call. So "evaluate whether actively working" doesn't map to an observable, and the result will be inconsistent across orchestrator agents.

**3. `dev_orchestration.md`: loss of a concrete failure signal**

> Old: If no PR was opened, then the Programmer did not finish, even if it claimed otherwise. Assign another to finish the job.
> New: If an agent has failed to complete its work, and shows no sign of continued effort, assign a replacement to finish the job.

The old version had a concrete, falsifiable signal ("no PR was opened"). The new version uses "failed to complete" + "no sign of continued effort," both of which are subjective. The replacement reads as strictly weaker guidance. Consider keeping the explicit signal as one example/criterion of "failed to complete," rather than dropping it.

**4. Minor: "reluctantly"**

> Replace subagents, reluctantly and when necessary, to complete a workflow

"Reluctantly" is unusually emotional for a process doc. "When necessary" already conveys the threshold. (Not blocking — just noting that the doc otherwise reads in a fairly clinical voice.)

**Note on CI gating this review**

The new rule in `pr_participation.md` says I should wait for CI green before approving. CI is currently in progress on this PR (build-and-test and Analyze java-kotlin both in_progress). Since the review here recommends changes rather than approving, I'm sending this now rather than waiting — that's consistent with the new rule's "You need not wait if other required changes are outstanding."

---

aunger:
@claude You had a good run with those PRs, but my experiences skew very differently:
- CI-triggered resumption happens half of the time
- Orchestrators repeatedly (more than half of the time) launched “cleanup” agents the moment an Author wrote an uncommitted file or an unpushed commit, so the Author’s in-progress work was deleted, pushed, or taken over by a replacement Author.

I appreciate how Claude is *intended* to work, but the harness is a mess.

PR #129, opened 2026-05-14
Remove always-failing PIL diagnostic from CI pre-flight failure path

### LINKED ISSUE #124: CI pre-flight: PIL diagnostic in failure path always silently no-ops
## Context

When `check_green_feed.py` fails, the pre-flight step runs a secondary diagnostic:

```bash
python3 -c "from PIL import Image; from collections import Counter; ..." 2>/dev/null || echo "(PIL not available)"
```

## Problem

`Pillow` is not installed on the GitHub Actions `ubuntu-*` runners by default. This block will always print `(PIL not available)` and produce no additional information. The appearance of a useful diagnostic that silently does nothing adds noise without value.

Meanwhile, `check_green_feed.py` already prints the dominant color of the sampled region on failure:

```
FAIL: coverage 12.3% < 90%. Actual dominant color: #1A1A1A (R=26, G=26, B=26)
```

## What to do

Remove the PIL diagnostic block entirely. The existing output from `check_green_feed.py` is sufficient. If richer diagnostics are needed in the future, either install `Pillow` explicitly (`pip install Pillow`) before the step, or extend `check_green_feed.py` itself (which is pure stdlib and already present).

aunger:
## Summary

- Removes the `python3 -c "from PIL import Image ..."` block from the CI pre-flight failure path in `.github/workflows/build.yml`.
- Pillow is not installed on GitHub Actions `ubuntu-*` runners, so this block always fell through to `|| echo "(PIL not available)"`, printing noise with zero diagnostic value.
- `check_green_feed.py` already prints the dominant color on failure (e.g. `FAIL: coverage 12.3% < 90%. Actual dominant color: #1A1A1A (R=26, G=26, B=26)`), making the PIL block entirely redundant.

Fixes #124

## Test plan

- [x] Confirmed the YAML file still parses correctly (`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/build.yml'))"` exits 0).
- [x] Diff is a pure 2-line deletion; no logic changes.

https://claude.ai/code/session_01Xn5pP8vxJNsME7vFceawXR

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xn5pP8vxJNsME7vFceawXR)_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -226,8 +226,6 @@ jobs:
 
           echo "==> Running green-feed smoke check..."
           if ! python3 scripts/check_green_feed.py preflight.png; then
-            echo "=== pre-flight failed: check_green_feed dominant color ==="
-            python3 -c "from PIL import Image; from collections import Counter; img = Image.open('preflight.png').convert('RGB'); w, h = img.size; cx, cy = w//2, h//2; region = img.crop((cx-100, cy-100, cx+100, cy+100)); pixels = list(region.getdata()); top = Counter(pixels).most_common(3); print('Top 3 colors in center 200x200:', top)" 2>/dev/null || echo "(PIL not available)"
             exit 1
           fi
 


aunger (commented):
**Review: LGTM — would approve if the API permitted self-review.**

The change does exactly what it claims:

- The diff is a pure 2-line deletion: the PIL diagnostic banner and the always-failing `python3 -c "from PIL import Image..."` one-liner (with its `|| echo "(PIL not available)"` fallback). Nothing else was touched.
- `exit 1` is preserved immediately after the removed block, so the failure path is still correct.
- The surrounding step logic (screenshot trap, install, launch, screencap, green-feed check) is untouched.
- The resulting YAML is structurally valid; the `if/then/fi` block is complete and well-formed.
- `check_green_feed.py` already prints the dominant color on failure (e.g. `FAIL: coverage 12.3% < 90%. Actual dominant color: #1A1A1A (R=26, G=26, B=26)`), so no diagnostic capability is lost by removing the dead PIL block.
- Issue #124 is fully addressed: the always-silent no-op is gone.

**CI status:** All checks completed green — `build-and-test` passed, all three CodeQL analyses passed, `dev-build` correctly skipped (PR-only trigger). No concerns.

---

PR #142, opened 2026-05-15
Skip CI build for documentation-only changes

### LINKED ISSUE #141: Don’t rebuild the app for documentation-only changes

aunger:
Adds a `paths-ignore: ['**/*.md']` filter to both the `push` and `pull_request` triggers in `.github/workflows/build.yml`. With this filter in place, CI is skipped automatically when a commit or PR only touches markdown files, avoiding unnecessary build runs for documentation-only changes.

Fixes #141

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,18 +5,10 @@ on:
     branches: [main]
     paths-ignore:
       - '**/*.md'
-      - 'docs/**'
-      - 'agents/**'
-      - 'AGENTS.md'
-      - 'CLAUDE.md'
   pull_request:
     branches: [main]
     paths-ignore:
       - '**/*.md'
-      - 'docs/**'
-      - 'agents/**'
-      - 'AGENTS.md'
-      - 'CLAUDE.md'
 
 permissions:
   contents: read
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,8 +3,20 @@ name: Build & Test
 on:
   push:
     branches: [main]
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+      - 'agents/**'
+      - 'AGENTS.md'
+      - 'CLAUDE.md'
   pull_request:
     branches: [main]
+    paths-ignore:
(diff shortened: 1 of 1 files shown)


aunger (commented):
The `paths-ignore` filters are correctly placed in the `on:` block and cover the right doc-only paths. Mixed doc+code PRs still trigger the build as expected.

CI failure (`build-and-test`) is a pre-existing flaky emulator issue: the green-feed smoke check timed out before `MockCameraActivity` rendered its green view. The run lasted 4m 9s vs. the usual 7–12 minutes for a passing run, and the failure step is unrelated to the `paths-ignore` YAML change. This PR's diff cannot affect job execution logic. LGTM.

---

aunger:
- Why specify individual files? Doesn't `**/*.md`already cover them?
- What are you trying to guard against by excluding the directories? It seems safer not to skip those and add them later if needed.

aunger (commented):
The simplification is correct and an improvement over the first iteration. My analysis:

**Correctness of `**/*.md`**

The glob is accurate. It matches every markdown file at any depth — `AGENTS.md`, `CLAUDE.md`, anything under `docs/` or `agents/` — without needing those directories enumerated separately. Redundant entries in the previous version are gone.

**Removing `docs/**` and `agents/**` is the right call**

The old directory globs would have suppressed CI for any file added to those directories, not just markdown. If someone later drops a script or config file under `agents/`, CI should run. The current filter leaves that protection intact.

**Scope**

`**/*.md` is deliberately minimal and matches the issue exactly. Expanding to `.txt`, images, etc. would be scope creep; those can be addressed separately if they become a concern.

**YAML**

Four lines, symmetric across both triggers, valid glob syntax, correct indentation. Nothing to add.

**One nit: the PR description is stale.** It still lists `docs/**`, `agents/**`, `AGENTS.md`, and `Claude.md` as ignored paths, which are no longer in the diff. Worth updating before merge to avoid misleading future readers.

**Verdict:** The code change itself is correct and I would approve it pending CI passing. The stale description is the only outstanding issue — it should be corrected, but it does not affect the behavior of the workflow.

---

PR #148, opened 2026-05-16
build-and-test: move path filtering to job level so docs-only PRs clear the merge gate

### LINKED ISSUE #147: build-and-test: move path filtering to job level so docs-only PRs still clear the merge gate
## Problem

The `build-and-test` workflow currently uses a workflow-level path filter (e.g. `paths:` or `paths-ignore:` on the `on:` trigger). When a PR changes only `.md` files, the entire workflow is skipped and no status is ever posted to the commit. Because `build-and-test` is a required status check in branch protection, GitHub shows it as "Expected — Waiting for status to be reported" indefinitely, and the PR cannot be merged.

This was observed in practice on PR #146 (a documentation-only PR): all CodeQL checks passed, but `mergeable_state` remained `blocked` solely because `build-and-test` never posted.

## Root cause

Workflow-level path filters prevent the workflow from running at all. GitHub then has no status to evaluate, so the required check never becomes green. The PR is permanently blocked with no actionable error.

## Fix

Convert the path check from workflow-level to job-level:

1. **Remove** the `paths:`/`paths-ignore:` filter from the `on: pull_request:` trigger so the workflow always starts.
2. **Add a job-level step** that checks the git diff directly:
   ```yaml
   - name: Check for non-docs changes
     id: diff_check
     run: |
       git fetch origin ${{ github.base_ref }} --depth=1
       CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -v '\.md$' || true)
       echo "has_code_changes=$([ -n "$CHANGED" ] && echo true || echo false)" >> $GITHUB_OUTPUT
   ```
3. **Gate the build/test steps** on `steps.diff_check.outputs.has_code_changes == 'true'` using `if:` conditions or a separate job dependency.
4. The job always **completes** (posting a status), but skips the expensive build and test steps when only `.md` files changed.

This ensures:
- Docs-only PRs get a green `build-and-test` status immediately without running the emulator.
- Code PRs continue to run the full build and test suite.
- The required status check is always satisfied, unblocking docs-only merges.

## Why not use `paths-ignore` on the job?

Job-level `if:` conditions that reference `github.event.pull_request` path lists are not directly available. Inspecting the diff via `git diff` at runtime is more reliable and does not depend on GitHub's path-filter evaluation, which has edge cases (e.g. first commit to a branch, force-pushes).

aunger:
## Problem

The `build-and-test` workflow used a workflow-level `paths-ignore` filter on the `on: pull_request:` trigger. When a PR changed only `.md` files, GitHub skipped the workflow entirely and never posted a status. Because `build-and-test` is a required status check, GitHub showed it as "Expected — Waiting for status to be reported" indefinitely, blocking the PR from merging.

## Fix

Fixes #147

1. **Removed** the `paths-ignore` filter from both `on: push:` and `on: pull_request:` so the workflow always starts.
2. **Added a `Check for non-docs changes` step** (`id: diff_check`) as the second step in `build-and-test`. It computes a `has_code_changes` output (`true`/`false`) by diffing the PR head against `origin/${{ github.base_ref }}` and checking whether any non-`.md` file changed. Direct pushes to `main` always set `has_code_changes=true` since `github.base_ref` is empty on push events.
3. **Gated every expensive step** (JDK setup, Android SDK, KVM, Gradle cache, emulator install/start/wait, build, unit tests, E2E tests, artifact uploads) on `steps.diff_check.outputs.has_code_changes == 'true'`.

The job always completes and posts a green (or failed) status check, so the GitHub merge gate is satisfied regardless of what files a PR touches. Docs-only PRs see a fast job that skips all build and emulator work.

## Test plan

- [ ] Open a PR that changes only `.md` files — `build-and-test` should start, run only the checkout + diff-check steps, and complete green in seconds.
- [ ] Open a PR that changes a source file — `build-and-test` should run the full build + unit + E2E pipeline as before.
- [ ] Merge to `main` — `build-and-test` always runs the full pipeline (push event path).

https://claude.ai/code/session_01Fg2pLXPoQMHTUifWPEJz3Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fg2pLXPoQMHTUifWPEJz3Y)_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,12 +3,8 @@ name: Build & Test
 on:
   push:
     branches: [main]
-    paths-ignore:
-      - '**/*.md'
   pull_request:
     branches: [main]
-    paths-ignore:
-      - '**/*.md'
 
 permissions:
   contents: read
@@ -36,17 +32,34 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for non-docs changes
+        id: diff_check
+        run: |
+          if [[ "${{ github.event_name }}" == "push" ]]; then
+            # Always build on direct pushes to main (e.g. merge commits).
+            echo "has_code_changes=true" >> $GITHUB_OUTPUT
+          else
+            git fetch origin ${{ github.base_ref }} --depth=1
+            CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -v '\.md$' || true)
+            echo "has_code_changes=$([ -n "$CHANGED" ] && echo true || echo false)" >> $GITHUB_OUTPUT
+          fi
 
       - name: Set up JDK 17
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         uses: actions/setup-java@v5
         with:
           java-version: '17'
           distribution: 'temurin'
 
       - name: Set up Android SDK
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         uses: android-actions/setup-android@v4
 
       - name: Enable KVM
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         run: |
           echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | \
             sudo tee /etc/udev/rules.d/99-kvm4all.rules
@@ -54,6 +67,7 @@ jobs:
           sudo udevadm trigger --name-match=kvm
 
       - name: Cache Gradle packages
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         uses: actions/cache@v5
         with:
           path: |
@@ -64,6 +78,7 @@ jobs:
             ${{ runner.os }}-gradle-
 
       - name: Install emulator, system image, and create AVD
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         run: |
           sdkmanager --install \
             "emulator" \
@@ -90,9 +105,11 @@ jobs:
           ANDROID_AVD_HOME="$AVD_HOME" $ANDROID_HOME/emulator/emulator -list-avds
 
       - name: Grant execute permission for gradlew
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         run: chmod +x gradlew
 
       - name: Build and run unit tests
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         env:
           BUILD_NUMBER: ${{ github.run_number }}
         run: >
@@ -100,7 +117,7 @@ jobs:
           testDebugUnitTest -PversionName=dev.${{ github.run_number }} --no-daemon
 
       - name: Upload unit test results
-        if: always()
+        if: always() && steps.diff_check.outputs.has_code_changes == 'true'
         uses: actions/upload-artifact@v4
         with:
           name: unit-test-results
@@ -108,6 +125,7 @@ jobs:
           retention-days: 14
 
       - name: Start emulator
+        if: steps.diff_check.outputs.has_code_changes == 'true'
         run: |
           echo "=== $(date -u '+%Y-%m-%d %H:%M:%S UTC') === KVM device:"
           ls -la /dev/kvm || echo "WARNING: /dev/kvm not found"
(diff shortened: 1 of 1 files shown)


aunger (commented):
## Review

The overall approach is correct. Removing `paths-ignore` at the trigger level and replacing it with a job-level `diff_check` step is the right solution to the "Expected — Waiting for status to be reported" problem. The three-dot diff semantics, `|| true` on grep, the `always() &&` guards on cleanup steps, the push-event short-circuit, and the `fetch-depth: 0` are all handled correctly.

One issue requires a fix before this merges.

---

### Shell injection: unquoted `${{ github.base_ref }}` in a `run:` script

Lines 45–46 of `.github/workflows/build.yml`:

```yaml
git fetch origin ${{ github.base_ref }} --depth=1
CHANGED=$(git diff --name-only origin/${{ github.base_ref }}...HEAD | grep -v '\.md$' || true)
```

`${{ github.base_ref }}` is interpolated by the Actions template engine directly into the shell source before the shell parses it. There are no surrounding quotes. If the base branch name ever contained shell metacharacters — spaces, `$`, backticks, semicolons — the shell would interpret them, not treat them as a literal branch name. While the current base branch is `main` and GitHub's branch-name rules make exploitation unlikely in practice, this is a well-known injection class and the safe pattern is trivial: assign the context value to a shell variable first, then quote the variable in subsequent uses.

The same pattern on line 41 (`"${{ github.event_name }}"`) is already quoted correctly. The `base_ref` uses on 45–46 need the same treatment.

---

CI note: `build-and-test` was still `in_progress` at the time this review was posted (running the full emulator pipeline, as expected, because `build.yml` is a non-docs file). No failures were observed during polling.


---

aunger:
The shell injection finding is valid — `${{ github.base_ref }}` was indeed substituted literally by the Actions template engine before shell parsing, with no quoting protection.

Fixed in commit 3249cbf: `github.base_ref` is now assigned to `BASE_REF` as the first thing in the `else` branch, and `"$BASE_REF"` is quoted in both the `git fetch` and `git diff` calls. This matches the pattern already used correctly for `github.event_name` on line 41.

---

aunger (commented):
## Review

The approach is sound and the implementation is correct. The shell injection fix (commit `3249cbf`) is properly applied: `${{ github.base_ref }}` is assigned to `BASE_REF` before use, and `"$BASE_REF"` is double-quoted in all subsequent shell expansions. All other logic details remain correct — three-dot diff semantics, `|| true` on `grep`, `always() &&` guards on cleanup/upload steps, push-event short-circuit, `fetch-depth: 0`.

**Verdict: no issues with the PR's changes. Would approve if GitHub permitted self-review.**

---

### `build-and-test` failure: pre-existing emulator flake, not caused by this PR

The failed step is **CI pre-flight — green-feed smoke check** (exit code 1). The unit-test-results artifact was produced (build and unit tests passed); E2E artifacts are missing because the pipeline died at the preflight step before E2E ran.

This PR's changes do not touch the preflight step, the emulator lifecycle steps, or the green-feed smoke check script. The only code changed is the `diff_check` step and the `if:` guards added to each existing step. The workflow executed the full pipeline because `build.yml` itself changed — `diff_check` correctly set `has_code_changes=true` — so the failure is not a gating logic error.

The workflow already documents this kind of emulator fragility (see the inline comment about `sys.boot_completed` binder-thread saturation and exit code 224). Main-branch runs show a consistent green history with no recent failures. This is a known emulator flake on the runner, unrelated to this PR's changes.

The PR is ready to merge once CI is re-run and passes (or once a second reviewer approves).


---

aunger:
The name `has_code_changes` doesn’t match the meaning, “needs full build”. If you roll in the “true if direct push” predicate, you should rename.

aunger (commented):
**Review verdict: LGTM — ready to merge.** (GitHub prevented a formal APPROVE because this account is the PR author, but my assessment is a full approval.)

CI is fully green: build-and-test success, all CodeQL analyses success, dev-build skipped as expected.

**Shell injection fix** — `github.base_ref` is assigned to `BASE_REF` before use; both `git fetch origin "$BASE_REF"` and `git diff "origin/$BASE_REF"...HEAD` quote it correctly. Sound.

**Rename** — every output assignment uses `needs_full_build` and every `if:` guard references `steps.diff_check.outputs.needs_full_build == 'true'`. No `has_code_changes` residue anywhere in the diff.

**Logic** — the three-dot diff with a shallow fetch of the base ref is the right approach. `|| true` on `grep` prevents a false non-zero exit when all changes are docs-only. `always() && needs_full_build == 'true'` on teardown and upload steps correctly handles failure paths without running teardown when the build was skipped entirely. The push path unconditionally setting `needs_full_build=true` is correct for merge commits to main.

---

PR #165, opened 2026-05-17
Clarify ci_watcher sleep must be foreground

aunger:
Agents were calling sleep 30 with run_in_background: true, causing the
poll loop to execute all 5 iterations without any actual delay.

https://claude.ai/code/session_01CQjZ8BYjwYMpcZDCfP1Gdw


aunger (commented):
**Verdict: Approved** (posted as COMMENT because the reviewer identity matches the PR author; GitHub blocks self-approval)

The three changes are correct and the document remains coherent.

**sleep clarification** — The addition of "in the foreground (set run_in_background: false)" at the exact call site is unambiguous. An agent following the pseudocode literally cannot miss it.

**loop wording** — "repeat 5 times (unless broken)" is a cleaner description than "repeat up to 5 times" because the loop _always_ tries 5 unless an early-return fires; the parenthetical makes that explicit without being verbose.

**Haiku directive** — The phrasing "Create a Haiku sub-agent unless the user requested otherwise." is a character-for-character match of the pattern used for Programmers and Reviewers in dev_orchestration.md (substituting model name only). Consistent.

**No unintended changes** — 3 additions, 3 deletions, single file, all touching exactly the described lines.

The old Lifecycle bullet ("CiWatcher is spawned by the Orchestrator after a Reviewer exits") carried no actionable instruction for the CiWatcher itself; replacing it with the model-selection directive is a net improvement in clarity.

---

Issue #156, opened 2026-05-16
E2E Visual Tests: plan tracking


aunger:
Tracking issue for the 7-phase E2E visual test plan described in [`PLAN_E2E_VISUAL_TESTS.md`](https://github.com/aunger/gallery-button-for-pixel-camera/blob/18c752d70e118230ddca11e101c0cf37a19789c3/PLAN_E2E_VISUAL_TESTS.md).

The plan adds a visually-evaluated emulator test suite that verifies GB4PC's overlay button renders correctly, opens the gallery on tap, and behaves correctly in Android's secure-camera mode from a locked screen.

## Phases

- [x] Phase 1 — `:testgallery` module
- [x] Phase 2 — Mock camera and GREEN feed
- [x] Phase 3 — Visual-assertion library
- [x] Phase 4 — `E2EFixture` extensions
- [x] Phase 5 — `GalleryButtonVisualE2ETest`
- [x] Phase 6 — Build & CI wiring
- [x] Phase 7 — Commit organization

aunger:
## Final integration review — E2E Visual Tests (#156)

Reviewer pass over the full 7-phase plan as it exists at `main@d693d64` plus the two open PRs (#173 Phase 6, #174 Phase 7). Treating this as a "is the whole thing actually delivered" review, not a per-PR review.

### TL;DR — verdict

The implementation is **substantially correct and well-engineered** in the parts I reviewed. The visual library, fixture, test class, mock-camera Alternative 1 implementation, and mock-gallery are all faithful to the plan (with Alternative 1 explicitly chosen and documented). The unit tests for `ShapeMatcher` are thorough.

However, there are **three blocking integration issues** that mean Phase 6 is not actually done, and one of them means the entire E2E suite has never run in CI in the green path. PR #173 in its current form does not fix the most important of these, and its title/body misrepresent what it changes.

---

### Blocking issues

#### B1. The E2E test step never runs on green CI (pre-existing bug in `main`)

`.github/workflows/build.yml`, step `Run E2E setup and tests`:

```yaml
- name: Run E2E setup and tests
  if: failure() && steps.diff_check.outputs.needs_full_build == 'true'
```

The condition is `if: failure()`. This step only executes when an **earlier step has already failed**. On a clean run (pre-flight passes), `failure()` is false, so the entire E2E suite — the whole point of issues #157–#163 — is silently skipped. The build then passes without ever invoking `connectedE2EAndroidTest`.

[...]

PR #175, opened 2026-05-17
Fix E2E test Alternative-1 consistency: error messages, test4a docs, displaySize dedup

aunger:
$(cat <<'EOF'
## Summary

Closes part of #156 integration review findings (N1, N3, N6).

- **N1**: `test0_smokeGreenFeedVisible` failure message and doc comment referenced the virtualscene poster, which doesn't exist under Alternative 1 (MockCameraActivity renders a solid-green View). Updated both to say "check that the mock-camera APK is installed and MockCameraActivity is in the foreground."
- **N3**: `test4a_secureCameraLockedEmptyGalleryNoGreen` doc comment claimed the test "passes regardless of the regression." Under Alternative 1 that reasoning is wrong: when the overlay is blocked, the green MockCameraActivity stays on screen (~100% GREEN coverage), causing the test to fail. Updated the comment to document test4a as a red-light test under Alternative 1, and to explain that the assertion itself is still correct.
- **N6**: `displaySize()` was duplicated in both `E2EFixture.kt` (private) and `GalleryButtonVisualE2ETest.kt` (private). Made the fixture's copy `internal`, removed the duplicate from the test class, and updated the one call site to use `fixture.displaySize()`. Removed the now-unused `android.os.Build` import. Pure behavior-preserving refactor.

## Test plan

- [x] `./gradlew assembleDebug` — BUILD SUCCESSFUL
- [x] `./gradlew testDebugUnitTest` — BUILD SUCCESSFUL (all unit tests pass)

https://claude.ai/code/session_01SoSXuP8bw3ED3fw12ckFFS
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01SoSXuP8bw3ED3fw12ckFFS)_

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -397,7 +397,7 @@ class E2EFixture(
         return count
     }
 
-    private fun displaySize(): Pair<Int, Int> {
+    internal fun displaySize(): Pair<Int, Int> {
         return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
             val wm = context.getSystemService(Context.WINDOW_SERVICE) as android.view.WindowManager
             val bounds = wm.currentWindowMetrics.bounds
--- a/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
@@ -275,9 +275,28 @@ class GalleryButtonVisualE2ETest {
 
     /**
      * Verifies that tapping the overlay in secure-camera mode (screen locked) with an empty
-     * camera roll does not show GREEN content. Whether or not the overlay renders in
-     * secure-camera mode, an empty gallery produces no GREEN — so this test passes regardless
-     * of the secure-camera-overlay regression tracked in issue #156.
+     * camera roll does not show GREEN content (coverage < 10%).
+     *
+     * **Alternative 1 note — this is a RED-LIGHT test, not a baseline pass.**
+     *
+     * The original plan claimed this test "passes regardless of the regression" because an
+     * empty gallery produces a black empty state → no GREEN. That reasoning assumed the camera
+     * background was black (e.g. real hardware or a virtualscene with a dark scene). Under
+     * Alternative 1, MockCameraActivity's background is solid green (#00C853).
+     *
+     * When the secure-camera overlay regression is present (overlay is blocked / not tappable),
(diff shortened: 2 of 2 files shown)


aunger (commented):
**Verdict: would APPROVE (GitHub rejects self-approval, so posting as COMMENT).** All three fixes land correctly.

**N1 — `test0_smokeGreenFeedVisible` message (accurate and clear).**
The old failure message ("emulator virtualscene poster is configured") was a leftover from the primary approach and would have misdirected a future debugger toward emulator AVD config instead of the actual root causes under Alternative 1. The new message — "Check that the mock-camera APK is installed under $MOCK_CAMERA_PACKAGE and MockCameraActivity is in the foreground" — names the two real failure modes for an Alt-1 setup. The KDoc update on the same method also accurately describes Alt 1 ("MockCameraActivity renders a solid-green (#00C853) View — no camera hardware and no virtualscene poster are required"). Clear, accurate, and aligned with the harness as it actually exists.

**N3 — `test4a_secureCameraLockedEmptyGalleryNoGreen` doc comment (correct).**
The new doc comment correctly walks through the Alt-1 mechanics:
- Old plan's "passes regardless of regression" reasoning depended on the camera background being black; under Alt 1 it is solid green.
- Regression present → `tapOverlay()` is a no-op → green MockCameraActivity stays on-screen → coverage ≈ 100% → `coverage < 10%` assertion fails (red light, as intended).
- Regression absent → tap opens the empty gallery → black empty state → coverage < 10% → assertion passes.

The assertion `coverage(GREEN) < 10%` is unchanged and remains correct — it accidentally produces the right signal under Alt 1 because the empty-gallery screen is black and the MockCameraActivity screen is green, which gives clean separation across the threshold. The explicit "Do not change the assertion" note and the "Do NOT skip, ignore, or quarantine this test" note plus the #156 cross-reference are good defenses against a future contributor "fixing" the failing test the wrong way.

Minor non-blocking nit: the comment says the empty-gallery screen is "black" — strictly it just needs to be non-green; if mock-gallery's empty state ever changes to a non-black-but-still-non-green palette the assertion would still hold. Not worth changing.

**N6 — `displaySize()` deduplication (clean refactor).**
Body of `E2EFixture.displaySize()` is semantically identical to the removed copy (same API-30 branch using `WindowManager.currentWindowMetrics.bounds`, same pre-R branch using `defaultDisplay.getMetrics(DisplayMetrics)` with `@Suppress("DEPRECATION")`). Visibility widened from `private` to `internal`, which is the right scope for cross-class use within the androidTest source set without exposing it to production. The single caller (`test1a_…`) is correctly updated to `fixture.displaySize()`, and the now-unused `android.os.Build` import is removed from the test class. No behavior change.

**No new issues found.** Not blocking on CI per the reviewer guide.

---

Issue #179, opened 2026-05-18
E2E: visual overlay checks fail — BLUE pixel mismatch and adaptive-icon shape is CIRCLE not SQUIRCLE (tests 1a, 1c, 2a)


aunger:
## Problem

Three `GalleryButtonVisualE2ETest` tests that do pixel-level screenshot analysis fail on the CI emulator. 

### test1a — no BLUE pixels found (`pixelCount=0`)

`ColorMatch.mask(screen, Rgb.BLUE)` returns an empty mask even though `OverlayService.isOverlayActive` is true (confirmed by `PixelCameraOverlayE2ETest` passing). 

### test1c — adaptive-icon outer shape is CIRCLE, expected SQUIRCLE

`ShapeMatcher.requireShape(outer, Shape.SQUIRCLE)` fails with `winner=CIRCLE (IoU=0.993)`. The Android launcher on the `google_apis` API-35 emulator clips adaptive icons to a circle, not the squircle mask used on Pixel devices. 

### test2a — 87 % GREEN coverage after tap (overlay not tappable or wrong position)

After `waitForOverlayActive()` and `tapOverlay()`, the screen is still ~87 % green. 

## Investigation needed

1. Upload and inspect the screenshot artifacts from a CI failure run (enable the `Upload E2E screenshots on failure` step to capture `1a-screen.png`, `1a-blue-mask.png`, `1c-screen.png`, `2a-s1.png`, `2a-s2.png`).
2. Check `Rgb.BLUE` tolerance against the actual icon pixel values.
3. For test1c: configure the emulator to use a squircle icon shape, or create it in GB4PC code.
4. For test2a: check `OverlayService`/`OverlayManager` window flags and verify the tap coordinate calculation against the actual overlay position in the screenshot.

aunger:
For test1c, no, circle is **not** an acceptable shape.

aunger:
This issue is unlikely to produce any good work. Closing.

Issue #178, opened 2026-05-18
E2E: lockScreen() fails in CI — emulator keyguard never engages without a PIN (test 4a)


aunger:
## Problem

`E2EFixture.lockScreen()` sends `KEYCODE_SLEEP` (223) and then polls `KeyguardManager.isKeyguardLocked` for up to 5 s. On the CI emulator (`system-images;android-35;google_apis;x86_64`) there is no lock screen PIN configured, so `KEYCODE_SLEEP` turns the display off but the keyguard never engages. The poll times out and the test fails.

**Failing test:** `test4a_secureCameraLockedEmptyGalleryNoGreen`

**First observed:** PR #176 — first CI run after the `failure()` guard on the E2E step was removed.

## Fix

Add `adb shell locksettings set-pin 1234` to `scripts/setup-e2e-emulator.sh` (both the full and `--post-boot` paths) so the emulator has a PIN before the test suite runs. After tests complete, the PIN can be cleared with `locksettings clear --old 1234`.

Alternatively, increase the `lockScreen()` timeout, but that won't help if the keyguard fundamentally won't engage without a PIN.

## Re-enable gate

Once fixed, remove `if: false` from the "Run E2E setup and tests" step in `.github/workflows/build.yml` (see comment there) and re-run CI.

PR #180, opened 2026-05-18
Export shutter receiver and drop setPackage to unblock cross-UID broadcast

### LINKED ISSUE #177: E2E: captureOnePhoto() blocked by RECEIVER_NOT_EXPORTED cross-UID broadcast (tests 3a, 5a)
## Problem

`E2EFixture.captureOnePhoto()` sends `ACTION_SHUTTER` from the test process (`com.gb4pc` UID) to `MockCameraActivity` in `com.google.android.GoogleCamera` via `context.sendBroadcast()`. On API 33+, the receiver is registered with `Context.RECEIVER_NOT_EXPORTED`, which blocks cross-UID broadcasts. The broadcast is silently dropped, so `MockCameraActivity` never writes a photo to MediaStore, and the test fails after a 15 s timeout.

The same problem exists in the reverse direction: `MockCameraActivity.broadcastShutterDone()` sends `ACTION_SHUTTER_DONE` with `setPackage("com.google.android.GoogleCamera")`, which restricts delivery to within that package — the test's latch receiver (in `com.gb4pc`) never fires.

**Failing tests:** `test3a_populatedGalleryShowsGreenAfterTap`, `test5a_secureCameraLockedPopulatedGalleryShowsGreen`

**First observed:** PR #176 — first CI run after the `failure()` guard on the E2E step was removed.

## Fix options

1. Register the shutter BroadcastReceiver in `MockCameraActivity` with `RECEIVER_EXPORTED` (lowest-risk; this is a test-only APK, and `ACTION_SHUTTER` is a private action string).
2. Drop `setPackage(packageName)` from `broadcastShutterDone()` so `ACTION_SHUTTER_DONE` can reach the test's latch receiver.
3. Change `captureOnePhoto()` to trigger the shutter via `uiAutomation.executeShellCommand("am broadcast …")` using an explicit component intent, which bypasses the exported check.

Option 1 + 2 together is the cleanest path.

## Re-enable gate

Once fixed, remove `if: false` from the "Run E2E setup and tests" step in `.github/workflows/build.yml` (see comment there) and re-run CI.

aunger:
## Summary

Fixes #177.

`E2EFixture.captureOnePhoto()` (running in the `com.gb4pc` test UID) sends `ACTION_SHUTTER` to `MockCameraActivity` (running in `com.google.android.GoogleCamera`). On API 33+ the receiver was registered with `Context.RECEIVER_NOT_EXPORTED`, so the broadcast was silently dropped across the UID boundary; the activity never wrote a photo to MediaStore and the test stalled until its 15 s timeout.

The reverse direction had the mirror-image problem: `MockCameraActivity.broadcastShutterDone()` called `setPackage(packageName)`, restricting `ACTION_SHUTTER_DONE` delivery to the mock-camera package — the test's latch receiver in `com.gb4pc` never fired.

This PR applies fix options 1 + 2 from the issue:

1. Register the `ACTION_SHUTTER` receiver with `Context.RECEIVER_EXPORTED` so the test process can deliver shutter triggers across UIDs.
2. Drop the `setPackage(packageName)` call from `broadcastShutterDone()` so the completion broadcast can reach a receiver in another package.

This APK is only installed on test emulators and the action strings are private, so exporting the receiver carries no production risk.

A follow-up commit removes the resolved `#177` reference from the `Run GalleryButtonVisualE2ETest` workflow comment. `continue-on-error: true` stays in place because #178 and #179 are still open.

## Test plan

- [x] `./gradlew assembleDebug assembleDebugAndroidTest :e2e-mock-camera:assembleDebug :e2e-mock-gallery:assembleDebug testDebugUnitTest` — passes locally.
- [ ] CI runs `test3a_populatedGalleryShowsGreenAfterTap` and `test5a_secureCameraLockedPopulatedGalleryShowsGreen` against an API-35 emulator and they complete within their 15 s captureOnePhoto budget instead of timing out.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -292,9 +292,9 @@ jobs:
           }
 
       - name: Run GalleryButtonVisualE2ETest
-        # Known failures tracked in #177 (RECEIVER_NOT_EXPORTED), #178 (emulator
-        # keyguard not engaging), #179 (adaptive-icon colour/shape). continue-on-error
-        # keeps the job green until those are fixed.
+        # Known failures tracked in #178 (emulator keyguard not engaging) and
+        # #179 (adaptive-icon colour/shape). continue-on-error keeps the job
+        # green until those are fixed.
         if: steps.diff_check.outputs.needs_full_build == 'true'
         continue-on-error: true
         run: |
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -178,7 +178,16 @@ class E2EFixture(
         }
         val filter = IntentFilter(actionShutterDone)
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
-            context.registerReceiver(receiver, filter, Context.RECEIVER_NOT_EXPORTED)
+            // RECEIVER_EXPORTED is required so MockCameraActivity (running under the
+            // Pixel Camera UID, com.google.android.GoogleCamera) can deliver
+            // ACTION_SHUTTER_DONE across UIDs back into this test process (com.gb4pc).
+            // On API 33+ RECEIVER_NOT_EXPORTED silently drops cross-UID broadcasts,
+            // which would cause the latch to time out and the test to fall back to
+            // MediaStore/MediaScanner polling — defeating the intent of the
+            // ACTION_SHUTTER_DONE handshake. The receiver is only registered inside
+            // the instrumented E2E test process and ACTION_SHUTTER_DONE is a private
+            // action string, so exporting the receiver carries no production risk.
+            context.registerReceiver(receiver, filter, Context.RECEIVER_EXPORTED)
         } else {
             @Suppress("UnspecifiedRegisterReceiverFlag")
             context.registerReceiver(receiver, filter)
--- a/e2e-mock-camera/src/main/java/com/gb4pc/mockcamera/MockCameraActivity.kt
+++ b/e2e-mock-camera/src/main/java/com/gb4pc/mockcamera/MockCameraActivity.kt
@@ -125,10 +125,15 @@ class MockCameraActivity : Activity() {
         super.onResume()
         startCameraThread()
         openCamera()
+        // RECEIVER_EXPORTED is required so the E2E test process (com.gb4pc UID) can
+        // deliver ACTION_SHUTTER across UIDs into this activity (com.google.android.GoogleCamera).
+        // On API 33+ RECEIVER_NOT_EXPORTED silently drops cross-UID broadcasts. This APK is
+        // installed only on test emulators and ACTION_SHUTTER is a private action string, so
+        // exporting the receiver carries no production risk.
         registerReceiver(
             shutterReceiver,
             IntentFilter(ACTION_SHUTTER),
-            Context.RECEIVER_NOT_EXPORTED,
+            Context.RECEIVER_EXPORTED,
         )
     }
 
@@ -253,8 +258,10 @@ class MockCameraActivity : Activity() {
     }
 
     private fun broadcastShutterDone(uri: Uri?) {
+        // Intentionally do NOT call setPackage(packageName): the listener is the E2E test
+        // process (com.gb4pc), not this APK. setPackage(com.google.android.GoogleCamera)
+        // would restrict delivery to this package and starve the test's latch receiver.
         val intent = Intent(ACTION_SHUTTER_DONE).apply {
-            setPackage(packageName)
             if (uri != null) putExtra(EXTRA_IMAGE_URI, uri)
         }
         sendBroadcast(intent)
--- a/e2e-mock-camera/src/main/java/com/gb4pc/mockcamera/MockCameraActivity.kt
+++ b/e2e-mock-camera/src/main/java/com/gb4pc/mockcamera/MockCameraActivity.kt
@@ -82,6 +82,13 @@ class MockCameraActivity : Activity() {
         }
     }
 
+    /**
+     * True only after a successful [registerReceiver]. Tracked so [onPause] does not call
+     * [unregisterReceiver] when registration threw, which would itself throw
+     * `IllegalArgumentException: Receiver not registered` and mask the original error.
+     */
+    private var shutterReceiverRegistered = false
+
     // -------------------------------------------------------------------------
     // CameraDevice state callback
     // -------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


aunger (commented):
## Review (changes requested — posted as COMMENT because the review API blocks REQUEST_CHANGES on one's own PR)

Thanks for the clear PR description and the explanatory comments at both edit sites — they make the intent obvious and will save future readers a trip back to #177.

The two `MockCameraActivity` edits and the workflow-comment edit each match the plan in #177 (and the PR body correctly explains that the issue's "remove `if: false`" instruction maps to "update the comment" because `main` never actually had an `if: false` gate — only `continue-on-error: true` with a stale `#177` reference). So far so good.

### Blocking: the reverse-direction symmetry isn't actually fixed

Issue #177 identifies two cross-UID broadcast failures. This PR addresses one side of each:

1. `MockCameraActivity` now uses `RECEIVER_EXPORTED` so the test UID can deliver `ACTION_SHUTTER` to the mock-camera UID. Good.
2. `broadcastShutterDone()` no longer calls `setPackage(packageName)` so the intent is no longer restricted to `com.google.android.GoogleCamera`. Good.

But the *latch receiver in `E2EFixture.captureOnePhoto()`* — which is the actual consumer of `ACTION_SHUTTER_DONE` — is still registered with `RECEIVER_NOT_EXPORTED` (`app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt:181`):

```kotlin
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
    context.registerReceiver(receiver, filter, Context.RECEIVER_NOT_EXPORTED)
} else {
    @Suppress("UnspecifiedRegisterReceiverFlag")
    context.registerReceiver(receiver, filter)
}
```

By exactly the reasoning your own commit message uses for the `MockCameraActivity` side — "On API 33+ `RECEIVER_NOT_EXPORTED` silently drops cross-UID broadcasts" — this side will also drop the broadcast: the broadcast is sent from `com.google.android.GoogleCamera`'s UID and received by `com.gb4pc`'s UID, so it is cross-UID just like the other direction. Dropping `setPackage` on the sender doesn't undo the `RECEIVER_NOT_EXPORTED` filter on the receiver.

The likely observable effect is that `latch.await(10, TimeUnit.SECONDS)` always times out, and the test then falls through to the MediaStore polling / `MediaScannerConnection` fallback. The 15 s budget you cite in the PR test plan may still be met because of that polling fallback, so it's possible the tests appear green; but the latch is then doing nothing useful, the test is silently relying on the slow fallback path, and the "fix option 2" the issue describes (`ACTION_SHUTTER_DONE` reaches the test latch) is not actually achieved.

Please flip the latch receiver in `E2EFixture.captureOnePhoto()` to `RECEIVER_EXPORTED` on API 33+ as well, so both halves of the round-trip work as designed.

### Minor

- The new comment block above the `registerReceiver` call in `MockCameraActivity` describes only the forward direction. Once the `E2EFixture` side is also exported, please add a similarly short comment there so both halves of the explanation live next to the code that needs them.
- The "this APK is installed only on test emulators … so exporting the receiver carries no production risk" justification is good. If you also export the latch receiver, please keep the wording consistent there (instrumentation APK, never shipped).

### Scope

No scope-creep concerns from me; the diff is appropriately minimal. #178 and #179 are correctly left for separate PRs and `continue-on-error: true` is correctly retained.

Verdict: changes requested for the receiver-side asymmetry above. Happy to re-review once it's addressed.

---

aunger (commented):
LGTM — approving in spirit (GitHub blocks a formal APPROVE because the reviewer account also authored the PR). The asymmetric receiver flag I flagged is now fixed.

Verified in this revision:

- `E2EFixture.kt`: the `ACTION_SHUTTER_DONE` latch receiver now registers with `RECEIVER_EXPORTED` on API 33+, matching the cross-UID delivery requirement (the broadcaster runs under `com.google.android.GoogleCamera`, the listener under `com.gb4pc`). Without this, API 33+ would silently drop the broadcast and force the test onto the MediaStore polling fallback — exactly the issue from #177.
- `MockCameraActivity.kt`: the symmetric path is also corrected — the `ACTION_SHUTTER` receiver is exported, and `broadcastShutterDone` no longer calls `setPackage(packageName)` (which would have pinned delivery to `com.google.android.GoogleCamera` and starved the test listener). Good catch keeping both directions consistent.
- `build.yml`: the `#177` reference is removed from the known-failures comment, which is appropriate now that the fix is in.

The inline comments justifying `RECEIVER_EXPORTED` (test-only process, private action strings, no production exposure) are clear and accurate. No further concerns from me.

---

aunger (commented):
**Verdict: APPROVE** (posted as COMMENT because GitHub blocks self-approval on this PR; the substance is an approval of follow-up commit `9e2d473`).

## Correctness review

The defensive try/catch around `registerReceiver(..., RECEIVER_EXPORTED)` in `MockCameraActivity.onResume()` plus the `shutterReceiverRegistered` guard in `onPause()` are correct and appropriately scoped.

1. **try/catch is appropriate here, not masking.** The activity's primary CI job is to render the #00C853 view so the smoke check (and the overlay / ForegroundDetector path tests) can run. An uncaught `RuntimeException` escaping `onResume()` after `super.onResume()` returns leaves the window stuck on the system background — exactly the 0% green symptom from run #405. Catching and logging keeps the green View on-screen for the smoke check while preserving a clear diagnostic in logcat (`Log.e(TAG, "onResume: shutter receiver registration failed: ${e.message}", e)`). The catch is bounded to `RuntimeException`, so genuine fatal `Error` subclasses still propagate.

2. **Shutter handshake degradation is acceptable.** If registration fails, ACTION_SHUTTER will not be delivered, `triggerCapture()` won't run, and tests 3a/5a will fail loudly at the 15 s `captureOnePhoto()` budget — which currently has `continue-on-error: true` and is still gated by #178 / #179 anyway. The smoke check and the overlay / test4a paths don't depend on the shutter handshake, so the prioritization is right. The trade-off is acknowledged inline.

3. **onPause unregister guard is correct.** `shutterReceiverRegistered = true` is set only after `registerReceiver` returns normally; `onPause` checks the flag before calling `unregisterReceiver`, then clears it. This avoids the secondary `IllegalArgumentException: Receiver not registered` that would otherwise mask the original cause. Lifecycle callbacks run on the main thread, so no synchronization on the flag is needed.

4. **Scope is tight.** Only the registration site is touched; no unrelated edits, no changes to the (correct) `broadcastShutterDone()` or the symmetric `E2EFixture` receiver.

## Non-blocking note

The commit message frames the root cause as a hypothesis ("SecurityException for protected-broadcast collisions or IllegalArgumentException for filter conflicts have both been reported"). The fix is defensive enough that the precise platform bug doesn't matter for shipping this PR — but if CI still shows 0% green after this lands, the `Log.e` line is the next thing to grep in the logcat artifact to pin down the actual exception class. No code change required.

LGTM.

---

PR #181, opened 2026-05-18
Set lock-screen PIN in E2E emulator setup so keyguard engages on sleep

### LINKED ISSUE #178: E2E: lockScreen() fails in CI — emulator keyguard never engages without a PIN (test 4a)

aunger:
## Summary

Fixes #178.

`E2EFixture.lockScreen()` sends `KEYCODE_SLEEP` and polls `KeyguardManager.isKeyguardLocked` for up to 5 s. On the CI emulator (`system-images;android-35;google_apis;x86_64`) there is no lock-screen credential, so `KEYCODE_SLEEP` turns the display off but the keyguard never engages — the poll times out and `test4a_secureCameraLockedEmptyGalleryNoGreen` fails.

This PR adds a "Step 7" to `scripts/setup-e2e-emulator.sh` that runs `adb shell locksettings set-pin 1234`, which configures a PIN so the keyguard becomes secure and engages whenever the display sleeps. The step lives in the common path of the script, so it runs in both full and `--post-boot` modes. The tests never enter the PIN themselves — they only rely on `isKeyguardLocked` flipping to true.

For idempotency on persistent local AVDs, the command falls back to `locksettings set-pin --old 1234 1234` if a PIN is already configured. The header documents the corresponding `locksettings clear --old 1234` for tearing the PIN down locally.

## Why not also re-enable the E2E gate?

Issues #177 and #179 are still open and both also reference the same `continue-on-error: true` (the workflow comment lists all three issues as blockers). Removing `continue-on-error` while #177 and #179 remain unresolved would re-introduce red CI for unrelated reasons, so the gate must stay open until all three issues are fixed. This PR fixes only #178.

## Test plan

- [x] `./gradlew testDebugUnitTest --no-daemon` — BUILD SUCCESSFUL
- [x] `bash -n scripts/setup-e2e-emulator.sh` — script parses cleanly
- [ ] CI: with continue-on-error still in place, observe in the CI logs that `test4a_secureCameraLockedEmptyGalleryNoGreen` no longer times out in `lockScreen()` (it may still fail later on the green-coverage assertion if #156 is unresolved, but the keyguard-timeout symptom should be gone).

---

### DIFF
--- a/scripts/setup-e2e-emulator.sh
+++ b/scripts/setup-e2e-emulator.sh
@@ -2,8 +2,8 @@
 # setup-e2e-emulator.sh — Prepare an Android emulator for E2E tests.
 #
 # Usage:
-#   scripts/setup-e2e-emulator.sh            # Full local setup (steps 1–6)
-#   scripts/setup-e2e-emulator.sh --post-boot # CI post-boot setup only (steps 4–6)
+#   scripts/setup-e2e-emulator.sh            # Full local setup (steps 1–7)
+#   scripts/setup-e2e-emulator.sh --post-boot # CI post-boot setup only (steps 4–7)
 #
 # Full setup (local):
 #   1. Create AVD (API 35, Google APIs, x86_64, Pixel_6 skin)
@@ -12,11 +12,15 @@
 #   4. Grant GET_USAGE_STATS to GB4PC
 #   5. Grant SYSTEM_ALERT_WINDOW to GB4PC
 #   6. Disable animations
+#   7. Set lock-screen PIN (so the keyguard actually engages on sleep)
 #
 # Post-boot setup (CI): the emulator is already running and all system services
-# have been verified ready by the workflow; this script performs steps 4–6 only.
+# have been verified ready by the workflow; this script performs steps 4–7 only.
 # Mock Pixel Camera (e2e-mock-camera) is installed separately by the CI workflow
 # and by the connectedE2EAndroidTest Gradle task.
+#
+# To clear the PIN after the test run (e.g. for a clean local AVD), run:
+#   adb shell locksettings clear --old 1234
 #
 # Prerequisites:
 #   - ANDROID_HOME (or ANDROID_SDK_ROOT) must be set
@@ -135,6 +139,24 @@ echo "==> Disabling animations..."
 "$ADB" shell settings put global window_animation_scale 0
 "$ADB" shell settings put global transition_animation_scale 0
 "$ADB" shell settings put global animator_duration_scale 0
+
+# ── Step 7: Set lock-screen PIN ─────────────────────────────────────────────
+# Without a lock-screen credential the keyguard never engages on the CI
+# emulator: KEYCODE_SLEEP turns the display off but KeyguardManager.isKeyguardLocked
+# remains false, so E2EFixture.lockScreen() times out (issue #178).
+# Setting a PIN makes the keyguard secure, so it engages whenever the display
+# sleeps. The PIN itself is never entered by the tests — they only rely on the
+# keyguard being locked. Clear the PIN with `locksettings clear --old 1234`
+# if you need to remove it later (the AVD is otherwise idempotent).
+#
+# `locksettings set-pin` rejects a new PIN if one is already configured unless
+# `--old` is supplied. For idempotency across re-runs of --post-boot on a
+# persistent local AVD, try the bare form first and fall back to `--old 1234`
+# (which is a no-op when the PIN is already 1234). Either path leaves the AVD
+# with PIN=1234, which is all the keyguard needs to engage on sleep.
+echo "==> Setting lock-screen PIN (1234) so keyguard engages on sleep..."
+"$ADB" shell locksettings set-pin 1234 || \
+    "$ADB" shell locksettings set-pin --old 1234 1234
 
 echo ""
 echo "==> E2E emulator setup complete. Run E2E tests with:"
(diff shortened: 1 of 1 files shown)


aunger (commented):
Verdict: **approve** (submitted as COMMENT because the review API rejects self-approval — same identity as the PR author).

The fix is tightly scoped and lands in the right place.

Checked:

- **Placement.** Step 7 sits in the common path of the script (after the `POST_BOOT_ONLY` branch closes), so it runs in both full and `--post-boot` invocations. Issue #178 explicitly required both paths — satisfied.
- **Idempotency fallback.** `set-pin 1234 || set-pin --old 1234 1234` is correct: on a fresh AVD the first command succeeds; on a persistent local AVD where the PIN is already `1234`, the first fails (because `locksettings set-pin` requires `--old` when a credential exists) and the fallback re-sets the same PIN as a no-op. End state in both cases is PIN=1234, which is all the keyguard needs.
- **Teardown.** Issue #178 phrased the post-test `locksettings clear --old 1234` as "the PIN can be cleared" — informational, not a hard requirement. CI runners are ephemeral so it doesn't matter there, and the script header documents the command for local developers. Not a blocker.
- **Scope.** Correctly avoids touching the test, `E2EFixture.lockScreen()`, or the `continue-on-error` gate. The PR body explains the gate stays open until #177/#179 are also resolved — agreed.
- **Comment quality.** The Step 7 comment correctly notes the tests rely only on `isKeyguardLocked`, not on PIN entry.

Minor non-blocking observation: the fallback only succeeds if the existing PIN is exactly `1234`. If a local AVD has been hand-configured with a different PIN, both invocations will fail and `set -euo pipefail` will abort the script. This is a niche local-dev case (CI is always fresh) and the header already documents `locksettings clear --old 1234` as the manual remedy — not worth fixing here.

---

aunger (commented):
**Verdict: Approve** (posted as COMMENT because GitHub blocks the formal APPROVE action on one's own PR.)

The Step 8 follow-up correctly addresses the keyguard-engaged-too-early failure mode introduced by Step 7.

**Sequence is correct.** `KEYCODE_WAKEUP` → `wm dismiss-keyguard` → `input text 1234` → `KEYCODE_ENTER` is the canonical adb pattern for dismissing a *secure* keyguard on API 30+. `wm dismiss-keyguard` against a secure keyguard only raises the bouncer; the PIN must still be supplied, which is what the next two lines do. Without the leading `KEYCODE_WAKEUP`, `dismiss-keyguard` could be a no-op on a display that has slipped to AOD between `locksettings set-pin` returning and Step 8 running.

**Ordering of set-pin → dismiss → stay-on is sound.** Step 7 makes the keyguard secure (and engages it); Step 8 dismisses it; the workflow's pre-existing `stay_on_while_plugged_in 7` then keeps the display on and the keyguard down until `E2EFixture.lockScreen()` explicitly re-engages via `KEYCODE_SLEEP`.

**Minor / non-blocking observations (do not address in this PR — spawn follow-ups if desired):**

1. *Potential focus race in the bouncer.* There is no sleep or `wait-for-idle` between `wm dismiss-keyguard` and `input text 1234`. If the bouncer hasn't focused its PIN field yet, the digits could land in the void and ENTER would fire on an empty entry, leaving the keyguard up. In practice, adb round-trip latency (50–200 ms per shell exec) is usually enough cover on API-35 emulators, and there is no evidence this is currently failing — but if Step 8 ever flakes, the first remediation to try is a short `sleep 0.5` (or `uiautomator dump`/`waitForIdle`) between `dismiss-keyguard` and `input text`.

2. *Idempotency of Step 8 on persistent local AVDs.* On a re-run where the keyguard is already down, `input text 1234` types "1234" into whatever has focus (launcher search bar, an open app), and ENTER fires on that. Benign on a CI ephemeral emulator; slightly noisy on a long-lived local AVD. Worth a one-line note for future readers, but not a blocker.

3. *No post-condition check.* The script does not verify `KeyguardManager.isKeyguardLocked == false` before exiting. If Step 8 silently fails, the symptom reappears in the first E2E test. A `dumpsys window | grep -i keyguard` assertion (or polling `dumpsys window policy` for `mShowingLockscreen=false`) would make the script fail loudly at setup time instead of midway through the suite. Out of scope here.

None of the above is required. The fix matches the diagnosis in the commit message, the inline rationale is accurate, and the change is tightly scoped to unblocking CI for #178. LGTM.

---

PR #189, opened 2026-05-19
e2e: capture diagnostic screenshot on failure via shared FailureScreenshotRule

### LINKED ISSUE #187: E2E: capture diagnostic screenshot on failure in all E2E test classes
## Problem

When an E2E test fails there is no visual artifact to diagnose what was on screen at the moment of failure. This applies to both test classes:

- `PixelCameraOverlayE2ETest` — purely assertion-based, never writes screenshots (see #186)
- `GalleryButtonVisualE2ETest` — already saves comparison screenshots for its own assertions, but a failure during setup or a non-visual assertion would still leave no diagnostic image

The CI pull-and-upload step already collects everything from `/sdcard/Android/data/com.gb4pc/files/screenshots/`, so any screenshot saved there is automatically uploaded as an artifact with no workflow changes needed.

## Fix

Extract a shared `FailureScreenshotRule` (e.g. in `e2e/src/androidTest/.../FailureScreenshotRule.kt`) and apply it in both classes:

```kotlin
class FailureScreenshotRule : TestWatcher() {
    override fun failed(e: Throwable?, description: Description) {
        val dir = File(
            InstrumentationRegistry.getInstrumentation().targetContext
                .getExternalFilesDir(null), "screenshots"
        )
        dir.mkdirs()
        val className = description.testClass?.simpleName ?: "UnknownClass"
        UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
            .takeScreenshot(File(dir, "$className-${description.methodName}-failure.png"))
    }
}
```

Then in each test class:

```kotlin
@get:Rule
val screenshotOnFailure = FailureScreenshotRule()
```

### Notes

- `description` in `TestWatcher.failed()` is a method-level description, so `description.methodName` is never null here — it is only null on class- or suite-level descriptions, which `TestWatcher` never passes to `failed()`. The `@Nullable` annotation on the `Description` parameter itself is why `failed()` above overrides the non-null form.
- `description.testClass?.simpleName` extracts the class name from the description directly; no constructor parameter is needed.
- `GalleryButtonVisualE2ETest` already saves its own named screenshots during the test body; this rule adds a catch-all for any failure that occurs before or outside those saves.

## Why not video

A still at the moment of failure is sufficient for timing and overlay-presence assertions. Video can be added later if stills prove insufficient.

aunger:
$(cat <<'EOF'
## Summary

- Adds `FailureScreenshotRule` — a JUnit `TestWatcher` that automatically captures a screenshot to `/sdcard/Android/data/com.gb4pc/files/screenshots/<ClassName>-<methodName>-failure.png` whenever a test fails.
- Applies `@get:Rule val screenshotOnFailure = FailureScreenshotRule()` in both `PixelCameraOverlayE2ETest` (which previously had no screenshot capture at all) and `GalleryButtonVisualE2ETest` (where it covers failures in setup and non-visual assertions not already covered by `Screenshot.saveForArtifact`).
- The CI artifact upload step already collects the `screenshots/` directory, so failure images are automatically available as build artifacts with no CI changes needed.

Fixes #187

## How the changes fix the issue

Before this PR, a failed assertion in `PixelCameraOverlayE2ETest` left no visual artifact — there was no way to tell what was on screen when the test failed. `GalleryButtonVisualE2ETest` saved named screenshots for its own visual assertions but had no fallback for failures in `@Before` setup or non-visual assertions.

`FailureScreenshotRule` is a `TestWatcher` whose `failed()` hook fires after any test method exits with an uncaught exception or assertion failure. It captures the screen at the exact moment of failure and writes it to the directory the CI upload step already monitors. No changes to CI workflow are needed.

## Test plan

- [x] `./gradlew assembleDebug` — BUILD SUCCESSFUL
- [x] `./gradlew test` — all unit tests pass
- [ ] E2E: intentionally fail one test in each class on an emulator and verify `<ClassName>-<methodName>-failure.png` appears in the CI artifacts

EOF
)

---

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/FailureScreenshotRule.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/FailureScreenshotRule.kt
@@ -1,0 +1,36 @@
+package com.gb4pc.e2e
+
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.uiautomator.UiDevice
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+import java.io.File
+
+/**
+ * JUnit [TestWatcher] rule that captures a diagnostic screenshot whenever a test fails.
+ *
+ * The screenshot is saved to the app's external files directory under a "screenshots"
+ * subdirectory — the same location used by [com.gb4pc.e2e.visual.Screenshot.saveForArtifact] —
+ * so that CI artifact pickup collects it automatically alongside any other screenshots produced
+ * during the test.
+ *
+ * File name format: `<ClassName>-<methodName>-failure.png`
+ *
+ * Apply in any E2E test class with:
+ * ```kotlin
+ * @get:Rule
+ * val screenshotOnFailure = FailureScreenshotRule()
+ * ```
+ */
+class FailureScreenshotRule : TestWatcher() {
+    override fun failed(e: Throwable?, description: Description) {
+        val dir = File(
+            InstrumentationRegistry.getInstrumentation().targetContext
+                .getExternalFilesDir(null), "screenshots"
+        )
+        dir.mkdirs()
+        val className = description.testClass?.simpleName ?: "UnknownClass"
+        UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+            .takeScreenshot(File(dir, "$className-${description.methodName}-failure.png"))
+    }
+}
--- a/app/src/androidTest/java/com/gb4pc/e2e/FailureScreenshotRule.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/FailureScreenshotRule.kt
@@ -1,5 +1,6 @@
 package com.gb4pc.e2e
 
+import android.util.Log
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.uiautomator.UiDevice
 import org.junit.rules.TestWatcher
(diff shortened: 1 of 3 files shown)


aunger (commented):
The structure is sound — `TestWatcher` is the right hook, the file path matches the existing artifact location, and the rule is applied cleanly in both test classes. Two issues need to be addressed before this merges.

## 1. `takeScreenshot` return value is silently discarded

`UiDevice.takeScreenshot(File)` returns `false` when it cannot write the screenshot (permission denied, storage unavailable, etc.). The current implementation discards that result with no log message. When this happens in CI:

- The test fails.
- No screenshot artifact appears.
- There is no indication anywhere that the capture was even attempted, let alone that it failed.

This defeats the diagnostic purpose of the rule. The return value should be checked, and a failure should produce at minimum a logcat warning that names the file path that could not be written. That way a developer inspecting a failing run can distinguish "screenshot was captured" from "screenshot capture itself also failed."

## 2. `getExternalFilesDir(null)` can return null

`Context.getExternalFilesDir()` is documented to return `null` when external storage is not available. Passing `null` as the first argument to `File(File?, String)` throws `NullPointerException`. JUnit suppresses exceptions thrown from `TestWatcher` callbacks so this would not mask the original test failure, but the screenshot attempt would crash silently — same observability problem as issue 1, with no artifact and no log message.

A null check (with a logged warning on the null path) before constructing `dir` would make the rule robust against this case.

---

Both issues are variations of the same pattern: silent failure in a tool whose entire job is to make failures visible. Neither requires structural changes to the rule.

**Verdict: changes needed** (posted as COMMENT because GitHub does not allow REQUEST_CHANGES on your own PR).

---

aunger (commented):
Both issues from the first pass are correctly resolved.

**Issue 1 (takeScreenshot return value):** Fixed. The `success` variable captures the return value and a `Log.w` with the full file path is emitted on `false`. The message is specific enough to distinguish a capture failure from a missing artifact in CI logs.

**Issue 2 (getExternalFilesDir null):** Fixed. The null check is in place before `dir` is constructed, the method returns early, and a `Log.w` names the failing test method. No NullPointerException path remains.

The implementation is correct. This is ready to merge.

_(Posting as COMMENT; GitHub does not permit APPROVE on your own PR.)_

---

PR #193, opened 2026-05-19
Apply squircle clip via SquircleDrawable, bypassing launcher icon-mask

### LINKED ISSUE #188: make squircular overlays
test1c — adaptive-icon outer shape is CIRCLE, expected SQUIRCLE

`ShapeMatcher.requireShape(outer, Shape.SQUIRCLE)` fails with `winner=CIRCLE (IoU=0.993)`. The Android launcher on the google_apis API-35 emulator clips adaptive icons to a circle, not the squircle mask used on Pixel devices.

This isn't a test error. Update the GB4PC app to make squircular overlays even if the host device doesn't use them in its launcher.

aunger:
$(cat <<'EOF'
## Summary

- Introduces `SquircleDrawable`, a `Drawable` wrapper that clips its content to a superellipse path (n=4), matching the `ShapeTemplates.squircle` formula used by the E2E tests
- `OverlayManager` now wraps all gallery icons and photo thumbnails in `SquircleDrawable` and switches to `FIT_XY` scale so the drawable fills the full view bounds
- For `AdaptiveIconDrawable`, `SquircleDrawable` draws the background and foreground layers **directly**, bypassing `AdaptiveIconDrawable.draw()`'s internal launcher mask entirely

## Why clipToOutline was insufficient

The E2E test `test1c` failed with `winner=CIRCLE (IoU=0.993)` on the `google_apis` API-35 emulator. The root cause: `AdaptiveIconDrawable.draw()` applies the device launcher's icon mask **before** drawing to the canvas. On `google_apis` the mask is a circle, so the icon content (YELLOW background + BLUE foreground) is already circular by the time `clipToOutline` gets a chance to clip it. A circular bitmap fitting inside a rounded-rect outline leaves the outer boundary circular — the `clipToOutline` rounded-rect had no visible effect.

## Fix

`SquircleDrawable` intervenes at the drawable level:
1. It clips the canvas to a superellipse path (same formula as `ShapeTemplates.squircle`, i.e. `|2x/w−1|^4 + |2y/h−1|^4 ≤ 1`).
2. For `AdaptiveIconDrawable` (API 26+): draws `background` and `foreground` layers directly, skipping the `draw()` method that applies the launcher mask.
3. For any other drawable: draws it with only the superellipse clip applied.

This guarantees a squircle outer boundary regardless of what the device launcher's icon-mask shape is.

## Test plan

- [ ] Build locally: `./gradlew assembleDebug` — passes (no compile errors)
- [ ] Unit tests: `./gradlew :app:testDebugUnitTest` — all pass (21/21 overlay tests, 0 failures)
  - `SquircleDrawableTest` (6 new tests): superellipse path classifies as SQUIRCLE, not CIRCLE
  - `OverlayManagerRobolectricTest` (updated 3 tests): overlay view holds a `SquircleDrawable`; thumbnail also wrapped in `SquircleDrawable`; adaptive icon wrapped inside `SquircleDrawable`
- [ ] E2E: `test1c_overlayOuterIsSquircle` should now pass — the outer silhouette is clipped by `SquircleDrawable` to a superellipse, independent of the `google_apis` circular launcher mask

Fixes #188

EOF
)

---


aunger (commented):
## Review

The core approach is sound: `SquircleDrawable` correctly bypasses the launcher mask by drawing `AdaptiveIconDrawable` layers directly, and the superellipse path formula matches `ShapeTemplates.squircle`. The tests are well-constructed. One real regression and one dead-code issue need to be addressed before merging.

---

### Required: `FIT_XY` distorts non-square thumbnails

```kotlin
imageView.scaleType = ImageView.ScaleType.FIT_XY
```

`FIT_XY` stretches the drawable to fill the full view bounds without preserving aspect ratio. For the gallery icon (square by convention) this is harmless, but for the photo thumbnails loaded by `showLatestPhotoThumbnail()`, landscape or portrait photos will be distorted into a square before the squircle clip is applied. `FIT_CENTER` was correct before and should be retained. `SquircleDrawable` clips the canvas — it does not require the content to fill every pixel of the view bounds.

---

### Minor: `paint` field is declared but never used

```kotlin
@Suppress("unused")
private val paint = Paint(Paint.ANTI_ALIAS_FLAG)
```

This field allocates a `Paint` on every `SquircleDrawable` construction and is never referenced in `draw()` or anywhere else. The `@Suppress("unused")` masks the warning but leaves dead code. Either remove it, or add a comment explaining why it is being kept.

---

### Observation (no change required): redundant `lineTo` before `close()`

```kotlin
path.lineTo(firstX, firstY)
path.close()
```

`Path.close()` already draws a line back to the start of the current contour, so the explicit `lineTo(firstX, firstY)` is redundant. Not a bug, just noise.

---

The adaptive-icon masking bypass logic, the path caching on bounds change, the `canvas.save()`/`restoreToCount` guards, and the test coverage (both `SquircleDrawableTest` and the updated `OverlayManagerRobolectricTest`) are all correct. Once `FIT_XY` is reverted to `FIT_CENTER` (or `CENTER_CROP` if that is preferred for thumbnails) this should be ready to merge.

---

aunger (commented):
## Second-pass review — all first-pass issues resolved, implementation correct

All three first-pass issues are fully resolved. The implementation is correct and the tests are adequate. Verdict: **ready to merge**.

---

### First-pass issues — confirmed fixed

1. **`FIT_XY` → `FIT_CENTER`** — `imageView.scaleType = ImageView.ScaleType.FIT_CENTER` is present. Non-square thumbnails will not be distorted.
2. **Dead `paint` field** — no `paint` field, no `@Suppress("unused")`, no `import android.graphics.Paint` anywhere in `SquircleDrawable.kt`.
3. **Redundant `lineTo` before `close()`** — `buildSuperellipsePath` goes directly from the loop to `path.close()`. The `firstX`/`firstY` variables are gone.

---

### Implementation correctness

- **Superellipse path**: parametric form with `EXP = 0.5` (n = 4) and 256 steps matches `ShapeTemplates.squircle`. Correct.
- **Bounds / path caching**: `onBoundsChange` invalidates `pathWidth`/`pathHeight`; `rebuildPathIfNeeded` skips the rebuild when dimensions are unchanged. Correct.
- **`draw()`**: early return on `b.isEmpty`; translates to `(b.left, b.top)` so the path origin aligns with the drawable's position in the canvas coordinate space; sets layer/inner bounds to a local `Rect(0, 0, w, h)` after translation. Correct.
- **Canvas save/restore**: `canvas.save()` + `canvas.restoreToCount(save)` inside a `finally` block — the clip and translate are always undone even on exceptions. Correct.
- **AdaptiveIconDrawable layer bypass**: `bg.bounds = localBounds; bg.draw(canvas)` / `fg.bounds = localBounds; fg.draw(canvas)` with null-safe `?.let` calls. This correctly skips `AdaptiveIconDrawable.draw()` and its internal launcher mask. Correct.
- **`setAlpha` / `setColorFilter`**: delegate to `inner` and call `invalidateSelf()`. Correct.
- **`getIntrinsicWidth/Height`**: delegate to `inner`. Correct.
- **`getOpacity()`**: returns `TRANSLUCENT`, appropriate for a shape-clipped drawable.

One observation (no change required): `draw()` sets `inner.bounds` (or layer bounds) on every call, mutating the inner drawable's bounds as a side effect. For this codebase the inner drawable is always dedicated to its `SquircleDrawable` wrapper, so there is no practical risk.

---

### Test adequacy

**`SquircleDrawableTest`** (6 tests): SQUIRCLE classification at 64×64 and 128×128, not-CIRCLE assertions at both sizes, non-empty path for 100×100, and pixel-count parity with `ShapeTemplates.squircle` within 5%. This directly validates the formula against the same template used by the E2E test that was failing. Good coverage.

**`OverlayManagerRobolectricTest`** (3 updated tests): verifies `SquircleDrawable` wrapping for the gallery icon, the photo thumbnail, and the adaptive-icon-specific case (inner must be `AdaptiveIconDrawable`, not a pre-masked `BitmapDrawable`). Removal of the zero-dimension `outlineProvider` test is appropriate — that guard now lives in `SquircleDrawable.draw()`'s `if (b.isEmpty) return`. The deleted `clipToOutline` / `Outline` infrastructure tests no longer map to any live code and were right to remove.

---

No remaining concerns.

---

---

PR #196, opened 2026-05-20
Add test-result pass/fail summary to CI job summary

### LINKED ISSUE #195: Test case pass-fail summary CI output step
## problem
Scanning the test execution steps to see which test-cases failed is difficult, because they're so verbose.
The test result transfer steps have more focused output, but require a download and unzip. 

## one possible solution
A workflow step that merely collects, formats, and displays the tree of test classes and tests, with simple PASS/FAIL status would allow an overview at a glance.

## but what else?
Is there a more typical or idiomatic way to show test red-green statuses?

## recommendation

**Use GitHub Job Summaries (`$GITHUB_STEP_SUMMARY`)** — the idiomatic GitHub-native approach for at-a-glance pass/fail status within a run.

The main alternative is a third-party action such as `dorny/test-reporter` or `EnricoMi/publish-unit-test-result-action`. These post per-test annotations directly on the PR diff, which is more visible — but they require `checks: write` (a non-trivial permission escalation) and silently fail for fork PRs due to GitHub's token restrictions. They also introduce third-party actions into the trusted supply chain.

Job Summaries avoid all of that: zero permission changes, no third-party actions, and fork PRs work identically to branch PRs. The output appears in the "Summary" tab of every CI run — one click from the status badge.

### implementation plan

**New file: `scripts/summarize_test_results.py`**

A small Python script (consistent with the existing `scripts/check_green_feed.py` pattern) that:
- Accepts one or more result directory paths and `--suite-label` names as arguments
- Parses `TEST-*.xml` JUnit files using stdlib `xml.etree.ElementTree` (no extra dependencies)
- Groups test cases by class name; marks a class PASS if all its tests pass, FAIL if any fail
- Writes a Markdown table to `$GITHUB_STEP_SUMMARY` (falls back to stdout when run locally)
- Exits 0 always — display only; failures are already surfaced by the test steps themselves
- Handles missing/empty directories gracefully (prints a note, does not fail the step)

Example output in the Summary tab:

```
## Test Results

### Unit Tests
| Status | Suite / Test |
|--------|--------------|
| ✅ PASS | **com.gb4pc.service.OverlayServiceLogicTest** |
| ✅ PASS |     `cameraOpenShowsOverlay` |
| ❌ FAIL | **com.gb4pc.data.PrefsManagerTest** |
| ❌ FAIL |     `readDefaultReturnsNull` |
| ✅ PASS |     `writeAndReadRoundtrip` |

**Total: 42 passed, 1 failed** (43 tests)
```

**Modified file: `.github/workflows/build.yml`**

Add a single consolidated step at the end of the `build-and-test` job (guarded with `if: always() && steps.diff_check.outputs.needs_full_build == 'true'`) that passes all three result directories to the script:

- `app/build/test-results/testDebugUnitTest` (unit tests)
- `app/build/outputs/androidTest-results/connected/debug` (instrumented + E2E tests)

No permissions changes required.

aunger:
## Summary

Fixes #195

- Adds `scripts/summarize_test_results.py`, which parses JUnit XML files from one or more result directories and writes a pass/fail Markdown table to `$GITHUB_STEP_SUMMARY` (falls back to stdout when the env var is absent).
- Adds a &#34;Write test-result summary&#34; step at the end of the `build-and-test` job that calls the script for both the unit test and instrumented test result directories.
- Adds `scripts/test_summarize_test_results.py` with 28 unit tests covering parsing, rendering, CLI argument validation, and summary-file writing.

## How it fixes the issue

CI output was too verbose to scan for failures, and the artifact-based test results required a download and unzip. The new step populates the GitHub Actions &#34;Summary&#34; tab with a compact table showing each test class and case as ✅ PASS or ❌ FAIL, plus a total line, making failures immediately visible without leaving the browser.

The step is guarded with `if: always() &amp;&amp; steps.diff_check.outputs.needs_full_build == &#39;true&#39;` so it runs even when earlier steps fail, and exits 0 always — actual failure gating remains with the existing test steps.

## Test plan

- [ ] Run `python3 -m unittest discover -s scripts -p &#34;test_summarize_test_results.py&#34;` — all 28 tests pass.
- [ ] Trigger a CI run and verify the &#34;Summary&#34; tab shows the pass/fail table.
- [ ] Confirm the step does not fail when result directories are absent (e.g. on docs-only PRs that skip the full build).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -344,6 +344,15 @@ jobs:
           path: /tmp/e2e-screenshots/
           retention-days: 14
 
+      - name: Write test-result summary
+        if: always() && steps.diff_check.outputs.needs_full_build == 'true'
+        run: |
+          python3 scripts/summarize_test_results.py \
+            app/build/test-results/testDebugUnitTest \
+              --suite-label "Unit Tests" \
+            app/build/outputs/androidTest-results/connected/debug \
+              --suite-label "Instrumented Tests"
+
   # Produce a signed release-config APK on every merge to main for easy sideloading.
   # Named gb4pc-dev.<run_number>.apk to distinguish from tagged releases.
   # Waits for the full test suite (unit + E2E) before producing an artifact.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,6 +105,9 @@ jobs:
           echo "==> emulator -list-avds:"
           ANDROID_AVD_HOME="$AVD_HOME" $ANDROID_HOME/emulator/emulator -list-avds
 
+      - name: Run Python script unit tests
+        run: python3 -m unittest discover -s scripts -p "test_*.py" -v
+
       - name: Grant execute permission for gradlew
         if: steps.diff_check.outputs.needs_full_build == 'true'
         run: chmod +x gradlew
--- a/scripts/summarize_test_results.py
+++ b/scripts/summarize_test_results.py
@@ -1,0 +1,234 @@
+#!/usr/bin/env python3
+"""Summarize JUnit XML test results as a GitHub Job Summary Markdown table.
+
+Reads TEST-*.xml files from one or more result directories, groups test cases
+by class name, and writes a pass/fail table to $GITHUB_STEP_SUMMARY (falling
+back to stdout when the env var is not set).
+
+Usage:
+    python3 scripts/summarize_test_results.py \\
+        path/to/unit-results       --suite-label "Unit Tests" \\
+        path/to/e2e-results        --suite-label "Instrumented Tests"
+
+Each directory path must be immediately followed by --suite-label <name>.
+Exit code is always 0 (display only; failures are surfaced by earlier steps).
+"""
+
+import os
+import sys
+import xml.etree.ElementTree as ET
+from dataclasses import dataclass, field
+from pathlib import Path
+
+
+# ---------------------------------------------------------------------------
+# Data model
+# ---------------------------------------------------------------------------
+
+@dataclass
+class TestCase:
+    name: str
+    passed: bool
+
+
+@dataclass
+class TestClass:
+    name: str
+    cases: list[TestCase] = field(default_factory=list)
+
+    @property
+    def passed(self) -> bool:
+        return all(tc.passed for tc in self.cases)
+
+
+# ---------------------------------------------------------------------------
+# Parsing
+# ---------------------------------------------------------------------------
+
+def parse_directory(directory: Path) -> dict[str, TestClass]:
+    """Parse all TEST-*.xml files in *directory* and return classes keyed by name."""
+    classes: dict[str, TestClass] = {}
+
+    xml_files = sorted(directory.glob("TEST-*.xml"))
+    for xml_path in xml_files:
+        try:
+            tree = ET.parse(xml_path)
+        except ET.ParseError as exc:
+            print(f"  Warning: could not parse {xml_path}: {exc}", file=sys.stderr)
+            continue
+
+        root = tree.getroot()
+        if root.tag == "testsuites":
+            suite_elements = root.findall("testsuite")
+        else:
+            suite_elements = [root]
+
+        for suite in suite_elements:
+            class_name = suite.get("name", xml_path.stem)
+            if class_name not in classes:
+                classes[class_name] = TestClass(name=class_name)
+
+            for tc in suite.findall("testcase"):
+                tc_name = tc.get("name", "<unknown>")
+                failed = (
+                    tc.find("failure") is not None
+                    or tc.find("error") is not None
+                )
+                classes[class_name].cases.append(
+                    TestCase(name=tc_name, passed=not failed)
+                )
+
+    return classes
+
+
+# ---------------------------------------------------------------------------
+# Markdown rendering
+# ---------------------------------------------------------------------------
+
+def render_suite(label: str, classes: dict[str, TestClass]) -> list[str]:
+    """Render one suite as Markdown lines (header + table rows + totals)."""
+    lines: list[str] = []
+    lines.append(f"### {label}")
+
+    if not classes:
+        lines.append("_No test results found._")
+        lines.append("")
+        return lines
+
+    lines.append("| Status | Suite / Test |")
+    lines.append("|--------|--------------|")
+
+    total_pass = 0
+    total_fail = 0
+
+    for cls in classes.values():
+        class_icon = "✅ PASS" if cls.passed else "❌ FAIL"
+        lines.append(f"| {class_icon} | **{cls.name}** |")
+        for tc in cls.cases:
+            tc_icon = "✅ PASS" if tc.passed else "❌ FAIL"
+            lines.append(f"| {tc_icon} |     `{tc.name}` |")
+            if tc.passed:
+                total_pass += 1
+            else:
+                total_fail += 1
+
+    total = total_pass + total_fail
+    lines.append("")
+    lines.append(f"**Total: {total_pass} passed, {total_fail} failed** ({total} tests)")
+    lines.append("")
+    return lines
+
+
+def build_markdown(suite_data: list[tuple[str, dict[str, TestClass]]]) -> str:
+    """Build the full Markdown document for all suites."""
+    lines: list[str] = ["## Test Results", ""]
+    for label, classes in suite_data:
+        lines.extend(render_suite(label, classes))
+    return "\n".join(lines)
+
+
+# ---------------------------------------------------------------------------
+# CLI
+# ---------------------------------------------------------------------------
+
+_USAGE = (
+    "usage: summarize_test_results.py "
+    "<dir> --suite-label <name> [<dir> --suite-label <name> ...]"
+)
+
+
+def parse_args(argv: list[str]) -> list[tuple[Path, str]]:
+    """Parse argv into a list of (directory, label) pairs.
+
+    Expected pattern: <dir> --suite-label <name> [<dir> --suite-label <name> ...]
+
+    Raises SystemExit(2) on bad input (matching argparse convention).
+    """
+    if argv and argv[0] in ("-h", "--help"):
+        print(__doc__)
+        print(_USAGE)
+        raise SystemExit(0)
+
+    pairs: list[tuple[Path, str]] = []
+    tokens = list(argv)
+    i = 0
+    while i < len(tokens):
+        directory = tokens[i]
+        # We need tokens[i], tokens[i+1], and tokens[i+2] to all be present.
+        if i + 3 > len(tokens):
+            print(
+                f"error: expected --suite-label <name> after '{directory}'\n{_USAGE}",
+                file=sys.stderr,
+            )
+            raise SystemExit(2)
+        flag = tokens[i + 1]
+        label = tokens[i + 2]
+        if flag != "--suite-label":
+            print(
+                f"error: expected --suite-label after '{directory}', got '{flag}'\n{_USAGE}",
+                file=sys.stderr,
+            )
+            raise SystemExit(2)
+        pairs.append((Path(directory), label))
+        i += 3
+
+    if not pairs:
+        print(
+            f"error: at least one <directory> --suite-label <name> pair is required\n{_USAGE}",
+            file=sys.stderr,
+        )
+        raise SystemExit(2)
+
+    return pairs
+
+
+# ---------------------------------------------------------------------------
+# Entry point
+# ---------------------------------------------------------------------------
+
+def main(argv: list[str] | None = None) -> int:
+    if argv is None:
+        argv = sys.argv[1:]
+
+    pairs = parse_args(argv)
+
+    suite_data: list[tuple[str, dict[str, TestClass]]] = []
+    for directory, label in pairs:
+        if not directory.exists():
+            print(
+                f"Note: result directory not found, skipping: {directory}",
+                file=sys.stderr,
+            )
+            suite_data.append((label, {}))
+            continue
+        if not directory.is_dir():
+            print(
+                f"Note: path is not a directory, skipping: {directory}",
+                file=sys.stderr,
+            )
+            suite_data.append((label, {}))
+            continue
+
+        classes = parse_directory(directory)
+        if not classes:
+            print(
+                f"Note: no TEST-*.xml files found in {directory}",
+                file=sys.stderr,
+            )
+        suite_data.append((label, classes))
+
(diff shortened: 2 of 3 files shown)


aunger (commented):
The implementation is correct and the 28 tests cover the meaningful paths well. A few notes:

**`parse_directory` merges same-named suites across files.** If two XML files both declare `<testsuite name="com.example.FooTest">`, their cases merge into one `TestClass` entry. This is reasonable behaviour for real JUnit output where a suite can be split across files, and the code handles it gracefully — just worth being aware of.

**Hand-rolled CLI parser.** Using a positional-interleaving pattern (`<dir> --suite-label <name>`) instead of `argparse` is unconventional but defensible: `argparse` doesn't natively express this repeating heterogeneous pattern without awkward workarounds. The error messages are clear, and the tests cover the failure modes.

**Workflow `if` guard is correct.** `always() && needs_full_build == 'true'` ensures the step runs on test failure (the important case) while still being skipped on docs-only builds. Exit code is always 0 — failure gating stays with the existing test steps. Good.

**Minor: empty `TestClass` (zero cases) shows ✅ PASS.** `all()` on an empty iterable is `True`, so a class parsed with no `<testcase>` children gets a green icon. This won't happen with well-formed JUnit XML, and the behaviour is documented by the existing test `test_empty_cases_passes`. Not a blocker.

Overall the script is clean, well-separated into parse/render/CLI layers, and the CI integration is minimal and non-intrusive. This is a solid implementation of #195 — ready to merge.

---

aunger (commented):
**Verdict: approve** (GitHub prevents self-approval, so posting as a comment.)

The implementation is correct and the test coverage is thorough. A few observations:

**`Run Python script unit tests` placement is sound.** The step has no `if:` guard, so it runs on every CI invocation — including docs-only PRs that skip the full build. That is the right choice: these are lightweight pure-Python tests with no Android toolchain dependency.

**The `Write test-result summary` guard is correct.** `if: always() && steps.diff_check.outputs.needs_full_build == 'true'` ensures the summary is written even when earlier build or test steps fail, while still skipping it on docs-only runs where no XML results exist. The script's own handling of missing directories (prints a note, appends an empty-results section, exits 0) means CI never fails because of this step.

**`build_markdown([])` is reachable only from tests, not from the CLI.** `parse_args` enforces at least one pair, so the empty-list path is dead code in production. The test calling it directly is fine as a unit boundary, but worth noting that this test exercises a path the CLI cannot reach.

**Append mode for `GITHUB_STEP_SUMMARY` is correct** and matches the GitHub Actions documentation for multi-step summaries.

**Minor: the PR description carries a "Generated by Claude Code" footer.** The repo's no-bylines rule covers reviews and commits; this does not block merging.

Overall the change is clean, well-scoped, and ready to merge.

---

aunger (commented):
The `summarize_test_results.py` script and its 28 tests are solid. The XML parsing handles both `<testsuite>` and `<testsuites>` roots, the always-0 exit is correct by design, the `$GITHUB_STEP_SUMMARY` append is the right approach, and test coverage is thorough. The CI YAML step guard (`if: always() && needs_full_build == 'true'`) is correct.

The emulator wake/unlock fix in commit 4 has two problems that need addressing before this merges.

---

**1. KEYCODE_MENU is not a reliable keyguard dismissal on modern Android**

`KEYCODE_MENU (82)` was historically used to dismiss the keyguard, but this stopped working reliably in Android 8 (API 26). On API 26+, the correct no-PIN keyguard dismissal is:

```
adb shell wm dismiss-keyguard
```

or, if you need to keep the `input keyevent` approach for compatibility, `KEYCODE_HOME` is what the platform actually acts on at the lock screen. `KEYCODE_MENU` on the lock screen is a no-op on API 26+ — the screen wakes up but the keyguard stays up, so the screencap still captures the keyguard overlay and the green-view check keeps failing.

What API level do the emulators in CI target? If it is 26 or above (virtually certain for a Pixel Camera target), this fix does not solve the problem it was introduced to solve.

**2. KEYCODE_MENU fired while the app is already in the foreground can disrupt the check**

The two keyevents are sent before *every* screencap, not just on the first attempt or only when the screen is actually locked. On attempts where the screen is already on and `MockCameraActivity` is in the foreground, `KEYCODE_MENU` may trigger the activity's options menu on Android ≤ 9, partially obscuring the green view. That would cause the coverage check to fail on those attempts even though the activity launched successfully.

A safer fix would gate the wake/unlock calls on the first attempt only, or use `adb shell wm dismiss-keyguard` which is a no-op when the screen is already unlocked.

**3. The new test does not verify ordering**

`test_wakeup_keyevents_sent_before_each_screencap` checks that a `keyevent 224` call and a `keyevent 82` call appear *somewhere* in the subprocess call list, but does not verify they appear *before* the `exec-out screencap` call. The test would pass even if the keyevents were sent after the screencap. The assertion should check relative position in `mock_run.call_args_list`.

---

aunger (commented):
## Prior concerns

All three are resolved:

1. `KEYCODE_MENU` is gone; `wm dismiss-keyguard` is in its place with the correct call order (wakeup → dismiss → screencap).
2. No risk of triggering the options menu — `wm dismiss-keyguard` is a no-op when no keyguard is present.
3. The test now asserts strict ordering: for each screencap call at index `i`, it checks that index `i-1` is `wm dismiss-keyguard` and index `i-2` is KEYCODE_WAKEUP (224). The two-iteration setup (fail once, pass once) exercises this across multiple retry cycles.

## One change needed — attribution byline in PR body

The PR description ends with:

> _Generated by [Claude Code](https://claude.ai/code/session_017fkV5M6GzguBV1Tq64hSE1)_

`pr_participation.md` is explicit: "Do not append any attribution byline, session URL, or 'Generated by Claude' footer to any PR review, comment, or reply." The same rule applies to the PR description. Please remove this footer before merging.

## Everything else looks good

The `summarize_test_results.py` logic is clean: the `parse_directory` / `render_suite` / `build_markdown` layering is clear, the CLI argument parser follows the right convention, the exit-0-always contract is sound, and all 28 new unit tests are thorough. The "Run Python script unit tests" step running unconditionally (no `needs_full_build` guard) is the right call for a cheap, fast script-level check.

---

aunger:
**Verdict: approve** (GitHub prevents self-approval, so posting as a comment.)

All prior concerns are resolved:

1. The attribution byline has been removed from the PR description.
2. `KEYCODE_MENU` is replaced with `wm dismiss-keyguard`, which is a no-op when no keyguard is present and works correctly on API 26+. No risk of accidentally triggering an options menu.
3. The ordering test now strictly verifies that for each screencap call at index `i`, the call at `i-1` is `wm dismiss-keyguard` and the call at `i-2` is `KEYCODE_WAKEUP (224)`. The two-iteration setup (fail once, pass once) exercises ordering across multiple retry cycles, not just the first.

The `summarize_test_results.py` script is well-structured across parse/render/CLI layers, the always-0 exit contract is correct, and all 28 unit tests are thorough. The unconditional `Run Python script unit tests` step is the right call for a cheap, toolchain-free check. The `Write test-result summary` guard (`if: always() && needs_full_build == 'true'`) is correct.

Ready to merge.

---

aunger (commented):
**Verdict: changes needed before merging.**

The branch is clean — three files changed, all on-topic, no `check_green_feed.py` changes, no emulator-fix code. The script design, CLI parsing, and markdown rendering are solid, and the test coverage is thorough. One correctness gap requires a fix.

## Required change: skipped tests are silently counted as passing

JUnit XML uses a `<skipped/>` child element to mark skipped tests. The parser's failure detection is:

```python
failed = (
    tc.find("failure") is not None
    or tc.find("error") is not None
)
```

A `<testcase>` with `<skipped/>` has neither `<failure>` nor `<error>`, so `failed` is `False` and the case is recorded as `passed=True`. The summary will show ✅ PASS for every skipped test and count them in the passing total.

This matters in practice: Android unit test runs frequently include `@Ignore`-annotated tests or tests skipped via `assumeTrue`, which Gradle reports with `<skipped/>`. A run that aborts mid-suite also produces `<skipped/>` entries for tests that never ran. Showing those as green is misleading and defeats the purpose of the summary.

The fix is to detect the `<skipped/>` element and represent those tests distinctly — either as a third outcome (⏭ SKIP) or by excluding them from both pass and fail counts while noting them in the totals line. A totals line of the form `2 passed, 0 failed, 1 skipped` is probably the clearest output and requires minimal changes.

A test covering the skipped-element path should accompany the fix.

---

## Minor observations (no changes required)

**Vacuously-passing empty class rows**: `TestClass.passed` returns `True` when `cases` is empty, so a `<testsuite>` with no `<testcase>` children gets a ✅ PASS class row. Logically correct and harmless in practice.

**Step ordering in `build.yml`**: The unconditional "Run Python script unit tests" step placed before the `gradlew` steps is fine and the commit message explains the intent clearly.

**`parse_args` accepts `--`-prefixed labels**: A label value starting with `--` would be accepted as-is. Not a practical concern given the workflow hardcodes the labels.

---

aunger:
Try to remove the FooTest, BarTest cases that test the tests themselves from the final summary report

aunger (commented):
The skipped-test fix is largely correct: skipped cases are now parsed with `passed=False, skipped=True`, the pass count is no longer inflated, and the new tests verify both properties at the `TestCase` level. A few issues remain.

**Correctness gap: `TestClass.passed` is now dead code in rendering**

`render_suite` correctly switched from `cls.passed` to `cls.any_failed` for the class-row icon (line 115). As a result, `TestClass.passed` is no longer called from any production path — only from the `TestTestClass` unit tests. If `passed` is no longer the decision-making predicate, it should be removed; otherwise the two properties can silently diverge as the model evolves and the unit tests for `passed` give false confidence in something that isn't wired to the output.

**Missing render-level test for skipped counts**

The two new tests (`test_skipped_element_not_counted_as_pass`, `test_skipped_cases_do_not_inflate_pass_count`) verify parse-time data structure but do not exercise `render_suite` with a skipped case. There is no test asserting that:
- the skipped case row shows `⏭ SKIP`
- the totals line reads `1 passed, 1 failed, 1 skipped` (the exact string the summary displays)
- the class-row icon is `✅ PASS` when the only failure is skipped

Given that the prior reviewer concern was specifically about inflated green counts in the rendered output, a render-level test against `SKIPPED_XML` would be the most direct regression guard.

**Minor: all-skipped class silently shows green**

A class whose every test is skipped returns `passed=True` / `any_failed=False` and displays ✅ PASS. If CI skips an entire test class (e.g. due to a missing device), the summary will show it green with no indication it ran. Worth either a comment explaining the intent or a test pinning the behavior.

The core parsing fix is correct. The two substantive issues are: (1) `TestClass.passed` is dead in production and should be removed rather than tested as if it matters, and (2) no render-level test covers the skipped-count summary line.

---

aunger (commented):
**Verdict: approve.**

Both prior blockers are resolved.

**Dead code removed.** `TestClass.passed` is gone from `summarize_test_results.py`. The class now exposes only `any_failed`, which is the predicate `render_suite` actually uses. No risk of the two properties silently diverging.

**Render-level skipped tests are present.** Three new tests in `TestRenderSuite` directly exercise `render_suite` with skipped cases:
- `test_skipped_case_shows_skip_icon` — confirms ⏭ SKIP appears and ❌ FAIL does not
- `test_skipped_case_counted_in_totals_not_as_pass` — pins the exact `1 passed, 1 failed, 1 skipped` / `(3 tests)` string
- `test_all_skipped_class_icon_is_pass` — documents the all-skipped class → ✅ PASS behavior

The `TestTestClass` suite now covers `any_failed` across the relevant cases (all pass, one failure, empty, all skipped). The four `TestRenderSuite` skipped-path tests combined with the `TestParseDirectory` skipped-element tests form a complete regression guard for the originally reported inflated-green issue.

Everything else noted in earlier rounds stands: parse/render/CLI layering is clean, the always-0 exit is correct, the `$GITHUB_STEP_SUMMARY` append mode is right, and the workflow guard (`if: always() && needs_full_build == 'true'`) is sound. Ready to merge.

---

aunger (commented):
**Verdict: approve** (GitHub prevents self-approval on this account, so posting as a comment.)

The env-isolation fix is correct across all cases.

**`setUp`/`tearDown` correctness**

`os.environ.pop("GITHUB_STEP_SUMMARY", None)` safely handles both the "var present" and "var absent" cases without risking a `KeyError`. The guard `if self._saved_summary is not None` in `tearDown` correctly avoids assigning `None` back into `os.environ` (which would raise `TypeError`). The logic is sound.

**New test meaningfulness**

`test_main_does_not_write_to_ci_summary_when_no_summary_var` is a worthwhile addition — it directly asserts the no-file-created property rather than relying on `test_falls_back_to_stdout_when_no_env_var` as a proxy. The two tests are complementary, not redundant.

**`test_writes_to_github_step_summary` interaction with `setUp`**

`patch.dict` adds `GITHUB_STEP_SUMMARY` for the duration of the `with` block and removes it afterward. Since `setUp` already cleared the key, `patch.dict`'s restore step will correctly leave the key absent. No issue here.

**Minor observation (not blocking)**

Several other `TestMain` tests that call `srt.main()` without patching stdout still let output through, while `test_main_does_not_write_to_ci_summary_when_no_summary_var` suppresses it. Consistent suppression across the class would be cleaner, but this is cosmetic.

**Workflow step**

The unconditional `Run Python script unit tests` step placement (before the guarded build steps) is appropriate — script tests should run on every push regardless of `needs_full_build`.

The implementation is correct. Good fix.

---

aunger:
> test_all_skipped_class_icon_is_pass — documents the all-skipped class → ✅ PASS behavior

no, don’t do this. The class should appear with the skipped icon, not PASS.

aunger (commented):
The three-way icon logic is correct.

- `any_failed` uses `not tc.passed and not tc.skipped`, so a skipped test can never trigger it.
- Empty class: `any_failed` is False (empty `any()`), `any_passed` is False → `⏭ SKIP`. Correct.
- All-skipped class: same path → `⏭ SKIP`. Correct.
- Mixed pass + skip: `any_passed` is True → `✅ PASS`. Correct.
- Any failure wins immediately → `❌ FAIL`. Correct.

The renamed test `test_all_skipped_class_icon_is_skip` asserts both that `⏭ SKIP` is present in the class row and that `✅ PASS` is absent. That directly covers the bug that was fixed.

One minor coverage gap: no test explicitly checks that the class-row icon for a mixed pass+skip class is `✅ PASS`. `test_skipped_case_shows_skip_icon` checks the combined output for absence of `❌ FAIL` but does not inspect the class row's icon specifically. Given the simplicity of the branch, this is low risk. The PR is correct as written.

---

PR #197, opened 2026-05-20
Use upward swipe to dismiss swipe keyguard before each screencap in the adb retry loop

aunger:
CI logs showed every screencap in the `check_green_feed.py` pre-flight retry loop returning the lock-screen wallpaper color (#E9E8EF) instead of the expected green View color. The keyguard re-appears after the `adb install` step, so a one-time dismiss before the script runs is not enough.

## What changed and why

Both approaches tried previously have problems:

- **`wm dismiss-keyguard`**: CI evidence showed all 30 retry attempts returning the lock-screen wallpaper, meaning it did not reliably dismiss the swipe keyguard on the API-35 emulator.
- **`KEYCODE_MENU (82)`**: Once `MockCameraActivity` is in the foreground, `KEYCODE_MENU` dispatches `KeyEvent.KEYCODE_MENU` to the activity, which can open the options/overflow menu via `ActionBar`, obscuring the green view.

The fix uses `adb shell input swipe 300 1000 300 300` — an upward swipe gesture that matches what a user would do on the swipe-based emulator lock screen. This approach:

1. Reliably dismisses the swipe-type lock screen that appears before the E2E PIN setup script runs.
2. Has no side-effects on a foregrounded activity — a swipe gesture over the activity surface does not open any menus or dialogs.

The wake sequence in every retry iteration is now: `KEYCODE_WAKEUP (224)` → upward swipe → screencap.

The same change is applied to the pre-launch and post-launch wake sequence in the CI preflight step of `build.yml`, keeping both in sync.

## Tests

The unit test `test_wakeup_and_swipe_sent_before_each_screencap` verifies that every screencap call is immediately preceded by a swipe call, which is itself immediately preceded by `KEYCODE_WAKEUP`, covering the correct ordering across multiple retry iterations.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -228,6 +228,7 @@ jobs:
           echo "==> Waking display and dismissing keyguard..."
           "$ADB" shell input keyevent KEYCODE_WAKEUP
           "$ADB" shell input swipe 300 1000 300 300
+          sleep 2
 
           # Launch MockCameraActivity and wait for it to be displayed.
           # am start -W blocks until the activity is on screen, so no polling needed.
@@ -242,6 +243,7 @@ jobs:
           # same swipe unlock to stay consistent.
           "$ADB" shell input keyevent KEYCODE_WAKEUP
           "$ADB" shell input swipe 300 1000 300 300
+          sleep 2
 
           python3 scripts/check_green_feed.py --adb "$ADB" preflight.png
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -219,15 +219,15 @@ jobs:
           "$ADB" shell pm grant com.google.android.GoogleCamera \
             android.permission.CAMERA
 
-          # Wake the display and dismiss the keyguard before launching the activity.
-          # am start -W waits for the framework to launch the activity, but on an
-          # emulator with screen timeout the keyguard may still be covering the window.
-          # wm dismiss-keyguard (API 26+) removes the keyguard immediately and is more
-          # reliable than a keyevent sequence for this purpose.
+          # Wake the display and dismiss the swipe keyguard before launching the
+          # activity. An upward swipe is used instead of wm dismiss-keyguard or
+          # KEYCODE_MENU: wm dismiss-keyguard did not reliably dismiss the
+          # swipe-type lock screen in CI (all retries returned #E9E8EF), and
+          # KEYCODE_MENU can open the options menu once MockCameraActivity is
+          # foregrounded, obscuring the green view.
           echo "==> Waking display and dismissing keyguard..."
           "$ADB" shell input keyevent KEYCODE_WAKEUP
-          "$ADB" shell wm dismiss-keyguard
-          sleep 2
+          "$ADB" shell input swipe 300 1000 300 300
 
           # Launch MockCameraActivity and wait for it to be displayed.
           # am start -W blocks until the activity is on screen, so no polling needed.
@@ -235,12 +235,13 @@ jobs:
           "$ADB" shell am start -W -n \
             com.google.android.GoogleCamera/com.gb4pc.mockcamera.MockCameraActivity
 
-          # Re-wake and re-dismiss keyguard after am start -W: adb install above
-          # can take 10-30 s and the screen may sleep in that window. am start -W
-          # and setTurnScreenOn(true) usually recover, but this is a cheap extra
-          # guarantee that the green view is visible before screencap.
+          # Re-wake after am start -W: adb install above can take 10-30 s and
+          # the screen may sleep in that window. am start -W and
+          # setTurnScreenOn(true) usually recover, but this is a cheap extra
+          # guarantee that the green view is visible before screencap. Use the
+          # same swipe unlock to stay consistent.
           "$ADB" shell input keyevent KEYCODE_WAKEUP
-          "$ADB" shell wm dismiss-keyguard
+          "$ADB" shell input swipe 300 1000 300 300
 
           python3 scripts/check_green_feed.py --adb "$ADB" preflight.png
 
--- a/scripts/check_green_feed.py
+++ b/scripts/check_green_feed.py
@@ -221,6 +221,20 @@ def main() -> int:
         # Sleep first so the display has time to render before the first check
         # and between retries; the caller has already waited for am start -W.
         time.sleep(RETRY_DELAY_SECONDS)
+        # Wake the display and dismiss the keyguard before every screencap.
+        # The screen may dim or the keyguard may re-appear during a long retry
+        # loop (e.g. while the APK was being installed). KEYCODE_WAKEUP (224)
+        # turns the screen on; KEYCODE_MENU (82) dismisses the keyguard when
+        # no PIN is set (which is the case at pre-flight time, before the E2E
+        # setup script installs a PIN).
+        subprocess.run(
+            [adb_path, "shell", "input", "keyevent", "224"],
+            check=False,
+        )
+        subprocess.run(
+            [adb_path, "shell", "input", "keyevent", "82"],
+            check=False,
+        )
         with open(path, "wb") as out:
             subprocess.run(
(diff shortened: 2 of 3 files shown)


aunger (commented):
**Verdict: approved** (GitHub prevents approving your own PR, so posting as a comment.)

The fix is correct and the test is robust.

**`wm dismiss-keyguard` is the right call.** It is the documented API-26+ way to programmatically dismiss the keyguard, and AOSP guarantees it is a no-op when no keyguard is present. The PR description's reasoning for replacing `KEYCODE_MENU (82)` is accurate: `KEYCODE_MENU` is ineffective on the lock screen for API 26+ and has a side effect (opening the options menu) when the activity is foregrounded on older releases. `KEYCODE_WAKEUP (224)` + `wm dismiss-keyguard` is the right pairing.

**`check=False` on both new calls is appropriate.** These are best-effort; a failure to wake or dismiss should not abort the retry loop when the subsequent screencap will surface the problem anyway.

**The ordering test is correct.** `c.args[0]` is the first positional argument to `subprocess.run` — the command list — because all three call sites pass the command positionally. The `sc_idx - 1` / `sc_idx - 2` strict-position checks confirm the intended sequence within every iteration, not just global presence. The `side_effects = [1, 0]` pattern is fine: the loop exits on the first success, so `check_image` is called exactly twice and `StopIteration` is never reached regardless of `MAX_ATTEMPTS`.

**One minor observation (not a blocker):** The positional check (`sc_idx - 2`, `sc_idx - 1`) will silently break if a future change inserts another `subprocess.run` call between the wakeup/dismiss and the screencap — it would produce a confusing failure message. A brief comment noting that the offsets are intentionally strict about adjacency would help the next author understand why `sc_idx - 2` is load-bearing rather than a magic number. Cosmetic suggestion only; the invariant being tested is exactly right.

---

aunger (commented):
**Verdict: changes requested** (GitHub prevents requesting changes on your own PR, so posting as a comment.)

Two problems block this, one technical and one documentation.

---

**1. The `KEYCODE_MENU` "no-op when unlocked" claim is incorrect**

The comment at lines 228–231 states:

> KEYCODE_MENU (82) dismisses the swipe-based lock screen that the emulator shows before the E2E PIN setup script runs. It is a no-op when the screen is already unlocked, so it is safe to call unconditionally in the pre-flight retry loop.

This is wrong for the scenario where the screen is already unlocked. `KEYCODE_MENU` is a no-op *on the lock screen itself* — it dismisses the swipe lock screen because it sends a key event that the keyguard layer consumes. But once MockCameraActivity is in the foreground with the screen fully unlocked, `KEYCODE_MENU` dispatches `KeyEvent.KEYCODE_MENU` to the foreground activity. Android has not deprecated the delivery of that event; `Activity.onKeyDown` will receive it, and activities that delegate to an `ActionBar` will have the options/overflow menu opened. The claim that this is unconditionally a no-op in the retry loop is inaccurate and the same side-effect concern that motivated the previous rewrite applies here.

---

**2. The technical rationale for reverting `wm dismiss-keyguard` does not hold**

The previous reviewer's approval of `wm dismiss-keyguard` was correct. `adb shell wm dismiss-keyguard` is the documented programmatic API for dismissing the keyguard on API 26+. It dismisses the keyguard if it is showing, which includes insecure (swipe-based) keyguards — the command has no restriction to PIN/password keyguards. The revert's stated reason ("does not reliably dismiss swipe-based keyguards") contradicts the documented behavior and is not supported by any known AOSP limitation. The emulator CI target is API 26+ so this applies directly.

---

**3. PR title and description are stale**

The PR title still reads "Send KEYCODE_WAKEUP and wm dismiss-keyguard before each screencap" and the body describes `wm dismiss-keyguard` as the fix. The code uses `KEYCODE_MENU (82)`. These need to match the actual implementation regardless of which approach is chosen.

---

**Suggested path forward:** Restore `wm dismiss-keyguard` (the approach the previous reviewer approved), update the test's helper functions and docstring accordingly (`is_menu` → `is_dismiss_keyguard`, matching the shell command form), and update the PR title and description. The `wm dismiss-keyguard` approach correctly handles swipe keyguards on the CI emulator API level and avoids the menu-dispatch side effect.

---

aunger (commented):
The swipe approach is sound: `input swipe` is the right tool for a swipe-type keyguard, the coordinates are reasonable for a standard emulator, and it is genuinely safe as a no-op over a foregrounded activity. The test is precise and strong — checking strict adjacency rather than just presence is the right choice. PR title and description are accurate.

One problem requires a fix before merging.

**`sleep 2` was removed from `build.yml` without a replacement.**

The original code had:
```
"$ADB" shell wm dismiss-keyguard
sleep 2
```

The replacement is:
```
"$ADB" shell input swipe 300 1000 300 300
```

`adb shell input swipe` returns as soon as the input event is injected into the input queue — it does not wait for the swipe animation or the keyguard dismiss animation to complete. The old `sleep 2` was the settling delay that prevented the next command (`am start -W` in the first block, `python3 scripts/check_green_feed.py` in the second) from running before the display was fully in the expected state.

Without that delay there is a window where `am start -W` fires while the keyguard is still animating away, or where `check_green_feed.py` starts its first retry before the display settles. The retry loop inside `check_green_feed.py` has `time.sleep(RETRY_DELAY_SECONDS)` before each per-iteration wake sequence, so the in-script case is more tolerant — but the two `build.yml` inline wake sequences now have no settling pause at all.

A `sleep 1` (or `sleep 2`) after each `input swipe` in `build.yml` closes this gap. The value does not need to match the original exactly; 1 s is likely enough for a swipe animation, but the original 2 s is safe.

---

aunger (commented):
The approach is sound and the implementation is consistent across all three locations (preflight pre-launch, preflight post-launch, and the `check_green_feed.py` retry loop). A few observations:

**`build.yml` preflight change (ff67259)**

The replacement of both `wm dismiss-keyguard` calls with `input swipe 300 1000 300 300` is correct and the motivation is well-documented in the inline comments. Removing the `sleep 2` that followed the first `wm dismiss-keyguard` call is fine — the swipe command is synchronous and `am start -W` blocks until the activity is on screen, so the sleep was only compensating for the unreliable dismiss. The post-launch re-wake is unchanged in structure and now uses the same gesture, keeping both in sync with the retry loop.

**`check_green_feed.py` retry loop**

The `KEYCODE_WAKEUP` + swipe pair is placed before the `open(path, "wb")` / `exec-out screencap` block, which is the right position — the screen is woken before every capture attempt including the first, not just between retries.

**Test coverage**

`test_wakeup_and_swipe_sent_before_each_screencap` checks strict positional ordering (wakeup at `sc_idx - 2`, swipe at `sc_idx - 1`, screencap at `sc_idx`) across multiple retry iterations. One minor observation: the swipe predicate `"300" in args` would also match an unrelated `adb` call whose argument list happened to contain `"300"`. This is not a correctness problem given the current call set, but a more specific equality check would be more robust. Not a blocker.

Overall the change correctly addresses the failure mode, is consistent across the workflow and script, and is well-tested.

---

aunger (commented):
**Verdict: approved** (GitHub prevents approving your own PR, so posting as a comment.)

The prior concern — missing `sleep 2` after the post-launch swipe in `build.yml` — is resolved. The diff now shows `sleep 2` after both `input swipe 300 1000 300 300` invocations in the preflight step, matching the original settling delay.

Everything else holds up:

- The swipe approach correctly handles swipe-type keyguards without the `wm dismiss-keyguard` reliability issue or the `KEYCODE_MENU` options-menu side effect.
- `check=False` on both new `subprocess.run` calls is appropriate; a wake/unlock failure is surfaced by the subsequent screencap.
- The retry loop in `check_green_feed.py` already leads with `time.sleep(RETRY_DELAY_SECONDS)`, so the in-script wake sequence has a natural settle window before the first capture.
- The test checks strict positional adjacency (wakeup at `sc_idx - 2`, swipe at `sc_idx - 1`, screencap at `sc_idx`) across two iterations, which is the right invariant to enforce.

---

aunger:
Nope! #200

PR #200, opened 2026-05-20
ANR (Application Not Responsive) watcher script for CI

### LINKED ISSUE #194: Preflight: “Pixel Launcher isn't responding”

|---|
| ❌ Close app |
| 🕓 Wait |

## Root cause

---

### `scripts/dismiss_anr.sh`

### `build.yml` changes

```yaml
SETTLE_PID=$!
```

```yaml
wait "$SETTLE_PID" || true
```

---

```yaml
sleep 15
```

aunger:
# ANR (Application Not Responsive) watcher script for CI

Fixes #194

## Problem

After `adb install -r` in the CI pre-flight step, the Pixel Launcher receives an `ACTION_PACKAGE_REPLACED` broadcast and reconciles its icon grid. On slow CI emulators this saturates the Launcher's main thread long enough to trigger the broadcast ANR timeout (~10 s), producing a modal "Pixel Launcher isn't responding" dialog. That dialog overlays `MockCameraActivity` and causes the green-feed check to fail 25–50% of the time.

### `scripts/dismiss_anr.sh` (new)

A best-effort background watcher that polls every 3 s for up to 30 s:

1. If `dumpsys window windows` contains `AppNotRespondingDialog`, sends `KEYCODE_BACK` to dismiss the dialog and exits immediately.
2. Checks `dumpsys cpuinfo` for the `nexuslauncher` process. If the Launcher's CPU usage is below 5% for two consecutive readings (or the process is absent), exits — the system has settled.
3. After 30 s without either condition, exits anyway (timeout — the green-feed check is the real correctness gate).

Always exits 0 so it never blocks CI on its own.

Accepts an optional `--adb <path>` argument; falls back to `$ANDROID_HOME/platform-tools/adb`.

### `scripts/test_dismiss_anr.sh` (new)

Shell-based tests covering:
- (a) `--adb` argument parsing — verifies the supplied path is actually used
- (b) ANR-detection branch — mock adb emitting `AppNotRespondingDialog`, checks `KEYCODE_BACK` is sent
- (c) idle-count exit path — mock adb with low Launcher CPU, checks two polls trigger early exit
- (d) absent Launcher process treated as idle

### `.github/workflows/build.yml` (modified)

In the `CI pre-flight — MockCameraActivity smoke check` step:

- Immediately after `adb install -r`: launches `dismiss_anr.sh` in the background and captures its PID as `SETTLE_PID`.
- After `am start -W` and before `check_green_feed.py`: `wait "$SETTLE_PID" || true` blocks until the watcher has either dismissed the dialog or confirmed the Launcher is idle, maximising the chance the screen is clear when the screenshot is taken.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -215,6 +215,10 @@ jobs:
           echo "==> Installing mock Pixel Camera for pre-flight check..."
           "$ADB" install -r \
             e2e-mock-camera/build/outputs/apk/debug/e2e-mock-camera-debug.apk
+          # Adaptive ANR watcher: waits for Pixel Launcher to settle after install (Issue #194)
+          chmod +x scripts/dismiss_anr.sh
+          scripts/dismiss_anr.sh --adb "$ADB" &
+          SETTLE_PID=$!
           echo "==> Granting CAMERA permission..."
           "$ADB" shell pm grant com.google.android.GoogleCamera \
             android.permission.CAMERA
@@ -241,6 +245,8 @@ jobs:
           # guarantee that the green view is visible before screencap.
           "$ADB" shell input keyevent KEYCODE_WAKEUP
           "$ADB" shell wm dismiss-keyguard
+
+          wait "$SETTLE_PID" || true
 
           python3 scripts/check_green_feed.py --adb "$ADB" preflight.png
 
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -1,0 +1,90 @@
+#!/usr/bin/env bash
+# dismiss_anr.sh — Best-effort ANR watcher for CI pre-flight.
+#
+# After `adb install -r` the Pixel Launcher receives ACTION_PACKAGE_REPLACED and
+# reconciles its icon grid. On slow CI emulators this can saturate the Launcher's
+# main thread long enough to hit the broadcast ANR timeout (~10 s), producing a
+# modal "Pixel Launcher isn't responding" dialog that overlays MockCameraActivity
+# and causes the green-feed check to fail (Issue #194).
+#
+# This script runs in the background during the install → green-feed window and:
+#   1. Dismisses the ANR dialog immediately if it appears.
+#   2. Waits for Pixel Launcher CPU usage to fall below 5% (two consecutive
+#      readings) before exiting, so the caller knows the system has settled.
+#   3. Always exits 0 — the green-feed check is the real correctness gate.
+#
+# Usage:
+#   scripts/dismiss_anr.sh [--adb <path>]
+#
+# Arguments:
+#   --adb <path>   Path to the adb binary. Defaults to
+#                  $ANDROID_HOME/platform-tools/adb.
+
+# Run the real logic in a subshell with strict error handling.
+# The outer script catches any failure from the subshell and still exits 0.
+(
+  set -euo pipefail
+
+  # ── Resolve adb ─────────────────────────────────────────────────────────────
+  ADB=""
+  while [[ $# -gt 0 ]]; do
+    case "$1" in
+      --adb)
+        ADB="$2"
+        shift 2
+        ;;
+      *)
+        shift
+        ;;
+    esac
+  done
+
+  if [[ -z "$ADB" ]]; then
+    ADB="${ANDROID_HOME:-}/platform-tools/adb"
+  fi
+
+  # ── Poll loop ────────────────────────────────────────────────────────────────
+  POLL_INTERVAL=3
+  TIMEOUT=30
+  elapsed=0
+  idle_count=0
+
+  while [[ $elapsed -lt $TIMEOUT ]]; do
+    # 1. Check for the ANR dialog.
+    window_dump="$("$ADB" shell dumpsys window windows 2>/dev/null || true)"
+    if echo "$window_dump" | grep -q "AppNotRespondingDialog"; then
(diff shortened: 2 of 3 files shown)


aunger (commented):
The implementation matches the plan precisely. A few notes before signing off.

**Plan compliance**

| Requirement | Result |
|---|---|
| Poll every 3 s | `POLL_INTERVAL=3` + `sleep "$POLL_INTERVAL"` — correct |
| Up to 30 s | `TIMEOUT=30`, loop exits when `elapsed` reaches 30 — correct |
| Two-consecutive-idle guard | `idle_count` resets on any busy reading, exits at `≥ 2` — correct |
| Always exits 0 | Subshell uses `exit 0` everywhere; outer script appends `\|\| true` — correct |
| `--adb` argument | Argument-parsing loop in the subshell handles it; inherits `$@` correctly — correct |
| build.yml wiring | Background launch immediately after `adb install -r`; `wait` immediately before `check_green_feed.py` — correct |

**Shell safety**

`set -euo pipefail` is inside the subshell rather than at the top level, which is the right approach given the `|| true` escape hatch on the outer call. All variable expansions involving user-supplied paths are double-quoted. The `|| true` guards on the `adb` calls and `grep` are correct.

**`pct` extraction**

`grep -oE '^[[:space:]]*[0-9]+'` on a line like `  2% com.google...` extracts the leading digits only, and `tr -d ' '` strips the spaces, producing a clean integer. A decimal CPU value like `2.5%` would truncate to `2`, still below the 5% threshold — the truncation is safe.

**Test coverage**

All three required branches are present and structurally sound: `--adb` path (a), ANR-detection (b), idle-count (c). The absent-Launcher variant (d) is a valuable addition. The `case "$*"` pattern in the mock adb scripts correctly matches the full argument strings passed by the real script.

One small observation: tests (c) and (d) each incur at least one real `sleep 3` call because the script needs two idle readings and `POLL_INTERVAL` is not overridable via argument. This makes the suite take ~6–9 s. Not a defect, but worth knowing as the test suite grows.

Overall the code is clean, well-commented, and the fix is ready to merge.

---

aunger (commented):
This PR is already merged. Three substantive problems need follow-on fixes; one concern from the Orchestrator can be closed.

---

## 1. The dominant root cause is not addressed (blocking)

The CI log evidence cited in issue #194 — #E9E8EF (gray/wallpaper) on **all 30 retries** — is the signature of a persistent lock-screen obstruction, not an ANR dialog. An ANR dialog appears as a translucent modal; it would not cause every retry to capture the wallpaper color. The conclusion is that `wm dismiss-keyguard` silently fails on the API-35 CI emulator.

The merged PR adds the ANR watcher (commits 1–2) but leaves both `wm dismiss-keyguard` calls in `build.yml` untouched. Commits `fd072ed` and `b018853`, which exist on this branch beyond the merge point, contain the keyguard fix and per-retry wake sequence — but were not merged.

**Required action:** Replace the two `adb shell wm dismiss-keyguard` calls in the preflight step with `adb shell input swipe 300 1000 300 300` + `sleep 2`, as authored in commit `fd072ed`. Also apply the per-retry KEYCODE_WAKEUP + swipe in `check_green_feed.py` from commit `b018853`. Without these, the intermittent failure will continue at the same rate.

---

## 2. `dismiss_anr.sh` exits immediately after KEYCODE_BACK without confirming the dialog is gone (blocking)

In the merged script, when `AppNotRespondingDialog` is detected:

```bash
"$ADB" shell input keyevent KEYCODE_BACK 2>/dev/null || true
sleep 1
exit 0          # exits without verifying the dialog is gone
```

Sleeping 1 s and exiting is not sufficient on a slow CI emulator. The dialog may still be on screen when `check_green_feed.py` runs. Commit `71907dc` (on the branch, not merged) fixes this by replacing `exit 0` with `continue`, resetting `idle_count`, and requiring the loop to subsequently confirm the dialog is gone and the Launcher is idle before exiting.

---

## 3. Merged idle-exit path can exit while an ANR dialog is still visible (blocking)

An ANR dialog stops consuming CPU once rendered. If the dialog appears *after* the first window-dump poll but *during* CPU-polling iterations, `idle_count` will reach 2 and the script will exit — with the dialog still on screen and KEYCODE_BACK never sent. The merged idle-exit path does not re-check `dumpsys window windows` before exiting. Commit `71907dc` adds that re-check and needs to be applied.

---

## 4. Test case (b) is coupled to the flawed `exit 0` behavior

The merged `test_dismiss_anr.sh` test case (b) uses a mock that permanently returns `AppNotRespondingDialog`. This test passes today only because the merged script does `exit 0` immediately after KEYCODE_BACK. Applying the corrected `continue` behavior (commit `71907dc`) without also updating the test (commit `c802eda`) would cause the test to hang forever. Both commits must be applied together.

---

## 5. Per-retry wake+swipe does not risk disrupting MockCameraActivity (Orchestrator concern closed)

Commit `b018853` adds KEYCODE_WAKEUP + `input swipe 300 1000 300 300` before every screencap in the retry loop. MockCameraActivity renders a static solid-color View with no interactive UI elements. An upward swipe over it is a no-op from the application's perspective and does not affect the camera preview. The concern is unfounded; this commit is correct and safe.

---

**In summary:** commits `fd072ed`, `b018853`, `71907dc`, and `c802eda` from this branch need to land in a follow-on PR. The ANR watcher as merged is incomplete without them, and the primary failure mode (keyguard) is unaddressed.

---

PR #202, opened 2026-05-21
Fix ANR watcher premature exit and harden keyguard dismissal

### LINKED ISSUE #194: Preflight: “Pixel Launcher isn't responding”

aunger:
Closes #194

## Problem

Post-merge CI confirmed that `dismiss_anr.sh` (landed in #200) was exiting prematurely via the `idle_count >= 2` CPU-idle path while the ANR dialog was still on screen. CPU drops to idle as soon as Android renders the dialog — it does not stay elevated — so the idle threshold was firing immediately, leaving the dialog blocking the green-feed check for all 30 retries.

Failing run: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/26220173038/job/77152821075

## Changes

**`ci: replace wm dismiss-keyguard with input swipe in pre-flight step`** (`fd072ed`)
Switches the pre-flight keyguard dismissal from `wm dismiss-keyguard` (which silently fails on locked-screen builds) to a reliable `input swipe` gesture that actually wakes and unlocks the display.

**`fix: wake display and swipe-dismiss keyguard before each screencap in retry loop`** (`b018853`)
Applies the same wake-and-swipe logic inside the retry loop in the green-feed check, so that a display that re-locks between retries does not cause screencap to return a black frame.

**`fix(dismiss_anr): guard idle-exit behind ANR-dialog re-check`** (`71907dc`)
Core fix. Before exiting via `idle_count >= 2`, the script now re-runs `dumpsys window windows` and only exits clean if no `AppNotRespondingDialog` is present. If the dialog is still on screen, it is dismissed, `idle_count` is reset to zero, and polling continues. This closes the race where CPU goes idle at dialog-render time before the dismiss tap has landed.

**`Fix test case (b) in test_dismiss_anr.sh to avoid infinite loop`** (`c802eda`)
Updates the unit test stub for the re-check path so that the simulated ANR dialog disappears after the dismiss tap, preventing the test harness from hanging indefinitely in the new loop.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -223,14 +223,14 @@ jobs:
           "$ADB" shell pm grant com.google.android.GoogleCamera \
             android.permission.CAMERA
 
-          # Wake the display and dismiss the keyguard before launching the activity.
-          # am start -W waits for the framework to launch the activity, but on an
-          # emulator with screen timeout the keyguard may still be covering the window.
-          # wm dismiss-keyguard (API 26+) removes the keyguard immediately and is more
-          # reliable than a keyevent sequence for this purpose.
+          # Wake the display and dismiss the swipe keyguard before launching the
+          # activity. An upward swipe is used instead of wm dismiss-keyguard:
+          # wm dismiss-keyguard did not reliably dismiss the swipe-type lock
+          # screen on the API-35 CI emulator (all retries returned #E9E8EF).
+          # sleep 2 lets the swipe animation complete before am start -W fires.
           echo "==> Waking display and dismissing keyguard..."
           "$ADB" shell input keyevent KEYCODE_WAKEUP
-          "$ADB" shell wm dismiss-keyguard
+          "$ADB" shell input swipe 300 1000 300 300
           sleep 2
 
           # Launch MockCameraActivity and wait for it to be displayed.
@@ -239,12 +239,14 @@ jobs:
           "$ADB" shell am start -W -n \
             com.google.android.GoogleCamera/com.gb4pc.mockcamera.MockCameraActivity
 
-          # Re-wake and re-dismiss keyguard after am start -W: adb install above
-          # can take 10-30 s and the screen may sleep in that window. am start -W
-          # and setTurnScreenOn(true) usually recover, but this is a cheap extra
-          # guarantee that the green view is visible before screencap.
+          # Re-wake after am start -W: adb install above can take 10-30 s and
+          # the screen may sleep in that window. am start -W and
+          # setTurnScreenOn(true) usually recover, but this is a cheap extra
+          # guarantee that the green view is visible before screencap. Use the
+          # same swipe unlock to stay consistent.
           "$ADB" shell input keyevent KEYCODE_WAKEUP
-          "$ADB" shell wm dismiss-keyguard
+          "$ADB" shell input swipe 300 1000 300 300
+          sleep 2
 
           wait "$SETTLE_PID" || true
 
--- a/scripts/check_green_feed.py
+++ b/scripts/check_green_feed.py
@@ -221,6 +221,24 @@ def main() -> int:
         # Sleep first so the display has time to render before the first check
         # and between retries; the caller has already waited for am start -W.
         time.sleep(RETRY_DELAY_SECONDS)
+        # Wake the display and dismiss the keyguard before every screencap.
+        # The screen may dim or the keyguard may re-appear during a long retry
+        # loop (e.g. while the APK was being installed).
+        # KEYCODE_WAKEUP (224) turns the screen on on all API levels.
+        # An upward swipe then dismisses the swipe-based lock screen that the
+        # emulator shows before the E2E PIN setup script runs.  Using a swipe
+        # gesture avoids the side-effect of KEYCODE_MENU (82), which dispatches
+        # KeyEvent.KEYCODE_MENU to a foregrounded activity and can open the
+        # options/overflow menu, obscuring the green view.  When the screen is
+        # already unlocked the swipe is a harmless gesture over the activity.
+        subprocess.run(
+            [adb_path, "shell", "input", "keyevent", "224"],
+            check=False,
+        )
+        subprocess.run(
+            [adb_path, "shell", "input", "swipe", "300", "1000", "300", "300"],
+            check=False,
+        )
         with open(path, "wb") as out:
             subprocess.run(
                 [adb_path, "exec-out", "screencap", "-p"],
--- a/scripts/check_green_feed.py
+++ b/scripts/check_green_feed.py
@@ -239,6 +239,8 @@ def main() -> int:
             [adb_path, "shell", "input", "swipe", "300", "1000", "300", "300"],
             check=False,
         )
+        # Wait for any swipe animation to settle before capturing the screen.
+        time.sleep(RETRY_DELAY_SECONDS)
         with open(path, "wb") as out:
             subprocess.run(
                 [adb_path, "exec-out", "screencap", "-p"],
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -8,7 +8,7 @@
 # and causes the green-feed check to fail (Issue #194).
 #
 # This script runs in the background during the install → green-feed window and:
-#   1. Dismisses the ANR dialog immediately if it appears.
+#   1. Dismisses the ANR dialog if it appears (detected via logcat, not polling).
 #   2. Waits for Pixel Launcher CPU usage to fall below 5% (two consecutive
 #      readings) before exiting, so the caller knows the system has settled.
 #   3. Always exits 0 — the green-feed check is the real correctness gate.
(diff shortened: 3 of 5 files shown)


aunger (commented):
All 6 shell tests pass locally. The core fix (`71907dc`) is the right idea, but it introduces a broken timeout guarantee. One additional concern exists in the per-retry wake+swipe (`b018853`). Details below.

---

## Blocking issue: timeout guard is broken for a persistent ANR dialog

In `dismiss_anr.sh`, `elapsed` is only incremented at the bottom of the loop body, after `sleep "$POLL_INTERVAL"`. The ANR-detection branch now uses `continue` to restart the loop without reaching that increment:

```bash
if echo "$window_dump" | grep -q "AppNotRespondingDialog"; then
  "$ADB" shell input keyevent KEYCODE_BACK 2>/dev/null || true
  idle_count=0
  sleep 1
  continue          # jumps back to while-test; elapsed is never advanced
fi
```

If the ANR dialog persists (e.g., KEYCODE_BACK doesn't clear it, or the system keeps re-raising it), `elapsed` stays at 0 forever and the `while [[ $elapsed -lt $TIMEOUT ]]` guard never fires. The script spins in a tight 1-second loop indefinitely rather than timing out at 30 s.

The fix is to advance `elapsed` before `continue`, or restructure so every path through the loop increments it. The idle_count ≥ 2 ANR-recheck branch does not have this problem — it falls through to `sleep "$POLL_INTERVAL"` and `elapsed += POLL_INTERVAL`.

The existing test for case (b) does not catch this because the mock adb clears the ANR after the first call. A test whose mock keeps returning `AppNotRespondingDialog` indefinitely would currently hang the test suite — which is itself evidence the timeout protection is absent.

---

## Concern: swipe fired immediately before screencap with no settling time

In `check_green_feed.py` (commit `b018853`), the per-retry sequence is:

```python
time.sleep(RETRY_DELAY_SECONDS)   # top of iteration
subprocess.run([..., "keyevent", "224"], ...)
subprocess.run([..., "swipe", "300", "1000", "300", "300"], ...)
with open(path, "wb") as out:
    subprocess.run([..., "screencap", "-p"], stdout=out, ...)   # immediately after swipe
```

There is no sleep between the swipe and the screencap. If the swipe triggers even a short animation (lock-screen-dismiss or otherwise) on the `google_apis` API-35 emulator, the screencap can capture the transition and the green-pixel check fails. The pre-flight step in `build.yml` correctly adds `sleep 2` after each swipe before proceeding — the same discipline is absent here.

The comment "when the screen is already unlocked the swipe is a harmless gesture over the activity" is plausible but is an untested assumption for this specific emulator image.

---

## Commits 1 and 2 (keyguard fixes) — not harmful

Replacing `wm dismiss-keyguard` with `input swipe` responds to the observed emulator behaviour, and adding `sleep 2` after the second pre-flight re-wake (which the original lacked) is a genuine improvement. No objection to these.

---

## Test fix (commit `c802eda`) — correct as far as it goes

The updated mock correctly exercises the "ANR appears then clears" path and verifies `KEYCODE_BACK` was sent. The test is sound for the scenario it covers. Once the `elapsed` accounting is fixed, a test case that verifies timeout-and-exit when the dialog never clears would give meaningful coverage of the persistent-ANR timeout path.

---

aunger (commented):
Both issues are fully resolved. I cannot approve my own PR, so I am leaving this as a comment review instead.

**`dismiss_anr.sh` — elapsed counter in ANR branch**

The ANR branch now correctly does `sleep "$POLL_INTERVAL"` followed by `elapsed=$((elapsed + POLL_INTERVAL))` before `continue`. The script cannot loop forever on a persistent dialog; it will time out after TIMEOUT=30 s and exit 0. Test case (e) covers this path by running the script under `timeout 35` with a mock adb that always returns the ANR dialog, confirming it exits 0 within the wall-clock budget.

One minor observation: inside the `idle_count >= 2` branch (line 90), the "ANR dialog present despite idle Launcher" sub-path does `sleep 1` without incrementing `elapsed`, then falls through to `sleep "$POLL_INTERVAL"` + `elapsed=$((elapsed + POLL_INTERVAL))` at the loop bottom — so `elapsed` still advances correctly on each full iteration. Not a bug.

**`check_green_feed.py` — settling sleep before screencap**

`time.sleep(RETRY_DELAY_SECONDS)` is now inserted between the swipe call and the `open`/`screencap` block. The ordering per iteration is: `sleep` → `keyevent 224` → `swipe` → `sleep` → `screencap`. `test_sleep_between_swipe_and_screencap` verifies this ordering across multiple retry cycles using an interleaved call log, and `test_wakeup_and_swipe_sent_before_each_screencap` checks the subprocess call ordering independently. Both are solid.

No new issues introduced. This is ready to merge.

---

aunger (commented):
The logcat-trigger design is cleaner than the old polling approach, and the `elapsed` counter, `trap EXIT`, and `SLEEP_AFTER_ANR_DETECTED` env-var mechanism are all correct. However there are two real bugs and one coverage gap that need attention before this merges.

---

## Bug 1: `grep -m1` makes a second ANR undetectable (high severity)

`scripts/dismiss_anr.sh` lines 59–61:

```sh
( "$ADB" logcat ActivityManager:E '*:S' 2>/dev/null \
    | grep -m1 'ANR in com.google.android.apps.nexuslauncher' > /dev/null \
  && touch "$ANR_FLAG" ) &
```

`grep -m1` exits after the first match. Once that subshell exits, the write end of the pipe to `adb logcat` is closed, and the `adb logcat` grandchild receives SIGPIPE and exits too. No watcher is running for the rest of the script's lifetime.

The PR description itself notes the scenario: "If a second ANR fires after the first is dismissed but before CPU settles…" and answers "the logcat stream catches it." That answer is wrong — the stream is gone by then.

On slow CI emulators the Launcher _can_ ANR a second time after the first dismissal, leaving the dialog on screen while the script waits for CPU to settle. The old `dumpsys window windows` re-check before the idle-exit (now deleted at lines 104–113 in the original) was the guard against exactly this. Now there is no guard.

Fix: either remove `-m1` and loop inside the subshell to re-touch the flag file on each match, or restart the background watcher after each dismissal.

---

## Bug 2: orphaned `adb logcat` grandchild process (medium severity)

`LOGCAT_PID` is the PID of the background subshell `( adb logcat | grep -m1 ... ) &`. The `trap EXIT` targets that PID.

In the no-ANR path (the common case), `grep -m1` never matches so the subshell stays alive. `trap EXIT` kills the subshell, but `adb logcat` was started _inside_ the subshell — it is not a direct child of the outer process. Killing `$LOGCAT_PID` sends the signal to the subshell wrapper; once that wrapper exits, `adb logcat` may or may not receive SIGHUP depending on the shell implementation. In practice, on many systems it becomes orphaned and runs until the adb server drops the connection.

Fix: capture the adb PID explicitly inside the subshell and write it to a temp file so the trap can kill it directly, or use a process group (`setsid` + `kill -- -$LOGCAT_PID`).

---

## Coverage gap: no test for a second ANR after first dismissal

None of the 5 tests exercises the case where logcat fires, KEYCODE_BACK is sent, and then a second ANR occurs before CPU settles. Given that Bug 1 makes that scenario silently fail, this case should be explicitly tested. Without it, the regression from removing the `dumpsys window windows` re-check goes undetected by the suite.

---

## Minor: `mktemp` + immediate `rm` is an unusual pattern

```sh
ANR_FLAG="$(mktemp)"
rm -f "$ANR_FLAG"
```

This creates the file and immediately deletes it just to obtain a unique path. `mktemp -u` (dry-run, no creation) or a simple `$TMPDIR/dismiss_anr_flag.$$` assignment communicates the intent more clearly. Not a blocker.

---

## Summary

The `elapsed` counter is correct in all branches. The `SLEEP_AFTER_ANR_DETECTED` override and the overall structure are clean. The main issue before merging is Bug 1: a second ANR after the first dismissal is silently missed because the logcat stream has already exited, and the previous safety net for this case was removed without a replacement.

---

aunger (commented):
Re-review of commit 64676ba — all three previously flagged bugs are addressed correctly. Verdict: **approve**.

**Bug 1 (grep -m1 killed the stream):** Fixed. The `while IFS= read -r line; do case … done < "$LOGCAT_FIFO"` consumer stays alive for the script's lifetime and re-touches `$ANR_FLAG` on every matching line. The flag is cleared by the poll loop before sleeping and sending KEYCODE_BACK, so a second ANR arriving during that window will correctly re-set the flag and trigger a second dismissal on the next iteration.

**Bug 2 (grandchild not killed):** Fixed. `adb logcat` is backgrounded first (it blocks on the fifo until the reader opens), its PID is written synchronously to `$ADB_LOGCAT_PID_FILE`, and the `trap EXIT` reads and kills that PID directly before killing the consumer subshell and removing the fifo and temp files. The ordering — kill writer, kill reader, remove fifo — is correct.

**Bug 3 (missing second-ANR test):** Test (f) covers the scenario well. The 5-second gap between the two mock ANR lines gives the poll loop (SLEEP_AFTER_ANR_DETECTED=1 + POLL_INTERVAL=3 = ~4 s per cycle) enough headroom to consume and clear the first flag before the second line arrives. The cpuinfo mock returning high CPU until two KEYCODE_BACK events have been sent forces both dismissals before the script exits.

One minor observation, not blocking: test (f) has a wall-clock timing dependency. If a heavily loaded CI runner causes the first poll cycle to take longer than 5 seconds end-to-end, both ANR lines could already be queued in the fifo when the consumer processes them. In that case the consumer touches the flag twice nearly simultaneously; depending on scheduling, the poll loop might catch both touches as one event and send KEYCODE_BACK only once, causing the `keyevent_count -ge 2` assertion to fail non-deterministically. Widening the gap to 8–10 s would eliminate that risk. Not a correctness bug in the production script — only a test robustness concern.

---

PR #215, opened 2026-05-23
Fix ANR dismiss: skip unknown-CPU idle count and use KEYCODE_ENTER

### LINKED ISSUE #214: Preflight: ANR dismiss still fails — KEYCODE_BACK ineffective and dismiss_anr.sh exits too early
## Evidence

CI log from 2026-05-23 09:56 shows the ANR dialog persists across all 30 outer retry attempts (× 5 KEYCODE_BACK sends each). The preflight exits 1 on every attempt: `#E9E8EF` (dialog background) instead of `#00C853` (green feed).

### Bug 1 — `dismiss_anr.sh` exits in ~4 seconds when CPU is unreadable

The log shows:
```
[dismiss_anr] 10:58:37.395 [t=0s poll=1] idle_count=1 cpu=(unknown)%
[dismiss_anr] 10:58:40.643 [t=3s poll=2] idle_count=2 cpu=(unknown)%
[dismiss_anr] ... EXIT trap: cleanup complete (elapsed 3976ms).
```

When `dumpsys cpuinfo` returns no line for `nexuslauncher`, `cpu` is set to `(unknown)%`. The script currently increments `idle_count` for any reading below 5% — but `(unknown)%` triggers the same branch because the integer comparison fails. Two consecutive unknowns = `idle_count ≥ 2` = premature exit.

The ANR dialog then appears at 10:58:51 — 10 seconds after start and 7 seconds after `dismiss_anr.sh` has already exited — with no watcher to catch it.

**Fix:** Only increment `idle_count` when the CPU is a *known* value below 5%. When the reading is `(unknown)%`, either skip the iteration (don't change `idle_count`) or reset it to 0.

### Bug 2 — `KEYCODE_BACK` does not dismiss Android ANR dialogs

The log shows every single one of 150 KEYCODE_BACK sends (30 outer × 5 inner) failing:
```
[check_green_feed] 10:58:51 ANR dialog detected before screencap — sending KEYCODE_BACK.
[check_green_feed] 10:58:53 ANR dialog still present after KEYCODE_BACK #1.
...
[check_green_feed] 10:59:02 ANR dialog still present after KEYCODE_BACK #5.
[check_green_feed] 10:59:02 ANR dialog persisted after 5 KEYCODE_BACK sends — proceeding to screencap anyway.
```

Android ANR dialogs are `AlertDialog` instances set `setCancelable(false)`. KEYCODE_BACK is explicitly ignored. The "Wait" button must be pressed.

**Fix:** Replace `KEYCODE_BACK` with a key or tap that activates the focused "Wait" button. Options in order of preference:
1. `KEYCODE_ENTER` — presses the focused button, which defaults to "Wait" on AOSP ANR dialogs
2. `input tap <x> <y>` — coordinates of the Wait button (fragile but works if the emulator resolution is stable)
3. A `uiautomator dump` to locate and click the button by text

Both `dismiss_anr.sh` (logcat-triggered path) and `check_green_feed.py` (per-retry path) need the fix.

## Acceptance criteria

- `dismiss_anr.sh`: `cpu=(unknown)%` does not increment `idle_count`; the script stays alive for its full polling window when the launcher process is absent from cpuinfo
- Both dismiss paths (sh + py) replace `KEYCODE_BACK` with a key event that actually dismisses the ANR dialog
- Existing `scripts/test_dismiss_anr.sh` (13 tests) and `python3 -m pytest scripts/` (54+ tests) continue to pass; new tests cover the unknown-cpu and effective-dismiss cases

aunger:
Fixes #214

### Bug 1 — Unknown CPU reading prematurely incremented `idle_count`

When `dumpsys cpuinfo` returned a `nexuslauncher` line whose percentage
could not be parsed as a leading integer (e.g. `(unknown) com.google…`),
the old code fell into the `[[ -z "$pct" ]]` branch and incremented
`idle_count` as if the launcher were idle. After two such readings (~6 s)
the script exited before the ANR dialog had even appeared (it appears
~10 s after install).

Fix: when `pct` is empty the iteration is now skipped without touching
`idle_count`. The script only exits after two consecutive *known* readings
below 5%.

### Bug 2 — `KEYCODE_BACK` cannot dismiss ANR dialogs

Android ANR dialogs are created with `setCancelable(false)`. The Back key
is explicitly ignored by the dialog, so every `KEYCODE_BACK` send was a
no-op. CI logs confirmed this: 150 consecutive sends (30 outer retries × 5
inner retries) all failed to dismiss the dialog.

Fix: replace `KEYCODE_BACK` with `KEYCODE_ENTER`, which activates the
focused button. The "Wait" button is focused by default on AOSP ANR
dialogs, so Enter dismisses the dialog reliably. The fix is applied in
both dismiss paths:
- `scripts/dismiss_anr.sh` — logcat-triggered path and the `dumpsys window`
  safety-check fallback.
- `scripts/check_green_feed.py` — per-retry `_dismiss_anr_if_present()`.

## Test results

Shell suite (`bash scripts/test_dismiss_anr.sh`): **15 passed, 0 failed**
(added test (h): unknown-CPU readings do not increment `idle_count`).

Python suite (`python3 -m pytest scripts/ -v`): **57 passed, 0 failed**
(added `TestDismissAnrUsesKeycodeEnter`: asserts `KEYCODE_ENTER` is sent
and `KEYCODE_BACK` is never sent).

---

### DIFF
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -134,12 +134,16 @@
     else
       # Extract the leading integer percentage (e.g. "  12% com.google.android.apps...")
       pct="$(echo "$launcher_line" | grep -oE '^[[:space:]]*[0-9]+' | tr -d ' ' || true)"
-      if [[ -z "$pct" ]] || [[ "$pct" -lt 5 ]]; then
+      if [[ -z "$pct" ]]; then
+        # Percentage not parseable — skip this iteration without changing idle_count.
+        echo "[dismiss_anr] $(_ts) [t=${elapsed}s poll=${poll_n}] idle_count=${idle_count} cpu=(unknown)%" >&2
+      elif [[ "$pct" -lt 5 ]]; then
         idle_count=$((idle_count + 1))
+        echo "[dismiss_anr] $(_ts) [t=${elapsed}s poll=${poll_n}] idle_count=${idle_count} cpu=${pct}%" >&2
       else
         idle_count=0
+        echo "[dismiss_anr] $(_ts) [t=${elapsed}s poll=${poll_n}] idle_count=${idle_count} cpu=${pct}%" >&2
       fi
-      echo "[dismiss_anr] $(_ts) [t=${elapsed}s poll=${poll_n}] idle_count=${idle_count} cpu=${pct:-(unknown)}%" >&2
     fi
 
     if [[ $idle_count -ge 2 ]]; then
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -66,7 +66,7 @@
   if [[ -n "$LAUNCHER_PKG" ]]; then
     echo "[dismiss_anr] $(_ts) Launcher detected: $LAUNCHER_PKG" >&2
   else
-    echo "[dismiss_anr] $(_ts) Launcher detection failed — using auto-detected via pattern fallback (pixellauncher|launcher3)." >&2
+    echo "[dismiss_anr] $(_ts) Launcher detection failed — using auto-detected via pattern fallback (nexuslauncher|launcher3)." >&2
   fi
 
   # ── Phase 1: logcat trigger ──────────────────────────────────────────────────
@@ -107,8 +107,8 @@
               *"ANR in $_LAUNCHER_PKG_LOCAL"*) touch "$ANR_FLAG" ;;
             esac
           else
-            # Pattern fallback: match pixellauncher or launcher3 (case-insensitive).
-            if echo "$line" | grep -qiE 'ANR in [^ ]*(pixellauncher|launcher3)'; then
+            # Pattern fallback: match nexuslauncher or launcher3 (case-insensitive).
+            if echo "$line" | grep -qiE 'ANR in [^ ]*(nexuslauncher|launcher3)'; then
               touch "$ANR_FLAG"
             fi
           fi
@@ -157,7 +157,7 @@
     if [[ -n "$LAUNCHER_PKG" ]]; then
       launcher_line="$(echo "$cpu_dump" | grep -iF "$LAUNCHER_PKG" | head -1 || true)"
     else
-      launcher_line="$(echo "$cpu_dump" | grep -iE "pixellauncher|launcher3" | head -1 || true)"
+      launcher_line="$(echo "$cpu_dump" | grep -iE "nexuslauncher|launcher3" | head -1 || true)"
     fi
 
     if [[ -z "$launcher_line" ]]; then
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -22,7 +22,11 @@
 #
 # Environment:
 #   SLEEP_AFTER_ANR_DETECTED   Seconds to wait after logcat fires before sending
-#                              KEYCODE_BACK (default: 7). Override in tests.
+#                              KEYCODE_ENTER (default: 7). Override in tests.
+#   LAUNCHER_PACKAGE           Package name of the launcher to watch.
+#                              Default: com.google.android.apps.nexuslauncher
+#                              (matches google_apis emulator images and Pixel devices).
+#                              AOSP images use com.android.launcher3.
 
 # Run the real logic in a subshell with strict error handling.
 # The outer script catches any failure from the subshell and still exits 0.
@@ -79,10 +83,15 @@
   echo $! > "$ADB_LOGCAT_PID_FILE"
   echo "[dismiss_anr] $(_ts) Logcat stream started (pid=$(cat "$ADB_LOGCAT_PID_FILE"))." >&2
 
+  # Launcher package to watch for ANR events and CPU usage.
+  # Default matches google_apis emulator images and Pixel devices.
+  # Override to com.android.launcher3 for AOSP images.
+  LAUNCHER_PACKAGE="${LAUNCHER_PACKAGE:-com.google.android.apps.nexuslauncher}"
+
   # Consumer subshell: reads from the fifo and touches ANR_FLAG on every match.
   ( while IFS= read -r line; do
       case "$line" in
-        *'ANR in com.google.android.apps.nexuslauncher'*)
+        *"ANR in ${LAUNCHER_PACKAGE}"*)
           touch "$ANR_FLAG" ;;
       esac
     done < "$LOGCAT_FIFO" ) &
@@ -123,9 +132,9 @@
       # Fall through to the CPU check on this same iteration.
     fi
 
-    # Phase 3: check Pixel Launcher CPU usage.
+    # Phase 3: check launcher CPU usage.
     cpu_dump="$("$ADB" shell dumpsys cpuinfo 2>/dev/null || true)"
-    launcher_line="$(echo "$cpu_dump" | grep -i "nexuslauncher" | head -1 || true)"
+    launcher_line="$(echo "$cpu_dump" | grep -iF "$LAUNCHER_PACKAGE" | head -1 || true)"
 
     if [[ -z "$launcher_line" ]]; then
       # Launcher process not present — treat as idle.
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -117,8 +117,8 @@
       rm -f "$ANR_FLAG"
       echo "[dismiss_anr] $(_ts) Logcat ANR trigger fired — waiting ${SLEEP_AFTER_ANR_DETECTED}s for dialog to render." >&2
       sleep "$SLEEP_AFTER_ANR_DETECTED"
-      echo "[dismiss_anr] $(_ts) Sending KEYCODE_BACK to dismiss ANR dialog." >&2
-      "$ADB" shell input keyevent KEYCODE_BACK 2>/dev/null || true
+      echo "[dismiss_anr] $(_ts) Sending KEYCODE_ENTER to dismiss ANR dialog." >&2
+      "$ADB" shell input keyevent KEYCODE_ENTER 2>/dev/null || true
       idle_count=0
       # Fall through to the CPU check on this same iteration.
     fi
@@ -153,8 +153,8 @@
       # is the fallback that catches those cases.
       window_dump=$("$ADB" shell dumpsys window 2>/dev/null) || true
       if echo "$window_dump" | grep -q "Application Not Responding"; then
-        echo "[dismiss_anr] $(_ts) dumpsys window safety check: ANR dialog found — sending KEYCODE_BACK." >&2
-        "$ADB" shell input keyevent KEYCODE_BACK 2>/dev/null || true
+        echo "[dismiss_anr] $(_ts) dumpsys window safety check: ANR dialog found — sending KEYCODE_ENTER." >&2
+        "$ADB" shell input keyevent KEYCODE_ENTER 2>/dev/null || true
         idle_count=0
         # Continue the loop instead of exiting.
       else
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -23,10 +23,6 @@
 # Environment:
 #   SLEEP_AFTER_ANR_DETECTED   Seconds to wait after logcat fires before sending
 #                              KEYCODE_ENTER (default: 7). Override in tests.
-#   LAUNCHER_PACKAGE           Package name of the launcher to watch.
-#                              Default: com.google.android.apps.nexuslauncher
-#                              (matches google_apis emulator images and Pixel devices).
-#                              AOSP images use com.android.launcher3.
 
 # Run the real logic in a subshell with strict error handling.
 # The outer script catches any failure from the subshell and still exits 0.
@@ -57,6 +53,22 @@
   SCRIPT_START_TS="$(date +%s%3N)"
   echo "[dismiss_anr] $(_ts) Script started (adb=$ADB)." >&2
 
+  # ── Detect the home/launcher package at startup ───────────────────────────────
+  # Try the intent resolution approach first; fall back to pattern matching.
+  LAUNCHER_PKG=""
+  _resolve_out="$("$ADB" shell cmd package resolve-activity --brief \
+      -a android.intent.action.MAIN -c android.intent.category.HOME \
+      2>/dev/null | head -1 || true)"
+  if [[ -n "$_resolve_out" ]] && [[ "$_resolve_out" == */* ]]; then
+    LAUNCHER_PKG="${_resolve_out%%/*}"
+  fi
+
+  if [[ -n "$LAUNCHER_PKG" ]]; then
+    echo "[dismiss_anr] $(_ts) Launcher detected: $LAUNCHER_PKG" >&2
+  else
+    echo "[dismiss_anr] $(_ts) Launcher detection failed — using auto-detected via pattern fallback (pixellauncher|launcher3)." >&2
+  fi
+
   # ── Phase 1: logcat trigger ──────────────────────────────────────────────────
   # Clear the logcat buffer and start a background stream that sets a flag file
   # the moment an ANR for Pixel Launcher appears in the log.  This costs nothing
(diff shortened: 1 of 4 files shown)


aunger:
## Code Review — PR #215

**Verdict: Approved**, with one minor stale-comment nit noted below.

---

### Bug 1 — idle-count fix (dismiss_anr.sh)

The fix is correct. The old `if [[ -z "$pct" ]] || [[ "$pct" -lt 5 ]]` collapsed two distinct cases (unparseable and genuinely idle) into a single branch that always incremented `idle_count`. The new three-branch structure — skip/no-change on empty `pct`, increment on `pct < 5`, reset on `pct >= 5` — is the right shape. The log messages also move inside each branch so the log now accurately reflects whether the value was unknown, idle, or active.

Shell test (h) covers this correctly. The mock feeds two unparseable readings then switches to low-CPU readings, and the test asserts `cpuinfo` is polled at least four times before the script exits. That's a sound behavioural proxy for confirming unknown readings don't advance `idle_count`.

### Bug 2 — KEYCODE_BACK → KEYCODE_ENTER

Both dismiss paths in `dismiss_anr.sh` (logcat-triggered path at the ANR flag and the dumpsys-window safety-check fallback) are updated. The Python path in `check_green_feed.py` is updated. All three are covered by updated or new tests.

The new `TestDismissAnrUsesKeycodeEnter` class adds two focused tests: one asserting ENTER is sent exactly once and BACK is never sent on detection, and one asserting BACK is never sent across a retry cycle. These are clean and non-redundant alongside the updated existing tests.

No tests were deleted or disabled.

---

### Nit — one stale comment in dismiss_anr.sh

Line 105 (in main; unchanged in this PR) reads:

```
# How long to wait after logcat fires before sending KEYCODE_BACK.
```

This refers to `SLEEP_AFTER_ANR_DETECTED` and still says `KEYCODE_BACK`. It should say `KEYCODE_ENTER`. Not a correctness issue, but it will confuse the next reader. Worth a follow-up squash or a tiny fixup commit.

---

### Minor naming inconsistency (not blocking)

In `test_check_green_feed.py`, the method `test_no_anr_dialog_means_no_keycode_back` was not renamed (only its docstring was updated). After this PR, `keyback_calls` inside that method tracks ENTER sends, not BACK sends, so the variable name and the method name both refer to `keycode_back` while the logic checks `KEYCODE_ENTER`. The assertion is still correct, but the naming is a bit confusing. Again, not a blocking concern.

---

Both root causes are fixed correctly. Tests cover both new behaviours. The stale comment is cosmetic and does not affect correctness.

---

scripts/dismiss_anr.sh
aunger: Is this the launcher used both by AOSP and Pixel? I’d like the tests to work on both platforms.

aunger:
## Follow-up review — PR #215

**Verdict: Approved.**

All four requested changes are addressed correctly.

---

### Item 1 — Duplicate echo (dismiss_anr.sh)

Fixed correctly. The `elif`/`else` branches no longer each contain an echo. A single `if [[ -n "$pct" ]]` guard placed after `fi` covers both parseable paths with one log line. The `if [[ -z "$pct" ]]` branch retains its own distinct echo (`cpu=(unknown)%`), matching the specification exactly.

### Item 2 — Stale comment (SLEEP_AFTER_ANR_DETECTED)

Fixed in commit `052190728a7d`. The comment now says `KEYCODE_ENTER`.

### Item 3 — LAUNCHER_PACKAGE variable

Implemented correctly. The variable is set with the right default before the consumer subshell is spawned, so it is inherited without needing to be exported — plain shell variables are visible in child subshells. It drives both the logcat `case` pattern match and the `grep -iF "$LAUNCHER_PACKAGE"` CPU check. The switch to `-F` (literal matching) is a genuine improvement: the old `-i` alone would have treated the dots in the package name as "any character" in `grep`'s BRE mode. Header and inline comments explain the default and the AOSP alternative. Shell tests exercise the default package name correctly throughout.

### Item 4 — Test rename

Done in commit `bc0b1203`. `test_no_anr_dialog_means_no_keycode_back` → `test_no_anr_dialog_means_no_dismiss_keyevent`; `keyback_calls` → `dismiss_calls`. No logic changes.

---

### Test counts

No tests were deleted or disabled. The new commits add shell test (h) and `TestDismissAnrUsesKeycodeEnter` (two focused Python tests). The Author-reported totals of 15 shell and 57 Python are consistent with these additions.

---

One cosmetic note carried forward from the prior review (not blocking): `keyback_back_calls` in `test_anr_dialog_on_first_attempt_is_dismissed_before_screencap` still tracks ENTER sends under a name suggesting BACK. The assertion is correct; only the local variable name is stale. Worth a cleanup, but not a merge blocker.

---

aunger:
I see what you did in the update, but making the launcher app ID configurable isn’t what I meant. This should be detectable at runtime. At the least, you can grep for `/nexuslauncher|launcher3/`. Better yet would be looking up the app that’s set as the launcher or attached to the home key.

**However**, I suspect that just watching one process isn’t the best, even though we know which one usually ANRs. Would watching overall resource availability be better?

aunger:
Good question on overall resource availability. The tradeoff is: per-launcher CPU is a targeted signal — less noise, and the script exits as soon as that one process settles — but it could miss an ANR caused by a different process starving the launcher. Watching overall CPU/load would catch any heavy system activity, but the settle gate becomes "the whole system is quiet," which takes longer and may keep CI blocked by unrelated activity.

For this specific problem (Pixel Launcher ANRing on `ACTION_PACKAGE_REPLACED`), the bottleneck is the launcher's own main thread, so watching that process is the right signal — we care specifically that *it* has recovered, not that the system at large is idle. Now that the launcher is detected at runtime rather than hardcoded, the per-launcher approach also covers AOSP devices running Launcher3. The logcat ANR trigger remains the fast path; the CPU check is just the settle gate confirming the launcher thread has drained before we hand off to the green-feed check.

---

aunger:
## Code Review — commit `7fa42ba` ("Detect home launcher at runtime instead of env var")

**Verdict: CONDITIONAL APPROVAL — one stale comment needs fixing; everything else is correct.**

---

### Detection logic

The `cmd package resolve-activity --brief` invocation is correct. The output format of `--brief` is `package/activity` (e.g. `com.google.android.apps.nexuslauncher/.NexusLauncherActivity`), so the guard `[[ "$_resolve_out" == */* ]]` and extraction `${_resolve_out%%/*}` are both right. If the command returns nothing, exits nonzero, or returns text without a `/`, `LAUNCHER_PKG` stays empty and the fallback triggers — the logic is sound.

One mild note on `--brief` portability: on some Android versions the flag may emit a header line like `priority=0 preferredOrder=…` on stdout before the component string, which would cause `head -1` to grab the header instead. That header has no `/`, so the slash check catches it and falls back cleanly. Not a bug, but worth knowing.

### Subshell variable inheritance

`_LAUNCHER_PKG_LOCAL="$LAUNCHER_PKG"` is assigned in the outer subshell before `( while IFS= read -r line; ... ) &` forks. The forked consumer inherits the value correctly. No issue here.

### Logcat consumer fallback path

The double-`case` construction is slightly verbose but correct. The outer `case` pre-filters to lines containing `ANR in `, then the inner branch either does an exact string match on `$_LAUNCHER_PKG_LOCAL` or falls to the `grep -qiE 'ANR in [^ ]*(pixellauncher|launcher3)'` pattern. The pattern is case-insensitive and correctly matches `com.android.launcher3`.

### CPU check fallback

`grep -iF "$LAUNCHER_PKG"` (fixed-string, case-insensitive) when a package is known, and `grep -iE "pixellauncher|launcher3"` in fallback mode. Both are correct.

### Unknown-CPU handling (section h)

The rework that separates "unparseable percentage" from "low percentage" is correct: an empty `pct` now skips the `idle_count` increment without resetting it to 0, and the log line is still emitted. The test in (h) verifies this by requiring at least 4 `cpuinfo` polls before exit.

### Section (i) — fallback test

The new test correctly simulates a device where `cmd package resolve-activity` returns nothing (mock exits 0 with no output). It verifies both the logcat ANR path (`com.android.launcher3` matched via pattern) and the CPU-idle exit path. The 17-test total is reached.

---

### Bug: stale "KEYCODE_BACK" references in the section (f) comment block

`test_dismiss_anr.sh` section (f) still contains two stale references to `KEYCODE_BACK` in its prose comments (not in executable code):

```
# The mock cpuinfo returns high CPU until KEYCODE_BACK has
# been sent twice, then returns low CPU so idle_count reaches 2.
# We verify that KEYCODE_BACK is sent at least twice and the script exits 0.
```

The rest of the PR correctly updated every other reference. These two lines should read `KEYCODE_ENTER`. They are comments only, so they do not affect correctness or test outcomes, but they are misleading.

This is the only issue I found. Fix those two comment lines and the commit is ready to merge.

---

aunger:
**Sanity check on commit `fe6079b` — do not merge as described.**

The commit message claims this changes exactly two comment lines in `scripts/test_dismiss_anr.sh` section (f). The actual diff is dramatically larger:

- **`scripts/check_green_feed.py`**: 5 hunks of functional changes — log message strings and a live `subprocess.run` argument all change from `KEYCODE_BACK` to `KEYCODE_ENTER`.
- **`scripts/dismiss_anr.sh`**: Extensive functional rewrites — new launcher-detection block (~16 lines), logcat consumer logic rewritten to use dynamic `LAUNCHER_PKG`, CPU-check logic restructured, `sleep` replaced with `exec sleep`, new `logcat -d` case arms added throughout.
- **`scripts/test_check_green_feed.py`**: Test method renamed, variables renamed, assertion text updated, and a new 80-line test class (`TestDismissAnrUsesKeycodeEnter`) added.
- **`scripts/test_dismiss_anr.sh`**: Changes across sections b, c, d, e, f, g, and two entirely new test sections h and i added (136 new lines), plus new `cmd package resolve-activity` and `logcat -d` mock arms in every existing section.

The commit description is not accurate. This is a substantial feature/refactor commit that affects production logic in at least two scripts and adds significant new test coverage. It should be described and reviewed as such — not characterized as a two-line comment fix.

---

aunger:
Correction to my review of commit `fe6079b`: I made an error by reviewing the full cumulative PR diff rather than the individual commit. I'm withdrawing the "do not merge" verdict from that comment.

Looking at `fe6079b` in isolation (`git diff fe6079b~1 fe6079b`), the commit does exactly what it claims: two comment lines changed in `scripts/test_dismiss_anr.sh`, nothing else.

Corrected verdict: **Approved.**

---

scripts/dismiss_anr.sh
```diff
@@ -83,16 +95,24 @@
...
-        *"ANR in ${LAUNCHER_PACKAGE}"*)
-          touch "$ANR_FLAG" ;;
+        *'ANR in '*)
+          if [[ -n "$_LAUNCHER_PKG_LOCAL" ]]; then
+            case "$line" in
+              *"ANR in $_LAUNCHER_PKG_LOCAL"*) touch "$ANR_FLAG" ;;
+            esac
+          else
+            # Pattern fallback: match pixellauncher or launcher3 (case-insensitive).
+            if echo "$line" | grep -qiE 'ANR in [^ ]*(pixellauncher|launcher3)'; then
```
aunger: I’m pretty sure an agent took my earlier typo in a rough sketch idea and ran with it. Is it pixellauncher or still nexuslauncher? Do some research, stop screwing around, and get this simple thing working.

scripts/test_dismiss_anr.sh
```diff
@@ -238,15 +253,18 @@ echo "=== (f) Second ANR after first is dismissed ==="
...
-# We verify that KEYCODE_BACK is sent at least twice and the script exits 0.
+# We verify that KEYCODE_ENTER is sent at least twice and the script exits 0.
 
 SECOND_ANR_KEYEVENT_COUNT_FILE="$TMPDIR_TESTS/second_anr_keyevent_count"
 echo 0 > "$SECOND_ANR_KEYEVENT_COUNT_FILE"
 
 mock_adb_second_anr="$(make_mock_adb "adb_second_anr" "
 case \"\$*\" in
+  *'cmd package resolve-activity'*)
+    echo 'com.google.android.apps.nexuslauncher/.NexusLauncherActivity'
```
aunger, line 267: All these tests that use nexuslauncher are obviously either doing nothing or their results are being ignored.

aunger:
## Review — commit `2634127` ("fix: correct fallback pattern from pixellauncher to nexuslauncher") + test-quality audit

**Verdict: Approved.** The fix is correct. But the owner's instinct about test quality is right: the fallback path is only *partially* exercised, and that is precisely why the `pixellauncher` typo slipped through unnoticed.

### 1. The `pixellauncher` fix is complete

`grep -rn pixellauncher scripts/` returns nothing. `nexuslauncher` is now correct in all three fallback sites in `dismiss_anr.sh` (line 69 log message, line 110 comment + line 111 logcat grep, line 160 CPU grep) and in the section (i) comment in `test_dismiss_anr.sh` (line 464). Suite runs green: 17 passed, 0 failed.

### 2. Which path does each section actually exercise?

I checked every mock's `resolve-activity` arm. Sections **a–h all mock** `cmd package resolve-activity` and return `com.google.android.apps.nexuslauncher/.NexusLauncherActivity` (a:51-52, b:82-83, c:132-133, d:177-178, e:205-206, f:265-266, g:331-332, h:413-414). So a–h all run the **primary path** — `LAUNCHER_PKG` is set and the CPU check uses `grep -iF "$LAUNCHER_PKG"` (line 158), not the fallback alternation. The runtime logs confirm this ("Launcher detected: com.google.android.apps.nexuslauncher" in g and h).

Only section **(i)** returns empty (line 473 `exit 0`) and runs the **fallback path** ("Launcher detection failed — using ... pattern fallback" in the log).

So this is *not* a case of a–g silently falling through to fallback — they genuinely cover the primary path, which is the path real Pixel CI hits. That part is fine and not vacuous.

### 3. The real gap: the `nexuslauncher` fallback arm has zero coverage

Section (i) — the only fallback test — uses `com.android.launcher3` for both its ANR line (484) and cpuinfo (492/494). `launcher3` matches under **both** the old `pixellauncher|launcher3` and the new `nexuslauncher|launcher3` alternation. I confirmed this directly:

```
$ echo '  60% com.android.launcher3: launcher' | grep -iE "pixellauncher|launcher3"   # matches
$ echo 'E/ActivityManager: ANR in com.android.launcher3' | grep -qiE 'ANR in [^ ]*(pixellauncher|launcher3)'  # matches
```

**Consequence:** section (i) passed *identically* before and after this commit. No test in the suite would have caught the `pixellauncher` typo, because no test ever feeds a `nexuslauncher` process through the fallback path — every section that uses `nexuslauncher` cpuinfo (a–h) has `LAUNCHER_PKG` set and so takes the `grep -iF` branch, never the alternation. The fallback's `nexuslauncher` arm is dead-untested.

This is the test-quality hole worth closing. Recommend either:
- adding a section that returns empty from `resolve-activity` **and** uses `nexuslauncher` cpuinfo/ANR lines (directly exercises the regressed arm), or
- changing section (i) to do this instead of `launcher3` (but then `launcher3` loses coverage — so a separate section is better).

Suggest spawning a follow-up issue rather than expanding this PR's scope.

### 4. Other observations (non-blocking)

- Section (h)'s `(unknown)` cpuinfo line at 431 correctly produces no leading integer, so the unparseable-skip branch is genuinely exercised — that test is sound.
- Section (d) "absent launcher" uses `some.other.process` cpuinfo while `LAUNCHER_PKG` is set to nexuslauncher, so `grep -iF` finds nothing and the absent-as-idle branch is correctly hit. Sound.
- Minor: with the primary path, section (i) is the *only* line that depends on the alternation pattern at all, yet it's named/described as the general "pattern fallback" test. After adding nexuslauncher coverage, consider clarifying that (i) covers the AOSP/`launcher3` case specifically.

No correctness blocker in the commit itself; approving. The coverage gap is a pre-existing test weakness this commit exposes rather than introduces.

---

aunger:
**Verdict: Approved**

Reviewed commit `ecd1d50`. Section (j) does exactly what it should and earns its place next to (i).

Walkthrough:
- `resolve-activity` returns nothing, so `LAUNCHER_PKG` stays empty and both fallback branches engage (`dismiss_anr.sh` lines 110-114 for logcat, 159-161 for cpuinfo).
- The logcat line `ANR in com.google.android.apps.nexuslauncher` matches `grep -qiE 'ANR in [^ ]*(nexuslauncher|launcher3)'` only through the `nexuslauncher` arm.
- cpuinfo reports nexuslauncher busy (60%) twice then idle (1%), driving the idle-count exit path to 0.
- Both assertions (exit 0, KEYCODE_ENTER sent) are correct.

Section (i) is untouched — the commit is purely additive (66 insertions, 0 deletions).

I confirmed the test bites: copying the repo and flipping the pattern to `pixellauncher|launcher3` makes (j)'s KEYCODE_ENTER assertion fail (18 passed, 1 failed) while (i) still passes. With the correct pattern the suite is 19/19. So (j) genuinely guards the nexuslauncher arm rather than duplicating (i).

Minor, non-blocking observation: under the `pixellauncher` typo the "script exits 0" assertion still passes (empty cpuinfo match is treated as idle), so only the KEYCODE_ENTER assertion is doing the discriminating work. Fine as-is, just worth knowing the exit-0 check alone carries no fallback-arm coverage.

No regressions.

---

PR #216, opened 2026-05-24
Handle conditional approval in Orchestrator workflow (#210)

### LINKED ISSUE #210: Agents: can approvals contain instructions?
Sometimes a reviewer agent wants to give approval plus a final instruction. "Approved. Just fix this constant before you merge."

Our agentic workflow doesn't handle this; the final instruction is never delivered to the Author. What to do?

1. Instruct Reviewers not to do this
2. Instruct the Orchestrator to permit it (haha, no, I don't trust an agent to make even a simple change without review)
3. Ignore the Reviewer's approval and run one more edit/review cycle
4.  As above, but invoke the Reviewer with, *The Author made the final change you requested. Do a quick sanity check to confirm nothing else changed.*
5. As above, but with a Haiku agent, maybe with only limited context

---

**Recommendation: Option 5**

When a Reviewer gives conditional approval ("Approved, but fix X before merging"), the Orchestrator should treat it as "changes requested" for workflow purposes — the Author must still act. After the Author makes only the specific change(s) requested, the Orchestrator spawns a Haiku sanity-check agent (not a full review cycle) with narrowed context: the Reviewer's specific instruction and the Author's diff responding to it. The Haiku agent's only question is "did the Author address the requested change and nothing else?" If it confirms yes, treat as approved and proceed to CI. If it raises a new concern, escalate to the user rather than spinning another full cycle.

Why option 5 over the others:
- Option 1 gags the Reviewer, losing valuable last-minute feedback.
- Option 3 discards the approval signal and runs a wasteful full review cycle.
- Option 4 is nearly identical but uses a Sonnet for a task that's genuinely lightweight — "did exactly one thing change?" doesn't warrant it.
- A Haiku with limited context is the right fit: cheap, focused, and resistant to scope creep from re-reading the full diff.

aunger:
## Summary

- Adds a new **Handling conditional approval** section to `agents/dev_orchestration.md` that defines how the Orchestrator routes and verifies a Reviewer's conditional approval (approval + a final required change).
- Updates `agents/pr_participation.md` to document that Reviewers may give conditional approval and must phrase the approval signal and the instruction as distinct, unambiguous sentences.

## Behavior

When a Reviewer gives conditional approval the Orchestrator:
1. Treats it as "changes requested" — routes the Author to make only the named change(s).
2. After the Author commits, spawns a Haiku sanity-check agent with narrowed context (Reviewer instruction + Author's responding diff only).
3. If Haiku confirms the change was addressed → proceeds directly to CI Monitor loop (no full review cycle).
4. If Haiku raises a new concern → escalates to the user; stops.

## Test plan

- [ ] Read both modified files and confirm new sections are coherent and consistent with existing workflow rules.
- [ ] Verify no bylines, session URLs, or attribution footers appear in the PR or commits.
- [ ] Confirm the branch targets the default branch.

Closes #210

---
_Generated by [Claude Code](https://claude.ai/code/session_016FRYYoQ9ebhrD2dir3LdR6)_


aunger (commented):
The structure here is mostly right, and `pr_participation.md` is clean — the phrasing examples and placement are good. The Haiku narrowed-context constraint is explicit and correct. But there's a logic gap in `dev_orchestration.md` that needs fixing before this merges.

**The Haiku outcome branches are incomplete.**

The pseudocode gives two outcomes:

```
if Haiku confirms yes  → treat as approved; proceed to CI Monitor loop
if Haiku raises a new concern → escalate to user; stop
```

This is missing a third case: **Haiku says "no, the Author did not address the requested change"** — without raising a new concern. Right now, that falls into `"raises a new concern"` by default, which would escalate to the user and stop. But the right response is to re-route the Author to fix what they missed, not to escalate. Escalation is appropriate when there's a surprising new problem, not when the Author simply didn't complete the task.

The two cases — *didn't address it* and *addressed it but introduced something new* — need separate branches in the pseudocode.

**The Haiku prompt reinforces this ambiguity.**

`"Did the Author address this and only this?"` is a compound question. "Only this" is asking whether the Author stayed in scope. A "no" answer could mean either "didn't finish the job" or "did extra things the Reviewer didn't ask for." Both collapse into the same branch right now, but they call for different Orchestrator responses. Consider separating the prompt into two distinct questions, or at minimum giving the Orchestrator explicit guidance on how to interpret each failure mode.

**Minor: PR body byline.** The PR description ends with `_Generated by Claude Code_ (session URL)`. The no-bylines rule in `pr_participation.md` applies to PR reviews and comments, and while the PR body is a slightly different artifact, it's inconsistent and worth cleaning up regardless.

---

**Decision for the Orchestrator: changes requested.** The missing "not addressed" branch is a behavioral gap, not a style nit — an Orchestrator following this spec as written would escalate to the user whenever the Author fails to complete a conditional change, instead of routing them back to fix it. That's the wrong default and needs an explicit branch before this ships.

---

aunger (commented):
The fix addresses exactly what was requested — nothing more, nothing less. The third Haiku outcome branch is now present and correct, and the A/B/C labeling makes the prompt unambiguous.

**What changed (commit c55422c):**

The pseudocode now has three explicit branches where it previously had two. Outcome B — "Author did not address the requested change (incomplete or missing work, no new concerns raised)" — is now a distinct re-route-Author path rather than falling through into the escalate-to-user branch. The Haiku prompt names all three outcomes explicitly, and the constraint block below the pseudocode restates the B-vs-C distinction in plain prose as a second layer of guidance. That's good defensive documentation: the pseudocode gives the Orchestrator its decision tree, and the constraints block gives it the conceptual distinction it needs to interpret edge cases correctly.

**Checklist against Option 5 requirements:**
1. Detect conditional approval — covered in both files.
2. Treat as "changes requested" — explicit.
3. Spawn Haiku with narrowed context — constraints are specific and correctly exclude full PR diff and prior review history.
4. Three Haiku outcomes A/B/C — now correct and unambiguous.
5. `pr_participation.md` updated with conditional-approval phrasing guidance — clean and well-placed.

**Scope discipline:** Only `dev_orchestration.md` and `pr_participation.md` were touched. No unrelated changes snuck in.

**Still noting:** The PR body still has the `_Generated by Claude Code_` byline I flagged last round. The no-bylines rule in `pr_participation.md` is written to cover reviews and comments, so the PR body is adjacent rather than directly in scope — but it's inconsistent. Not blocking.

**Decision: approved.** The behavioral gap is closed. This is ready to merge.

---

aunger (commented):
The state at the prior approval (`c55422c`) was correct. The post-approval commit (`cfd3943`, "Refine Reviewer and Haiku agent guidelines") walks back the most important fix this PR introduced.

**The B/C distinction was re-merged, and that's a regression.**

The previous review explicitly required separate branches for "Author didn't address it" (B) and "Author introduced a new concern" (C), because the two cases call for different Orchestrator responses. The Author's fix at `c55422c` got that right:

```
if Haiku answers B → re-route Author to complete the task; treat as "changes requested" again
if Haiku answers C → escalate to user; stop
```

The new commit collapses both into:

```
if Haiku answers B or C → the PR hasn't yet converged; resume the normal cycle by routing to the full-fledged Reviewer.
```

This is wrong in both directions. When the Author simply didn't complete the task (B), routing to a full Reviewer cycle is wasteful — the Author hasn't done anything new for the Reviewer to evaluate. When the Author's response raises a surprising new concern (C), routing to the Reviewer instead of the user buries the signal; Option 5 specifically says C should escalate to the user and stop. Both cases now get the same handling, which is the same behavioral gap the prior Reviewer flagged.

The constraint block compounds this. The old version stated the distinction in plain prose:

> If Haiku returns B, re-route the Author to complete the task — do not escalate to the user.
> If Haiku returns C, do not spawn a new review cycle — escalate to the user instead.

The new version replaces both lines with one:

> If the Haiku agent responds with anything other than a clear-cut A, then return to the normal PR cycle: next invoke the full Reviewer.

The prose distinction that was the second layer of guidance for edge cases is now gone.

**The "instruct Author to make only" → "route to Author to consider" change is also a step backward.**

"Consider" is meaningfully weaker than "make only the specific change(s) named." The conditional approval workflow depends on the Author being directed to a specific, bounded task. "Consider" leaves room for interpretation; the previous phrasing was correctly prescriptive.

**The deleted Reviewer-exit guidance.**

Removing "The Reviewer posts its review immediately upon completing its analysis and then exits..." from the Assigning a Reviewer section is fine only if that instruction lives somewhere the Orchestrator will reliably see it when assigning a Reviewer. It does appear in `pr_participation.md` under the Reviewer role. Whether the Orchestrator reads that file when dispatching is a question of how subagents are briefed — it's not obviously broken, but the deletion trades an explicit reminder in the dispatch instructions for something the Orchestrator has to find in a separate document.

**The "When to abort" deletion.**

Removing "If the Reviewer agrees the PR may be merged" loses a legitimate abort condition. The remaining two bullets cover deadlock (three rounds) and programmer giving up, but not the success path in the abort section. Minor, but the original sentence was useful.

**The delegation-rule tweaks are fine.** Splitting uncommitted-work handling into two cleaner rules and the "separate subagents per ticket" rewrite are improvements with no downsides I can see.

**Decision: changes requested.** The B/C branch collapse is a behavioral regression that directly contradicts the issue's Option 5 specification. The three Haiku outcomes need to be restored as separate, correctly-routed branches — B back to Author, C to user escalation — in both the pseudocode and the constraint prose.

---

aunger:
The Author and Reviewer are adversarial, and each must have the freedom to concede, compromise, or contend for their position. I’m concerned that what you’re arguing for gives the Reviewer the final word.

aunger:
The concern about Reviewer authority is legitimate, and I want to engage with it directly rather than back down without reason. But I think it misidentifies the problem with outcome C.

**The Haiku check is not a review cycle.** At the point it fires, the Reviewer has already approved — conditionally, but approved. The Author–Reviewer adversarial dynamic has already played out. The Haiku agent is not adjudicating the merit of the original change; it's doing a narrow mechanical check: *did the Author do the one specific thing the Reviewer named as a condition of their approval?*

Given that framing, here's why the two outcomes require different handling:

- **Outcome B** (Author didn't address the change): The current code routes this to a full Reviewer. That's actually where I'm most willing to concede ground. If you believe the Author deserves a chance to argue "I deliberately didn't make that change and here's why," then routing back to a full review cycle preserves the adversarial dynamic you want. I can accept that, even though it reintroduces the full review overhead for what is often just incomplete work.

- **Outcome C** (Author's response introduces a new concern): This is where I hold the line. A full Reviewer cycle is the wrong tool here, and it doesn't protect the Author's position — it just picks up a new dispute that the Reviewer never raised in the original review. The Reviewer doesn't "get the final word" by having the Orchestrator stop and ask a human what to do. That's precisely the opposite: the human retains the final word.

**The current merged text conflates B and C into the same action.** That's the specific thing I'm objecting to — not the routing of B, but the equation of B with C. If a new concern emerged that no one foresaw, escalating to the user is the conservative, human-in-the-loop response. Routing it to a Reviewer instead means an automated agent is adjudicating a surprise without the human knowing it happened.

To summarize what I'd propose: keep routing B to a full review cycle if you want the Author to be able to contest it. But split C back out and escalate it to the user. That respects the adversarial design without removing the human from unexpected situations.

---

aunger (commented):
**Reviewer verdict: changes requested**

The core fix is in. Outcome C is now correctly split out and escalated to the user:

> if Haiku answers C → the PR is unstable; stop the PR cycle and escalate to the User.

The B/C distinction is restored and unambiguous — B routes to a full Reviewer cycle, C escalates to the human. The constraints paragraph reinforces this three-way split explicitly. That part of the earlier objection is fully addressed.

However, the updated "When to abort" section introduced a new problem:

> **If the Author introduces new ideas after the Reviewer gives conditional approval** or claims the issue cannot be solved as stated

Two issues here:

1. **Conflict with the Haiku logic.** If the Author introduces new ideas after conditional approval, the updated flow already handles that — Haiku outcome C fires and escalates to the user. Listing this as an *abort* condition duplicates that logic with a different outcome (abort vs. escalate). An Orchestrator following both rules would have no clear authority on which to apply.

2. **Garbled secondary clause.** "or claims the issue cannot be solved as stated" is copied from the bullet immediately above it ("If the Programmer gives up or claims the issue cannot be solved as stated"). The duplication appears accidental and makes the sentence incoherent — the Author introducing new ideas is unrelated to claiming the issue is unsolvable.

The replaced bullet ("If the Reviewer agrees the PR may be merged") described a success exit, not an abort, so removing it from this list is defensible. But the replacement text conflicts with the Haiku logic and should be corrected or removed. Dropping this bullet entirely is a clean fix, since new-concern-after-conditional-approval is now handled by Haiku outcome C.

---

aunger:
I’m glad you caught the copy paste mistake. 

I consider escalating to the human user the same as aborting the review-edit orchestration. Clarified that in the docs now, I believe.

aunger:
**Review verdict: Approved.**

The garbled bullet is gone. The updated "When to abort" reads:

> If the Author introduces new ideas after the Reviewer gives conditional approval. That is, if the "sanity check" Haiku agent does not answer A or B.

This is consistent with the Haiku block, where outcome C triggers "stop the PR cycle and escalate to the User." "Does not answer A or B" correctly collapses to "answers C" in the three-outcome system, so there is no conflict with the Haiku logic. The owner's stated intent — that escalating to the user is semantically equivalent to aborting the orchestration loop — is now accurately expressed.

Two minor observations, neither blocks approval:

1. The "When to abort" list doesn't cover the separate ambiguous-Haiku case (unclear response, not A/B/C). That case is handled inline in the Haiku block itself, so the abort list being non-exhaustive is fine.
2. The round-count bump from three to four is an independent change with no stated rationale. Benign, but worth a note in commit history if there was a reason for it.

The Option 5 implementation is otherwise complete: the Haiku sanity-check flow is clearly specified in `dev_orchestration.md`, conditional-approval guidance is added to `pr_participation.md`, and the Author-reply expectation is documented. This is ready to merge.

---

PR #219, opened 2026-05-25
Prefix screenshot filenames with the test name using ScreenshotTestRule

### LINKED ISSUE #201: Associate screenshots with the test that captured them
## Problem

Screenshots captured during a test run are currently not reliably linked to the specific test that produced them. This makes post-hoc analysis — including OCR in CI (#198) and any future AI image processing (#199) — harder, because there is no guaranteed mapping from image file to test case.

### Option A — Filesystem snapshot diff per test

At the **start of the test class** (a `@BeforeClass` / `@BeforeAll` hook), record the set of files already present in the screenshot output directory. At the **end of each test** (an `@AfterEach` hook), diff the current directory contents against that baseline and against the snapshot from the previous test. Any new files that appeared during the test are attributed to it.

- **Pros:** No changes required to how screenshots are taken; works even for third-party code that saves screenshots.
- **Cons:** Slightly racy under parallel test execution; the baseline must be captured before any test in the class starts to avoid missing files written during `@BeforeEach`; adds file-system I/O on every test teardown.

### Option B — Test fixture wraps screenshot saving

Modify the test base class or a shared test rule/extension to intercept or replace the screenshot-capture call. The wrapper prepends the current test name to the filename before saving.

Example: `failure_screenshot.png` → `PixelCameraOverlayE2ETest_overlayAppearsWhenViewfinderOpens_failure_screenshot.png`

- **Pros:** Naming is deterministic; no diff logic needed; files are self-describing in the artifact browser; plays well with `<system-out>` JUnit XML injection (see #198).
- **Cons:** Requires all screenshot-saving code to go through the fixture; screenshots saved outside the fixture (e.g. by the test runner itself on crash) are not covered.

## Recommendation

Option B is cleaner for screenshots the test code controls, and Option A is a good safety net for runner-generated screenshots. A combined approach — Option B as the primary mechanism, Option A as a sweep at teardown to catch stragglers — gives the best coverage.

## Relationship to other issues

- #198 (OCR on failure screenshots) becomes significantly more useful once screenshots are named per-test, because OCR output can be linked directly to the failing test case.
- #199 (AI image processing) has the same dependency.

aunger:
## Summary

Fixes #201

Screenshots captured during a test run are now reliably associated with the specific test that produced them.

- **Option B (primary):** `ScreenshotTestRule` sets a thread-local prefix (`_`) at the start of each test. `Screenshot.saveForArtifact()` reads this prefix and prepends it to every filename. For example, `&#34;0-screen.png&#34;` becomes `&#34;GalleryButtonVisualE2ETest_test0_smokeGreenFeedVisible_0-screen.png&#34;`, and the failure screenshot becomes `&#34;GalleryButtonVisualE2ETest_test0_smokeGreenFeedVisible_failure.png&#34;`.

- **Option A (safety net):** `ScreenshotTestRule` snapshots the screenshot directory before each test and, after the test completes, renames any new files that are not already prefixed with the test name. This catches screenshots written outside `saveForArtifact()` (e.g. by a crash handler or the test runner).

`ScreenshotTestRule` supersedes `FailureScreenshotRule` — both E2E test classes now use the new rule, which covers the failure-screenshot behavior plus the naming enhancements. The pure naming logic lives in `ScreenshotNaming` (sharedTest), which has no Android dependencies and is covered by JVM unit tests.

## Test plan

- [ ] `ScreenshotNamingTest` passes (JVM unit tests covering `buildPrefix`, `resolvedName`, and the combined round-trip)
- [ ] `./gradlew :app:testDebugUnitTest` passes
- [ ] `./gradlew :app:compileDebugAndroidTestKotlin` passes
- [ ] On a connected device/emulator: `./gradlew connectedE2EAndroidTest` produces screenshot files prefixed with the test class and method name

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
@@ -49,7 +49,7 @@ import kotlin.math.sqrt
 class GalleryButtonVisualE2ETest {
 
     @get:Rule
-    val screenshotOnFailure = FailureScreenshotRule()
+    val screenshotRule = ScreenshotTestRule()
 
     private val instrumentation = InstrumentationRegistry.getInstrumentation()
     private val context = instrumentation.targetContext
--- a/app/src/androidTest/java/com/gb4pc/e2e/PixelCameraOverlayE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/PixelCameraOverlayE2ETest.kt
@@ -39,7 +39,7 @@ import org.junit.runner.RunWith
 class PixelCameraOverlayE2ETest {
 
     @get:Rule
-    val screenshotOnFailure = FailureScreenshotRule()
+    val screenshotRule = ScreenshotTestRule()
 
     private val instrumentation = InstrumentationRegistry.getInstrumentation()
     private val fixture = E2EFixture(
--- a/app/src/androidTest/java/com/gb4pc/e2e/ScreenshotTestRule.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/ScreenshotTestRule.kt
@@ -1,0 +1,131 @@
+package com.gb4pc.e2e
+
+import android.util.Log
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.uiautomator.UiDevice
+import com.gb4pc.e2e.visual.Screenshot
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+import java.io.File
+
+/**
+ * JUnit [TestWatcher] rule that associates every screenshot produced during a test with
+ * the test that produced it.
+ *
+ * ## Option B — deterministic prefix (primary)
+ *
+ * At the start of each test [starting] sets [Screenshot.currentTestPrefix] to
+ * `"<ClassName>_<methodName>"`. [Screenshot.saveForArtifact] reads this prefix and
+ * prepends it to every filename, so `"0-screen.png"` becomes e.g.
+ * `"GalleryButtonVisualE2ETest_test0_smokeGreenFeedVisible_0-screen.png"`.
+ * The prefix is cleared in [finished] so screenshots saved outside of any test body
+ * are never mistakenly attributed.
+ *
+ * ## Option A — straggler detection (safety net)
+ *
+ * In [starting], a snapshot of all files currently in the screenshot directory is
+ * recorded. In [finished], any new files that appeared during the test but were NOT
+ * saved through [Screenshot.saveForArtifact] (i.e. already prefixed by Option B) are
+ * renamed with the test name prefix so they too carry a test association. This catches
+ * screenshots written directly by the test runner or other out-of-band paths (e.g. a
+ * crash handler).
+ *
+ * ## Failure screenshot
+ *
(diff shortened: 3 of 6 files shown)


aunger (commented):
**Review — Changes requested**

Two issues require changes before this can be merged.

---

**1. Attribution bylines (blocking)**

The PR body ends with:

```
_Generated by [Claude Code](https://claude.ai/code/session_01BUu7w5AfqZgJyrVdBwUjh2)_
```

Both commits end with a bare session URL:

```
https://claude.ai/code/session_01BUu7w5AfqZgJyrVdBwUjh2
```

`agents/pr_participation.md` is unambiguous: "Do not append any Claude attribution byline, session URL, or footer." These must be removed from the PR description and from both commit messages.

---

**2. DRY violation in `ScreenshotTestRule.failed()` (blocking)**

`Screenshot.screenshotDir()` was extracted specifically to avoid duplicating directory-resolution logic. `failed()` ignores it and rebuilds the path inline:

```kotlin
val externalFilesDir = InstrumentationRegistry.getInstrumentation().targetContext
    .getExternalFilesDir(null)
if (externalFilesDir == null) { ... return }
val dir = File(externalFilesDir, "screenshots")
dir.mkdirs()
```

This is a DRY violation against code introduced in the same PR for exactly this purpose. `failed()` should call `Screenshot.screenshotDir()`, wrapping it in a try/catch to preserve the graceful-degradation behaviour it currently provides for unavailable external storage.

---

**Minor (no change required):** The KDoc on `ScreenshotTestRule` states that `currentTestPrefix` "is cleared before `failed` is called by JUnit's default watcher order." This is backwards — in `TestWatcher`, `failed()` fires *before* `finished()`, so the prefix is still set when `failed()` runs. The code itself is correct (it does not use the thread-local inside `failed()`), but the comment is misleading and should be corrected when the other changes are made.

---

The core logic is correct, commit organisation is good, and the `ScreenshotNaming` unit-test coverage is solid. The two blocking items above are the only obstacles to approval.

---

aunger:
**Review — follow-up commit `92f4711`**

All three requested changes are correctly addressed.

**Verification of fixes**

1. **Attribution bylines** — PR body is clean; no footer or session URL present.

2. **DRY violation in `failed()`** — The method now delegates to `Screenshot.screenshotDir()` and wraps it in a try/catch. The inline directory-resolution block is gone. Fix is correct and consistent with how `renameStragglersForTest` acquires the directory.

3. **KDoc lifecycle order** — The updated comment accurately states that `currentTestPrefix` is still live when `failed()` runs and is cleared in `finished()`, which `TestWatcher` calls afterwards. Matches the actual execution order.

**Full review**

The overall design is sound. A few observations:

- `finished()` clears the ThreadLocal prefix *before* calling `renameStragglersForTest`. This is safe because `renameStragglersForTest` (and `failed()`) derive the prefix via `testPrefix(description)` — reading from the `Description` object — rather than from the ThreadLocal. The logic is self-consistent.

- The straggler guard (`name !in baselineFiles && !name.startsWith(prefix)`) correctly avoids double-prefixing files already attributed by Option B.

- Separating pure naming logic into `ScreenshotNaming` (sharedTest) is a clean architectural choice: it removes Android dependencies from the naming concern and makes it trivially testable on the JVM.

- `ScreenshotNamingTest` covers `buildPrefix`, `resolvedName` (with and without prefix), and the round-trip. Coverage is adequate for the scope.

- `screenshotDir()` and `currentTestPrefix` being `internal` is the right visibility.

No issues found. **Approved.**

---

PR #220, opened 2026-05-25
Run OCR on E2E failure screenshots and surface results in job summary

### LINKED ISSUE #198: Run OCR on screenshots when tests fail
## Problem

When E2E or instrumented tests fail, CI captures screenshots of the emulator screen, but diagnosing the failure requires downloading and manually inspecting each image. There is no automated extraction of the visible text content from those screenshots.

## Proposed solution

Add a CI workflow step (guarded with `if: failure()`) that runs OCR on screenshots collected during a failed test run, writing a companion `.ocr.txt` file alongside each image before the artifact is uploaded.

**Why companion files (recommended for this stage):**
The screenshot and its OCR output travel together in the artifact, scale to any number of failures, and there is no size concern. The artifact browser lets you open image and text side by side. A concise index (one row per screenshot, first ~80 chars of OCR) can still be appended to `$GITHUB_STEP_SUMMARY` for quick scanning without bloating it.

**OCR tool:** Tesseract (`tesseract-ocr` package, available on Ubuntu runners via `apt`), or the `pytesseract` Python wrapper.

### Other options considered

| Option | Pros | Cons |
|---|---|---|
| **`$GITHUB_STEP_SUMMARY` only** | One place to look, no artifact needed | Becomes unwieldy with many screenshots; whole-job scope, not per-test |
| **`<system-out>` in JUnit XML** | Per-test; picked up naturally by test reporters | Requires matching screenshot filenames to test case names; fragile without a naming convention |
| **`::error::` step annotations** | Appears inline on PR diff | 256-char limit; only useful for a short label, not full OCR text |
| **Step log stdout** | Zero setup | Buried in raw logs; not surfaced in summary or artifact browser |

The `<system-out>` approach becomes much more attractive once screenshots are reliably named per-test (tracked in a separate issue).

## Out of scope

- Processing screenshots from passing test runs.
- Replacing the uploaded screenshot artifacts (still useful for visual inspection).
- AI-based image understanding (tracked in #199).

aunger:
## What this PR does

Fixes #198

When E2E tests fail, CI now automatically runs Tesseract OCR on the captured screenshots and surfaces the results directly in the GitHub Actions job summary, so engineers can read screen text without downloading and opening image files.

### Changes

**`scripts/ocr_screenshots.sh`** — new script that:
- Accepts a screenshot directory path
- Runs `tesseract` on every `.png` file (recursively), writing a companion `.ocr.txt` alongside each image
- Appends a markdown table to `$GITHUB_STEP_SUMMARY` with one row per screenshot (filename + first 80 chars of OCR output)
- Is non-fatal per file: if tesseract fails on one image, the row shows `(OCR failed)` and the script continues

**`.github/workflows/build.yml`** — two new steps:
- **OCR on E2E screenshots**: installs `tesseract-ocr` via apt and runs the script. Gated on `steps.pull_e2e_screenshots.outcome == &#39;success&#39;` so it only fires when screenshots were actually captured (i.e. when tests failed). Runs *before* the artifact upload step so `.ocr.txt` companions are included in the `e2e-screenshots` artifact.
- **Run shell script unit tests**: discovers and runs all `scripts/test_*.sh` files. This also begins exercising the pre-existing `test_dismiss_anr.sh`, which was not previously run in CI.

**`scripts/test_ocr_screenshots.sh`** — 14 test cases covering: normal operation, GITHUB_STEP_SUMMARY content and format, 80-char preview truncation, empty directory, missing directory, wrong argument count, and non-fatal tesseract failure.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -347,6 +347,13 @@ jobs:
           path: app/build/outputs/androidTest-results/
           retention-days: 14
 
+      - name: OCR on E2E screenshots
+        if: always() && steps.pull_e2e_screenshots.outcome == 'success'
+        run: |
+          sudo apt-get install -y tesseract-ocr > /dev/null 2>&1
+          chmod +x scripts/ocr_screenshots.sh
+          scripts/ocr_screenshots.sh /tmp/e2e-screenshots
+
       - name: Upload E2E screenshots on failure
         if: always() && steps.pull_e2e_screenshots.outcome == 'success'
         uses: actions/upload-artifact@v7
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -108,6 +108,13 @@ jobs:
       - name: Run Python script unit tests
         run: python3 -m unittest discover -s scripts -p "test_*.py" -v
 
+      - name: Run shell script unit tests
+        run: |
+          for test_script in scripts/test_*.sh; do
+            echo "==> Running $test_script"
+            bash "$test_script"
+          done
+
       - name: Grant execute permission for gradlew
         if: steps.diff_check.outputs.needs_full_build == 'true'
         run: chmod +x gradlew
--- a/scripts/ocr_screenshots.sh
+++ b/scripts/ocr_screenshots.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+# ocr_screenshots.sh — Run Tesseract OCR on every PNG in a directory and
+# append a markdown summary table to $GITHUB_STEP_SUMMARY.
+#
+# Usage: ocr_screenshots.sh <screenshot-dir>
+#
+# For each *.png found (recursively) this script:
+#   1. Runs `tesseract` to produce a companion <name>.ocr.txt file.
+#   2. Appends one table row (filename + first 80 chars of OCR text) to
+#      $GITHUB_STEP_SUMMARY when that variable is set.
+#
+# Exits 0 in all cases — OCR failure on individual images is non-fatal.
+
+set -euo pipefail
+
+if [[ $# -ne 1 ]]; then
+  echo "Usage: $(basename "$0") <screenshot-dir>" >&2
+  exit 1
+fi
+
+SCREENSHOT_DIR="$1"
+
+if [[ ! -d "$SCREENSHOT_DIR" ]]; then
+  echo "Directory not found: $SCREENSHOT_DIR" >&2
+  exit 1
+fi
+
+shopt -s globstar nullglob
+PNG_FILES=("$SCREENSHOT_DIR"/**/*.png)
+
+if [[ ${#PNG_FILES[@]} -eq 0 ]]; then
+  echo "No screenshots found in $SCREENSHOT_DIR; skipping OCR." >&2
+  exit 0
+fi
+
+# Write the summary header only when there are images to report.
+if [[ -n "${GITHUB_STEP_SUMMARY:-}" ]]; then
+  {
+    echo "## E2E Screenshot OCR"
+    echo ""
+    echo "| Screenshot | OCR preview (first 80 chars) |"
+    echo "| --- | --- |"
+  } >> "$GITHUB_STEP_SUMMARY"
+fi
+
+for PNG in "${PNG_FILES[@]}"; do
+  [[ -f "$PNG" ]] || continue
+
+  # Derive output base: strip .png, append .ocr so tesseract writes <name>.ocr.txt
+  OCR_BASE="${PNG%.png}.ocr"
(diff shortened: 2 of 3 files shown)


aunger:
## Code Review — PR #220

### Summary

The core OCR feature is correctly implemented: `ocr_screenshots.sh` runs Tesseract on every PNG, writes companion `.ocr.txt` files, and appends a markdown summary to `GITHUB_STEP_SUMMARY`. The step is ordered before the artifact upload (so `.ocr.txt` files are included), gated correctly on `steps.pull_e2e_screenshots.outcome == 'success'`, and the 14-case test suite covers the main behaviours with a clean mock-Tesseract approach.

### Issues

**1. PR description contains an attribution byline (blocking)**

The PR body ends with:

> _Generated by [Claude Code](https://claude.ai/code/session_01BUu7w5AfqZgJyrVdBwUjh2)_

Per project protocol, no Claude attribution byline, session URL, or footer should appear in PR description text. This must be removed before merging.

**2. Two commit messages contain session URLs**

Commits `58459d24` ("Associate screenshots with the producing test…") and `414e5a75` ("Add ScreenshotNaming…") both include `https://claude.ai/code/session_01BUu7w5AfqZgJyrVdBwUjh2` in their message bodies. Same protocol violation as above, in the commit history.

**3. Scope expansion not tied to issue #198**

Issues #201 (per-test screenshot naming), `ScreenshotNaming`, `ScreenshotTestRule`, and `ScreenshotNamingTest` are bundled into this PR. That's four new files and ~280 lines addressing a different issue. These are not required by #198 (OCR can operate on any filenames), and the issue requirements explicitly listed out-of-scope items without mentioning test attribution. The prerequisite argument is reasonable but this work should have been tracked separately (as indeed it was: #201 exists). Please justify why these are included here rather than in a dedicated #201 PR, or split them out.

**4. Minor: truncation test passes vacuously if regex finds no X's**

In test case (c), `PREVIEW_LEN` is extracted by grepping for `X+` in the summary row. If the mock tesseract output text is not embedded in the row as expected (e.g. the row format changes), the grep finds nothing, `wc -c` returns `0`, and `0 -le 80` passes silently without catching the regression. Adding a lower-bound assertion (`[[ "$PREVIEW_LEN" -ge 1 ]]`) would close this gap.

### Verdict

**Changes requested.** Items 1 and 2 (attribution bylines) are clear protocol violations and must be fixed. Item 3 requires a justification or a split. Item 4 is a test quality improvement worth addressing but not blocking on its own.

---

aunger (commented):
## Follow-up Review — PR #220

### Verification of the three requested fixes

**1. Attribution byline removed from PR description** — confirmed. The body now ends with the sentence about non-fatal tesseract failure; no Claude byline or session URL present.

**2. Branch rebased onto `main`; #201 scope removed** — confirmed. The diff now shows exactly 3 files: `.github/workflows/build.yml`, `scripts/ocr_screenshots.sh`, and `scripts/test_ocr_screenshots.sh`. No `ScreenshotNaming`, `ScreenshotTestRule`, or related files appear.

**3. Truncation test uses `-eq 80`** — confirmed. The comparison on `PREVIEW_LEN` is now exact equality (`-eq 80`). Importantly, this also closes the degenerate-case gap I noted: if the grep found no X's, `PREVIEW_LEN` would be 0, and `0 -eq 80` fails, so the test is now correctly self-auditing in both directions.

### Full review pass

`ocr_screenshots.sh` is correct:
- `set -euo pipefail` with `|| true` on the tesseract call is the right combination for per-file non-fatal failure.
- `shopt -s nullglob` means the array is empty (not a literal glob string) when no PNGs exist; the empty-array guard correctly exits 0.
- `GITHUB_STEP_SUMMARY` is guarded with `[[ -n "${GITHUB_STEP_SUMMARY:-}" ]]` so the script works cleanly outside CI.
- The `tr -d '\n' | tr -s ' ' | cut -c1-80` preview pipeline is sensible.

`build.yml` additions are correct:
- The OCR step's `if: always() && steps.pull_e2e_screenshots.outcome == 'success'` condition is the right idiom for a conditional step inside an `always()`-scoped job.
- Placement before the artifact upload step is correct; `.ocr.txt` companions will be included in the artifact.
- `sudo apt-get install -y tesseract-ocr > /dev/null 2>&1` swallows stderr, so an apt failure would produce no diagnostic output — a minor operational nuisance, not a correctness bug.

`test_ocr_screenshots.sh` is clean: mock tesseract approach is sound, `mktemp -d` / `trap` cleanup is correct, `set +e` / `set -e` brackets for exit-code assertions are correct.

### Verdict

**Approved.** All three requested changes are correctly implemented. No new issues introduced.

(Note to Orchestrator: GitHub does not permit approving one's own PR via the review API, so this approval is recorded as a comment.)

---

PR #221, opened 2026-05-25
Try to quiet agents and simplify instructions

aunger:


aunger (commented):
Reviewer here. Verdict: **Request changes** (informal — same-account, so leaving this as a comment per pr_participation.md). Only item 2 below needs action; everything else is an improvement.

**1. Reorderings — good.**
- Moving the "create a dedicated per-issue branch" bullet up under "Assigning a Programmer" so it precedes "pass the branch name to the subagent" reads better: the branch now exists conceptually before you're told to pass its name. No change in meaning.
- Reordering the Delegation rules (parallel / one-branch / separate-subagents) is harmless.

**2. The new "prefer resuming" bullet duplicates the existing exit bullet — please reconcile.**
The new bullet:
> For follow-up work ... **prefer resuming the existing Author or Reviewer over spawning a replacement**. Use SendMessage with the original agent's ID to resume it with its full prior context intact, no reconstruction needed.

is nearly word-for-word the same guidance as the already-present bullet a few lines down:
> **If an agent has exited without completing its task**, prefer resuming it over spawning a replacement. Use SendMessage with the original agent's ID to resume it with its full prior context intact. Spawn a replacement if the original agent's ID is unavailable or resumption fails.

By this document's own rule — "Agent completion and exit are the same event ... there is no idle/suspended state between 'completed' and 'exited'" — the "follow-up work after a completed turn" case and the "exited without completing its task" case are the *same* operational event. So we now have two bullets prescribing "resume via SendMessage with the original ID; full context intact" for what the doc insists is one transition.

Two concrete problems with leaving both:
- It's redundant, which works against the PR's stated goal of simplifying.
- It's mildly contradictory in completeness: the new bullet omits the "Spawn a replacement if the original agent's ID is unavailable or resumption fails" fallback and the two caveats (time-window, ID-availability) that the existing bullet carries. A reader who follows only the new bullet has no fallback instruction.

Recommendation: merge these into a single rule (or make the new bullet a one-line cross-reference to the existing one with its caveats), rather than maintaining two near-identical sources of truth. The "follow-up work / subsequent rounds of edits or reviews" framing is a genuinely useful addition — keep that phrasing, but fold it into the single canonical bullet.

**3. Conditionalising "dispatch in parallel" to "If requested by the user" — good and consistent.**
Demoting parallel dispatch from an unconditional top rule to user-gated matches the conservative posture elsewhere in the file (e.g. "Replace subagents, reluctantly and when necessary"). No contradiction introduced, and the "must each have their own branch and worktree" requirement is preserved.

**4. Language simplifications — intent preserved.**
Dropping "quietly" in favour of "Continue waiting without updating the User" is actually clearer: it states the concrete behavior (don't notify the User) rather than a vague adverb, which aligns with the PR's "quiet agents" goal. The semicolon/dash swaps are stylistic and fine.

**5. `### Haiku agent constraints` heading change — correct.**
Promoting the bolded pseudo-heading to a real `###` heading is right; it parallels the sibling `### Monitor bash script` and `### Outcome vocabulary` headings in the same document and makes the section navigable.

Net: a solid simplification. The one blocker is the duplicated resume guidance in item 2 — please consolidate it.

---

aunger (commented):
Follow-up review — my blocking issue is resolved. Approving.

The consolidation works:

1. **Both scenarios covered.** The single bullet names "subsequent rounds of edits or reviews" (follow-up work) and "an agent has exited without completing its task" (incomplete task), so the prior duplication is gone — there's now exactly one resume-vs-spawn directive.
2. **Fallback retained.** The ID-unavailable fallback and context-reconstruction sources (PR, issue, prior comments) are preserved in the second sub-bullet.
3. **Dropped "time window" caveat — acceptable loss.** It explained *why* a prompt resume matters (sessions may be kept alive only briefly after exit). Dropping it removes rationale but not correctness: the ID-unavailability fallback still gives a working recovery path.

Two non-blocking nits, optional:
- The original triggered fallback on "ID unavailable **or resumption fails**." The new version only triggers on ID unavailability, so a resume that fails for another reason has no stated fallback. Consider re-adding "or resumption fails."
- The lead-in is grammatically uneven — "such as subsequent rounds of edits or reviews, **or an agent has exited**" mixes a noun phrase with a clause. "...or cases where an agent has exited..." would read cleaner.

Neither blocks. Good to merge.

---

PR #238, opened 2026-05-26
Fix auto-filed issue format: title, branch, PR link, failure message

### LINKED ISSUE #237: Auto filed issue modifications
## Title

Currently, the titles of issues filed from test failures look like this:

> [Test Failure] com.gb4pc.e2e.PixelCameraOverlayE2ETest.overlayAppearsWhenViewfinderOpens @ 260525-1737-4855c4b

Move the test class name into the brackets instead of “Test Failure” and remove the prefix, package/namespace, date, and SHA, so the above title would be like this:

> [PixelCameraOverlayE2ETest] overlayAppearsWhenViewfinderOpens

## Branch

In the body, the branch name isn’t available in the example I have seen. If this is a bug, repair it.

## PR

Add a link to the body of the PR that the CI run is connected to, if possible.

## Failure message

Make the failure message word wrap instead of scroll horizontally. If no better way exists, simply use the `inline` code markup—or even no markup at all—instead of the code block notation.

(Don’t make this change to the stack trace.)

aunger:
Fixes #237.

Addresses all four format problems in the CI issue-filer script:

**1. Title format** — `make_issue_title` now produces `[SimpleClassName] methodName` instead of `[Test Failure] ClassName.method @ timestamp-sha`. The package prefix is stripped from the class name (e.g. `com.gb4pc.e2e.PixelCameraOverlayE2ETest` → `PixelCameraOverlayE2ETest`) and the timestamp/SHA suffix is removed. The duplicate-search query is updated to match the new title format.

**2. Branch name missing** — `WORKFLOW_RUN_BRANCH` was absent from the `env:` block in the `file-issues` workflow step, so the branch always showed as `_unknown_`. Fixed by passing `github.head_ref || github.ref_name`.

**3. PR link** — Added `WORKFLOW_RUN_PR_URL` (set from `github.event.pull_request.html_url`) to the workflow step. The issue body now includes a `[PR]` link in the Links section when the variable is non-empty (omitted for push-to-main runs with no associated PR).

**4. Failure message formatting** — The failure message is now rendered as plain text instead of a fenced code block, so it word-wraps in GitHub's issue view. The stack trace section retains its fenced code block.

All changes are covered by tests; the full test suite passes (65 tests).

---

### DIFF
--- a/scripts/file_test_failure_issues.py
+++ b/scripts/file_test_failure_issues.py
@@ -145,6 +145,7 @@ def make_issue_body(
     github_repository: str,
     github_run_id: str,
     workflow_run_branch: str,
+    pr_url: str = "",
 ) -> str:
     """Build the Markdown body for a test-failure issue."""
     ts_str = timestamp.strftime("%Y-%m-%d %H:%M UTC")
@@ -172,6 +173,7 @@ def make_issue_body(
         ocr_section = f"\n### OCR text from screenshot\n\n```\n{ocr_text}\n```\n"
 
     branch_info = f"`{workflow_run_branch}`" if workflow_run_branch else "_unknown_"
+    pr_line = f"- [PR]({pr_url})\n" if pr_url else ""
 
     body = f"""\
 ## Test Failure
@@ -187,9 +189,7 @@ def make_issue_body(
 
 ### Failure message
 
-```
 {failure.failure_message}
-```
 
 ### Stack trace
 
@@ -201,7 +201,7 @@ def make_issue_body(
 
 - [CI run]({run_url})
 - [Test artifact: {failure.artifact_name}]({artifact_url})
-{ocr_section}
+{pr_line}{ocr_section}
 ---
 _Filed automatically by CI on failure of `{failure.class_name}.{failure.method_name}`._
 """
@@ -314,6 +314,7 @@ def process_failure(
     github_server_url: str,
     github_run_id: str,
     workflow_run_branch: str,
+    pr_url: str = "",
 ) -> None:
     """File or update a GitHub issue for a single test failure."""
     title = make_issue_title(failure.class_name, failure.method_name)
@@ -326,6 +327,7 @@ def process_failure(
         github_repository=repository,
         github_run_id=github_run_id,
         workflow_run_branch=workflow_run_branch,
+        pr_url=pr_url,
     )
 
     existing = find_existing_issue(token, repository, failure.class_name, failure.method_name)
@@ -425,6 +427,7 @@ def main(argv: list[str] | None = None) -> int:
     github_run_id = os.environ.get("GITHUB_RUN_ID", "")
     sha = os.environ.get("WORKFLOW_RUN_SHA", "")
     workflow_run_branch = os.environ.get("WORKFLOW_RUN_BRANCH", "")
+    pr_url = os.environ.get("WORKFLOW_RUN_PR_URL", "")
 
     if not token:
         print("Warning: GITHUB_TOKEN not set — skipping issue filing.", file=sys.stderr)
@@ -469,6 +472,7 @@ def main(argv: list[str] | None = None) -> int:
                     github_server_url=github_server_url,
                     github_run_id=github_run_id,
                     workflow_run_branch=workflow_run_branch,
+                    pr_url=pr_url,
                 )
             except Exception as exc:  # noqa: BLE001
                 print(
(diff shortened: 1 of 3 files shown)


aunger:
**Review — Approved**

All four requirements from #237 are correctly addressed.

**Title format** — `make_issue_title` drops its `timestamp` and `sha` parameters, strips the package prefix with `.split(".")[-1]`, and produces `[SimpleClassName] methodName`. The duplicate-search query in `find_existing_issue` is updated to match. Tests are replaced with ones that validate the new format and explicitly assert the absence of `@`, timestamps, and the "Test Failure" prefix.

**Branch name** — `WORKFLOW_RUN_BRANCH: ${{ github.head_ref || github.ref_name }}` is added to the workflow `env:` block. The fallback to `ref_name` handles push-to-main runs where `head_ref` is empty.

**PR link** — `WORKFLOW_RUN_PR_URL` is added to the workflow env, read in `main()`, and passed down through `process_failure` and `make_issue_body` (defaulting to `""`). The body template emits `- [PR]({pr_url})\n` only when non-empty, so push-to-main runs produce no spurious PR line. Tests cover both the present and absent cases.

**Failure message** — The fenced code block around `failure_message` is removed; the text renders as plain Markdown and will word-wrap. The stack trace fenced block is untouched. Tests assert both properties explicitly.

No regressions observed. This PR is approved.

---

scripts/file_test_failure_issues.py
```diff
@@ -143,6 +145,7 @@ def make_issue_body(
     github_repository: str,
     github_run_id: str,
     workflow_run_branch: str,
+    pr_url: str = "",
```
aunger: Follow existing code style.

scripts/file_test_failure_issues.py
```diff
@@ -170,6 +173,7 @@ def make_issue_body(
         ocr_section = f"\n### OCR text from screenshot\n\n```\n{ocr_text}\n```\n"
 
     branch_info = f"`{workflow_run_branch}`" if workflow_run_branch else "_unknown_"
+    pr_line = f"- [PR]({pr_url})\n" if pr_url else ""
```
aunger: Follow existing pattern, use _unknown_

scripts/file_test_failure_issues.py
```diff
@@ -235,7 +237,8 @@ def find_existing_issue(
         print("  Error: 'requests' library not available.", file=sys.stderr)
         return None
 
-    query = f'repo:{repository} is:issue is:open label:test-failure "{class_name}.{method_name}" in:title'
+    simple_class = class_name.split(".")[-1]
+    query = f'repo:{repository} is:issue is:open label:test-failure "[{simple_class}] {method_name}" in:title'
```
aunger: Don't require is:open

scripts/file_test_failure_issues.py
aunger:
```Markdown
### Failed on {sha-7} @ {timestamp}

{body}
```

scripts/file_test_failure_issues.py
```diff
@@ -311,9 +314,10 @@ def process_failure(
     github_server_url: str,
     github_run_id: str,
     workflow_run_branch: str,
+    pr_url: str = "",
 ) -> None:
```
aunger, line 319: Follow established pattern

scripts/file_test_failure_issues.py
aunger: No back compatibility is needed.

... Do search this directory to be sure, though!

aunger:
**Review — Approved**

All five change requests are correctly addressed.

**Thread 1 (code style / back-compat):** `pr_url: str = ""` default removed; `pr_url: str` is now a required parameter in both `make_issue_body` and `process_failure`, matching the style of every other parameter in those signatures.

**Thread 2 (_unknown_ pattern):** `pr_info = f"[PR]({pr_url})" if pr_url else "_unknown_"` mirrors `branch_info` exactly.

**Thread 3 (is:open):** The search query no longer includes `is:open`. The title format in the query was also correctly updated to `"[{simple_class}] {method_name}"` to match the new issue title format.

**Thread 4 (recurrence header):** Comment header is now `### Failed on {short_sha} @ {ts_str}`, matching the owner's template. `short_sha` and `ts_str` are computed in `process_failure` before use, and the test asserts both the SHA fragment and the date are present.

**Thread 5 (established pattern):** `pr_url: str` added as a required positional parameter; no default. Matches all sibling parameters.

Tests cover the new behavior thoroughly: PR link present/absent, recurrence header format, failure message not in a fenced block, stack trace still fenced. No regressions observed.

**Decision: Approved.**

---

PR #239, opened 2026-05-26
Elide base64 blobs from E2E logcat output

### LINKED ISSUE #224: Elide base64 encoded logcat lines
The CI system sometimes dumps from logcat into the step output. This sometimes includes base-64-encoded binaries, which reduce the output’s usability. A truncated example is below. 

I suggest `s/(Bundle{bitmap_url=)[^}]+(})/\1\3/`, but this is just pseudo-code or pseudo-regex, and is likely wrong both in syntax and in tactic.

```
BUILD FAILED in 2m 2s
=== LOGCAT (last 300 lines) ===
05-25 14:08:16.131  1204  1204 D SearchTargetUtil: layoutType=extra_tall_icon_row resultType=131072 packageName=com.google.android.googlequicksearchbox extras=Bundle{bitmap_url=data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wCEAAkGBwgHBgkIBwgKCgkLDRYPDQwMDRsUFRAWIB0iIiAdHx8kKDQsJCYx(…continues a long time…)
```

aunger:
Fixes #224

## What changed

Added `scripts/filter_logcat.sh`, a one-liner `sed` filter that replaces any base64 payload (text after `;base64,`) with `[elided]`, keeping the surrounding context (key name, data URI type, closing delimiter) intact.

The two E2E test failure logcat dumps in `.github/workflows/build.yml` now pipe through `filter_logcat.sh` before `grep` and `tail`, so a line like:

```
extras=Bundle{bitmap_url=data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wCEA...}
```

becomes:

```
extras=Bundle{bitmap_url=data:image/jpeg;base64,[elided]}
```

## Design notes

- The minimum match length is 8 base64 characters, so short incidental tokens after `;base64,` are not elided.
- Multiple blobs on a single line are each independently elided.
- The script is self-contained (`sed` only, no external dependencies) and is invoked via `bash scripts/filter_logcat.sh` so no `chmod +x` step is needed in CI.

## Tests

`scripts/test_filter_logcat.sh` covers six scenarios: pass-through of normal lines, blob elision, context preservation, multiple blobs per line, short-token exclusion, and end-of-line blobs. All pass locally alongside the existing shell and Python test suites.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -310,7 +310,6 @@ jobs:
             echo "=== LOGCAT (last 300 lines) ==="
             "$ADB" logcat -d | \
               bash scripts/filter_logcat.sh | \
-              grep -E "AndroidRuntime|FATAL|Process crashed|com\.gb4pc|OverlayService|MockCamera|CameraService|E/\w" | \
               tail -300
             exit 1
           }
@@ -327,7 +326,6 @@ jobs:
             echo "=== LOGCAT (last 300 lines) ==="
             "$ADB" logcat -d | \
               bash scripts/filter_logcat.sh | \
-              grep -E "AndroidRuntime|FATAL|Process crashed|com\.gb4pc|OverlayService|MockCamera|CameraService|E/\w" | \
               tail -300
             exit 1
           }
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -309,6 +309,7 @@ jobs:
             -Pe2eClass=com.gb4pc.e2e.PixelCameraOverlayE2ETest || {
             echo "=== LOGCAT (last 300 lines) ==="
             "$ADB" logcat -d | \
+              bash scripts/filter_logcat.sh | \
               grep -E "AndroidRuntime|FATAL|Process crashed|com\.gb4pc|OverlayService|MockCamera|CameraService|E/\w" | \
               tail -300
             exit 1
@@ -325,6 +326,7 @@ jobs:
             -Pe2eClass=com.gb4pc.e2e.GalleryButtonVisualE2ETest || {
             echo "=== LOGCAT (last 300 lines) ==="
             "$ADB" logcat -d | \
+              bash scripts/filter_logcat.sh | \
               grep -E "AndroidRuntime|FATAL|Process crashed|com\.gb4pc|OverlayService|MockCamera|CameraService|E/\w" | \
               tail -300
             exit 1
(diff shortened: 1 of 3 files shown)


aunger:
**Review — Approved**

The implementation is correct and well-tested. All six test cases pass when run locally. Findings below.

**Regex delimiter safety.** The sed expression uses `/` as the s-command delimiter and also has `/` inside the character class `[A-Za-z0-9+/=]`. GNU sed (ubuntu-latest) handles this without ambiguity, and the test suite confirms it works end-to-end with realistic JPEG blobs that contain `/`. This would break on BSD sed (macOS), but CI runs on Ubuntu so it is not a defect here.

**Short-token threshold boundary.** Test (e) exercises the 7-character case (`abc1234`) but not the 8-character boundary (`abc12345`). The regex is `\{8,\}` so the boundary is correct, but there is no test that proves the exact edge. Minor gap; not a blocker.

**Test wiring.** `test_filter_logcat.sh` is named `test_*.sh` and the existing CI step already globs `scripts/test_*.sh`, so CI will pick it up automatically without any additional workflow change.

**Pipeline ordering.** The filter is inserted before `grep`, which is correct — grep operates on already-elided lines so its match patterns still fire normally and blobs don't produce spurious grep hits.

**Verdict: Approved.**

---

aunger:
Move the grep into filter_logcat.sh

aunger:
**Review — Second round, Approved**

The change request ("Move the grep into filter_logcat.sh") was addressed correctly.

**What was done:**
- `filter_logcat.sh` now has two pipeline stages: the base64 `sed` elision followed by the `grep -E` relevance filter.
- Both E2E logcat blocks in `build.yml` are reduced to a clean `bash scripts/filter_logcat.sh` invocation; `tail -300` remains after it in the caller pipeline.
- `test_filter_logcat.sh` adds tests (g) and (h) covering all seven keep-patterns and the drop case.

**`|| true` is correct.** Without it, a grep with no matches returns exit 1, which would kill the CI step on a quiet log. The placement is right — it is scoped to the grep, not the sed.

**Pipeline ordering is still correct.** Elision happens before filtering, so the grep patterns fire on already-clean lines with no risk of a base64 blob body accidentally matching (or not matching) a tag pattern.

**Test (g) boundary check.** The `E/\w` case tests `E/SomeTag`, which works because `\w` matches the first character of the tag. This is fine; the test is faithful to the regex.

No regressions. The implementation is clean and the change request is fully satisfied.

**Verdict: Approved.**

---

PR #247, opened 2026-05-26
Re-open closed issues when CI detects a recurring failure

### LINKED ISSUE #246: When CI appends to an existing issue, open it
If commenting on a matching issue that is closed, re-open it.

aunger:
Fixes #246

When the CI issue filer found a matching issue that was already closed, it would append a comment but leave the issue closed. This change re-opens the issue at the same time the comment is posted.

## How it works

- `find_existing_issue` now searches both open **and** closed issues (previously the GitHub Search API default returned only open issues). It returns a `(number, state)` tuple instead of just a number.
- A new `reopen_issue` function sends a `PATCH /repos/{repo}/issues/{number}` request with `{"state": "open"}`.
- `process_failure` unpacks the tuple and calls `reopen_issue` before posting the comment whenever `state == "closed"`.

## Tests

Seven new tests cover:
- `find_existing_issue` returning the state for open and closed issues
- `reopen_issue` success, API error, and unavailable-requests cases
- `process_failure` calls `reopen_issue` for closed issues and skips it for open ones

---

### DIFF
--- a/scripts/file_test_failure_issues.py
+++ b/scripts/file_test_failure_issues.py
@@ -229,16 +229,19 @@ def find_existing_issue(
     repository: str,
     class_name: str,
     method_name: str,
-) -> int | None:
-    """Search open issues for an existing failure report.
+) -> tuple[int, str] | None:
+    """Search open and closed issues for an existing failure report.
 
-    Returns the issue number if found, else None.
+    Returns a (issue_number, state) tuple if found, else None.
+    *state* is the string returned by the GitHub API, e.g. ``"open"`` or
+    ``"closed"``.
     """
     if requests is None:
         print("  Error: 'requests' library not available.", file=sys.stderr)
         return None
 
     simple_class = class_name.split(".")[-1]
+    # Omit is:open / is:closed so both states are searched.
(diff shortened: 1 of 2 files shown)


aunger:
**Review — Approved**

The implementation is correct and complete.

`find_existing_issue` now searches both open and closed issues by dropping the old `is:open` restriction, and the `(number, state)` return type threads cleanly through `process_failure`. The `PATCH /repos/{repo}/issues/{number}` call with `{"state": "open"}` is the right API shape.

One minor note on the PR description: the GitHub Search API does not default to open-only; the previous code had an explicit `is:open` token, which this change removes. That is the right fix either way, and the inline code comment ("Omit is:open / is:closed so both states are searched") accurately describes the intent.

The soft-failure ordering — reopen, then always comment regardless of reopen success — is consistent with the rest of the script's error-tolerant design, and preferable to silently dropping the comment if the reopen call fails.

All 150 tests pass locally. The seven new tests cover the meaningful paths: open issue (no reopen), closed issue (reopen called), reopen API failure, and requests-unavailable cases.

No issues found. **Approved.**

---

PR #251, opened 2026-05-26
Make requests import hard in file_test_failure_issues.py

### LINKED ISSUE #250: Make requests import hard in file_test_failure_issues.py
The `requests` import is wrapped in a `try/except ImportError` that sets `requests = None` on failure. Each API function then silently returns `None`/`False` if `requests is None`, so a missing dependency produces exit code 0 and no issues are filed.

Since `requests` is reliably available in CI, remove the `try/except` and make it a direct import. Remove the four `if requests is None` guards as well.

aunger:
Fixes #250

Replace the soft `try/except ImportError` block (which silently set `requests = None`) with a plain `import requests`. Remove all four `if requests is None` guard blocks from `find_existing_issue`, `create_issue`, `add_issue_comment`, and `reopen_issue`. If `requests` is not installed the script will now fail loudly at import time rather than silently no-oping at runtime.

The four `test_returns_*_when_requests_unavailable` tests that exercised the now-removed guards are also removed.

### DIFF
--- a/scripts/file_test_failure_issues.py
+++ b/scripts/file_test_failure_issues.py
@@ -28,7 +28,6 @@ import xml.etree.ElementTree as ET
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
-
 import requests
 
 
--- a/scripts/file_test_failure_issues.py
+++ b/scripts/file_test_failure_issues.py
@@ -29,10 +29,7 @@ from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
 
-try:
-    import requests
-except ImportError:
-    requests = None  # type: ignore[assignment]
+import requests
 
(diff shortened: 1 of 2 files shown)


aunger:
The code change is correct and complete. The soft import is gone, all four `if requests is None` guards are removed from `find_existing_issue`, `create_issue`, `add_issue_comment`, and `reopen_issue`, and the four matching `test_returns_*_when_requests_unavailable` tests are deleted. The file is clean — no leftover None-path artefacts, no new comments added. Net count of 1 addition and 40 deletions is exactly right.

One issue: the PR description contains a Claude attribution footer (`_Generated by [Claude Code](...)_`). The project instructions prohibit attribution bylines in PR description text. Please remove that line from the description before merging.

Approved — please remove the Claude attribution footer from the PR description before merging.

---

Issue #257, opened 2026-05-26
[#253 Stage 1] Build emits per-test result markers to the CI job log


aunger:
Part of #253 (CI Monitor should stream per-test failures/passes).

**Dependency:** none. Enables Stage 2.

## Goal

Have the build print a stable, greppable, one-line marker per test as each test finishes, so the
CI job log becomes the single source of truth for per-test results — independent of the overall
check conclusion.

## Marker format

One line per test, JSON payload, JSON-escaped:

```
##GB4PC_TEST## {"suite":"com.gb4pc.FooTest","name":"doesBar","outcome":"FAIL","ms":42,"msg":"expected:<1> but was:<2>","trace":"at com.gb4pc.FooTest.doesBar(FooTest.kt:31)\n…(truncated)"}
```

- `outcome` ∈ `PASS` / `FAIL` / `SKIP`.
- On `FAIL`, include a short `msg` and a truncated `trace` (first ~10 stack frames).
- `suite` + `name` is the dedup key the Monitor will use.
- Printed for **every** test regardless of outcome (so passes are available for Stage 4).

## Work

- **Unit tests:** add a Gradle test listener (`afterTest`) to the `testDebugUnitTest` task in
  `app/build.gradle.kts` (via `tasks.withType<Test>` / `testOptions`) that prints one marker per
  test to stdout, including failure message + truncated trace.
- **E2E tests:** the custom `connectedE2EAndroidTest` task already parses per-test
  `INSTRUMENTATION_STATUS` blocks (`app/build.gradle.kts`, ~lines 175–290). Emit the same marker
  per test from that loop.

## Verify

- `./gradlew testDebugUnitTest` locally and confirm markers appear with correct outcomes.
- Add a deliberately-failing test and confirm it emits a `FAIL` marker with a trace.

aunger:
Implemented by PR #262 (merged to `main` as `e097b99`, "Emit per-test result markers to CI job log"). Both sides emit spec-compliant `##GB4PC_TEST##` markers in `app/build.gradle.kts`:
- **Unit:** an `afterTest(TestDescriptor, TestResult)` listener on the `Test` tasks (build.gradle.kts:130–145) — emits PASS/FAIL/SKIP with `ms`, and `msg`+truncated `trace` on failure.
- **E2E:** the `connectedE2EAndroidTest` `INSTRUMENTATION_STATUS` parser (build.gradle.kts:~295–307) emits the same marker per test.

[...]

PR #261, opened 2026-05-26
Suppress in_progress heartbeat unless silent for >120s

### LINKED ISSUE #259: [#253 Stage 3] CI Monitor heartbeat: suppress in_progress unless silent >120s
Part of #253 (CI Monitor should stream per-test failures/passes).

**Dependency:** none (independent of Stages 1–2; can land anytime). Composes with Stage 2's
streamed output once that lands.

## Goal

Make `in_progress` a pure liveness signal for otherwise-quiet stretches.

## Work

In the Monitor bash script in `agents/dev_orchestration.md`, replace the current rule —
"emit an `in_progress` heartbeat every 5th iteration" — with:

> Suppress `in_progress` heartbeats unless **no output of any kind** has been emitted for **>120s**.

- Track a `last_output_ts`. Any emitted line resets it — including the streamed test results from
  Stage 2 (#258).
- Update the surrounding prose / outcome-vocabulary table in `dev_orchestration.md` that currently
  describes the every-5th-iteration heartbeat.

## Verify

> **Moved to #272.** This live-verification step had no recorded run before #259 was closed (the implementation is in `scripts/ci_monitor.sh`). It is carried into #272 — which rewrites this script in Python — to be run against the port. Original text, struck through:

~~Run the Monitor against a quiet PR and confirm `in_progress` appears only after >120s with no other output, and that any emitted line resets the timer.~~

aunger:
Fixes #259

## What changed

The CI Monitor's `in_progress` heartbeat was previously emitted on every 5th polling iteration, regardless of how much other output had already been produced. This made it a periodic noise signal rather than a true liveness indicator.

This PR replaces the `iter % 5` rule with a `last_output_ts` timestamp. The heartbeat is now suppressed unless **no output of any kind** has been emitted for **more than 120 seconds**. Any line the script emits (including error messages and "still computing" status lines) resets the timer, so the heartbeat is purely a signal that the monitor is alive but has nothing new to report.

## Changes in `agents/dev_orchestration.md`

- **Script**: Replace `iter=0` / `iter % 5` logic with `last_output_ts=$(date +%s)` and a `>120 s` silence check. Reset `last_output_ts` on every non-in_progress output line (including the "could not fetch SHA" error path and the "still computing" all_passed branch).
- **Prose** (Monitor loop instructions): Updated to say "suppresses these unless no other output has been emitted for over 120 seconds" instead of "emitting one in_progress heartbeat every few minutes".
- **Outcome-vocabulary table**: Updated the `in_progress` row to say "emitted only after >120 s of silence (no other output)".

---
_Generated by [Claude Code](https://claude.ai/code/session_01KbfU9NqAAUT3XSkkp4Pw1q)_

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -89,7 +89,7 @@ After the Reviewer exits and delivers its decision, the Orchestrator acts as fol
   if Reviewer gave approval:
     Orchestrator launches a Monitor tool call (run_in_background: true, timeout_ms: 1800000)
     Each stdout line arrives as a task-notification event
-    Act only on lines containing Clear, Blocked, or Infra. Relay in_progress lines to the user as brief status updates (the script throttles these down, emitting one in_progress heartbeat every few minutes).
+    Act only on lines containing Clear, Blocked, or Infra. Relay in_progress lines to the user as brief status updates (the script suppresses these unless no other output has been emitted for over 120 seconds).
     if Monitor emits a Blocked line  → goto newAuthor
     if Monitor emits an Infra line   → escalate to user; stop
     if Monitor emits a Clear line    → PR may be merged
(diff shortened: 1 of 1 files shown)


aunger:
**Review: Approved.**

The implementation is correct and well-scoped.

- `last_output_ts` is initialized before the loop and reset on every non-`in_progress` output path: the "could not fetch SHA" error line, the "still computing" branch, and the catch-all `else` branch. The terminal `Clear`/`Blocked`/`Infra` lines break immediately after emitting, so no reset is needed there.
- The silence check (`now - last_output_ts > 120`) is arithmetically sound; no unbound-variable risk.
- When the heartbeat fires, `last_output_ts` is set to `$now` (already captured) rather than a fresh `date +%s` call — a small correctness nicety that avoids drift.
- Prose and outcome-vocabulary table are updated consistently; no stale "every 5th" or "every few minutes" language remains.
- The diff is strictly confined to #259; no streaming/markers work from #257/#258/#260 is included.

---

Issue #260, opened 2026-05-26
[#253 Stage 4] CI Monitor: optional regex filters for which PASS/FAIL/SKIP tests are reported


aunger:
Part of #253 (CI Monitor should stream per-test failures/passes). **Nice-to-have.**

**Dependency:** Stages 1 (#257) and 2 (#258).

## Goal

Let the Orchestrator control *which* per-test markers the CI Monitor streams, so it can validate the current task early without flooding the stream — and so that a skipped task-relevant test (a false-validation trap: the code path was never exercised) is never silently swallowed.

The build already emits three distinct outcomes per test (`app/build.gradle.kts`: Gradle `SUCCESS→PASS`, `FAILURE→FAIL`, `SKIPPED→SKIP`). This stage adds independent filtering for each.

## API

Add **three independent flag pairs** to `scripts/ci_monitor.sh`. Each `--include-*` flag takes an **optional regex**; supplied without a pattern it includes *all* markers of that outcome. The regex is matched against `<TestClass>.<TestMethod>` (i.e. the marker's `<suite>.<name>`).

| Outcome | Default | Flags |
|---|---|---|
| FAIL | report **all** | `--include-fail [pattern]` (default) / `--no-include-fail` |
| SKIP | report **all** | `--include-skip [pattern]` (default) / `--no-include-skip` |
| PASS | report **none** | `--include-pass [pattern]` / `--no-include-pass` (default) |

**Default behavior (no flags): report _all failed tests_, _all skipped tests_, and _none of the passing tests_.**

Notes:
- The three outcomes keep their distinct markers. `--include-pass` never relabels a SKIP as PASS. (The existing skip≠pass handling in `summarize_test_results.py` is about *display* — not painting a skipped class with a green ✅ — not about hiding skips.)
- `--include-pass` with no pattern emits every PASS; with a pattern it emits only matching passes. Same for `--include-fail` / `--include-skip`.
- `--no-include-*` suppresses that outcome entirely, overriding the default.
- We deliberately reject a single whole-line regex: it would leak the rendered output grammar (the monitor prints `PR#N: FAIL [suite] name: …`, not the raw `##GB4PC_TEST##` markers), conflate *which outcome* with *which test*, and break on multi-line FAIL+trace entries.

## Work

[...]

Issue #258, opened 2026-05-26
[#253 Stage 2] CI Monitor surfaces per-test failures via step-conclusion deltas + per-step result artifacts


aunger:
Part of #253. Consumer side of **approach A** (see the design note on #253). **Depends on #264** (producer: per-group steps + per-step result artifacts).

> **Design changed.** This issue originally proposed scraping the in-progress job log for markers. That approach is abandoned: `GET /actions/jobs/{job_id}/logs` does not serve partial logs for an in-progress job, so markers only become readable in one clump at job end (confirmed empirically). PR #263 implemented the old approach and is superseded.

## Goal
Teach the CI Monitor (the bash poll loop in `agents/dev_orchestration.md`) to surface per-test failures **with message + stack trace** as each test group finishes — near-real-time at **step granularity**, independent of the overall check conclusion (so E2E failures surface even when the check stays green via `continue-on-error`).

## Work (`agents/dev_orchestration.md` Monitor script)
On each poll iteration, in addition to the existing check-runs polling:

1. **Live step signal.** `GET /actions/runs/{run_id}/jobs`; for the `build-and-test` job, read `steps[]`. When a step transitions to `completed`, emit one delta line with its name + `conclusion` (e.g. `step "unit: model" -> failure`), deduped by step name/number.
2. **Per-test detail.** `GET /actions/runs/{run_id}/artifacts`; when a new `testresults-` artifact appears (produced by #264), download it (follow `archive_download_url`, unzip), parse the `results/.ndjson` `##GB4PC_TEST##` markers, and emit **new `FAIL` entries only** with `msg` + truncated `trace`, deduped across iterations via a seen-set keyed on `suite#name`.
3. Both signals are emitted **independent of the overall check conclusion**. Any emitted line resets `last_output_ts` (the 120s silence timer from #259). The existing `Clear`/`Blocked`/`Infra`/`in_progress` terminal lines stay.

- `FAIL`/step-failure lines are **informational** — they do not by themselves start a new Author round; only a terminal `Blocked` does. Document this and update the outcome-vocabulary table.
- Remove the live-log-scraping path entirely; add a one-line note on *why* (logs aren't partial-readable mid-run).

## Out of scope
- Producer-side workflow/build changes (#264).
- Selective PASS streaming / task-relevance regex (#260).

## Verify
[...]

PR #262, opened 2026-05-26
Emit per-test result markers to CI job log

### LINKED ISSUE #257: [#253 Stage 1] Build emits per-test result markers to the CI job log

aunger:
Fixes #257.

Adds stable, greppable `##GB4PC_TEST##` markers — one per test, emitted as each test finishes — so the CI job log becomes the single source of truth for per-test results. The marker format is a JSON payload on a single line:

```
##GB4PC_TEST## {"suite":"com.gb4pc.FooTest","name":"doesBar","outcome":"FAIL","ms":42,"msg":"expected:<1> but was:<2>","trace":"java.lang.AssertionError: ...\n\tat ..."}
```

`outcome` ∈ `PASS` / `FAIL` / `SKIP`. On `FAIL`, `msg` carries the exception message and `trace` carries the first ≤10 stack frames. Both fields are empty strings on `PASS`/`SKIP`.

## Changes in `app/build.gradle.kts`

- **`jsonEscape()` and `buildTrace()` helpers** added above `android {}`, shared by both test paths.
- **Unit tests:** `testOptions.unitTests.all { addTestListener { afterTest { … } } }` registers a `TestListener` on the `testDebugUnitTest` task that prints one marker per test.
- **E2E tests:** the `INSTRUMENTATION_STATUS_CODE` branch in `connectedE2EAndroidTest` now emits the same marker immediately when each test's final status code arrives (`code != 1`), using the already-captured `curClass`/`curName`/`curStack` state. The marker format is identical to the unit-test path so a downstream parser can treat both uniformly.

Verified locally: `./gradlew testDebugUnitTest` prints `PASS` markers for all passing tests; a deliberately-failing test (added and then removed before this commit) produced a `FAIL` marker with the correct exception message and a 10-frame truncated trace.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KbfU9NqAAUT3XSkkp4Pw1q)_

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -24,6 +24,25 @@ val buildNumber: Int = when {
     else -> error("BUILD_NUMBER env var is set but not a valid integer: '$envBuildNumber'")
 }
 
+// ── Test-result marker helpers ────────────────────────────────────────────────
+// Shared by the unit-test listener (tasks.withType<Test>) and the E2E parser.
+// Produces the stable ##GB4PC_TEST## line format consumed by the CI Monitor.
+
+/** JSON-escapes a string value (no surrounding quotes). */
+fun jsonEscape(s: String): String = s
+    .replace("\\", "\\\\")
+    .replace("\"", "\\\"")
+    .replace("\n", "\\n")
+    .replace("\r", "\\r")
+    .replace("\t", "\\t")
+
+/** Returns the first ≤10 stack frames of [ex] as a single newline-separated string. */
+fun buildTrace(ex: Throwable): String {
+    val frames = ex.stackTrace.take(10).joinToString("\n") { "\tat $it" }
+    val header = ex.toString()
+    return if (frames.isEmpty()) header else "$header\n$frames"
+}
+
 android {
     namespace = "com.gb4pc"
     compileSdk = 35
@@ -103,6 +122,29 @@ android {
         unitTests {
             isIncludeAndroidResources = true
             isReturnDefaultValues = true
+            all { testTask ->
+                testTask.addTestListener(object : TestListener {
+                    override fun beforeSuite(suite: TestDescriptor) {}
+                    override fun afterSuite(suite: TestDescriptor, result: TestResult) {}
+                    override fun beforeTest(test: TestDescriptor) {}
+                    override fun afterTest(test: TestDescriptor, result: TestResult) {
+                        val outcome = when (result.resultType) {
+                            TestResult.ResultType.SUCCESS -> "PASS"
+                            TestResult.ResultType.FAILURE -> "FAIL"
+                            TestResult.ResultType.SKIPPED -> "SKIP"
+                        }
+                        val suite = test.className ?: ""
+                        val name = test.name
+                        val ms = result.endTime - result.startTime
+                        val ex = result.exception
+                        val msg = if (ex != null) jsonEscape(ex.message ?: ex.javaClass.name) else ""
+                        val trace = if (ex != null) jsonEscape(buildTrace(ex)) else ""
+                        println(
+                            """##GB4PC_TEST## {"suite":"${jsonEscape(suite)}","name":"${jsonEscape(name)}","outcome":"$outcome","ms":$ms,"msg":"$msg","trace":"$trace"}"""
+                        )
+                    }
+                })
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


aunger:
**Reviewer verdict: Request changes** — one blocking format-parity bug in the E2E `trace` field.

### Blocking: E2E trace uses a literal `\n`, breaking parity with the unit path

`app/build.gradle.kts` line 302 (E2E path):

```kotlin
val frames = stackStr.lines().take(10).joinToString("\\n")
```

`"\\n"` in Kotlin is a **literal backslash + n** (two characters), not a newline. The frames are joined by literal `\n` text, and that string is then passed through `jsonEscape`, which doubles the backslash to `\\`. The emitted JSON therefore contains `frame1\\nframe2`, which a downstream JSON parser decodes to `frame1\nframe2` — a literal backslash-n, **not** a real newline.

The unit-test path does the opposite (and correct) thing via `buildTrace`:

```kotlin
val frames = ex.stackTrace.take(10).joinToString("\n") { "\tat $it" }   // real newline
```

There, `jsonEscape` turns the real newline into `\n` in the JSON → decodes back to a real newline. So the two paths produce semantically different `trace` content for the same conceptual data. That violates the Stage-1 contract that both paths emit an identical, uniformly-parseable format (issue #257; the #258 parser must treat them the same).

Fix: join E2E frames with a real newline (`"\n"`) and let `jsonEscape` handle the escaping, matching the unit path. The change is confined to that one line.

### Non-blocking observations (your call)

- `jsonEscape` escapes `\ " \n \r \t` but not other C0 control chars (e.g. NUL, `\b`, `\f`, vertical tab). #257 mentions "control chars". A raw control char reaching the value would produce technically-invalid JSON. Very unlikely in stack traces, so not blocking — but if the #258 parser is strict, consider a `\uXXXX` fallback for remaining `< 0x20` chars. Could also be deferred to a follow-up.
- E2E markers emit `"ms":0` (no per-test timing available from the instrument output). Acceptable — the field still parses; just noting the asymmetry vs. the unit path's real `ms`.

### Verified good

- Emission point and sourcing for E2E are correct: marker fires when each test's final `INSTRUMENTATION_STATUS_CODE` arrives (`code != 1`), using captured `curClass`/`curName`/`curStack`; `curStack` is reset afterward so no leak into the next test.
- `outcome` covers PASS/FAIL/SKIP on both paths; markers print for every non-`1` outcome. Unit `when(result.resultType)` over SUCCESS/FAILURE/SKIPPED is exhaustive and compiles without an `else`.
- Unit-path `msg`/`trace` go through `jsonEscape`, so no raw newlines reach stdout — single-line guarantee holds. `jsonEscape` orders the backslash replacement first (correct).
- Sentinel `##GB4PC_TEST##` is distinctive and unlikely to collide with Gradle/instrumentation output.
- Scope is clean: no #258 Monitor-side parsing or #260 pass-streaming/selection logic leaked in.

Please fix line 302 and I'll approve.

---

aunger:
Addressed the line-302 finding: changed the E2E stack trace `joinToString("\\n")` to `joinToString("\n")` so the separator is a real newline character, letting `jsonEscape` produce the correct `\n` JSON escape — matching the unit-test path on line 41. One-line change, no other modifications. Build and unit tests pass locally (`./gradlew testDebugUnitTest` → BUILD SUCCESSFUL).

---

Issue #265, opened 2026-05-27
Six instrumented (androidTest) classes are never executed in CI


aunger:
## Problem
The CI workflow (`.github/workflows/build.yml`) only ever runs **two** of the eight instrumented `@Test` classes. The E2E job invokes the custom `connectedE2EAndroidTest` Gradle task twice, each filtered to a single class via `-Pe2eClass=`:

- `com.gb4pc.e2e.PixelCameraOverlayE2ETest` (build.yml:301)
- `com.gb4pc.e2e.GalleryButtonVisualE2ETest` (build.yml:317)

The remaining instrumented classes are **never invoked by CI**:

- `com.gb4pc.e2e.E2ETest` (appears to be a base/fixture in the `com.gb4pc.e2e` package)
- `MainActivityTest` (`ui`)
- `SetupActivityTest` (`ui`)
- `PickerActivityTest` (`ui.picker`)
- `ThemeTest` (`ui.theme`)
- `SessionTrackerIntegrationTest` (`viewer`)

The workflow never calls `connectedAndroidTest` / `connectedDebugAndroidTest` / `connectedCheck`. Note also that `app/build.gradle.kts:61` sets `testInstrumentationRunnerArguments["notPackage"] = "com.gb4pc.e2e"` for the *standard* connected-test config — i.e. the standard run is explicitly scoped to the non-e2e classes — but that standard run is never triggered in CI. So these `ui`/`viewer` instrumented tests are dormant.

## Why it matters
These tests exist and presumably encode real expectations, but they provide **zero signal** in CI. Regressions they would catch can merge undetected.

## Possible resolutions (to decide)
- Wire the non-e2e instrumented classes into the existing emulator job (they can run against the already-booted AVD, so no extra boot cost), e.g. a `connectedDebugAndroidTest` invocation alongside the two E2E steps; **or**
- Confirm they are intentionally local-only / obsolete and delete or clearly document them.

## Out of scope
This is independent of #253 / #264 (per-group CI steps + result artifacts), which only touches how the *already-running* tests are grouped and reported.

### Summary

[...]

PR #266, opened 2026-05-27
Upload per-step test marker artifacts for near-real-time CI signal

### LINKED ISSUE #264: [#253] Split tests into per-group CI steps and upload per-step result artifacts
Part of #253. Producer side of **approach A** (see the design note on #253). **Blocks the rewritten #258.**

> **Scope note (decided 2026-05-27):** This issue / PR #266 is **plumbing-only**. The `##GB4PC_TEST##` markers it tees on are emitted by **#257**, which is not yet implemented, so the uploaded `.ndjson` artifacts will be **empty until #257 lands**. End-to-end verification (non-empty artifacts, anchor/format agreement) is tracked in **#267**. The Acceptance below is therefore only fully met once #257 + #267 are done.

## Why this exists
We abandoned mid-run **log scraping**: the public REST endpoint `GET /actions/jobs/{job_id}/logs` does not serve partial logs for an in-progress job (it 404s until the job completes), and within a single big test step the `##GB4PC_TEST##` markers only flush to the log at step end — they arrive in one clump. Confirmed empirically.

Approach A instead gets near-real-time results from two public, pollable signals:
1. **Per-step conclusions** via `GET /actions/runs/{run_id}/jobs` (`jobs[].steps[]` carries live `status`/`conclusion`).
2. **Per-step result artifacts** carrying the #257 markers, uploaded right after each test step so they appear in the run&#39;s artifact list mid-job.

This issue is the **producer** side (`.github/workflows/build.yml` + `app/build.gradle.kts`). The Monitor/consumer side is the rewritten #258.

## Current state of the workflow (corrected)
An earlier draft of this issue assumed test steps needed splitting and that **E2E should stay one step to avoid repeated emulator boot**. Inspection shows that premise is stale:

- **E2E is already two separate, well-named steps** — `Run PixelCameraOverlayE2ETest` and `Run GalleryButtonVisualE2ETest` — each invoking the custom `connectedE2EAndroidTest` task filtered to one class via `-Pe2eClass=`, **both against the same already-booted emulator** (the AVD boots once in setup; the E2E steps use `--post-boot` and do not re-boot). So there is no repeated-boot cost, and E2E already yields per-class step conclusions.
- **Unit tests run in one step.** Of eight instrumented `@Test` classes, only the two E2E classes above run in CI at all (the dormant six are tracked separately in #265).

## Agreed grouping (do NOT re-group)
- **Unit: keep as a single step.**
- **E2E: keep the two existing per-class steps exactly as they are** — do not merge them, do not add more.

The remaining producer work is therefore just per-step marker capture + per-step artifact upload on the three existing test steps; no test re-grouping is needed.

## Work
- For each of the three existing test steps (1 unit + 2 E2E), tee its `##GB4PC_TEST##` markers (from #257) to a per-step `results/*.ndjson` (`results/unit.ndjson`, `results/e2e-overlay.ndjson`, `results/e2e-gallery.ndjson`). Teeing must not swallow the non-zero exit that drives the step `conclusion`, and must preserve existing behavior (E2E `continue-on-error: true` and the logcat-on-failure blocks).
- After each test step, add an `actions/upload-artifact` step with `if: always()` that uploads `testresults-<group>` (`testresults-unit`, `testresults-e2e-overlay`, `testresults-e2e-gallery`). `if: always()` so results upload even when the group has failures / E2E is `continue-on-error`. Combine with the existing `needs_full_build` gating so uploads only happen when the tests actually ran.
- No new token scope required — artifact upload works under the existing top-level `permissions: contents: read`.
- Keep overhead low: coarse grouping only, rely on warm Gradle daemon + config cache; do not shard per-test-class.

## Out of scope
- Marker emission itself (#257) — this PR only consumes markers, it does not produce them.
- End-to-end verification that artifacts are non-empty / anchor matches emitted format (#267).
- Monitor-side parsing of steps/artifacts (#258).
- Selective PASS streaming / task-relevance (#260).
- Any `checks: write` / check-run publishing (that was approach B, not chosen).
- Wiring the dormant non-e2e instrumented classes into CI (#265).

## Acceptance (fully met only once #257 + #267 land)
- On a run containing a deliberate failure, a `testresults-<group>` artifact appears in `GET /actions/runs/{run_id}/artifacts` **after that group&#39;s step completes** (i.e. mid-job, not only at job end), and it contains the failing test&#39;s marker with `msg` + truncated `trace`.
- `jobs[].steps[]` shows each test group&#39;s `conclusion` (success/failure) as it finishes.
- Overhead added by the extra steps/uploads is modest (target: well under a minute total).

aunger:
Fixes #264.

## Scope: plumbing-only

This PR is **plumbing-only** by decision. The `##GB4PC_TEST##` markers it tees on are emitted by **#257**, which is **not yet implemented**, so the uploaded `.ndjson` artifacts will be **empty until #257 lands**. Merging this first is intentional: it puts the producer wiring (tee + per-step uploads) in place so that once #257 emits markers the artifacts immediately carry content. End-to-end verification — non-empty artifacts, and confirming the `^##GB4PC_TEST##` grep anchor matches #257's emitted format (watch for Gradle log line-prefixing) — is tracked in **#267**.

## What this does

Each of the three test steps now tees its `##GB4PC_TEST##` markers to a per-step ndjson file under `results/`, using `set -o pipefail` + `tee >(grep '^##GB4PC_TEST##' > results/<group>.ndjson)` so the Gradle exit code still drives the step conclusion:

- `results/unit.ndjson` — **Build and run unit tests** (single step, unchanged grouping)
- `results/e2e-overlay.ndjson` — **Run PixelCameraOverlayE2ETest**
- `results/e2e-gallery.ndjson` — **Run GalleryButtonVisualE2ETest**

Immediately after each test step, an `actions/upload-artifact` step with `if: always()` uploads a distinctly-named artifact (`testresults-unit`, `testresults-e2e-overlay`, `testresults-e2e-gallery`) so artifacts appear in the run's artifact list as soon as that step finishes, mid-job. Each upload is also gated on `steps.diff_check.outputs.needs_full_build == 'true'` so uploads only occur when tests actually ran.

The existing `continue-on-error: true` on both E2E steps and all logcat-on-failure blocks are preserved unchanged. No new token scope is required; artifact upload works under the existing `permissions: contents: read`.

## Correction to the issue's premise

The issue's original text said "Keep E2E as a single step initially (avoid repeated emulator boot)." That premise was already false before this PR. The workflow already had two separate, well-named E2E steps — `Run PixelCameraOverlayE2ETest` and `Run GalleryButtonVisualE2ETest` — each running `connectedE2EAndroidTest` filtered to one class via `-Pe2eClass=`, both against the same already-booted emulator (AVD booted once in an earlier setup step; the E2E steps use `--post-boot` and do not re-boot). There is no repeated-boot cost and E2E already provides per-class step conclusions. (#264 has been updated to reflect this.)

Per the repo owner's decision:
- **Unit tests: kept as a single step** (not split further).
- **E2E: the two existing per-class steps left exactly as they are.** No merging, no additions.

## Files changed

- `.github/workflows/build.yml` — 39 insertions, 6 deletions: added `mkdir -p results` + tee-to-ndjson in the unit test step; added three `Upload *` artifact steps (one after each test step, `if: always() && needs_full_build == 'true'`).

## Related issues

- Depends on #257 (marker emission) to make artifacts non-empty.
- #267 — end-to-end verification once #257 + this land.
- #265 — dormant instrumented classes (independent).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,10 +123,13 @@ jobs:
         if: steps.diff_check.outputs.needs_full_build == 'true'
         env:
           BUILD_NUMBER: ${{ github.run_number }}
-        run: >
-          ./gradlew assembleDebug assembleDebugAndroidTest :e2e-mock-camera:assembleDebug
-          :e2e-mock-gallery:assembleDebug
-          testDebugUnitTest -PversionName=dev.${{ github.run_number }}
+        run: |
+          mkdir -p results
+          set -o pipefail
+          ./gradlew assembleDebug assembleDebugAndroidTest :e2e-mock-camera:assembleDebug \
+            :e2e-mock-gallery:assembleDebug \
+            testDebugUnitTest -PversionName=dev.${{ github.run_number }} \
+            | tee >(grep '^##GB4PC_TEST##' > results/unit.ndjson)
 
       - name: Upload unit test results
         if: always() && steps.diff_check.outputs.needs_full_build == 'true'
@@ -136,6 +139,14 @@ jobs:
           path: |
             app/build/test-results/testDebugUnitTest/
             app/build/reports/tests/
+          retention-days: 14
+
+      - name: Upload unit test markers
+        if: always() && steps.diff_check.outputs.needs_full_build == 'true'
+        uses: actions/upload-artifact@v7
+        with:
+          name: testresults-unit
+          path: results/unit.ndjson
           retention-days: 14
 
       - name: Start emulator
@@ -305,14 +316,25 @@ jobs:
         continue-on-error: true
         run: |
           ADB="$ANDROID_HOME/platform-tools/adb"
+          mkdir -p results
+          set -o pipefail
           ./gradlew connectedE2EAndroidTest \
-            -Pe2eClass=com.gb4pc.e2e.PixelCameraOverlayE2ETest || {
+            -Pe2eClass=com.gb4pc.e2e.PixelCameraOverlayE2ETest \
+            | tee >(grep '^##GB4PC_TEST##' > results/e2e-overlay.ndjson) || {
             echo "=== LOGCAT (last 300 lines) ==="
             "$ADB" logcat -d | \
               bash scripts/filter_logcat.sh | \
               tail -300
             exit 1
           }
+
+      - name: Upload overlay E2E markers
+        if: always() && steps.diff_check.outputs.needs_full_build == 'true'
+        uses: actions/upload-artifact@v7
+        with:
+          name: testresults-e2e-overlay
+          path: results/e2e-overlay.ndjson
+          retention-days: 14
 
       - name: Run GalleryButtonVisualE2ETest
         # Failures auto-tracked as issues. continue-on-error lets subsequent steps run even on failure.
@@ -321,14 +343,25 @@ jobs:
         continue-on-error: true
         run: |
           ADB="$ANDROID_HOME/platform-tools/adb"
+          mkdir -p results
+          set -o pipefail
           ./gradlew connectedE2EAndroidTest \
-            -Pe2eClass=com.gb4pc.e2e.GalleryButtonVisualE2ETest || {
+            -Pe2eClass=com.gb4pc.e2e.GalleryButtonVisualE2ETest \
+            | tee >(grep '^##GB4PC_TEST##' > results/e2e-gallery.ndjson) || {
             echo "=== LOGCAT (last 300 lines) ==="
             "$ADB" logcat -d | \
               bash scripts/filter_logcat.sh | \
               tail -300
             exit 1
           }
+
+      - name: Upload gallery E2E markers
+        if: always() && steps.diff_check.outputs.needs_full_build == 'true'
+        uses: actions/upload-artifact@v7
+        with:
+          name: testresults-e2e-gallery
+          path: results/e2e-gallery.ndjson
+          retention-days: 14
 
       - name: Pull and upload E2E screenshots on failure
         id: pull_e2e_screenshots


aunger:
## Review of PR #266

### Verdict: Request changes

---

### Blocking issue — markers do not exist yet

The entire purpose of this PR is to capture `##GB4PC_TEST##` lines from each test step into per-step `.ndjson` files. However, `##GB4PC_TEST##` is not emitted anywhere in the codebase. A search across all `.kt`, `.kts`, `.java`, `.py`, and `.sh` files finds zero occurrences of `GB4PC_TEST` except in `build.yml` itself (introduced by this PR). Issue #257, which defines the marker format and requires adding an `afterTest` listener to `testDebugUnitTest` and marker emission to `connectedE2EAndroidTest`, is still open and unimplemented.

Consequence: every run will produce three empty `.ndjson` files and three artifact uploads containing nothing. The acceptance criterion — "a `testresults-<group>` artifact appears…and it contains the failing test's marker with `msg` + truncated `trace`" — cannot be met until #257 is implemented.

**Required fix:** Either implement #257's marker emission in `app/build.gradle.kts` as part of this PR, or state clearly in the PR description and issue that this PR is intentionally partial (capture plumbing only, markers to follow in a separate PR for #257) and update the issue's acceptance criterion accordingly. Given that #264 explicitly lists `app/build.gradle.kts` as in-scope and the acceptance criterion requires per-test markers with `msg` + `trace`, the former is the correct resolution.

---

### Non-blocking observations (no changes required)

**Exit-code correctness** — sound. With `pipefail` active (redundant since the GitHub Actions Linux default shell is already `bash --noprofile --norc -eo pipefail`, but harmless), the pipeline exit code equals Gradle's exit code regardless of the `tee >(grep...)` process substitution. Confirmed: a non-zero left-hand-side of the pipe propagates through `tee` with a process substitution on the right, and the `|| { ... ; exit 1; }` logcat block in the E2E steps still triggers correctly. `continue-on-error: true` is preserved on both E2E steps.

**`if: always()` gating** — correct. `always() && steps.diff_check.outputs.needs_full_build == 'true'` runs uploads even when the test step failed, but not when tests were skipped due to a docs-only PR.

**Artifact naming** — all three names (`testresults-unit`, `testresults-e2e-overlay`, `testresults-e2e-gallery`) are unique and do not collide with pre-existing artifacts.

**YAML** — well-formed. The switch from `run: >` to `run: |` in the unit-test step is correct.

**Marker anchor** — `grep '^##GB4PC_TEST##'` relies on markers appearing at the start of a line in Gradle's stdout. This is consistent with how `print(output)` works in `connectedE2EAndroidTest` and how a Gradle `afterTest` listener would call `println`. No issue contingent on how #257 ultimately emits the markers.

**Permissions** — unchanged at `contents: read`. No `checks: write` introduced.

---

The sole required change is resolution of the missing marker emission (#257). The workflow plumbing itself is correct.

---

PR #273, opened 2026-05-27
Eliminate duplicate parser code: tests call ci_monitor.sh subcommands

aunger:
This is the conditional-approval fix from PR #271's review that missed the squash-merge into main.

## What happened

PR #271 (which delivered issue #258) was squash-merged. A reviewer granted conditional approval contingent on removing the duplicated parser strings from `scripts/test_ci_monitor.sh` — that fix landed on the branch *after* the squash-merge and was therefore not included in `90daa3f`.

## What this PR does

Adds two hidden subcommands to `scripts/ci_monitor.sh`:

- `__parse-steps <seen-file>` — exposes the step-delta parser
- `__parse-fails <seen-fails-file>` — exposes the failure-line parser

`scripts/test_ci_monitor.sh` previously duplicated both Python parser snippets verbatim (as `STEP_PARSER_CODE` and `FAIL_PARSER_CODE` shell variables) and ran them directly. It now invokes the real parsers via `bash ci_monitor.sh __parse-steps` and `bash ci_monitor.sh __parse-fails`. This means any future change to the parsers in `ci_monitor.sh` is automatically exercised by the tests, with no copy to keep in sync.

No behaviour changes — all 19 tests pass unchanged.

Note: two separate `ci_monitor.sh` robustness gaps found while dogfooding are tracked in #272 and are intentionally out of scope here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KbfU9NqAAUT3XSkkp4Pw1q)_

Issue #272, opened 2026-05-27
Rewrite CI Monitor in Python, then harden (closed/merged termination, SHA backoff, throttling, bailout PID)


aunger:
## Background
`scripts/ci_monitor.sh` (the CI Monitor poll loop, added in #258 via PR #271, deduped in #273) is **bash that shells into embedded Python** for its two parsers. Dogfooding it live on PRs #271/#273 surfaced several robustness gaps, and the bash↔Python split makes them awkward to fix and test. So this issue is sequenced: **port to a single Python script first, then layer the robustness work on top.**

The parser logic / signals themselves work as designed (covered by `scripts/test_ci_monitor.sh`, 19 cases) — this is about loop-termination, rate-limit resilience, output hygiene, and a bailout switch.

---

## Step 1 — Port to a single Python script (do this FIRST)
Replace the bash loop + embedded `python3 -c` heredocs with one `scripts/ci_monitor.py`:
- HTTP via stdlib `urllib` (no pip dependency — keep CI/tests hermetic); JSON handled natively instead of piping through `python3 -c`.
- **Behavior-preserving:** identical stdout vocabulary (`Clear` / `Blocked` / `Infra` / `in_progress` / `step "..." -> ...` / `FAIL [suite] name: ...`), identical loop/exit semantics, identical >120s `in_progress` heartbeat suppression (#259).
- Tests `import` and call the parser functions directly (no more subprocess `__parse-steps`/`__parse-fails` shims, no copy-paste — this supersedes #273's mechanism). Port the 19 existing cases; keep them hermetic (mock HTTP, no network/token).
- **Dedup state (`seen_steps`/`seen_arts`/`seen_fails`) becomes in-memory `set()`s** instead of `mktemp` files. Those files exist today only because bash subprocesses share state through the filesystem; in a single Python process they vanish — which also removes the need for any cleanup trap (see Gap D).
- The Monitor invocation in `agents/dev_orchestration.md` changes from `bash scripts/ci_monitor.sh ` to `python3 scripts/ci_monitor.py `.

> **Do not implement any of Step 2 before this port lands.** (Per owner.)

---

## Step 2 — Robustness (only after the port)

[...]

PR #285, opened 2026-05-29
Port CI Monitor from bash to Python (Step 1 of #272)

aunger:
Addresses #272 (Step 1 only).

## What changed

Replaces the bash poll loop that shelled into embedded `python3 -c` heredocs with a single `scripts/ci_monitor.py`. This is the **port** stage of the sequenced issue; it is behavior-preserving and lands before any Step 2 robustness work.

- **HTTP via stdlib `urllib`** — no pip dependency, JSON parsed natively instead of piping through `python3 -c`. Tests stay hermetic.
- **Identical stdout vocabulary** — `Clear` / `Blocked` / `Infra` / `in_progress` / `step "..." -> ...` / `FAIL [suite] name: ...`, with the same loop/exit semantics and the same `>120s`-of-silence `in_progress` heartbeat suppression.
- **In-memory dedup sets** — `seen_steps` / `seen_arts` / `seen_fails` are now plain `set()`s. The `mktemp` files (and the `trap ... EXIT` cleanup) existed only so bash subprocesses could share state through the filesystem; in a single process they are unnecessary and are gone.
- **Tests call the parsers directly** — `scripts/test_ci_monitor.py` imports `ci_monitor` and invokes the parser functions, replacing the subprocess `__parse-steps` / `__parse-fails` shims (this supersedes #273's mechanism). All 19 existing cases are ported, plus a mocked-HTTP smoke test for the `urllib` request helper (no network, no token).
- **Docs updated** — the Monitor invocation in `agents/dev_orchestration.md` now reads `python3 scripts/ci_monitor.py <PR_NUMBER>` (and the surrounding prose follows).
- **Deleted** `scripts/test_ci_monitor.sh`.

## Verification

`python3 scripts/test_ci_monitor.py` reports `20 passed, 0 failed` (the 19 ported cases plus the HTTP-mock test). Both new files pass `py_compile`.

## Scope

**Step 1 (port) is complete.** Step 2 robustness work — closed/merged termination (Gap A), throttling the "still computing" line (Gap B), SHA-fetch retry/backoff and rate-limit handling (Gap C), and the bailout PID kill-switch (Gap D) — will follow in a separate PR per the issue's "do not implement any of Step 2 before this port lands" instruction. The bash `scripts/ci_monitor.sh` is intentionally left in place for now.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NoARHaSjrPUoDipXHhNnHd)_


aunger:
**Reviewer verdict: Changes requested** — one behavioral regression to fix before merging.

## What works

- All 20 tests pass (`python3 scripts/test_ci_monitor.py` — ran locally).
- Stdout vocabulary is identical: `Clear`, `Blocked`, `Infra`, `in_progress`, `step "..." -> ...`, `FAIL [suite] name: ...` — all prefixed with `PR#N:` exactly as in the bash version.
- 120 s heartbeat suppression: `last_output_ts` is reset in every output path (including `could not fetch SHA` and `still computing`) — matches bash behavior exactly.
- `parse_check_result` logic is a clean, faithful transcription of the inline `python3 -c` from the bash script.
- `_request` catches `urllib.error.HTTPError` via the `URLError` parent — 4xx/5xx are handled correctly.
- stdlib-only (`urllib` + `json` + `zipfile`), no pip dependency.
- `agents/dev_orchestration.md` prose updated correctly; the invocation line, the section heading, and the explanatory sentence all reflect the switch to Python.
- Scope is clean: no Step 2 work introduced.

## The regression

**`seen_arts` is marked before the download attempt; bash marks it only on success.**

In `parse_new_artifacts`, the artifact ID is added to `seen` (which is `seen_arts`) before the function returns, so it is consumed before the caller ever tries to download. If the zip download fails (`zip_bytes is None` or raises `BadZipFile`), the artifact is silently dropped and will never be retried on the next poll.

The bash script does the opposite: it lists candidate IDs without touching `seen_arts`, and only appends `$aid` to the seen file after a successful `curl … && unzip`. A download blip causes a retry on the next poll.

The fix is to move the `seen_arts` mutation out of `parse_new_artifacts` and into the main loop, adding the ID only after `emit_block(parse_fails(...))` succeeds. `parse_new_artifacts` should filter by `seen` without mutating it, and the caller owns the write-back. This is a caller-side, single-location change.

---

The `result is None` case is a minor improvement over bash (Python skips silently; bash emits `PR#N: ` with an empty result string) — not worth reverting.

Fix the artifact dedup ordering and this is ready to merge.

---

aunger:
**Reviewer verdict: Approved.**

The fix is correct.

## What I checked

**The specific regression:** `seen_arts.add(aid)` is now at line 298, in the main poll loop body, after `emit_block(parse_fails(...))`. Both failure exits — `if not zip_bytes: continue` and `except (zipfile.BadZipFile, OSError): continue` — branch past it, so a transient download or unzip failure leaves the ID out of `seen_arts` and eligible for retry on the next poll. This exactly matches the bash `curl … && unzip` success-gate.

**`parse_new_artifacts` read-only contract:** The function no longer mutates `seen` at all. Its docstring explicitly documents this: "read-only here. The caller is responsible for adding an id to `seen` only after the artifact has been successfully downloaded and parsed." No mutation of the set inside the function.

**Tests:** All 20 pass locally — `python3 scripts/test_ci_monitor.py` reports `20 passed, 0 failed`.

**No new issues:** The rest of the implementation — stdout vocabulary, 120 s heartbeat suppression, `parse_check_result` logic, step and fail dedup, stdlib-only HTTP — is unchanged from what was reviewed in the prior round and found correct.

---

scripts/ci_monitor.py
```diff
@@ -0,0 +1,348 @@
...
+
+Outcome vocabulary (one terminal line ends the loop):
+    PR#N: Clear ...      All checks passed and mergeable_state is clean/unstable.
+    PR#N: Blocked ...    A check failed, or mergeable_state is behind/dirty.
+    PR#N: Infra ...      A CI infrastructure problem, or mergeable_state=blocked.
+    PR#N: in_progress    CI still running; emitted only after >120 s of silence.
+    PR#N: step "..." -> ...    A build-and-test step reached a conclusion (informational).
+    PR#N: FAIL [suite] name: ...   A per-test failure from a testresults artifact (informational).
+
+NOTE on error handling: like the bash predecessor, the poll loop must survive
```
aunger: Mentioning earlier versions here doesn’t help anyone

scripts/ci_monitor.py
```diff
@@ -0,0 +1,348 @@
...
+
+OWNER = "aunger"
+REPO = "gallery-button-for-pixel-camera"
+API_BASE = "https://api.github.com"
+
+TEST_MARKER = "##GB4PC_TEST##"
+
+
+# ── Parsers ───────────────────────────────────────────────────────────────────
+# These are the two real parsers exercised directly by test_ci_monitor.py.
```
aunger:
```suggestion
```

scripts/ci_monitor.py
```diff
@@ -0,0 +1,348 @@
...
+    Returns None on any HTTP/URL/JSON error so the poll loop can survive blips.
+    """
+    req = urllib.request.Request(url)
+    req.add_header("Authorization", "Bearer %s" % token)
+    req.add_header("Accept", "application/vnd.github+json")
+    try:
+        with urllib.request.urlopen(req) as resp:
+            data = resp.read()
+    except (urllib.error.URLError, OSError):
+        return None
```
aunger, line 210: Really? Not even a little curious how this fails? Write exceptions to stderr.

scripts/ci_monitor.py
```diff
@@ -0,0 +1,348 @@
+#!/usr/bin/env python3
+"""ci_monitor.py — Poll a PR's CI and stream a terminal outcome plus per-test signals.
+
+Invoked by the Orchestrator's Monitor tool call (see agents/dev_orchestration.md).
+Each stdout line is consumed as a task-notification event, so output is the
+interface: terminal outcome lines end the loop, while informational lines
+(in_progress heartbeat, per-step deltas, per-test FAILs) keep it alive.
+
+Usage:
+    python3 scripts/ci_monitor.py <PR_NUMBER>
```
aunger: Modify this to be a named parameter, `--pr <PR_NUMBER>`.

scripts/ci_monitor.py
```diff
@@ -179,24 +179,46 @@ def extract_ndjson_lines(zip_bytes):
...
+    if reason is not None:
+        return "%s: %s" % (type(e).__name__, reason)
+    return "%s: %s" % (type(e).__name__, e)
+
+
 def _request(url, token, raw=False):
     """Perform an authenticated GET. Returns parsed JSON (or raw bytes if raw).
 
     Returns None on any HTTP/URL/JSON error so the poll loop can survive blips.
+    The actual error is logged to stderr (not stdout, which is the event
```
aunger: Remove stderr explanation

scripts/ci_monitor.py
aunger: Done. The PR number is now a named `--pr <PR_NUMBER>` argument parsed with argparse; the usage docstring and the `dev_orchestration.md` invocation/prose are updated to match. (commit b7cd7af)

---

scripts/ci_monitor.py
aunger: Removed. Dropped the "like the bash predecessor" reference here, and also the other comments that mentioned the earlier version (the mktemp/trap note in `main`, the "mirroring the bash script" note on the artifact dedup, and the "former test_ci_monitor.sh" line in the test docstring). (commit b7cd7af)

---

scripts/ci_monitor.py
aunger: Deleted the "These are the two real parsers..." comment under the Parsers separator. (commit b7cd7af)

---

scripts/ci_monitor.py
aunger: Removed the stderr explanation from the `_request` docstring. The actual `sys.stderr.write` logging stays; only the verbose prose explaining it is gone. (commit b7cd7af)

---

aunger:
**Review of commit b7cd7af — Approved**

All four items are correctly addressed:

1. **Line 10 (argparse):** `--pr <PR_NUMBER>` named argument is in place. `argparse` is imported, `parser.add_argument("--pr", required=True, ...)` is used, and the old `argv[1]` positional access is gone.

2. **Line 26 (docstring):** "bash predecessor" and "earlier versions" references are removed from the NOTE block. The reworded note is accurate and clean.

3. **Line 50 (comment):** The "These are the two real parsers..." comment was deleted. The `# ── Parsers ──` section header remains without the now-stale annotation.

4. **Line 201 (stderr docstring):** The verbose three-line explanation of *why* errors go to stderr was removed. The actual stderr logging at lines 208 and 216 is untouched — behavior is unchanged.

`agents/dev_orchestration.md` is fully updated: both the inline command in the pseudocode block (line 101) and the prose + fenced example in the "Monitor script" section (lines 141–144) now read `python3 scripts/ci_monitor.py --pr <PR_NUMBER>`.

`test_ci_monitor.py` had one stale docstring sentence ("Ports the 19 cases from the former test_ci_monitor.sh") removed; the test suite itself is intact and all 20 tests pass.

**Approved — ready to proceed to CI check.**

---

aunger:
**Reviewer verdict: Changes requested** — three errors in the revised plan for tests (i) and (j).

## Test (i) — artifact dedup

The deque count (11 entries across two iterations), the zip bytes appearing once at position 6, and the non-inclusion of a zip call in iteration 2 are all correct: after iteration 1 adds `"42"` to `seen_arts`, `parse_new_artifacts` returns an empty list in iteration 2, so no zip call is made. The `PR#285:` prefix on every assertion string is correct. Count assertions checking for exactly 1 occurrence per signal line are correct.

The `Blocked` terminal in iteration 2 via `parse_check_result` returning `"Blocked"` is correct: the mock drives `result in ("Blocked", "Infra")` at line 353, which prints `PR#285: Blocked` and breaks. Steps and FAIL lines from iteration 1 are emitted before the terminal. No issues here.

## Test (j) — three interlocking errors

### Error 1: iteration numbers are off by one throughout

The plan says "First `in_progress` fires at end of iteration 5 (t=150, condition 150 - 0 > 120 ✓)". This is wrong.

Trace against the real code: `last_output_ts = time.time()` at startup consumes clock[0]=0. Iteration 1's body reads `now = time.time()` → 0; 0 - 0 = 0, not > 120. `time.sleep(30)` at the end of iteration 1 advances clock to 30. Continuing: iter 2 reads now=30, iter 3 now=60, iter 4 now=90, iter 5 now=120; `120 - 0 = 120`, which is **not** strictly greater than 120 (the condition is `> 120`, not `>= 120`). Sleep advances to 150. Iteration 6 reads `now=150`; `150 - 0 = 150 > 120` ✓. The first heartbeat fires **during iteration 6**, not iteration 5.

The plan's iteration numbering is off by one: what the plan calls "iteration 5" is actually iteration 6 in the code's execution order. Every subsequent iteration number in the plan (step line in "iteration 7", second heartbeat in "iteration 11", Blocked in "iteration 12") is similarly off by one — they occur in iterations 8, 12, and 13 respectively by the correct count. The deque must hold 13 × 5 = 65 entries, not 60.

### Error 2: startup `time.time()` call is not accounted for in `fake_time`

`main()` calls `last_output_ts = time.time()` before entering the loop (line 236). This is a `time.time()` call that will consume the first element of the `fake_time` side_effect list. The plan's tick schedule must include this extra call, or the clock reads in all subsequent iterations will be shifted by one value.

### Error 3: `emit_block`'s `time.time()` call in iteration 7 (step line) is not accounted for

`emit_block` calls `last_output_ts = time.time()` (line 252). When the step line fires in what the plan calls "iteration 7", this is an *additional* `time.time()` call within that iteration's body, consuming an extra slot from the `fake_time` side_effect. The plan's tick schedule does not account for this, which means subsequent `now = time.time()` reads in the `in_progress` branch will read the wrong clock values.

The fix: for every iteration where `emit_block` fires, add one extra `time.time()` entry to the `fake_time` side_effect sequence (returning the same clock value as the iteration's `now` read, since no sleep separates them).

### Corollary: second heartbeat never fires

With the corrected iteration numbering, the second `in_progress` check (condition `now - 180 > 120`) requires `now > 300`. After what the plan calls "iteration 7" (correct: iteration 8), the sleep advances clock to 8×30=240. Iter 9: now=240 (60 gap). Iter 10: now=270 (90). Iter 11: now=300 (120, not >). Iter 12: now=330 (150 > 120) — but the plan drives `Blocked` in iter 12. So the second heartbeat and the `Blocked` terminal cannot both fire: if `CHECK_BL` in slot 2 of iteration 12 causes `result="Blocked"`, the code takes the `elif result in ("Blocked", "Infra")` branch and never enters the `in_progress` block. The second `in_progress` is never emitted. The plan should either drop the second heartbeat assertion or place `Blocked` one iteration later (iteration 13).

## Summary

Test (i): correct and ready to implement.

Test (j): three corrections needed before implementation:
1. Shift all iteration numbers forward by one (13 iterations, 65 deque entries).
2. Add the startup `time.time()` call as the first entry in the `fake_time` side_effect.
3. Add one extra `fake_time` entry (same value as `now`) in each iteration where `emit_block` fires, to account for `emit_block`'s own `time.time()` call.
4. Choose whether the second heartbeat fires (requires Blocked to land in iteration 14+) or drop it; both are testable, but the current plan is internally inconsistent.

---

aunger (commented):
**Reviewer verdict: Changes requested** — one arithmetic error remains in the test (j) plan.

## What is correct

The iteration numbering and clock values in the corrected plan are verified:

- **Startup call at line 236:** `last_output_ts = time.time()` consumes clock=0 before the loop. ✓
- **First `in_progress` at iteration 6:** `now=150`, `150 - 0 = 150 > 120`. ✓ (Iteration 5 reads `now=120`; `120 - 0 = 120` is not strictly greater than 120, so it does not fire.)
- **Step line in iteration 7:** `emit_block` fires and calls `time.time()`, setting `last_output_ts = 180`. The `in_progress` check that follows reads `now = time.time()` = 180; `180 - 180 = 0`, no heartbeat. ✓
- **Second `in_progress` at iteration 12:** `now=330`, `330 - 180 = 150 > 120`. ✓ (Iteration 11 reads `now=300`; `300 - 180 = 120` is not strictly greater than 120.)
- **Blocked in iteration 13:** result drives the `elif result in ("Blocked", "Infra")` branch, prints and breaks. ✓

The account of `emit_block`'s extra `time.time()` call in iteration 7 consuming its own slot is correct.

## The remaining error: deque count is 66, not 67

The plan states "13 iterations × 5 + 1 emit_block = 67 total deque entries."

The arithmetic is wrong: 13 × 5 = 65, + 1 zip download = 66, not 67.

The 5 `_request` calls per iteration are: (1) `/pulls/{pr}` for SHA, (2) `/commits/{sha}/check-runs`, (3) `/actions/runs?head_sha=...`, (4) `/actions/runs/{id}/jobs`, (5) `/actions/runs/{id}/artifacts`. The zip download in iteration 7 is the +1. No iteration adds a sixth call: `result="in_progress"` has no extra `_request`, and `result="Blocked"` in iteration 13 also has no extra `_request`.

If the Author uses the stated count of 67 to size the `_request` side_effect deque, the test will either have one spurious unused entry (silently wrong) or fail with `StopIteration` if the mock is strict.

Correct the deque entry count to 66 before implementation.

---

PR #289, opened 2026-05-30
ci_monitor test groups (m) and (n): verification of #258 #259 #272

aunger:
Adds two automated test groups to `scripts/test_ci_monitor.py` that codify the live `main()` poll-loop verification checks carried over from #258 and #259, so they run hermetically (mocked HTTP, no network/token) instead of as manual checks.

- **Group (m) — staggered step-then-FAIL (from #258):** Exercises the full `main()` poll loop across separate polls. Confirms a `step "..." -> failure` delta is emitted once when the group's step completes (poll 1), and a `FAIL` line with message + indented trace is emitted once when that group's `testresults-*` artifact becomes available (poll 2) — both before the overall job concludes, deduped across polls, and with the 120s heartbeat correctly suppressed because each emitted signal resets the silence timer.

- **Group (n) — quiet passing PR (from #259):** Runs `main()` against a quiet PR and confirms `in_progress` appears only after silence crosses 120s, that exactly two adjacent heartbeats are emitted (no other PR# output between them), and that a single `Clear` terminal line is emitted last.

The full suite now runs at 69 passed, 0 failed.

Part of the verification work for #272.

---
_Generated by [Claude Code](https://claude.ai/code/session_019ifjHbMyNAK1htimKM3DVT)_

### DIFF
--- a/scripts/ci_monitor.py
+++ b/scripts/ci_monitor.py
@@ -45,6 +45,11 @@ API_BASE = "https://api.github.com"
 
 TEST_MARKER = "##GB4PC_TEST##"
 
+# Suppress repeated informational lines (in_progress heartbeat, "could not fetch
+# SHA", "still computing") until this many seconds have elapsed with no output of
+# any kind, so a quiet poll loop stays quiet. Every emitted line resets the timer.
+SILENCE_SECONDS = 120
+
 
 # ── Parsers ───────────────────────────────────────────────────────────────────
 
@@ -341,7 +346,7 @@ def main(argv):
             # Throttle the noise: only surface the failure after >120s of
             # silence, matching the in_progress heartbeat suppression.
             now = time.time()
-            if now - last_output_ts > 120:
+            if now - last_output_ts > SILENCE_SECONDS:
                 print("PR#%s: could not fetch SHA" % pr)
                 sys.stdout.flush()
                 last_output_ts = now
@@ -414,7 +419,7 @@ def main(argv):
 
         if result == "in_progress":
             now = time.time()
-            if now - last_output_ts > 120:
+            if now - last_output_ts > SILENCE_SECONDS:
                 print("PR#%s: in_progress" % pr)
                 sys.stdout.flush()
                 last_output_ts = now
@@ -441,7 +446,7 @@ def main(argv):
                 # Gap B — throttle "still computing" to >120s of silence, just
                 # like the in_progress heartbeat, so it does not print every poll.
                 now = time.time()
-                if now - last_output_ts > 120:
+                if now - last_output_ts > SILENCE_SECONDS:
                     print(
                         "PR#%s: all_passed mergeable_state=%s (still computing)"
                         % (pr, mergeable)
--- a/scripts/test_ci_monitor.py
+++ b/scripts/test_ci_monitor.py
@@ -17,6 +17,8 @@ Covers:
   (j) main(): in_progress heartbeat fires after >120s and resets on emission
   (k) Gap A: parse_pr_terminal() maps merged/closed/open + main() terminates on merged
   (l) Gap C: fetch_pr_with_retry() retry/backoff and _retry_after_seconds() header parsing
+  (m) main(): staggered step-then-FAIL emitted once each before terminal, heartbeat suppressed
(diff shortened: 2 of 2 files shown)


aunger:
**Review of groups (m) and (n)**

I ran the full test file against the current `ci_monitor.py` on main: 69/69 pass, 0 fail.

**Group (m) — staggered step-then-FAIL**

Request count is correct: poll 1 = 5 (no artifact yet), poll 2 = 6 (zip downloaded), poll 3 = 5 (artifact already in `seen_arts`, zip skipped). Total 16. The comment that "each signal resets the 120s timer" suppressing the heartbeat is accurate: `emit_block` calls `time.time()` and resets `last_output_ts`; with 30s-per-poll advances, the gap stays well under 120s after each emit. The `CHECK_BL_M` fixture uses `"conclusion": "failure"`, which maps to `Blocked` via `parse_check_result` and hits the terminal branch at line 451, continuing through runs/jobs/artifacts before breaking — consistent with the "no short-circuit" fact stated in the plan. All asserts (ordering, trace indent, deque drained, rc==0) are sound.

**Group (n) — quiet passing PR heartbeats**

Clock math is correct: 11 quiet in_progress polls at 30s each. Pre-loop `last_output_ts = time.time() = 0`. Strict `>` threshold means poll 5 (t=120) does not fire; poll 6 (t=150) does, resetting to 150. Polls 7–10 accumulate 30s each: at poll 10 the gap is exactly 120, which is not `>120`, so no heartbeat. Poll 11 (t=300, gap=150) fires the second heartbeat. Since polls 7–10 emit nothing (no steps, no FAIL, no heartbeat), the two heartbeats land on adjacent output lines. The 12th (final) poll supplies 6 deque entries including `MPR_CLEAN_N` for the `mpr_json` call. Total 11×5 + 6 = 61. The `pr_lines_n[0] == heartbeat_line_n` assertion is safe because the PID line does not start with `PR#272:`. All asserts hold.

**Minor observations** (no action required):

- `clear_lines_n` is defined but group (n) never checks whether `clear_lines_n` is the *last PR# line* specifically — it checks `lines_n[-1] == clear_lines_n[0]`, which is stronger and correct, since Clear is the final print before `return 0`.
- The `CHECK_PASS_N` fixture uses a single check run with `"conclusion": "success"`, which routes to `all_passed` in `parse_check_result`. That path then calls the extra `mpr_json` request. Correct.

**Verdict: Approved.**

---

aunger:
Reviewing the two new commits (refactor `SILENCE_SECONDS` + groups o/p/q). Tests confirmed at 86 passed, 0 failed.

---

**Commit 1 — `SILENCE_SECONDS` refactor**

Straightforward. All three `> 120` sites replaced with `> SILENCE_SECONDS` (`could not fetch SHA`, `in_progress`, `still computing`). Behavior-preserving: nothing else in the file changed. The `emit_block` docstring still says "reset the 120s timer" rather than naming the constant, but that's cosmetic and not a correctness concern.

---

**Commit 2 — Groups (o), (p), (q)**

**(o)** Subprocess / PID / SIGTERM: correct. Spawns the real script, reads the advisory line through a thread with a 10s guard, checks the printed PID matches the real PID, delivers SIGTERM, confirms `rc == -SIGTERM`. All five checks are sound.

**(p)** Real-artifact-shape fidelity: correct. The test builds its own zip at construction time using `_zf.writestr("unit.ndjson", ...)` — not the `make_zip_ndjson` helper (which still stores `testresults.ndjson` and is used only by the existing groups (i)/(m)). So the entry name actually exercises the real pipeline shape, and using the wrong name would have caused the deque to drain without a FAIL line, making the `FAIL line count != 1` check fail. Not a vacuous pass.

**(q)** Real-clock heartbeat: correct. `_REAL_SLEEP` is captured before `ci_monitor.time.sleep` is patched, avoiding the recursion. `time.time()` is deliberately left unpatched. `WINDOW_Q = 0.05s`, `SLEEP_Q = 0.25s` — a 5× margin that comfortably clears the window on every quiet poll without racing. The 5-poll schedule (quiet / quiet / step / quiet / Blocked) produces exactly 2 heartbeats with the step between them. The ordering check proves the real-time reset: if `emit_block` did not reset `last_output_ts`, poll 3 would also cross the window and a third heartbeat would be emitted, causing the `count != 2` check to fail.

---

Approved.

---

PR #290, opened 2026-05-30
Replace tail -300 with timestamp-based logcat capture

### LINKED ISSUE #268: Investigate: is `tail 300` really the best way?
## Problem

A couple of steps in the build workflow write out the last 300 filtered lines of logcat. Can this be done differently, not by an arbitrary number of lines, maybe including the entries since the step starts, either by timestamp or by a marker of any kind?

## Implementation

Replace the `adb logcat -d | ... | tail -300` failure-path block in each of the two E2E test steps in `.github/workflows/build.yml` with timestamp-based capture:

1. Immediately before the `./gradlew` invocation, record the device clock:
   `LOGCAT_SINCE=$("$ADB" shell date +'%m-%d %H:%M:%S.000')`
2. On failure, replace the logcat+tail block with:
   `"$ADB" logcat -d -T "$LOGCAT_SINCE" | bash scripts/filter_logcat.sh`
   (no tail — the timestamp filter bounds the output naturally)
3. Update the usage comment in `scripts/filter_logcat.sh` to drop the `tail -300` example and show the timestamp form.

**Key detail:** use `adb shell date` (device clock), not host `date`, to avoid clock skew between the CI runner and the emulator. `-T &lt;time&gt;` is supported on Android 4.1+ (API 16+); the emulator targets API 35.

aunger:
Fixes #268

## What changed

Both E2E test steps in `.github/workflows/build.yml` previously dumped the full device log buffer and truncated it to the last 300 filtered lines on failure. This could silently drop early failures and included pre-step noise from boot/setup.

The new approach records the device clock immediately before the Gradle invocation and passes that timestamp to `adb logcat -T` on failure, capturing all filtered lines from the start of the step with no arbitrary cutoff.

Key implementation detail: `adb shell date` (device clock) is used rather than host `date` to avoid clock skew between the CI runner and the emulator. `-T <time>` is supported on Android 4.1+ (API 16+); no compatibility concern with the API 35 target.

Also updates the usage comment in `scripts/filter_logcat.sh` to reflect the new recommended pattern.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BNa9d26UtzCM446jExQMFC)_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -318,13 +318,13 @@ jobs:
           ADB="$ANDROID_HOME/platform-tools/adb"
           mkdir -p results
           set -o pipefail
+          LOGCAT_SINCE=$("$ADB" shell date +'%m-%d %H:%M:%S.000')
           ./gradlew connectedE2EAndroidTest \
             -Pe2eClass=com.gb4pc.e2e.PixelCameraOverlayE2ETest \
             | tee >(grep '^##GB4PC_TEST##' > results/e2e-overlay.ndjson) || {
-            echo "=== LOGCAT (last 300 lines) ==="
-            "$ADB" logcat -d | \
-              bash scripts/filter_logcat.sh | \
-              tail -300
+            echo "=== LOGCAT (since test start) ==="
+            "$ADB" logcat -d -T "$LOGCAT_SINCE" | \
+              bash scripts/filter_logcat.sh
             exit 1
           }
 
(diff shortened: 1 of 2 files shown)


aunger:
**Review — Approved**

The approach is correct and the improvement is real. A few observations:

**Unbounded output**: `tail -300` is gone and nothing replaces it as a ceiling. If a verbose test failure produces a large number of lines that pass through `filter_logcat.sh`, the CI log could grow large. `filter_logcat.sh`'s tag filtering makes this unlikely to be a problem in practice, but it is a deliberate behavioral change the PR description does not mention.

**Empty `LOGCAT_SINCE` edge case**: If `adb shell date` fails (e.g., momentary adb connection issue before Gradle starts), `LOGCAT_SINCE` is empty and `adb logcat -d -T ""` behavior is implementation-defined — it may error or fall back to the full buffer. This is low severity because any adb failure at that point would likely surface through the test run itself, but there is no guard.

**`%S.000` millisecond hardcode**: Hardcoding `.000` for the sub-second component means the timestamp is always rounded down to the nearest second. This is safe (slightly over-inclusive rather than under-inclusive) and not a correctness problem.

**Usage comment in `filter_logcat.sh`**: The `&&`-chained single-line form in the comment is functional but differs in style from the two-line approach used in the workflow. Not a bug.

None of these rise to blocking issues. The core logic — capturing the device clock before Gradle, then filtering from that timestamp on failure — is sound and a genuine improvement over the `tail -300` approach.

**Decision: Approved.** (Ordinary comment used in place of GitHub approval because Author and Reviewer share the same account.)

---

PR #294, opened 2026-05-31
Add regex filters for FAIL/PASS/SKIP outcomes to CI monitor

### LINKED ISSUE #260: [#253 Stage 4] CI Monitor: optional regex filters for which PASS/FAIL/SKIP tests are reported

aunger:
## Summary

Fixes #260

Implements the three independent per-outcome filter flag pairs for `scripts/ci_monitor.py`, enabling the Orchestrator to control which per-test markers the CI monitor streams.

- `parse_fails()` now accepts an `outcome_filters` dict mapping each outcome (`FAIL`, `PASS`, `SKIP`) to an `(enabled, pattern)` tuple. When a pattern is supplied it is matched against the test `name` field via `re.search`. Default behavior is unchanged: all FAIL, all SKIP, no PASS.
- Six CLI flags parsed by `_parse_outcome_filters()` and threaded into the poll loop: `--include-fail [PATTERN]`, `--no-include-fail`, `--include-skip [PATTERN]`, `--no-include-skip`, `--include-pass [PATTERN]`, `--no-include-pass`.
- SKIP markers retain the `SKIP` label in output regardless of which `--include-pass` form is used.
- `agents/dev_orchestration.md` gains a "Per-test outcome filters" section, a task-relevance validation example (`--include-pass 'MyFeatureTest'`), and PASS/SKIP rows in the outcome vocabulary table.

## Test plan

- [x] `python3 scripts/test_ci_monitor.py` passes (128 tests, 0 failed)
- [x] New test group (r): `parse_fails()` default behavior, `--no-include-fail`, `--no-include-skip`, `--include-pass` with and without pattern, `--include-fail` pattern, `--include-skip` pattern, SKIP label preservation
- [x] New test group (s): `_parse_outcome_filters()` for all six flag combinations; `main()` end-to-end with `--include-pass ''`, no flags, `--no-include-fail`, `--no-include-skip`

---
_Generated by [Claude Code](https://claude.ai/code/session_014d7X7nZrANe4xhvXyFVA5U)_

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -161,6 +161,29 @@ python3 scripts/ci_monitor.py --pr <PR_NUMBER>
 
 `OWNER`/`REPO` default to this repo at the top of the script, and it reads `$GITHUB_TOKEN` from the environment. The script catches transient REST/parse failures per call so they cannot kill the resilient poll loop; the 30-minute escalation threshold is enforced by `timeout_ms: 1800000` on the Monitor call, not inside the script. Each stdout line is the interface (see the outcome vocabulary below): terminal lines (`Clear`/`Blocked`/`Infra`) end the loop, while informational lines keep it alive.
 
+### Per-test outcome filters
+
+By default the monitor reports **all FAIL markers**, **all SKIP markers**, and **no PASS markers**. The Orchestrator may supply independent filter flags to narrow or expand which per-test outcomes are streamed:
+
+| Flag | Effect |
+|---|---|
+| `--include-fail [PATTERN]` | Report FAIL markers (default); optionally restrict to those whose `name` matches PATTERN. |
+| `--no-include-fail` | Suppress all FAIL markers. |
+| `--include-skip [PATTERN]` | Report SKIP markers (default); optionally restrict to those whose `name` matches PATTERN. |
+| `--no-include-skip` | Suppress all SKIP markers. |
+| `--include-pass [PATTERN]` | Report PASS markers (not the default); optionally restrict to those whose `name` matches PATTERN. |
+| `--no-include-pass` | Suppress all PASS markers (explicit form of the default). |
+
+Each `--include-*` flag takes an **optional regex** matched against the marker's `name` field. Supplied without a pattern it includes *all* markers of that outcome. The three outcomes keep their distinct labels in output: `--include-pass` never relabels a SKIP as PASS.
+
+**Task-relevance validation.** To verify that a task-relevant test actually ran and passed — rather than being silently skipped — supply `--include-pass` with a regex matching the test(s) of interest:
+
+```bash
+python3 scripts/ci_monitor.py --pr <PR_NUMBER> --include-pass 'MyFeatureTest'
+```
+
+This emits a `PASS` line when the matching test passes and a `SKIP` line if it was skipped (which would be a false-validation trap: the code path was never exercised). With no pattern (`--include-pass ''`), every passing test is reported.
+
 ### Outcome vocabulary
 
 | Line emitted          | Meaning                                                                                               |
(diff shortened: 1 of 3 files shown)


aunger (commented):
**Review — APPROVE**

I reviewed the diff against issue #260 (Stage 4), ran the suite locally (`python3 scripts/test_ci_monitor.py` → 128 passed, 0 failed), and confirmed `parse_fails` has no callers outside `main()`. This is a clean, spec-faithful implementation.

**Spec conformance — verified:**
- Defaults are exactly right: all FAIL, all SKIP, no PASS (`_parse_outcome_filters` and `parse_fails`'s `outcome_filters=None` fallback agree).
- The three flag pairs behave per the issue table. `nargs="?"` with `const=""` / `default=None` correctly distinguishes the three states (not supplied → default; supplied with no pattern → match-all; supplied with pattern → filtered). `--no-include-*` correctly overrides the default.
- Regex is matched against the marker's `name` field via `re.search`, as specified. The whole-line regex the issue explicitly rejects is not introduced.
- SKIP is never relabeled as PASS: `outcome` is read verbatim from the marker and used both as the dedup-key suffix and the output label; `--include-pass` only gates PASS markers. Covered by an explicit test.

**Other correctness:**
- The dedup key change `suite#name` → `suite#name#outcome` is self-contained (`seen_fails` is created fresh in `main` and passed only here) and is in fact more correct now that multiple outcomes flow through.
- Commits are split by concern (code, then docs), per `agents/code_edit.md`.
- Docs in `dev_orchestration.md` add the filter section, the `--include-pass` task-relevance example, and the PASS/SKIP vocabulary rows the issue asked for.

**Test coverage:** Groups (r) and (s) cover default behavior, each `--include-*`/`--no-include-*`, pattern vs. no-pattern for all three outcomes, SKIP-label preservation, `_parse_outcome_filters` for all six flag forms, and `main()` end-to-end (`--include-pass ''`, no flags, `--no-include-fail`, `--no-include-skip`). Thorough.

**Minor, non-blocking (no action required):**
- Function-local `import re as _re` works but is slightly unusual; `re` could live at module top with the other imports.
- A malformed regex from the CLI would raise at `re.search`; the issue doesn't require validation and the poll loop is resilient to per-call exceptions, so this is acceptable.

No changes requested. CI was pending at review time; not blocking per our protocol.

Decision to Orchestrator: **Approved.**

---

aunger (commented):
Review of the latest commit (`0792cf5` — `--help`/`-h` and bad-invocation advice). The prior implementation was already approved; I scoped this pass to the new behavior only.

**Verified manually:**
- `--help` and `-h` both print the full module docstring (usage, args, filter flags, environment, outcome vocabulary) to stdout and exit 0. `-h` correctly routes through the overridden `print_help` since `add_help` defaults stay on.
- Missing `--pr` → `ci_monitor.py: error: the following arguments are required: --pr` then `Use --help for usage information.` on stderr, exit 2.
- Unrecognized flag (`--bogus`) → same error+advice pattern, exit 2.
- `python3 scripts/test_ci_monitor.py`: 128 passed, 0 failed.

The change does exactly what the User asked for, cleanly, in 16 lines.

**Non-blocking notes (your call, Author):**
1. No test coverage for the new paths. The rest of this file is meticulous — group (r)/(s) were added for the filter feature in this same PR — yet `--help`/`-h`/bad-invocation have no assertions. A couple of cheap checks (e.g. capture stdout on `main(["ci_monitor.py","--help"])` asserting the docstring is emitted and exit is 0; capture stderr on a missing `--pr` asserting the advice line and exit 2) would lock the behavior in. Given the change is small and I verified it by hand, I don't consider this blocking.
2. Minor fragility: `print_help` does `file.write(__doc__)` unguarded. Under `python -OO`, `__doc__` is `None` and `--help` would raise `TypeError`. Edge case, almost certainly out of scope — flagging only for awareness.

Approving. The new behavior is correct and matches the request; the notes above are optional follow-ups.

---

PR #298, opened 2026-05-31
Show instrumented tests as skipped when pre-flight fails

### LINKED ISSUE #255: CI bug: after pre-flight failure, all tests show green

**Setup**: The pre-flight check failed, so all tests were skipped.

**Actual**: Build-and-test run summary shows all tests green.

**Expected**: Run summary shows all test classes and/or test cases as skipped.

---

Scope: This issue **excludes** the pre-flight check failure itself. That error is tracked by issue 256.

aunger:
## Problem

When the CI pre-flight check fails, all instrumented E2E tests are skipped (never run), so no JUnit XML files are produced. The `Write test-result summary` step then shows **"No test results found"** for each instrumented suite — which looks clean/green in the run summary, hiding the fact that the entire test suite was skipped.

## Fix

Two changes:

**`scripts/summarize_test_results.py`** — adds a `--skipped` flag to the CLI.  
When `--skipped` is passed for a suite and no XML results are present, the summary now shows `_⏭ Skipped (tests did not run)_` instead of `_No test results found._`  
The flag is per-suite and is silently ignored if XML results are actually present (so a partial result set renders normally).

**`.github/workflows/build.yml`** — the `Write test-result summary` step now checks `steps.preflight.outcome`. When it is not `success` (i.e., the pre-flight failed or was skipped), `--skipped` is passed for the instrumented-test suites, making their skipped state visible in the job summary.

## Tests

13 new tests in `test_summarize_test_results.py` cover:
- `--skipped` flag parsing (per-suite, only applies to its own suite, defaults to false)
- `render_suite(..., skipped=True)` with no results shows skipped note
- `render_suite(..., skipped=True)` with results renders normally
- `main()` end-to-end with `--skipped` and missing directory

Fixes #255.

---
_Generated by [Claude Code](https://claude.ai/code/session_018iBAbk3JWQLsVKaAnRHRs1)_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -404,11 +404,18 @@ jobs:
       - name: Write test-result summary
         if: always() && steps.diff_check.outputs.needs_full_build == 'true'
         run: |
+          # When the pre-flight check did not succeed, E2E tests were skipped
+          # entirely. Pass --skipped for those suites so the summary shows them
+          # as "⏭ Skipped" rather than the misleading "No test results found."
+          SKIPPED_FLAG=""
+          if [[ "${{ steps.preflight.outcome }}" != "success" ]]; then
+            SKIPPED_FLAG="--skipped"
+          fi
           python3 scripts/summarize_test_results.py \
             app/build/test-results/testDebugUnitTest \
               --suite-label "Unit Tests" \
             app/build/outputs/androidTest-results/connected/debug \
-              --suite-label "Instrumented Tests"
+              --suite-label "Instrumented Tests" $SKIPPED_FLAG
 
   file-issues:
     needs: [build-and-test]
--- a/scripts/summarize_test_results.py
+++ b/scripts/summarize_test_results.py
@@ -8,9 +8,13 @@ back to stdout when the env var is not set).
 Usage:
     python3 scripts/summarize_test_results.py \\
         path/to/unit-results       --suite-label "Unit Tests" \\
-        path/to/e2e-results        --suite-label "Instrumented Tests"
+        path/to/e2e-results        --suite-label "Instrumented Tests" [--skipped]
 
 Each directory path must be immediately followed by --suite-label <name>.
+The optional --skipped flag (after --suite-label <name>) marks the suite as
+skipped: when no XML results are present the summary shows "⏭ Skipped" rather
+than "No test results found."  Use this when a pre-flight step failure caused
+tests to be skipped entirely.
 Exit code is always 0 (display only; failures are surfaced by earlier steps).
 """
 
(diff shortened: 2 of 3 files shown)


aunger:
## Review

### What this PR does

Adds a `--skipped` CLI flag to `summarize_test_results.py` and wires it into the CI workflow so that instrumented-test suites show "⏭ Skipped (tests did not run)" rather than the misleading "No test results found." when a pre-flight failure caused those tests to never run.

### Correctness

**Parser logic is sound.** The `parse_args` loop increments `i` by 3 to consume `<dir> --suite-label <name>`, then peeks at the next token for `--skipped`. Because an unmatched next token (e.g. the next suite's directory path) won't equal `"--skipped"`, multi-suite parsing is not disrupted. The positional logic is correct.

**`SKIPPED_FLAG` placement in the workflow is correct.** The variable is interpolated after `--suite-label "Instrumented Tests"`, which is exactly where the parser expects it. Unit Tests get no flag. An empty `$SKIPPED_FLAG` in bash expands to nothing, so the default path is unaffected.

**The `!= "success"` condition is the right choice** over `== "failure"`. It correctly covers the `skipped` outcome (e.g. when `needs_full_build` is false and the preflight step itself is skipped). No gap there.

**Silently ignoring `--skipped` when results are present** is the right design. A partial XML result set renders normally. Confirmed in the test and in `render_suite`.

**Step ordering in the workflow is consistent with the premise.** The preflight step has no `continue-on-error`, so a failure causes GitHub Actions to skip subsequent steps whose `if` condition is not wrapped in `always()`. The instrumented-test steps (`run_overlay_e2e`, `run_gallery_e2e`) use a plain `if: steps.diff_check.outputs.needs_full_build == 'true'` condition — no `always()` — so they are indeed skipped when preflight fails. The PR's premise is accurate.

### Tests

13 new tests cover: flag parsing per-suite, isolation between suites, default-false behavior, `render_suite` skipped vs. no-results vs. results-present, and `main()` end-to-end. Coverage is thorough. Existing tests are correctly updated for the 2→3-tuple change.

### Minor observations (non-blocking)

- The bash variable `$SKIPPED_FLAG` is unquoted in the `run:` block. This is safe here (the value is either empty or the single word `--skipped`), but it is a mild style inconsistency with how the rest of the script arguments are handled. Not a correctness issue.
- There is only one `SKIPPED_FLAG` variable in the workflow. If a second instrumented suite were ever added, the author would need to remember to extend the logic. A comment noting this assumption would reduce future confusion, but this is not a requirement given the current single-suite usage.

### Verdict

The logic is correct, the tests are solid, and the CI behavior matches the stated intent. Approving.

---

aunger:
Way too hacky

PR #299, opened 2026-05-31
Make preflight ANR dismissal survive adb command timeouts

### LINKED ISSUE #256: CI pre-flight check failure

You must read the preflight task logs for that run.

The pre-flight check failed.

**Not** the same symptoms as issue 214.

aunger:
Fixes #256

## Root cause

In Actions run 534, the `CI pre-flight — MockCameraActivity smoke check` step failed all 30 retry attempts. The screencaps never showed the green feed (dominant colors `#000000`, `#D7D6DC`, etc.) because an **"Application Not Responding" dialog from `com.android.systemui`** was overlaying `MockCameraActivity`:

```
mCurrentFocus=Window{... u0 Application Not Responding: com.android.systemui}
```

`_dismiss_anr_if_present` in `scripts/check_green_feed.py` *did* detect the dialog ("ANR dialog detected before screencap — sending KEYCODE_ENTER"), but the emulator was severely overloaded (`100% TOTAL: 42% user + 55% kernel`). Under that load the `adb shell input keyevent KEYCODE_ENTER` and `dumpsys window` commands repeatedly **timed out after 10s**:

```
Command '[... input keyevent KEYCODE_ENTER]' timed out after 10 s
```

The whole routine was wrapped in a single broad `try/except`, so the first `TimeoutExpired` aborted the entire dismiss loop. KEYCODE_ENTER was effectively sent zero or one time and never confirmed, so the dialog persisted across every attempt and each screencap captured the dialog instead of the green feed.

This is **not** the same as issue #214 (which was a *launcher* ANR plus an ineffective `KEYCODE_BACK`). Here the detection and the dismiss key were already correct; the new failure mode is per-command timeouts aborting the retry loop on a wedged emulator, and the ANR being SystemUI's rather than the launcher's.

## Fix

- Split the adb calls out of the monolithic `try/except`:
  - `_window_has_anr` runs `dumpsys window` and returns `True`/`False`, or `None` when the call times out (state undetermined). Treating a timed-out detect as "no ANR" would have skipped dismissal entirely.
  - `_send_keycode_enter` swallows its own timeout so a single failed send no longer aborts the loop.
- The dismiss loop now keeps retrying through transient timeouts, so once the `input` service recovers within the window the ENTER lands and the dialog is dismissed.
- Clarified docstrings/comments: the ANR dialog is not always the launcher's (SystemUI can ANR under load), and the stale `KEYCODE_BACK` comment was corrected to `KEYCODE_ENTER`.

## Tests

Added three tests to `scripts/test_check_green_feed.py`:
- `test_timeout_during_dismiss_does_not_abort_loop` — a `TimeoutExpired` on an ENTER send and on a confirm `dumpsys` no longer aborts; the loop keeps retrying and dismisses once the service recovers.
- `test_initial_detect_timeout_returns_without_sending_enter` — a timed-out initial detect returns cleanly without sending ENTER (undetermined state).
- `test_systemui_anr_is_dismissed` — a `com.android.systemui` ANR is detected and dismissed.

All 25 `test_check_green_feed.py` tests and all 19 `test_dismiss_anr.sh` tests pass locally.

---
_Generated by [Claude Code](https://claude.ai/code/session_018iBAbk3JWQLsVKaAnRHRs1)_

### DIFF
--- a/scripts/check_green_feed.py
+++ b/scripts/check_green_feed.py
@@ -237,7 +237,7 @@ def _dismiss_anr_if_present(adb: str) -> None:
             )
             return
         print(
-            f"[check_green_feed] {ts} ANR dialog still present after KEYCODE_ENTER #{attempt}.",
+            f"[check_green_feed] {ts} ANR dialog still present or undetermined after KEYCODE_ENTER #{attempt}.",
             file=sys.stderr,
         )
     print(
--- a/scripts/check_green_feed.py
+++ b/scripts/check_green_feed.py
@@ -153,27 +153,20 @@ def dominant_color(pixels_region: list[tuple[int, int, int]]) -> tuple[int, int,
 
 _ANR_DISMISS_MAX_RETRIES = 5
 # Intentionally separate from RETRY_DELAY_SECONDS: this poll interval governs how
-# long to wait between each KEYCODE_BACK send and the subsequent dumpsys window
+# long to wait between each KEYCODE_ENTER send and the subsequent dumpsys window
 # confirmation inside _dismiss_anr_if_present(), while RETRY_DELAY_SECONDS controls
 # the outer green-feed retry loop.  They happen to share the same value today but
 # may diverge if one needs tuning independently of the other.
 _ANR_DISMISS_POLL_SECONDS = 2
 
 
-def _dismiss_anr_if_present(adb: str) -> None:
-    """Check for the Pixel Launcher ANR dialog and dismiss it if present.
+def _window_has_anr(adb: str) -> bool | None:
+    """Return True/False if an ANR dialog is present, or None if undetermined.
 
-    This is a lightweight guard that runs before every screencap in the retry
-    loop.  dismiss_anr.sh exits as soon as Launcher CPU goes idle, but the ANR
-    dialog can appear during or after the `am start -W` call (which takes 3+ s),
-    after the watcher has already exited.  By checking here we ensure the retry
-    loop is self-defending against late-appearing ANR dialogs.
-
-    After sending KEYCODE_BACK, polls dumpsys window to confirm the dialog has
-    actually disappeared before returning, retrying up to _ANR_DISMISS_MAX_RETRIES
-    times.  This guards against cases where KEYCODE_BACK is dispatched but the
-    dialog does not dismiss immediately (e.g. slow emulator rendering, focus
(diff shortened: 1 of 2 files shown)


aunger:
**Reviewer: Approved — pending one log-message fix before merging.**

The fix is correct. The tri-state decomposition solves the root cause cleanly.

**Logic review**

- `_window_has_anr` returning `None` on timeout is the right design. The entry guard `is not True` correctly skips dismissal for both `False` and `None`, preventing speculative KEYCODE_ENTER sends when ANR state is unknown.
- The in-loop check `is False` is exactly right: only a confirmed-clear result exits the loop. A timed-out confirm (`None`) falls through to the retry iteration, which is the fix for Issue #256.
- `_send_keycode_enter` swallowing its own timeout is correct; it keeps the loop alive through the first problem case from CI run 534.

**One issue — misleading log message when the confirm times out**

In `_dismiss_anr_if_present`, the in-loop `print` after a `None` result from `_window_has_anr` says:

```
ANR dialog still present after KEYCODE_ENTER #{attempt}.
```

But when `_window_has_anr` returns `None` (timeout), the state is *undetermined*, not confirmed present. Someone reading CI logs will see "still present" when the real situation is "dumpsys timed out, retrying." This will make diagnosing future wedged-emulator failures harder.

The fix is a single string change: distinguish the two cases in the log. For example, log "undetermined (confirm timed out), retrying" when `_window_has_anr` returns `None`, and reserve "still present" for when it returns `True`. Alternatively, a single combined message like "ANR dialog still present or undetermined" is acceptable — the key requirement is that "still present" not be logged when the actual result was a timeout.

**Tests**

The three new tests are well-structured and cover the right scenarios. The main timeout-loop test correctly sequences 7 subprocess calls and verifies both call count and the dismissal log line. No gaps in the new tests; existing tests are unaffected.

---

**Decision: Approved, pending the log-message fix described above.** The behavior is correct as written; only the "still present" log text needs to accurately reflect the undetermined case before this merges.

---

aunger:
Done. Changed the in-loop print in `_dismiss_anr_if_present` from `"ANR dialog still present after KEYCODE_ENTER #{attempt}."` to `"ANR dialog still present or undetermined after KEYCODE_ENTER #{attempt}."` — single string edit, no other changes. All 25 existing tests pass.

---

PR #308, opened 2026-05-31
Fix adb shell date quoting so busybox receives a single argument

### LINKED ISSUE #307: E2E test regression
Fix the regression introduced by 290.
- #290

Now, the E2E tasks fail with this:

```
date: Max 1 argument (see "date --help")
Error: Process completed with exit code 1.
0s
```

aunger:
The format string '+%m-%d %H:%M:%S.000' contains a space. Passing it as
a bare argument to 'adb shell date' causes the device shell to split it
into two tokens, giving busybox date two arguments and triggering
"Max 1 argument". Quoting the entire command inside double quotes means
adb shell passes the string to the device sh as-is, which then gives
date a single quoted argument.

Fixes #307

https://claude.ai/code/session_01MsQvwKA11rxct9wjZh5CtE

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -318,7 +318,7 @@ jobs:
           ADB="$ANDROID_HOME/platform-tools/adb"
           mkdir -p results
           set -o pipefail
-          LOGCAT_SINCE=$("$ADB" shell date +'%m-%d %H:%M:%S.000')
+          LOGCAT_SINCE=$("$ADB" shell "date '+%m-%d %H:%M:%S.000'")
           ./gradlew connectedE2EAndroidTest \
             -Pe2eClass=com.gb4pc.e2e.PixelCameraOverlayE2ETest \
             | tee >(grep '^##GB4PC_TEST##' > results/e2e-overlay.ndjson) || {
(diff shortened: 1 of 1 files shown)


aunger:
The root-cause diagnosis is correct and the quoting fix works as described. Wrapping the command in double quotes makes bash pass `date '+%m-%d %H:%M:%S.000'` as a single string to `adb shell`, which hands it to `sh -c` on the device, where the single-quoted argument arrives as one token to busybox `date`. The two changed lines are identical in both test steps, so no instance was missed.

**However, there is a latent bug this PR does not address, and it will produce silent failures.**

`adb shell` is well-known for returning output with `\r\n` line endings instead of `\n`. If the device shell appends `\r` to the date string, `LOGCAT_SINCE` will contain a trailing carriage return, e.g. `05-31 14:23:45.000\r`. That value is then passed directly to:

```
"$ADB" logcat -d -T "$LOGCAT_SINCE"
```

`logcat -T` is strict about its time argument format. A trailing `\r` will cause it to reject the timestamp, meaning the logcat dump on test failure will either be empty or show all logs from the beginning of time--defeating the entire purpose of capturing `LOGCAT_SINCE`.

Compare line 198 in the same file, where this exact problem is already known and handled:

```sh
"$("$ADB" shell getprop sys.boot_completed 2>/dev/null | tr -d '\r')"
```

`LOGCAT_SINCE` gets no such treatment in either test step. The fix should add `| tr -d '\r'` (or equivalent) to both assignments, just as `getprop` does.

This is not a theoretical concern--the `\r` behavior of `adb shell` is consistent across Android versions and is the reason the `getprop` line was written that way. The quoting fix removes the crash but leaves the logcat capture broken in a way that will only be noticed when a test actually fails and someone wonders why the logcat output is useless.

Changes requested.

---

aunger:
You're right. The `\r` behavior of `adb shell` is consistent and well-documented, and the existing `getprop` line at 198 was already written defensively for exactly this reason. Not carrying that same treatment over to `LOGCAT_SINCE` was an oversight.

I've added `| tr -d '\r'` to both `LOGCAT_SINCE` assignments in the same style as line 198. All local unit tests pass. Pushed to this branch.

---

aunger:
The fix is correct. Both `LOGCAT_SINCE` assignments now pipe through `| tr -d '\r'`, in exactly the same style as the `getprop` line at 198. The quoting fix from the original commit is preserved. Both changed lines are symmetric and no instance was missed.

Approved.

---

PR #311, opened 2026-05-31
Gate CI on a red-to-green test allowlist instead of continue-on-error

### LINKED ISSUE #309: Remove continue-on-error, still allow individual failed tests
In light of
  - #307 

stop using `continue-on-error` CI workflow tasks, because they hide real errors.

Instead, find another way to allow a few tests to fail while I do this red-to-green thing.

aunger:
Fixes #309.

## Problem

The CI test steps used `continue-on-error: true`, which swallowed every error.
Real, unexpected failures (including non-test breakage like the broken `date` invocation in #307) went unnoticed because the build stayed green regardless.

## Approach

Replace the blanket `continue-on-error` on the instrumented and E2E test steps with an explicit, temporary allowlist gate.

- `scripts/check_allowed_failures.py` parses the JUnit XML for each suite and fails the build for any failing test, except those named in `.github/allowed-test-failures.txt`. The allowlist exists only to let a few known-red tests stay red during a deliberate red-to-green effort; it is empty by default.
- The gate also takes each test step's real exit status via `--outcome`. A step that reported `failure` without writing any failing test result is reported as an infrastructure failure and still fails the build. This is the class of breakage `continue-on-error` used to hide (e.g. #307).
- In `build.yml`, the instrumented and E2E run steps no longer use `continue-on-error`. Each records its exit status as a step output and lets later steps run, so artifacts, the test summary, and issue filing are never skipped. A final `Gate on test failures` step is the single authority on pass/fail.
- The on-failure E2E screenshot pull now reads the recorded step outputs rather than the native step outcome, since the run steps no longer fail at the shell level.
- `.github/workflows/CLAUDE.md` is rewritten to describe the new failing-with-allowlist design and how to allowlist a test temporarily.

The allowlist format uses `#` as the class/method separator, so inline comments are stripped only when preceded by whitespace and the separator is preserved.

## Testing

- `scripts/test_check_allowed_failures.py` covers allowlist parsing (including the `#` separator vs. inline comment distinction), failure parsing, exit codes, the `--outcome` infrastructure-failure path, and argument parsing.
- Ran the full project script test suite locally: `python3 -m unittest discover -s scripts -p "test_*.py"` (172 tests, all passing) and all `scripts/test_*.sh` shell suites (51 assertions, all passing).
- Validated `build.yml` as YAML and smoke-tested the gate end to end against fixture XML for the clean-pass, blocking-failure, allowlisted-failure, and infrastructure-failure cases.

No application (Kotlin) code changed, so the Android Gradle build is unaffected.

### DIFF
--- a/.github/allowed-test-failures.txt
+++ b/.github/allowed-test-failures.txt
@@ -1,0 +1,19 @@
+# Tests allowed to fail without failing CI.
+#
+# This file is the replacement for the old `continue-on-error: true` on the
+# test steps in build.yml (see issue #309).  `continue-on-error` swallowed
+# every error, including real infrastructure breakage, so the build went
+# green even when something genuinely broke.  Now any failing test fails the
+# build, EXCEPT tests listed here.
+#
+# Use this list only as a temporary measure during a deliberate red-to-green
+# effort, and remove entries as the tests are fixed.
+#
+# Format (one entry per line):
+#   com.gb4pc.e2e.SomeTest#someMethod    # tolerate one method
+#   com.gb4pc.e2e.SomeTest               # tolerate the whole class
+#
+# Blank lines and lines starting with `#` are ignored; trailing inline `#`
+# comments are stripped.
+#
+# (No tests are currently allowlisted.)
--- a/.github/allowed-test-failures.txt
+++ b/.github/allowed-test-failures.txt
@@ -16,4 +16,15 @@
 # Blank lines and lines starting with `#` are ignored; trailing inline `#`
 # comments are stripped.
 #
-# (No tests are currently allowlisted.)
+# These tests are red on main today; `continue-on-error` hid them before
+# issue #309 removed it.  Each already has a tracking issue and must be
+# removed from this list as it is fixed.
+
+com.gb4pc.ui.MainActivityRedirectTest#mainScreen_showsPixelCameraMissingCard_whenNotInstalled   # issue #303
+com.gb4pc.ui.MainActivityRedirectTest#setupScreen_showsGrantButton                              # issue #304
+com.gb4pc.ui.MainActivityRedirectTest#pickerScreen_loadsApps_andShowsSettingsApp                # issue #305
+
+com.gb4pc.e2e.GalleryButtonVisualE2ETest#test1a_overlayShowsBlueAtConfiguredPosition            # issue #229
+com.gb4pc.e2e.GalleryButtonVisualE2ETest#test2a_emptyGalleryNoGreenAfterTap                     # issue #230
+com.gb4pc.e2e.GalleryButtonVisualE2ETest#test3a_populatedGalleryShowsGreenAfterTap              # issue #231
+com.gb4pc.e2e.GalleryButtonVisualE2ETest#test5a_secureCameraLockedPopulatedGalleryShowsGreen    # issue #232
--- a/.github/workflows/CLAUDE.md
+++ b/.github/workflows/CLAUDE.md
@@ -1,57 +1,54 @@
 # CI Architecture
 
-## Non-Failing CI Design
+## Failing CI with an explicit red-to-green allowlist
 
-The CI pipeline in this repository is intentionally configured to **not fail the overall build** when tests fail. This is a deliberate temporary decision while test fixes are landing incrementally.
+The CI pipeline fails the build when a test fails.
+The only exception is a small, explicit allowlist of tests that are permitted to stay red during a deliberate red-to-green effort.
 
-### Why Non-Failing?
+This replaced the previous `continue-on-error: true` design (issue #309).
+`continue-on-error` swallowed *every* error, including non-test breakage (for example the broken `date` invocation in issue #307), so the build went green while genuine failures hid.
 
-During periods of active test stabilization:
-- Test failures should not block check-ins or code merges
-- All contributors should be able to land fixes incrementally
-- Failed tests are automatically filed as GitHub issues for tracking and prioritization
-- The pipeline continues to produce artifacts (APKs, screenshots, logcat) to support debugging
+### Why an allowlist instead of `continue-on-error`?
 
-This allows the development velocity to remain high while test suites are being stabilized.
+During periods of active test stabilization we still want to let a few known-red tests stay red without blocking unrelated work.
+The allowlist does that narrowly:
 
-### How It Works
+- A genuine, unexpected test failure fails the build, giving real signal.
+- Infrastructure or tooling breakage that aborts a test step (no test result written) fails the build.
+- Only the specific tests named in the allowlist are tolerated, and only temporarily.
+- Failed tests are still filed as GitHub issues, and the pipeline still produces artifacts (APKs, screenshots, logcat) for debugging.
 
-The build pipeline uses `continue-on-error: true` on test steps that can fail:
+### How it works
 
-1. **Unit tests and compile steps** run normally and fail the build if they encounter errors (fast feedback loop)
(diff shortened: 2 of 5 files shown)


aunger:
**Review: Approved.**

The approach is correct and well-executed. A few observations:

**Positive**

- The three-part design (record outcome as step output, defer pass/fail to one gate step, run gate with `if: always()`) is exactly right. It preserves artifact upload, summaries, and issue filing regardless of test results.
- The `--outcome` infra-failure path directly addresses the #307 class of breakage that `continue-on-error` hid.
- The `_INLINE_COMMENT` regex correctly distinguishes the `#` class/method separator (not preceded by whitespace) from an inline comment (preceded by whitespace). The unit test for that case is present.
- Test coverage is thorough: all documented behaviors have corresponding cases.

**Minor observations (not blocking)**

- There is no test for the combination of a missing result directory AND `outcome=failure`. The existing `test_missing_directory_is_skipped_not_fatal` only tests the no-outcome case. The actual behavior in that combination--triggering an infra failure--is arguably correct (missing directory + step failure = something is broken), but a test would document the intent. Not blocking.
- The unit test suite has no `--outcome` passed to the gate step. If the unit test Gradle task aborts before writing XML, the gate would not catch the infra failure for that suite. However, unit tests still fail the build natively (no `continue-on-error` was ever applied to them), so the build would already be failing before the gate runs. Coverage is adequate.

The gate correctly consolidates pass/fail authority, the allowlist mechanism is narrow and explicit, and the documentation in `CLAUDE.md` and the allowlist header is clear. Ready to merge.

---

aunger:
**Review verdict: Approved.**

Reviewing the new commit specifically (SHA 29f4476, "Allowlist seven tests already red on main").

**The new commit is exactly one change:** 12 lines added to `.github/allowed-test-failures.txt` (plus removal of the "empty by default" placeholder comment), listing 7 test methods with tracking issue citations. No production code, no test code, no workflow logic changed.

**The Author's "already red on main" claim is credible and verifiable:**

- Issues #229, #230, #231, #232 (the four GalleryButtonVisualE2ETest methods) were opened by CI on 2026-05-25, six days before this PR was created.
- Issues #303, #304, #305 (the three MainActivityRedirectTest methods) were opened by CI against PR #297 on 2026-05-31 at 04:39 UTC, more than five hours before this PR was opened at 10:19 UTC.

All seven tracking issues are open and reference CI runs and commits that are not part of this PR. The failures are not regressions introduced here.

**The allowlist entries are correctly formatted.** Each uses the `ClassName#methodName` form with a whitespace-preceded inline comment (`# issue #NNN`). The `_INLINE_COMMENT` regex in `check_allowed_failures.py` strips only whitespace-preceded `#` comments, so the class/method separator is preserved and the inline issue references are stripped. The unit test `test_method_separator_hash_is_not_treated_as_comment` covers exactly this pattern.

**The gate still blocks new failures.** Any test not on the allowlist that fails will fail CI. The allowlist is intentionally narrow: 7 named methods, each with an open issue, each expected to be removed as its issue is resolved.

**This change is in scope for #309.** The allowlist mechanism was described in the original PR as the intended escape hatch for known-red tests, and the PR description explicitly anticipated this use. Populating it with the pre-existing red tests is the correct application of that mechanism.

No concerns with this commit. The PR is sound.

---

aunger:
It seems like a Gradle solution would be simpler and less coupled to the current CI system. 

Thoughts?

```groovy
test {
    ignoreFailures = true  // prevents task from stopping early

    def allowedFailures = [
        'com.example.FlakyTest',           // entire class
        'com.example.OtherTest.myMethod',  // specific method
    ] as Set

    def unexpectedFailures = []

    afterTest { TestDescriptor desc, TestResult result ->
        if (result.resultType == TestResult.ResultType.FAILURE) {
            def classMatch  = allowedFailures.contains(desc.className)
            def methodMatch = allowedFailures.contains("${desc.className}.${desc.methodName}")
            if (!classMatch && !methodMatch) {
                unexpectedFailures << "${desc.className}.${desc.methodName}"
            }
        }
    }

    doLast {
        if (unexpectedFailures) {
            throw new GradleException(
                "Build failed — unexpected test failures:\n  " +
                unexpectedFailures.join("\n  ")
            )
        }
    }
}
```

aunger:
Thanks, this is a fair instinct and I tried hard to make it work, but after walking it against how this repo actually runs its tests I don't think the `test {}` hook can replace the gate. The short version: `afterTest` only fires for the JVM unit-test task, and that is the one suite that already fails the build on its own. The suites the allowlist actually exists for never flow through a Gradle `Test` task at all.

**Where `afterTest` can and cannot see results in this project**

1. **Unit tests** (`testDebugUnitTest`) are a real `org.gradle.api.tasks.testing.Test` task, so `afterTest` works. We already use exactly this hook, scoped under `testOptions.unitTests.all { ... }` in `app/build.gradle.kts` (the `##GB4PC_TEST##` listener). But these tests have never had `continue-on-error` and already fail the build natively, so they need no allowlist.

2. **Instrumented tests** run via `./gradlew connectedDebugAndroidTest`. That is AGP's `DeviceProviderInstrumentTestTask`, not a `Test` task. The JUnit run happens on the emulator via the on-device runner; Gradle only collects the resulting XML. `afterTest`/`ignoreFailures` are members of `Test` and do not apply to it, so a `test {}` block would not observe a single instrumented result.

3. **E2E tests** run via the custom `connectedE2EAndroidTest` task (`app/build.gradle.kts`, the `tasks.register(...) { doLast { ... } }` near the bottom). It shells out to `adb shell am instrument -r -w` and parses `INSTRUMENTATION_STATUS` from stdout itself. There is no Gradle `Test` instance anywhere in that path, so `afterTest` never fires.

So the allowlist entries we care about are all in suites #2 and #3 (`GalleryButtonVisualE2ETest`, `MainActivityRedirectTest`), and a `test {}` hook structurally cannot reach them. The snippet would silently police only the unit suite, which is the suite that needs no policing.

**The infrastructure-failure case (#307) cuts the other way**

`ignoreFailures = true` is the part I'm most wary of. The whole point of #309 was that `continue-on-error` hid non-test breakage (the broken `date` in #307: a step that fails *without writing any test result*). The current gate catches that because each step records its real exit status and `check_allowed_failures.py` treats "step failed but produced no failing test" as an infra failure. `ignoreFailures = true` does the opposite. It tells Gradle to swallow the task's failure, which reintroduces exactly the #307 blind spot. And for the E2E task specifically, the failure isn't even a JUnit assertion. It is a `GradleException` thrown from `doLast` on crash/empty-results/non-zero `am instrument`, which `ignoreFailures` (a `Test`-only flag) wouldn't govern anyway.

**On coupling**

I agree coupling to CI is a real cost and worth minimizing. But the gate is a plain Python script driven by JUnit XML plus a step outcome string; it isn't tied to GitHub Actions internals and is unit-tested in `scripts/test_check_allowed_failures.py`. The `test {}` approach would actually be *more* coupled to a specific runner: it only covers the JVM task, so we'd still need a separate mechanism for the instrumented and E2E suites, leaving the allowlist split across two systems with different semantics.

**What I'd happily adopt**

The good idea in your snippet, building the allowlist into the build rather than a sidecar file, is one I'd take for the unit suite if we want defense in depth there. But since unit failures already break the build, it would be redundant today, and it still wouldn't cover the two suites that motivated #309. Given that, I'd prefer to keep the single gate as the one authority over all three suites rather than introduce a second, narrower mechanism.

If I've misread how you intend `test {}` to reach the instrumented/E2E runs (for example via a wrapper `Test` task I'm not seeing), I'm glad to dig further. But as the build is wired today I believe the script gate is the only thing that can see all three suites and still catch the #307 class of failure, so I've kept it.

---

aunger:
Good info, thanks.

PR #312, opened 2026-05-31
Render skipped test suites as skipped, not green, in the CI summary

### LINKED ISSUE #255: CI bug: after pre-flight failure, all tests show green

aunger:
Fixes #255.

## Problem

When the CI pre-flight check fails, GitHub Actions skips the test steps (they have no `if: always()`), so those steps write no JUnit XML.
The "Write test-result summary" step still runs (`if: always()`) and calls `summarize_test_results.py` against the now-empty/absent result directories.
The summarizer rendered such empty suites with a neutral "_No test results found._" note that did not convey that the tests never ran, so the build-and-test job summary did not surface the skipped state.

Scope: this PR addresses only the misleading summary. The pre-flight failure itself is out of scope (tracked by #256).

## Fix

- Thread each producing step's GitHub Actions `outcome` through the summarizer via an optional `--outcome` flag, mirroring the existing `SuiteSpec` / `--outcome` pattern already used by `check_allowed_failures.py` (issue #309).
- When a suite has no results and its step `outcome` was `skipped`, the summary now renders the suite as `SKIPPED` (explaining the test step did not run) instead of as a plain no-results note. A skipped test phase can no longer read as green.
- Wire the workflow's "Write test-result summary" step to pass the instrumented and E2E step outcomes (using the step-level `steps.<id>.outcome` context, which reports `skipped` when the step never ran).

`parse_args` now returns `SuiteSpec` objects instead of `(dir, label)` tuples; the existing `parse_args` tests were updated to read the new attributes (no requirement was loosened).

## Tests

- New unit tests cover: `--outcome` parsing (present, optional between suites, missing-value error), `render_suite` rendering skipped vs. no-results, `build_markdown` for a skipped suite, and an end-to-end `main()` test reproducing the issue #255 scenario (missing directory + `--outcome skipped`) that asserts the output reads as skipped and never as passing.
- `scripts/test_summarize_preflight_integration.sh`: a shell integration test that runs the summarizer with skipped suites and asserts the rendered summary reports SKIPPED (and not the no-results note). It captures the rendered output with `env -u GITHUB_STEP_SUMMARY` so it works both locally and under CI (where `GITHUB_STEP_SUMMARY` is set).
- `.github/workflows/simulate-preflight-failure.yml`: a standalone regression workflow that reproduces the implicit preflight-skip summary scenario.
- Local run: full `python3 -m unittest discover -s scripts -p "test_*.py"` passes (179 tests). Both workflow YAMLs validated.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -459,8 +459,10 @@ jobs:
               --suite-label "Unit Tests" \
             app/build/outputs/androidTest-results/instrumented \
               --suite-label "Instrumented Tests" \
+              --outcome "${{ steps.run_instrumented.outcome }}" \
             app/build/outputs/androidTest-results/e2e \
-              --suite-label "E2E Tests"
+              --suite-label "E2E Tests" \
+              --outcome "${{ (steps.run_overlay_e2e.outcome == 'skipped' && steps.run_gallery_e2e.outcome == 'skipped') && 'skipped' || ((steps.run_overlay_e2e.outcome == 'failure' || steps.run_gallery_e2e.outcome == 'failure') && 'failure' || 'success') }}"
 
       # Single authority on whether the test phase passed (issue #309). Replaces
       # the old blanket `continue-on-error` on the test steps: any failing test
--- a/.github/workflows/simulate-preflight-failure.yml
+++ b/.github/workflows/simulate-preflight-failure.yml
@@ -1,0 +1,81 @@
+name: Simulate pre-flight failure
+
+# Manually-triggered regression check for issue #255: when a pre-flight failure
+# implicitly skips the test phase, the test-result summary must report SKIPPED
+# rather than a misleading "all green" / "no test results found" outcome.
+#
+# This workflow reproduces production's *implicit* skip mechanism: a failed
+# preflight step puts the job into a failed state, which skips all subsequent
+# non-`always()` steps (the same mechanism by which build.yml's test steps are
+# skipped). It does NOT use an explicit `if` guard on the test steps.
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  simulate:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      # Fails the job (no continue-on-error), so every subsequent step that is
+      # not marked `if: always()` is implicitly skipped.
+      - name: Pre-flight check
+        id: preflight
+        run: exit 1
+
+      # The three simulated test steps below carry NO `if` guard. They are
+      # implicitly skipped because the job is already in a failed state. This
+      # mirrors build.yml, whose real test steps are gated only on
+      # `needs_full_build`, not on the preflight outcome.
+      - name: Run instrumented tests
+        id: run_instrumented
+        run: echo "instrumented tests ran"
+
+      - name: Run PixelCameraOverlayE2ETest
+        id: run_overlay_e2e
+        run: echo "overlay e2e ran"
+
+      - name: Run GalleryButtonVisualE2ETest
+        id: run_gallery_e2e
+        run: echo "gallery e2e ran"
+
+      # DRIFT NOTE: the summarize_test_results.py invocation below is a copy of
+      # build.yml's "Write test-result summary" step and must be kept in sync
+      # with it (e.g. if a fourth suite is added, or the E2E outcome expression
+      # changes). It is pointed at guaranteed-absent result directories so the
+      # summary reflects the skipped outcomes, and its output is tee'd to
+      # /tmp/summary.md for the assertion step to read.
+      - name: Write test-result summary
+        if: always()
+        run: |
+          env -u GITHUB_STEP_SUMMARY python3 scripts/summarize_test_results.py \
+            /nonexistent/unit \
+              --suite-label "Unit Tests" \
+            /nonexistent/instrumented \
+              --suite-label "Instrumented Tests" \
+              --outcome "${{ steps.run_instrumented.outcome }}" \
+            /nonexistent/e2e \
+              --suite-label "E2E Tests" \
+              --outcome "${{ (steps.run_overlay_e2e.outcome == 'skipped' && steps.run_gallery_e2e.outcome == 'skipped') && 'skipped' || ((steps.run_overlay_e2e.outcome == 'failure' || steps.run_gallery_e2e.outcome == 'failure') && 'failure' || 'success') }}" \
+            | tee /tmp/summary.md
+
+      - name: Assert summary reports SKIPPED, not green
+        if: always()
+        run: |
+          echo "=== Captured summary ==="
+          cat /tmp/summary.md
+          echo "========================"
+          if ! grep -qF "⏭ SKIPPED" /tmp/summary.md; then
+            echo "FAIL: summary did not report '⏭ SKIPPED' for the skipped test phase."
+            exit 1
+          fi
+          if sed -n '/### Instrumented Tests/,$p' /tmp/summary.md | grep -qF "_No test results found._"; then
+            echo "FAIL: summary contains misleading '_No test results found._'."
+            exit 1
+          fi
+          echo "PASS: summary correctly reports the skipped test phase as SKIPPED."
--- a/scripts/summarize_test_results.py
+++ b/scripts/summarize_test_results.py
@@ -8,10 +8,16 @@ back to stdout when the env var is not set).
 Usage:
     python3 scripts/summarize_test_results.py \\
         path/to/unit-results          --suite-label "Unit Tests" \\
-        path/to/instrumented-results  --suite-label "Instrumented Tests" \\
-        path/to/e2e-results           --suite-label "E2E Tests"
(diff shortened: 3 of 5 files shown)


aunger:
**Review: APPROVED**

**Workflow wiring is correct.** The "Write test-result summary" step uses `steps.run_instrumented.outcome` (the built-in runner outcome, not `.outputs.outcome`). This is the right choice: when the step is skipped by its `if:` condition, `.outcome` is `skipped` while `.outputs.outcome` would be empty string. The E2E composite expression correctly yields `skipped` only when both sub-steps are skipped, which is the pre-flight-failure scenario issue #255 describes.

**Unit Tests have no `--outcome` -- correctly omitted.** "Build and run unit tests" executes before the pre-flight check, so it cannot be the victim of a pre-flight failure. No `--outcome` is needed there.

**`parse_args` test changes did not loosen requirements.** `test_single_pair` is stricter than before (it now also asserts `outcome == ""`). `test_two_pairs` is equivalent; the attribute access replaces tuple indexing but checks the same values.

**Scope is clean.** No pre-flight failure logic (issue #256) is touched. The fix is confined to rendering and the workflow step that calls the summarizer.

**Test coverage is thorough.** The new tests cover `--outcome` parsing (present, optional between suites, missing-value error), `render_suite` skipped vs. non-skipped outcomes, `build_markdown` for a skipped suite, and a full `main()` integration test reproducing the issue #255 scenario (missing directory + `--outcome skipped` never reads as passing).

---

aunger (commented):
Reviewing commits `aff5082` and `d974d1a` only.

## Shell test (`scripts/test_summarize_preflight_integration.sh`) -- Good

The scoping fix for the negative assertion is correctly implemented. `SKIPPED_PORTION` is extracted via `sed -n '/### Instrumented Tests/,$p'` and the `_No test results found._` check is applied only to that portion, not to the whole output. This correctly excludes Unit Tests (which legitimately emits `_No test results found._` because it has no `--outcome` and no XML) while still catching any regression where a skipped suite reverts to the no-results rendering. The same scoping is applied consistently in both section (A) (stdout) and section (B) (GITHUB_STEP_SUMMARY file). The `⏭ SKIPPED` positive assertion and the file-existence check are correct. Build.yml auto-discovery of `scripts/test_*.sh` will pick this up. No issues.

## Simulated workflow (`.github/workflows/simulate-preflight-failure.yml`) -- Bug

The "Assert summary reports SKIPPED, not green" step applies the negative grep to the entire `/tmp/summary.md` file, without scoping to the skipped suites:

```sh
if grep -qF "_No test results found._" /tmp/summary.md; then
  echo "FAIL: summary contains misleading '_No test results found._'."
  exit 1
fi
```

The summary step passes Unit Tests with no `--outcome` and a guaranteed-absent directory. With no outcome and no XML, `render_suite` falls through to the `else` branch and emits `_No test results found._` for Unit Tests. The grep above will find that string and the assertion step will always exit 1, turning every run of this workflow into a false failure regardless of whether the skipped-suite fix is working.

The shell test handles this correctly by scoping the negative check: `sed -n '/### Instrumented Tests/,$p' | grep -qF "_No test results found._"`. The workflow assertion must apply equivalent scoping--replacing the bare `grep -qF "_No test results found._" /tmp/summary.md` with `sed -n '/### Instrumented Tests/,$p' /tmp/summary.md | grep -qF "_No test results found._"`. The positive `⏭ SKIPPED` assertion is unaffected and correct.

The DRIFT NOTE comment is present. The E2E outcome expression matches build.yml exactly. The implicit-skip mechanism (failing preflight, no `continue-on-error`, no `if` guard on test steps) is correctly reproduced. The `if: always()` placement on both the summary and assertion steps is correct. Build.yml and summarize_test_results.py are untouched by these two commits.

**Verdict: changes requested.** Fix the negative assertion in the workflow's "Assert summary reports SKIPPED, not green" step to scope to only the skipped-suite portion of the summary file, matching the logic already present in the shell test.

---

aunger:
Good catch, fixed in 543140e.

You're right: "Unit Tests" is passed with no `--outcome` and no XML, so `render_suite` legitimately emits `_No test results found._` for it.
The unscoped `grep -qF "_No test results found._" /tmp/summary.md` matched that, so the assertion step would have exited 1 on every dispatch (a false failure).

The assertion now scopes the check to the skipped-suite portion, matching the scoping already used in `test_summarize_preflight_integration.sh`:

```
if sed -n '/### Instrumented Tests/,$p' /tmp/summary.md | grep -qF "_No test results found._"; then
```

The `⏭ SKIPPED` presence check is unchanged.

Validation: workflow YAML parses (PyYAML); the shell integration test passes (5/5); `python3 -m unittest discover -s scripts -p "test_*.py"` passes (179 tests). I also replayed the assertion logic against the summarizer's real output: the old unscoped grep matched (confirming the false failure), the new scoped check passes.

---

aunger:
Reviewer verdict on commit 543140e: **Approved.**

Commit 543140e makes exactly the one mechanical edit requested and nothing more.

The "Assert summary reports SKIPPED, not green" step now reads:

```sh
if sed -n '/### Instrumented Tests/,$p' /tmp/summary.md | grep -qF "_No test results found._"; then
```

This scopes the no-results check to the Instrumented Tests section onward, matching the scoping already used in `test_summarize_preflight_integration.sh`. The `⏭ SKIPPED` presence check against the full file is unchanged. No other lines were touched in any file.

This PR is approved.

---

aunger:
Reviewed commit 9e8964d.

The fix is correct and minimal. `env -u GITHUB_STEP_SUMMARY` unsets the variable for that single invocation, which causes `summarize_test_results.py` to fall back to stdout (confirmed in the script at lines 283-289). The `tee /tmp/summary.md` then captures that output, and the assertion step reads the same file it always has. Nothing else in the step changed.

The `--outcome` expressions, the implicit-skip mechanism, and the assertion logic are all untouched. The production summary step in `build.yml` (line 463) is correctly not modified--it still writes to `$GITHUB_STEP_SUMMARY` as intended. Scope is clean: only `simulate-preflight-failure.yml` changed.

Approved.

---

aunger (commented):
Reviewed commit 89ee79a.

The fix is correct and appropriately scoped.

`summarize_test_results.py` (lines 283-289) writes to `$GITHUB_STEP_SUMMARY` when set, and falls back to stdout otherwise.
Group A captured stdout via `$(...)`, so in CI--where the variable is always set--`$OUTPUT` was empty and the `⏭ SKIPPED` assertion failed.
`env -u GITHUB_STEP_SUMMARY` cleanly removes the variable for that one invocation, restoring the stdout path in both CI and local environments.

Verified:
- Only `scripts/test_summarize_preflight_integration.sh` is touched; `summarize_test_results.py` and all workflow files are untouched.
- Group A's two assertions are unchanged--no weakening.
- Group B, which deliberately exercises the file-write path by setting the variable explicitly, is untouched and correct.
- The added comment accurately explains the problem and the fix.
- The author's reported results (5 passed, 0 failed with `GITHUB_STEP_SUMMARY` set; same without; 179 unittest OK) are consistent with the change.

**Verdict: Approved.**

---

aunger (commented):
Final review pass on the cleaned-up branch (decce8a). **Verdict: Approved.**

**Scope check**

Diff touches exactly the 5 expected files (339+/43-). No `verify-255-acceptance` job, no broadened `on:` triggers, no `check_allowed_failures.py` / `allowed-test-failures.txt` re-introduced from #309. Clean.

**Core fix**

`render_suite` correctly routes an empty-results suite to "SKIPPED" only when `outcome == "skipped"`, falling through to `_No test results found._` for all other outcomes including `"failure"`. The `SuiteSpec` dataclass and `parse_args` refactor are straightforward; the old tuple-pair shape is fully replaced.

**build.yml outcome wiring**

The summary step uses `steps.run_instrumented.outcome` (the built-in step context property), not `.outputs.outcome` (a step output set by the run script). This is the right choice: the built-in `.outcome` is `"skipped"` when the step never ran, whereas `.outputs.outcome` would be empty. The E2E aggregation expression is correct: both-skipped => skipped, either-failed => failure, mixed skipped+failed => failure, otherwise success. Unit Tests carries no `--outcome` intentionally--its step has no `id:` and a failure there stops the job before the summary step runs anyway.

**Test coverage**

42 Python unit tests pass. Shell integration test passes 5/5. Both the shell test and the simulate workflow correctly use `env -u GITHUB_STEP_SUMMARY` to redirect summarizer output to stdout/tee rather than to the step-summary file, so the assertion steps can read what was rendered. The negative assertion for `_No test results found._` is correctly scoped to `### Instrumented Tests` and later, so the Unit Tests suite's expected "no results" note does not trigger a false failure.

**simulate-preflight-failure.yml**

Reproduces the implicit skip correctly (no `if:` guard on test steps, a bare `exit 1` pre-flight). Drift note is present and accurate.

**Commit messages**

Three commits, no attribution bylines, messages accurately describe the changes. "Fixes #255" is present on the first commit.

No issues found. Ready to merge.

---

PR #316, opened 2026-06-03
Fix ReleaseVersionTest to accept pre-release version suffixes

aunger:
The semver regex only matched X.Y.Z, so tags like v0.0.13-rc.1 (version
0.0.13-rc.1) failed the unit test and blocked the entire release workflow,
leaving the prerelease APK unbuilt and unattached.

Extend the regex to accept an optional -PRERELEASE label per semver spec.

https://claude.ai/code/session_01MGhVzu42qRPAQGJ1HG6fJP

### DIFF
--- a/app/src/test/java/com/gb4pc/ReleaseVersionTest.kt
+++ b/app/src/test/java/com/gb4pc/ReleaseVersionTest.kt
@@ -14,8 +14,8 @@ import java.time.format.DateTimeParseException
  * validate the build that produced this APK, not future builds.
  *
  * Verifies:
- *  1. versionName is either semver (MAJOR.MINOR.PATCH) for tagged releases, or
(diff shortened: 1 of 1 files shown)


aunger:
**Review**

The diagnosis is correct.
The workflow ran but the unit test `versionName is semver or dev build label` rejected `0.0.13-rc.1` because the regex only accepted bare `X.Y.Z`.
The extended regex `(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?` correctly matches the pre-release suffix and the fix is minimal.

One observation: the inner group `(\.[0-9A-Za-z-]+)*` is a capturing group whose captured value is never used.
`(?:\.[0-9A-Za-z-]+)*` would be more precise, but this has no functional impact on `Regex.matches()`.

**Approved.**
The fix is correct and the capturing-group imprecision is not worth a change cycle.

---

PR #322, opened 2026-06-06
Skip Python and shell unit tests on docs-only PRs

### LINKED ISSUE #321: Don’t run Python or shell unit tests on docs-only PRs
Skip these steps too when the Kotlin steps are skipped.

aunger:
Apply the same `needs_full_build` gate that already guards every Kotlin build and test step to the "Run Python script unit tests" and "Run shell script unit tests" steps, so all test execution is skipped when a PR touches only Markdown files.

Fixes #321

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,9 +106,11 @@ jobs:
           ANDROID_AVD_HOME="$AVD_HOME" $ANDROID_HOME/emulator/emulator -list-avds
 
       - name: Run Python script unit tests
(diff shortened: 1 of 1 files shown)


aunger (commented):
The change is correct and complete. Both "Run Python script unit tests" and "Run shell script unit tests" now carry `if: steps.diff_check.outputs.needs_full_build == 'true'`, which is the identical condition used on every other skippable step in the job. The `diff_check` step that defines this output runs unconditionally, so the condition is always available. Nothing else in the job needed adjustment.

Verdict: APPROVED.

---

PR #325, opened 2026-06-06
Log full list of foreground apps in ForegroundDetector summary

### LINKED ISSUE #324: Log foreground apps
change the logging to show the full list of current FG apps in foreground detector, not just the top one.

aunger:
Fixes #324

Previously the summary log line in `ForegroundDetector.getForegroundPackage()` showed only the single top (most-recent) foreground package:

```
ForegroundDetector: foreground=com.google.android.GoogleCamera (1 foreground event(s) of 3 total)
```

This change adds an `allForegroundPackages` list that accumulates every distinct foreground package seen in the query window (self-events excluded), and appends it to the summary line:

```
ForegroundDetector: foreground=com.google.android.GoogleCamera, all FG apps=[com.android.launcher3, com.google.android.GoogleCamera] (2 foreground event(s) of 3 total)
```

The "no foreground detected" branch is unchanged because `allForegroundPackages` is always empty in that case (there are no non-self foreground events to accumulate).

Three new tests in `ForegroundDetectorSelfFilterTest` verify:
- the summary line contains `all FG apps=` with all packages when multiple apps appear in the window
- the self package is excluded from the list
- a single-app window still emits the `all FG apps=` field


aunger:
**Review: APPROVED**

The implementation is minimal, correct, and tightly scoped to the issue.

A few notes:

- `allForegroundPackages` is populated after the self-event `continue`, so self-packages are correctly excluded from the list--confirmed by the dedicated test.
- The `.contains()` dedup on a `MutableList` is O(n), but this is diagnostic-only log code iterating over a small, bounded set of apps; no concern.
- The three new tests cover all the meaningful cases: multiple packages, self-exclusion, and single-package. The `DebugLog.clear()` call in `setUp()` is the right guard against cross-test log pollution.
- The `else` (no foreground detected) branch is correctly unchanged--`allForegroundPackages` would be empty there and is not referenced in that log line.

---

aunger:
Reviewed.

The change is correct and well-scoped. The `allForegroundPackages` list accurately accumulates distinct non-self foreground packages and is appended to the summary log line as required by Issue #324. The `else` branch is correctly left unchanged.

One minor note: `mutableListOf<String>()` with `contains()` before each `add()` is O(n) per event for deduplication. A `LinkedHashSet` would be the idiomatic Kotlin choice for "ordered distinct elements" and would read more clearly. In practice the event window is short and this is debug-only logging, so it has no impact--I raise it only for awareness.

Test coverage is solid. The three new tests cover multi-app listing, self-package exclusion, and the single-app case. The `DebugLog.clear()` in `setUp` is correct hygiene. All tests follow the existing conventions in the file.

**Approved.**

---

aunger (commented):
The production change is correct and minimal. The logic that populates `allForegroundPackages` sits after the `selfPackage` guard (`continue` fires before `.add()`), so self-exclusion works. The `LinkedHashSet` serialises to `[pkg1, pkg2]` format via Kotlin's default `toString()`, matching the example in the PR description. The "no foreground detected" branch is correctly unaffected because nothing is ever added to the set in that path.

The tests have two problems that need fixing before merge.

**1. Massive duplication between the two test files**

`ForegroundDetectorLogTest` (the new file, 176 lines) and the three tests appended to `ForegroundDetectorSelfFilterTest` cover the same ground. Mapping the overlap:

| ForegroundDetectorLogTest | ForegroundDetectorSelfFilterTest |
|---|---|
| "summary log includes single FG app" | "summary log all FG apps shows single package when only one foreground app" |
| "summary log lists all FG apps when multiple packages appear" | "summary log includes all FG apps when multiple packages appear" |
| "self package is not included in all FG apps list" | "summary log all FG apps excludes self package" |

That leaves only two tests in `ForegroundDetectorLogTest` that are not duplicated elsewhere: the deduplication test ("same package appearing multiple times is listed only once") and the "no foreground events produces 'no foreground app detected'" test. The three-app variant ("summary log lists three distinct FG apps") is also redundant with the existing three-app test in `ForegroundDetectorSelfFilterTest`.

This PR should ship one of the following, not both:
- Keep `ForegroundDetectorLogTest` and drop the three appended tests from `ForegroundDetectorSelfFilterTest`; or
- Keep the appended tests in `ForegroundDetectorSelfFilterTest` and drop `ForegroundDetectorLogTest` (moving only the deduplication and "no events" tests into `ForegroundDetectorSelfFilterTest`).

The second option is preferable because `ForegroundDetectorSelfFilterTest` already owns the `eventsOf` helper and the self-filtering context, and the new cases are a natural extension. The deduplication and "no events" cases should be added there rather than in a separate file.

**2. Inconsistent use of `.first {}` vs `.last {}`**

`ForegroundDetectorLogTest` uses `.first { it.startsWith("ForegroundDetector: foreground=") }` to find the summary line; `ForegroundDetectorSelfFilterTest` uses `.last { ... }`. With `DebugLog.clear()` in `setUp()`, both are equivalent for single-call tests, but `.first` and `.last` communicate different intent and will diverge if a test ever calls `getForegroundPackage()` more than once. All summary-line lookups should use `.last` for consistency, because the summary line is always the final log entry for a given detection pass.

---

aunger (commented):
**Review: Changes requested**

The production change is correct and minimal. `allForegroundPackages` is a `LinkedHashSet` populated only after the self-event `continue`, so self-exclusion is guaranteed. The `else` branch is untouched. The log format matches the PR description exactly.

The issue is with the tests.

**Duplicate test class must be removed**

`ForegroundDetectorLogTest` (new file, 176 lines) covers exactly the same scenarios as the three tests added to `ForegroundDetectorSelfFilterTest`:

| `ForegroundDetectorLogTest` test | Duplicate in `ForegroundDetectorSelfFilterTest` |
|---|---|
| `summary log includes single FG app in all-apps list` | `summary log all FG apps shows single package when only one foreground app` |
| `summary log lists all FG apps when multiple packages appear` | `summary log includes all FG apps when multiple packages appear` |
| `self package is not included in all FG apps list` | `summary log all FG apps excludes self package` |

The new file also duplicates the `eventsOf`/`setEventField` helper and the `@RunWith(RobolectricTestRunner::class)` setup that already exists in `ForegroundDetectorSelfFilterTest`. There is no structural reason to split these into two files--they test the same class, use the same runner, and are already in the same package.

The two tests in `ForegroundDetectorLogTest` that are NOT already in `ForegroundDetectorSelfFilterTest` (`summary log lists three distinct FG apps` and `same package appearing multiple times is listed only once`) are worth keeping, as is the `no foreground events` test. These three should be moved into `ForegroundDetectorSelfFilterTest` alongside the three already there, and `ForegroundDetectorLogTest` should be deleted.

To summarize: move the non-duplicate tests from `ForegroundDetectorLogTest` into `ForegroundDetectorSelfFilterTest`, then delete `ForegroundDetectorLogTest`. No production code changes needed.

---

PR #334, opened 2026-06-07
Post a sticky PR comment linking to the CI build-and-test summary

### LINKED ISSUE #331: Link PRs to their CI test summaries
When CI build runs for a PR, add a link from the PR to the build-test summary, similar to this:
https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27063845957#summary-79881146537

aunger:
Fixes #331.

CI already writes a rich pass/fail summary to the `build-and-test` job's "Summary" section (see `summarize_test_results.py`), but nothing on the PR points at it. Reviewers have to dig through the Checks tab to find it.

This adds a new `link-pr-to-ci-summary` job (depends on `build-and-test`, runs on `pull_request` events regardless of build outcome) that posts or updates a single sticky comment on the PR linking directly to the job's summary anchor, e.g.:

```
https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/<run_id>#summary-<job_id>
```

This matches the link format requested in the issue.

## How it works

`scripts/post_pr_ci_summary_link.py`:
- Looks up the numeric `build-and-test` job ID via the REST Jobs API (the anchor embeds the job ID, not the run ID), filtering by the current `run_attempt` so re-runs don't pick up a stale job from an earlier attempt.
- Builds the summary URL `#summary-<job_id>` (falling back to the bare run URL if the job ID can't be found).
- Creates a comment on the PR, or edits its prior comment in place (found via a hidden HTML marker), so repeated pushes update the same comment instead of piling up new ones.
- Checks `build-and-test`'s new `needs_full_build` job-level output (passed through as `SUMMARY_WRITTEN`): docs-only PRs skip the "Write test-result summary" step, so the job's "Summary" section has nothing to link to. In that case the comment links to the bare run and says plainly that no pass/fail summary was written, instead of claiming there's a summary to view.

The job is granted `pull-requests: write` (overriding the workflow's default `contents: read`), matching how `file-issues` already overrides with `issues: write`.

## Tests

Added `scripts/test_post_pr_ci_summary_link.py` (25 tests) covering PR-number extraction, job-ID lookup (including run-attempt disambiguation and API-error handling), comment body construction (both the normal and no-summary-written cases), sticky-comment lookup/create/edit, and the `main` entry point's environment handling and skip conditions. Full `python3 -m unittest discover -s scripts -p "test_*.py"` suite passes (204 tests).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,12 @@ jobs:
     # ubuntu-latest has KVM hardware acceleration (available since April 2024).
     runs-on: ubuntu-latest
 
+    outputs:
+      # Surfaced so link-pr-to-ci-summary (issue #331) can tell whether the
+      # "Write test-result summary" step actually ran--on docs-only PRs it is
+      # skipped, and the job's "Summary" section has nothing to link to.
+      needs_full_build: ${{ steps.diff_check.outputs.needs_full_build }}
+
     steps:
       - uses: actions/checkout@v6
         with:
@@ -543,6 +549,11 @@ jobs:
   # Reviewers can then jump straight to the test summary instead of hunting
   # through the Checks tab. Runs regardless of build outcome so the link is
   # always available, including on red builds.
+  #
+  # Docs-only PRs skip the "Write test-result summary" step entirely (see
+  # needs_full_build), so the job's "Summary" section has nothing to link to;
+  # SUMMARY_WRITTEN tells the script to say so rather than post a misleading
+  # link.
   link-pr-to-ci-summary:
     needs: [build-and-test]
     runs-on: ubuntu-latest
@@ -561,6 +572,7 @@ jobs:
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_RUN_ATTEMPT: ${{ github.run_attempt }}
           WORKFLOW_RUN_PR_URL: ${{ github.event.pull_request.html_url }}
+          SUMMARY_WRITTEN: ${{ needs.build-and-test.outputs.needs_full_build }}
         run: python3 scripts/post_pr_ci_summary_link.py
 
   # Produce a signed release-config APK on every merge to main for easy sideloading.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -537,6 +537,32 @@ jobs:
             downloaded-artifacts/e2e-test-results \
               --suite-label "E2E Tests"
 
+  # Surface the build-and-test job's pass/fail summary directly on the PR
+  # (issue #331): post or update a sticky comment linking to the job's
+  # "Summary" section, e.g. .../actions/runs/<run_id>#summary-<job_id>.
+  # Reviewers can then jump straight to the test summary instead of hunting
+  # through the Checks tab. Runs regardless of build outcome so the link is
+  # always available, including on red builds.
+  link-pr-to-ci-summary:
+    needs: [build-and-test]
+    runs-on: ubuntu-latest
+    if: always() && github.event_name == 'pull_request' && needs.build-and-test.result != 'cancelled'
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Post CI summary link on PR
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GITHUB_SERVER_URL: ${{ github.server_url }}
+          GITHUB_RUN_ID: ${{ github.run_id }}
+          GITHUB_RUN_ATTEMPT: ${{ github.run_attempt }}
+          WORKFLOW_RUN_PR_URL: ${{ github.event.pull_request.html_url }}
+        run: python3 scripts/post_pr_ci_summary_link.py
+
   # Produce a signed release-config APK on every merge to main for easy sideloading.
   # Named gb4pc-dev.<run_number>.apk to distinguish from tagged releases.
   # Waits for the full test suite (unit + E2E) before producing an artifact.
--- a/scripts/post_pr_ci_summary_link.py
+++ b/scripts/post_pr_ci_summary_link.py
@@ -1,0 +1,183 @@
+#!/usr/bin/env python3
+"""post_pr_ci_summary_link.py--Link a PR to its build-and-test job summary.
+
+CI produces a rich pass/fail summary in the `build-and-test` job's "Summary"
+section (see `summarize_test_results.py`), but nothing on the PR points at it:
+reviewers must hunt through the Checks tab to find it. This script posts (or
+updates) a single sticky comment on the triggering PR with a direct link to
+that job's summary anchor, e.g.:
+
+    https://github.com/<owner>/<repo>/actions/runs/<run_id>#summary-<job_id>
+
+It looks up the numeric job ID via the REST Jobs API (the anchor embeds the
+job ID, not the run ID), matching on job name and the current run attempt so
+re-runs don't pick up a stale job from a previous attempt.
+
+The comment is "sticky": a hidden HTML marker lets subsequent runs on the same
+PR find and edit the existing comment in place rather than piling up a new one
+per push.
+
+Usage:
+    python3 scripts/post_pr_ci_summary_link.py
+
+Environment:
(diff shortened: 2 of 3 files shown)


aunger:
Review

The overall approach (sticky comment with a hidden marker, job-ID lookup via the Jobs API, run-attempt disambiguation) is sound and well tested (22 passing tests). However, I found a substantive correctness problem that should be fixed before merge:

**The link can point at a summary that does not exist for a given run.**

`build-and-test` only writes the rich pass/fail summary (the `### Summary` content that the `#summary-<job_id>` anchor is meant to scroll to) when `steps.diff_check.outputs.needs_full_build == 'true'` (see the `Write test-result summary` step, gated the same as the rest of the heavy build/test steps). For docs-only PRs, `needs_full_build` evaluates to `false`, the summary step is skipped, and the job's "Summary" section contains nothing meaningful.

`link-pr-to-ci-summary`, however, runs unconditionally on every `pull_request` event (`if: always() && github.event_name == 'pull_request' && needs.build-and-test.result != 'cancelled'`) and always builds a `#summary-<job_id>` URL whenever it can resolve the job ID, regardless of whether a summary was actually written. Worse, `needs_full_build` is a step output of `build-and-test`, not a job-level `outputs:`, so the new job has no way to detect this case even if it wanted to.

The result: on a docs-only PR, reviewers get a sticky comment that says "View the build-and-test summary for this PR's latest run", linking to an anchor that has no real summary content behind it. That is misleading, and undermines the very purpose of this feature (helping reviewers jump straight to the test summary).

To fix this properly, `build-and-test` would need to expose `needs_full_build` (or an equivalent "summary was written" signal) as a job-level output that `link-pr-to-ci-summary` can check before posting (or before claiming there is a summary to view).

**Minor:**
- The comment body's wording ("View the build-and-test summary for this PR's latest run") is going to be subtly wrong even outside the docs-only case: once the comment is edited in place on a later push, the link always points at the *latest* run, which is consistent with what the body claims, so that part is fine. But combined with the above, on a docs-only PR the "latest run" claim and the "summary" claim can both be technically true (it is the latest run) while the linked content is empty, compounding the confusion rather than clarifying it.

Everything else looks solid: the marker-based sticky-comment lookup, the run-attempt filtering to avoid stale job IDs on re-runs, the graceful fallbacks (bare run URL when the job ID can't be resolved, no-op when not a `pull_request` event or token/run-id missing), and the always-exit-0 contract (a missing link must never fail the build) are all handled carefully and are well covered by tests.

Verdict: changes requested, primarily to address the docs-only-PR case where the posted link claims to point at a summary that was never written.

---

aunger:
Review

I independently re-checked `build.yml` and confirm the substantive issue raised in the prior review: `needs_full_build` is a *step* output of `build-and-test` (`steps.diff_check.outputs.needs_full_build`), gating the "Write test-result summary" step at line ~456-457, and is **not** exposed as a job-level `outputs:` (no job in this workflow currently exposes job-level outputs — this would be new plumbing). So on a docs-only PR (where `needs_full_build` evaluates to `false`), `link-pr-to-ci-summary` will still post "View the build-and-test summary for this PR's latest run" pointing at a `#summary-<job_id>` anchor whose "Summary" section contains nothing meaningful. That's a real, user-visible correctness gap that undercuts the feature's purpose, not a nit. I agree this needs to be fixed before merge — either by exposing a job-level "summary was written" signal that `link-pr-to-ci-summary` can check, or some other mechanism that avoids posting a misleading "view the summary" claim when no summary was written.

A few additional points from my own pass:

1. **Attribution byline in the PR description.** The PR description ends with `_Generated by [Claude Code](https://claude.ai/code/session_011KAycJTfAncDeiLZ1B6gAT)_`. `agents/pr_participation.md` is explicit that bylines (Claude attribution, links) are prohibited in issue/PR description text. Please remove it.

2. **Tests verified.** I ran `python3 -m unittest scripts.test_post_pr_ci_summary_link -v` myself — all 22 tests pass, matching the PR description's claim.

3. **Style/convention consistency (positive).** `_github_headers`, the `requests.get/.post/.patch` + `raise_for_status` pattern, and the `Authorization: Bearer` header all mirror `file_test_failure_issues.py`'s existing GitHub REST helpers closely — good consistency with the established conventions in `scripts/`.

4. **Minor: comment wording.** Even setting aside the docs-only case, "View the build-and-test summary for this PR's latest run" reads a little oddly inside a sticky comment that gets edited in place — "latest run" is always true by construction, so the phrase doesn't add information and could just say "View the build-and-test summary for this PR" (or similar). Not blocking, just a wording nit.

Verdict: changes requested — primarily for the docs-only-PR case where the link claims to point at a summary that was never written, and secondarily to remove the attribution byline from the PR description.

---

aunger:
Review

Re-reviewed after commit `4c24a6d` ("Avoid posting a misleading CI summary link on docs-only PRs"). Both substantive issues raised in the prior round are resolved:

1. **Docs-only-PR misleading link.** `build-and-test` now exposes `needs_full_build` as a job-level output (`outputs.needs_full_build: ${{ steps.diff_check.outputs.needs_full_build }}`), correctly mirroring the same step output that gates the "Write test-result summary" step (`steps.diff_check.outputs.needs_full_build == 'true'`). `link-pr-to-ci-summary` passes it through as `SUMMARY_WRITTEN`, and `post_pr_ci_summary_link.py` now branches on it: when no summary was written, it posts "View this PR's build-and-test run" with an explicit note ("This PR did not need a full build, so no pass/fail summary was written...") instead of claiming a summary exists. This is exactly the fix requested and it is correctly wired end to end.

2. **Attribution byline.** The PR description no longer contains the `_Generated by [Claude Code](...)_` footer.

3. **Wording nit.** "for this PR's latest run" has been dropped from the link text per feedback ("View the build-and-test summary for this PR" / "View this PR's build-and-test run").

I also re-ran the test suite independently: `python3 -m unittest scripts.test_post_pr_ci_summary_link -v` — all 25 tests pass, matching the updated PR description's count.

One small thing worth a sanity check before merge but not blocking: the existing sticky comment already posted on this PR (from before the fix) still shows the old "for this PR's latest run" / unconditional-summary wording — that's expected to self-correct on the next CI run that updates it in place, just flagging so it isn't mistaken for a live discrepancy.

Verdict: LGTM.

---

aunger (commented):
Review

I independently re-checked the current head (`4c24a6d`) against the concerns raised in the earlier review rounds, and they're all genuinely resolved:

1. **Docs-only-PR misleading link (the substantive issue).** `build-and-test` now exposes `outputs.needs_full_build: ${{ steps.diff_check.outputs.needs_full_build }}` at the job level, `link-pr-to-ci-summary` passes it through as `SUMMARY_WRITTEN`, and `post_pr_ci_summary_link.py` branches on it correctly: when `summary_written` is false it skips the (now unnecessary) job-ID lookup entirely, links to the bare run, and says plainly "This PR did not need a full build, so no pass/fail summary was written." `build_comment_body` and the `main` skip-condition tests (`test_links_to_bare_run_and_explains_when_summary_not_written`, `test_treats_missing_summary_written_as_false`) cover this cleanly, including asserting `find_job_id` is not called when there's nothing to look up.
2. **Attribution byline.** Confirmed gone from the PR description.
3. **Wording.** "for this PR's latest run" has been dropped; the link text now reads "View the build-and-test summary for this PR" / "View this PR's build-and-test run", which is clearer for a sticky comment that's edited in place.

I also ran the suite myself: `python3 -m unittest scripts.test_post_pr_ci_summary_link -v` — all 25 tests pass.

Other things I checked and found solid:
- `_github_headers`, the `requests.get/.post/.patch` + `raise_for_status` + `except Exception` warning pattern, and the `Authorization: Bearer` header match `file_test_failure_issues.py`'s established REST helper conventions closely.
- The `run_attempt` filtering in `find_job_id` correctly avoids picking up a stale job ID from an earlier attempt on a re-run, with a sensible fallback to an unfiltered match when the attempt is unknown.
- The sticky-comment marker/lookup/create-or-edit logic is straightforward and well tested, and the `if: always() && ... && needs.build-and-test.result != 'cancelled'` guard plus the script's always-exit-0 contract mean a failure here can't break the build.
- `find_job_id`/`find_existing_comment` cap at `per_page: 100` without pagination — for `build-and-test` (one job per name) and PR comment counts in this repo that's not a practical concern, and it matches the precedent in `file_test_failure_issues.py` (which also doesn't paginate), so I'm not flagging it as blocking.

Verdict: LGTM.

---

PR #338, opened 2026-06-07
Clarify Orchestrator/Verification-Planner handoff sequencing in agents docs

### LINKED ISSUE #335: Clarify Orchestrator/Verification-Planner manual-vs-automation handoff in agents docs
`agents/dev_orchestration.md`'s CI Monitor loop says, on a `Clear` line:

> Dispatch a Verification Planner sub-agent using the dispatch template.
> Relay the user's choice (manual testing or automation) to the planner verbatim.

This reads as if the Orchestrator must already have the user's manual-vs-automation choice in hand *before* dispatching the planner.

But `agents/verification_planning.md` describes a different sequencing: the planner itself builds the before-merging list, presents it to the user, and *then* (step 3d) asks the user "Do you want to run these tests manually, or have an agent plan automation for them?" — i.e., the planner solicits the choice itself, after the report, not before.

In practice the user wanted the before-merging report first and to decide on automation afterward, which only `verification_planning.md`'s flow supports — `dev_orchestration.md`'s wording suggests the Orchestrator should pre-collect that choice, which isn't possible without a report to base the decision on.

Please reconcile the two documents so the handoff sequencing is unambiguous (e.g. clarify that the Orchestrator does not need to pre-collect the choice when dispatching the planner, since the planner solicits it itself once it has findings to present).

aunger:
Fixes #335

`agents/dev_orchestration.md`'s CI Monitor loop said, on a `Clear` line:

> Dispatch a Verification Planner sub-agent using the dispatch template.
> Relay the user's choice (manual testing or automation) to the planner verbatim.

This read as if the Orchestrator must already have the user's manual-vs-automation choice in hand before dispatching the planner. But `agents/verification_planning.md` describes the opposite sequencing: the planner itself assembles the before-merging list, presents it to the user, and only then (step 3d) asks the user "Do you want to run these tests manually, or have an agent plan automation for them?"

This PR rewords the relevant step in `agents/dev_orchestration.md` to make the sequencing unambiguous: the Orchestrator dispatches the planner without a pre-collected choice, the planner solicits that choice itself once it has findings to present, and the Orchestrator relays the user's resulting choice back to the planner afterward.

---
_Generated by [Claude Code](https://claude.ai/code/session_011KAycJTfAncDeiLZ1B6gAT)_

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -158,8 +158,10 @@ Monitor output lines are relayed to the user verbatim; this is user-facing statu
       // Dispatch the verification planner (see verification_planning.md).
       // The Orchestrator does not scan the issue or PR itself.
       Dispatch a Verification Planner sub-agent using the dispatch template.
-      The planner assembles and presents the before-merging list, then solicits a choice from User: Will User perform manual validation, or should the planner plan an approach to automate the lost items?
-      Once the user responds, relay the user's choice (manual testing or automation) to the planner verbatim, so it can proceed per its own instructions.
+      The planner assembles the before-merging list, then presents it with a question: Will User perform manual validation, or should the planner plan an approach to automate the lost items?
+      If the Planner reports *no work remains*, then this *before-merging requirements* process is complete, and the Orchestrator should exit this step.
+      Relay any information and questions to the User.
+      Once the user responds, relay the user's choice(s) about the pending work to the planner verbatim, so it can proceed per its own instructions.
       If the user opts for automation and the planner produces a plan:
         Dispatch a Reviewer to evaluate the plan; follow the normal Reviewer loop.
         Once the plan is approved, dispatch an Author to implement it.
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -158,7 +158,7 @@ Monitor output lines are relayed to the user verbatim; this is user-facing statu
       // Dispatch the verification planner (see verification_planning.md).
       // The Orchestrator does not scan the issue or PR itself.
       Dispatch a Verification Planner sub-agent using the dispatch template.
-      The Orchestrator does not need a manual-vs-automation choice in hand before this dispatch: the planner first assembles and presents the before-merging list, then solicits that choice from the user itself (see verification_planning.md step 3d).
+      The planner assembles and presents the before-merging list, then solicits a choice from User: Will User perform manual validation, or should the planner plan an approach to automate the lost items?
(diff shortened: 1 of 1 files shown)


aunger:
Review

The core fix is correct: the new wording properly removes the implication that the Orchestrator must pre-collect the manual-vs-automation choice before dispatching the planner, and it correctly explains that the planner solicits the choice itself once it has a before-merging list to present (matching `verification_planning.md` step 3d). This resolves the sequencing contradiction described in #335.

One nit on the added wording in `dev_orchestration.md`:

> The Orchestrator does not need a manual-vs-automation choice in hand before this dispatch: the planner first assembles and presents the before-merging list, then solicits that choice from the user itself (see verification_planning.md step 3d).

"solicits that choice from the user itself" reads as if the planner converses with the user directly. But per the same file's "Orchestrator communication discipline" section (a few dozen lines above), "The Orchestrator is a mute message-passer between the user and the sub-agents," and relaying happens "in either direction." Sub-agents don't talk to the user directly — the planner's question in step 3d must be relayed to the user by the Orchestrator, and the user's answer relayed back. The phrase "from the user itself" obscures that the Orchestrator is still the relay for this exchange (which the very next sentence then describes — "Once the user responds, relay the user's choice... to the planner verbatim"). As written, the two sentences are slightly in tension: the first implies a direct planner-user conversation, the second clarifies the Orchestrator is actually relaying. Consider rewording the first sentence to make clear the planner's solicitation is also relayed through the Orchestrator (e.g., "...then asks the user (via the Orchestrator) for that choice..."), so the whole passage consistently describes one continuous relay-mediated exchange rather than reading like two different communication models.

This is a wording precision nit, not a blocker on the substance of the fix — the sequencing problem from the issue is genuinely resolved by this change.

---

aunger (commented):
Re-review

The latest revision (`ecf173a`) resolves the wording nit from my previous comment. The line now reads:

&gt; The planner assembles and presents the before-merging list, then solicits a choice from User: Will User perform manual validation, or should the planner plan an approach to automate the lost items?

This drops "from the user itself" entirely, so there's no longer any implication that the planner converses with the user directly — the whole passage now reads as one continuous Orchestrator-relayed exchange (assemble/present → solicit choice → "Once the user responds, relay the user's choice... to the planner verbatim"), consistent with the "Orchestrator communication discipline" section's mute-message-passer model.

The substance of the fix (removing the false implication that the Orchestrator must pre-collect the manual-vs-automation choice before dispatching the planner, and aligning with `verification_planning.md` step 3d's actual sequencing) remains correct and resolves #335.

LGTM.

---

PR #340, opened 2026-06-08
File tracking issues for before-merging requirements instead of consulting the user

### LINKED ISSUE #339: Change the verification-planner agent’s process
# Change the verification-planner agent’s process.
Follow up to #335 (because I changed my mind).

## What now?

The verification planner should still behave as current up to the point of presenting the remaining requirements to the User and offering to make an automated plan.

Instead of communicating with the User, it should take these steps:

1. If its list of *before merging* requirements is empty, it should report this success and exit.
2. It should open a GitHub issue titled, `(re PR #{number}) {required task title}` for each of the items on the *before merging* list.
3. Every new issue should be linked to the current PR in the following ways:
    - Title includes PR number as shown above
    - Description text includes a URL to the particular PR comment that called for this requirement
    - Blocks the PR
    - Is sub-issue of the PR (if this link type is valid in GitHub)
4. Report the *before-merge* list to the orchestrator and exit.

aunger:
Fixes #339. Follow-up to #335.

Changes the Verification Planner's process so that, once it has assembled the *before merging* list, it no longer presents that list to the user or asks whether to test manually or automate. Instead it acts autonomously:

1. If the *before merging* list is empty, it reports this success to the Orchestrator and exits.
2. Otherwise it opens one GitHub issue per outstanding requirement, titled `(re PR #{number}) {required task title}`.
3. Each new issue is linked to the current PR: the PR number is in the title, the description includes the URL of the specific PR comment that called for the requirement, the issue blocks the PR, and the issue is made a sub-issue of the PR when that link type is valid.
4. It reports the *before merging* list to the Orchestrator and exits. The PR may be merged once every filed issue is resolved.

## Files

- `agents/verification_planning.md`: rewrote Role, the before-merging list section, the What-to-do steps, and Boundaries to describe issue-filing instead of user consultation and automation planning. Removed the `verification needed` label step and the manual-vs-automate question.
- `agents/dev_orchestration.md`: reconciled the Monitor-loop handoff and the dispatch role-assignment line with the new process (planner files tracking issues and does not consult the user; merge is gated on those issues being resolved).

Docs-only change; no build or unit tests apply.

---
_Generated by [Claude Code](https://claude.ai/code/session_011KAycJTfAncDeiLZ1B6gAT)_

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -97,7 +97,7 @@ Role assignment statements (copy the applicable line exactly):
 
 - Programmer: "You are a Programmer resolving the linked issue. You *must* start your turn by re-fetching the description and all comments on the issue and, if it exists, on the PR."
 - Reviewer: "You are a Reviewer ensuring high quality and adherence to the development plan for the linked issue. You *must* start your turn by re-fetching the description and all comments on the issue and on the PR."
-- Verification planner: "You are a Verification Planner: scan the linked issue and PR for unautomated verification steps and produce an automation plan without implementing it. You *must* start your turn by re-fetching the description and all comments on the issue and on the PR."
+- Verification planner: "You are a Verification Planner: scan the linked issue and PR for outstanding before-merging requirements and file a tracking issue, linked to the PR, for each one. You *must* start your turn by re-fetching the description and all comments on the issue and on the PR."
 
 ## Decision-signal templates
 
@@ -158,15 +158,10 @@ Monitor output lines are relayed to the user verbatim; this is user-facing statu
       // Dispatch the verification planner (see verification_planning.md).
       // The Orchestrator does not scan the issue or PR itself.
       Dispatch a Verification Planner sub-agent using the dispatch template.
-      The planner assembles the before-merging list, then presents it with a question: Will User perform manual validation, or should the planner plan an approach to automate the lost items?
-      If the Planner reports *no work remains*, then this *before-merging requirements* process is complete, and the Orchestrator should exit this step.
-      Relay any information and questions to the User.
-      Once the user responds, relay the user's choice(s) about the pending work to the planner verbatim, so it can proceed per its own instructions.
-      If the user opts for automation and the planner produces a plan:
-        Dispatch a Reviewer to evaluate the plan; follow the normal Reviewer loop.
-        Once the plan is approved, dispatch an Author to implement it.
-        Follow the normal Author -> Reviewer -> CI Monitor cycle for the resulting changes.
-      → PR may be merged once every item on the planner's before-merging list is resolved: after the automation Author's work clears CI (or immediately if the user opts for manual testing or no automation), and after any changes outside the repo have been performed.
+      The planner assembles the before-merging list and, for each item, files a tracking issue linked to the PR (see verification_planning.md). It does not consult the user.
+      If the Planner reports its before-merging list is empty, then this *before-merging requirements* process is complete, and the Orchestrator should exit this step.
+      Otherwise, relay the planner's reported before-merging list to the user verbatim.
+      → PR may be merged once every issue the planner filed for its before-merging list is resolved.
 ```
 
 ### Monitor script
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -34,17 +34,19 @@ You file one tracking issue per item (see **What to do**) so that each requireme
    c. Record that the new issue **blocks** the PR, using GitHub's issue-dependencies feature.
       GitHub exposes this through the REST API, so use a `gh api` call (or an equivalent dependency tool if one is available to you) to add the relationship.
       To mark the PR as blocked by the new issue:
-      `gh api repos/{owner}/{repo}/issues/{number}/dependencies/blocked_by -f issue_id={new issue's id}`,
+      `gh api repos/{owner}/{repo}/issues/{number}/dependencies/blocked_by -F issue_id={new issue's id}`,
       where `{number}` is the current PR number and `{new issue's id}` is the new issue's internal id (the `id` field, not its number; obtain it from the issue-creation response or `gh api repos/{owner}/{repo}/issues/{new issue number}`).
+      Use `-F` (not `-f`) so the id is sent as a JSON integer, as the API requires, rather than a string.
       If the PR number is not accepted for an issue-dependency relationship, fall back to blocking the **parent issue** the PR resolves (the issue this PR fixes) instead:
-      `gh api repos/{owner}/{repo}/issues/{parent issue number}/dependencies/blocked_by -f issue_id={new issue's id}`.
+      `gh api repos/{owner}/{repo}/issues/{parent issue number}/dependencies/blocked_by -F issue_id={new issue's id}`.
       Only if neither call succeeds, state the blocking relationship in plain text in the new issue's description (for example, "Blocks PR #{number}") so it is not lost, and skip the formal link without failing.
    d. Make the new issue a **sub-issue** of the PR, using GitHub's sub-issues feature.
       GitHub exposes this through the REST API, so use a `gh api` call (or an equivalent sub-issue tool if one is available to you):
-      `gh api repos/{owner}/{repo}/issues/{number}/sub_issues -f sub_issue_id={new issue's id}`,
+      `gh api repos/{owner}/{repo}/issues/{number}/sub_issues -F sub_issue_id={new issue's id}`,
       where `{number}` is the current PR number and `{new issue's id}` is the new issue's internal id (not its number).
+      Use `-F` (not `-f`) so the id is sent as a JSON integer, as the API requires, rather than a string.
       If the PR number is not accepted for a sub-issue relationship, fall back to making the new issue a sub-issue of the **parent issue** the PR resolves instead:
-      `gh api repos/{owner}/{repo}/issues/{parent issue number}/sub_issues -f sub_issue_id={new issue's id}`.
+      `gh api repos/{owner}/{repo}/issues/{parent issue number}/sub_issues -F sub_issue_id={new issue's id}`.
       If neither call succeeds, skip this link without failing.
 4. Report the *before merging* list to the Orchestrator and exit.
    The PR may be merged once every one of these filed issues is resolved.
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -4,7 +4,8 @@
 
 You are a Verification Planner.
 You are the final validation reviewer.
-You scan the linked issue and PR for outstanding requirements that must be handled before merging, and, when the user opts in, produce a concrete automation plan without implementing anything.
+You scan the linked issue and PR for outstanding requirements that must be handled before merging, and you file a tracking GitHub issue for each one so none is lost.
+You do not communicate with the user and you do not implement anything.
 
 Two kinds of outstanding requirement are your responsibility:
 
@@ -15,30 +16,31 @@ Two kinds of outstanding requirement are your responsibility:
 ## The before-merging list
 
 Assemble a single *before merging* list that names every outstanding requirement you find, of either kind above.
-This list is the deliverable the merge decision depends on: the PR may be merged only once every item on it is resolved (by automation, by manual testing, or by performing the outside-the-repo action).
+This list is the deliverable the merge decision depends on: the PR may be merged only once every item on it is resolved.
+You file one tracking issue per item (see **What to do**) so that each requirement is followed up and blocks the PR until done.
 
 ## What to do
 
 1. Read the issue description, PR description, and all comments on both.
    Look for both kinds of outstanding requirement described under **Role**: unautomated verification steps, and changes outside the repo (such as an issue that needs to be filed).
-2. If none are found, report that to the user: the *before merging* list is empty; no unautomated steps or outside-the-repo requirements were identified; the PR may be merged.
-3. If any outstanding requirements are found:
-   a. Present the *before merging* list clearly to the user, labeling each item as either an unautomated verification step or a change outside the repo.
-   b. Apply the `verification needed` label to the PR and/or issue where the outstanding items were found.
-   c. For changes outside the repo, surface what must be done (for example, file the issue) so it is not lost; these are not automatable as tests.
-   d. For unautomated verification steps, ask the user: "Do you want to run these tests manually, or have an agent plan automation for them?"
-4. If the user chooses manual testing or no automation, report that back to the Orchestrator and stop.
-   The PR may be merged once every item on the *before merging* list is resolved: manual testing is complete and any outside-the-repo requirements have been performed.
-5. If the user opts for automation, produce a concrete automation plan:
-   - Describe what to automate and how.
-   - Reference the existing test infrastructure (test directories, CI config, test framework in use).
-   - Do NOT implement anything yet.
-   The plan is then reviewed by a Reviewer agent; if the Reviewer approves, an Author agent implements it.
+   Assemble the *before merging* list, labeling each item as either an unautomated verification step or a change outside the repo, and noting for each item the URL of the specific PR comment that called for it.
+2. If the *before merging* list is empty, report this success to the Orchestrator (no unautomated steps or outside-the-repo requirements were identified; the PR may be merged) and exit.
+3. Otherwise, open one GitHub issue for each item on the *before merging* list.
+   Do NOT communicate with the user, and do NOT ask whether to test manually or to automate.
+   For each item:
+   a. Title the issue `(re PR #{number}) {required task title}`, where `{number}` is the current PR number and `{required task title}` is a short title for the outstanding requirement.
+   b. In the issue description, include a URL to the particular PR comment that called for this requirement.
+      (If the requirement came from the PR or issue description itself rather than a comment, link to that description instead.)
+   c. Link the new issue to the current PR so that it **blocks** the PR.
+   d. Make the new issue a **sub-issue** of the PR, if that link type is valid in GitHub. If it is not valid, skip this link without failing.
+4. Report the *before merging* list to the Orchestrator and exit.
+   The PR may be merged once every one of these filed issues is resolved.
 
 ## Boundaries
 
-- Do not implement any automation yourself.
+- Do not communicate with, or ask questions of, the user. Your only conversational output is the report to the Orchestrator.
+- Do not produce an automation plan, and do not implement any automation yourself.
 - Do not modify source files.
 - Do not commit or push anything.
-- Do not perform the outside-the-repo actions yourself (for example, do not file the issue); only surface them on the *before merging* list so they are not lost.
 - Limit your reading to the issue, PR, and project test infrastructure references.
+- The only repository-changing action you take is filing the tracking issues described above and linking them to the PR.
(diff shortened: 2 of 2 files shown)


aunger:
**Review (Reviewer)**

Reviewed against issue #339's four steps. The empty-list exit, title format `(re PR #{number}) {required task title}`, the "report and exit" closing step, and the removal of the now-orphaned `verification needed` label step and manual-vs-automation branch in `dev_orchestration.md` all line up well with the requested process change. One substantive gap and a couple of smaller points:

**1. "Blocks the PR" link has no specified mechanism, and no fallback (substantive)**

Step 3c says: "Link the new issue to the current PR so that it **blocks** the PR." Unlike step 3d (sub-issue), which explicitly hedges with "if that link type is valid in GitHub... skip this link without failing," step 3c gives the planner no guidance on *how* to create a blocking relationship and no fallback if it can't. Looking at the GitHub tools available to agents in this repo, there's no dedicated "mark issue as blocking" tool — `sub_issue_write` exists for parent/sub-issue relations, but blocking/blocked-by relations are a separate GitHub feature (managed via the issue sidebar / a different GraphQL mutation) and may not be reachable through the exposed toolset at all. As written, the planner is as likely to improvise something that doesn't actually create a "blocks" relation (e.g., just writing "Blocks #340" in the body, which GitHub does not treat as a formal blocking link the way it treats "Fixes #340" for closing) as it is to fail outright. Please either point to a concrete mechanism the planner can use, or add the same kind of "if not possible, skip without failing" hedge you gave to the sub-issue step — otherwise this instruction risks being silently unfulfillable or producing a cosmetic, non-functional link.

**2. Sub-issue-of-a-PR is an unusual relation; worth a sanity note**

Step 3d asks the planner to make the new issue "a sub-issue of the PR." GitHub's sub-issues feature is documented and exposed (in `sub_issue_write`) as an issue-to-issue relation; PRs and issues share numbering under the hood, so this *might* mechanically work by passing the PR's number as `issue_number`, but it's not a relationship GitHub's UI is designed to display, and the planner has no way to know in advance whether the API will accept it versus erroring. The "if not valid, skip" hedge here is good and probably sufficient — just flagging that this is genuinely uncertain territory, so don't be surprised if this turns out to need iteration once a planner actually tries it.

**3. Minor: "final validation reviewer" line is now slightly stale**

`verification_planning.md`'s Role section retains "You are the final validation reviewer," held over from the old text. That's still roughly accurate, but combined with the new "you do not communicate with the user" framing two lines later, it reads a bit oddly — a "reviewer" that never reports findings to the party making the merge decision (the user), only to the Orchestrator. Not a blocker, just a phrase that could be tightened given the role has changed from presenting findings to filing tickets.

Net: the rewritten process logically matches what #339 asked for, and the orchestration-side edits are consistent with it. The one thing I'd want resolved before merge is point 1 — the "blocks the PR" instruction needs either a concrete mechanism or the same graceful-skip hedge given to the sub-issue step, otherwise the planner is set up to either fail or fake it.

---

aunger:
**Review (Reviewer)**

The PR is unchanged since the prior review comment (head is still `9e4e2f6`, which predates that comment), and its substantive point 1 remains unaddressed, so I'm requesting changes on the same basis.

**"Blocks the PR" has no specified mechanism and no fallback (still open)**

`verification_planning.md` step 3c says: "Link the new issue to the current PR so that it **blocks** the PR," with no guidance on how and no hedge. I checked the GitHub tools actually exposed to agents in this repo (`sub_issue_write`, `issue_write`, `add_issue_comment`, `pull_request_review_write`, etc.): none of them can create a "blocks" / "blocked by" relationship. That's a distinct GitHub feature from sub-issues (a different relation, managed through its own UI/GraphQL mutation) and from closing keywords like "Fixes #N". As written, a planner following this instruction literally has no real path to fulfill it. The likely outcomes are: it stalls trying to find a nonexistent tool, or it improvises something cosmetic like writing "Blocks #340" in the issue body, which GitHub does not register as a formal blocking link (unlike "Fixes #340", which it does wire up for auto-closing). Either way the instruction is set up to be silently unfulfillable or to produce a non-functional link that looks correct on the surface but does nothing.

Compare this to step 3d (sub-issue), which explicitly hedges: "if that link type is valid in GitHub... skip this link without failing." Step 3c needs the same kind of treatment: either point to a concrete, currently-reachable mechanism for creating a blocking relation, or add the same graceful-skip hedge so the planner doesn't fail outright or fabricate a non-functional link.

Everything else in the diff (the empty-list exit and report, the `(re PR #{number}) {title}` title format, the comment-URL requirement, the removal of the now-orphaned `verification needed` label step and the manual-vs-automation branch in `dev_orchestration.md`, and the reconciled Monitor-loop handoff text) reads as a faithful, internally consistent translation of issue #339's four steps. The "final validation reviewer" phrasing nit from the prior review is minor and not a blocker on its own.

Net: same verdict as the prior review for the same reason: point 1 needs a concrete mechanism or the same "skip if not possible" hedge given to the sub-issue step before this can merge.

---

aunger:
**Author**

Pushed `98f9546` addressing the review.

**Point 1 ("blocks the PR" has no mechanism and no fallback) -- agreed, fixed.** You're right that none of the exposed GitHub tools (`sub_issue_write`, `issue_write`, `add_issue_comment`, etc.) can create a formal "blocked by" / "blocks" relation, and the old step 3c gave neither a mechanism nor a hedge. Step 3c now:

- uses a formal "blocked by" / "blocks" tool *if one is available* to the planner, and
- otherwise states the blocking relationship in plain text in the issue body (e.g. "Blocks PR #{number}") so the requirement is not lost, and skips the formal link without failing.

That gives step 3c the same graceful-skip treatment that step 3d already had, so the planner can neither stall on a nonexistent tool nor silently fabricate a cosmetic link it believes is functional.

**Point 3 (stale "final validation reviewer" phrasing) -- fixed.** Replaced with "You are the final check that no before-merging requirement is dropped," which fits the new file-tickets role rather than the old report-to-user one.

**Point 2 (sub-issue-of-a-PR is unusual) -- acknowledged, no change.** Step 3d already carries the "if not valid, skip without failing" hedge, so if the API rejects an issue-to-PR sub-issue relation the planner degrades gracefully. Agree this is uncertain territory that may need iteration once a planner exercises it, but it isn't unfulfillable as written, so I've left it as-is.

Docs-only change; no build or unit tests apply.

---

aunger:
**Review (Reviewer)**

Re-reviewed at `98f9546`, which fixes the "blocks the PR" gap raised in the prior two rounds.

I independently checked the toolset exposed to agents in this repo (`sub_issue_write`, `issue_write`, `add_issue_comment`, `pull_request_review_write`, etc.) and confirm there is no formal "blocks"/"blocked by" relationship tool, separate from the sub-issue relation and from closing keywords. The fix in step 3c is sound: try a formal tool if one exists, otherwise record "Blocks PR #{number}" in plain text in the issue body and skip the formal link without failing. That gives the planner the same graceful-skip treatment step 3d already had for sub-issues, and it can no longer stall on a nonexistent tool or fabricate a cosmetic link it might mistake for a real one.

Walking through issue #339's four steps against the final diff:

1. Empty-list exit: "report this success to the Orchestrator... and exit" (line 27). Matches.
2. Issue title format `(re PR #{number}) {required task title}` (line 31). Matches verbatim.
3. Linking: title carries the PR number, description carries the URL of the calling PR comment (with a sensible fallback to the PR/issue description when the requirement didn't originate in a comment), blocking is now handled per the fix above, and sub-issue-of-PR keeps its existing "skip if not valid" hedge. All four sub-requirements are covered.
4. Report the before-merging list to the Orchestrator and exit (lines 38-39). Matches, and the Orchestrator-side text in `dev_orchestration.md` is reconciled with it (empty-list short-circuit, no more manual-vs-automate branch, merge gated on the filed issues).

The "final check that no before-merging requirement is dropped" rephrasing also resolves the earlier "final validation reviewer... but never reports to the user" oddity cleanly.

I don't see any remaining gaps against #339's spec. This is ready to merge from my side. (No CI signal yet on the latest push; this is a docs-only change so that's expected to be uneventful.)

LGTM.


---

aunger:
**Author**

Pushed `1fdc919` in response to the user's direction that the planner should use a `gh`/API mechanism for the blocking and sub-issue links, with a parent-issue fallback.

The earlier rounds concluded that no exposed tool can create a formal "blocks" relation and settled on a plain-text fallback. That conclusion was out of date. GitHub shipped issue dependencies (blocked-by/blocks) as a generally available feature with REST support, and sub-issues have a REST endpoint too, so an agent with `gh` is not stuck with cosmetic text. Both are reachable via `gh api`:

- Blocks: `POST repos/{owner}/{repo}/issues/{number}/dependencies/blocked_by` with `issue_id`.
- Sub-issue: `POST repos/{owner}/{repo}/issues/{number}/sub_issues` with `sub_issue_id`.

Both take the target issue's internal `id` (not its number), which the doc now calls out.

Changes to `verification_planning.md` steps 3c and 3d:

- 3c (blocks) now instructs the planner to add the relationship via a `gh api` call to the issue-dependencies endpoint, rather than assuming no tool exists.
- 3d (sub-issue) likewise points at the sub-issues `gh api` endpoint instead of a vague "if the link type is valid."
- Both add the fallback the user asked for: if the PR number is not accepted for the relationship, link the **parent issue the PR resolves** instead. Both relations are issue-to-issue, and a PR shares numbering with issues but is not guaranteed to be accepted, so blocking/parenting the parent issue keeps the requirement formally tracked. Plain text remains only as a last resort if neither call succeeds.

Docs-only change; no build or unit tests apply.

---

aunger:
**Review (Reviewer)**

Re-reviewed at `1fdc919`, the round that replaced the plain-text fallback with concrete `gh api` calls for both the "blocks" and "sub-issue" relations. I confirmed against current GitHub REST docs that both endpoints are real and GA:

- `POST /repos/{owner}/{repo}/issues/{issue_number}/dependencies/blocked_by` with body `{"issue_id": <integer>}`
- `POST /repos/{owner}/{repo}/issues/{issue_number}/sub_issues` with body `{"sub_issue_id": <integer>}`

Both correctly require the issue's internal database **id**, not its number, and the doc calls that out explicitly — good catch carrying that detail through.

**1. The literal `gh api` commands use the wrong flag for an integer body parameter (substantive)**

Steps 3c and 3d both give the planner this exact command form:

```
gh api repos/{owner}/{repo}/issues/{number}/dependencies/blocked_by -f issue_id={new issue's id}
gh api repos/{owner}/{repo}/issues/{number}/sub_issues -f sub_issue_id={new issue's id}
```

`gh api`'s `-f`/`--raw-field` always serializes the value as a JSON **string** (`{"issue_id": "12345"}`), but GitHub's documented schema requires `issue_id`/`sub_issue_id` to be a JSON **integer**. The flag that produces typed values (so a numeric literal is encoded as a JSON number) is `-F`/`--field`. As written, the planner would send a string where the API expects an integer — exactly the "looks right on the surface but doesn't actually work" failure mode the last two review rounds were trying to head off for the blocking link specifically. Please change both commands to use `-F issue_id={...}` and `-F sub_issue_id={...}` (capital F).

**2. Everything else checks out**

- The empty-list exit, title format, comment-URL requirement, and the orchestration-side reconciliation in `dev_orchestration.md` all remain faithful to #339 and unchanged in substance from the version I'd previously found acceptable.
- The parent-issue fallback added in this round (in case the PR number isn't accepted for either relation) is a sensible hedge given the genuine uncertainty about whether these issue-to-issue relations accept a PR number — and the plain-text last resort is preserved as a final safety net so the requirement is never silently lost.
- The "final check that no before-merging requirement is dropped" rephrasing still reads cleanly.

Net: this is very close, and the research behind the `gh api` mechanism is sound — but the literal command syntax given to the planner has a flag bug that would make the formal "blocks"/"sub-issue" link fail or be sent with the wrong JSON type, which is the precise risk this whole back-and-forth has been about. Swap `-f` for `-F` in both commands (3c and 3d) and I'd be glad to approve.

**Changes requested.**

---

aunger:
**Author**

Pushed `bb70768` addressing the review.

**Point 1 (`-f` sends a JSON string where the API wants an integer) -- agreed, fixed.** You're right: `gh api -f`/`--raw-field` always encodes the value as a JSON string, but the `dependencies/blocked_by` and `sub_issues` endpoints require `issue_id`/`sub_issue_id` to be a JSON integer, so the call would be sent with the wrong type. Switched both the primary commands and their parent-issue fallbacks (all four `gh api` calls in steps 3c and 3d) from `-f` to `-F`/`--field`, which sends a numeric literal as a JSON number. Added a one-line note next to each command stating why `-F` is used, so a future editor does not "simplify" it back to `-f`.

**Point 2 (everything else checks out) -- no change needed.** Agreed.

Docs-only change; no build or unit tests apply.

---

aunger:
**Review (Reviewer)**

Re-reviewed at `bb70768`. The `-f` → `-F` fix from the prior round is in place in all four `gh api` commands. But stepping back from that flag-level detail surfaces a more fundamental problem with the mechanism the last three rounds converged on:

**1. `gh` is not available to agents in this environment (substantive, and likely fatal to the instructions as written)**

Steps 3c and 3d both instruct the planner to run `gh api repos/{owner}/{repo}/issues/{number}/dependencies/blocked_by -F issue_id=...` and the equivalent `sub_issues` call. I checked: the `gh` CLI is not installed in the sandbox agents run in (`gh: command not found`), and nothing in `agents/*.md` establishes that agents have `gh` access elsewhere — `verification_planning.md` is the *only* place `gh` is mentioned anywhere in the agents docs. Meanwhile, this codebase already has an established precedent for hitting the GitHub REST API from agent-adjacent tooling: `scripts/ci_monitor.py`/`ci_monitor.sh` and `scripts/archive_stale_test_failures.py` all do it via `curl`/Python `urllib` plus `$GITHUB_TOKEN`, not `gh`.

So the literal instruction the planner is given — "use a `gh api` call" — is something a planner agent will likely be unable to execute at all (the binary doesn't exist), making the primary path of 3c/3d unreachable. The fallback hedges ("if the PR number is not accepted… fall back to the parent issue," "if neither call succeeds, [skip/state in plain text]") are written to handle *API rejection*, not *tool absence*. A planner hitting `command not found` may not recognize that as the "neither call succeeds" condition the doc anticipates, and could stall trying to find or install `gh`, or could misdiagnose the failure. The doc should either point at a mechanism actually reachable in the sandbox (e.g., `curl` + `$GITHUB_TOKEN` against the REST endpoints, mirroring `ci_monitor.py`'s approach, with the JSON body containing the integer id so there's no `-f`/`-F` ambiguity at all), or explicitly say "if `gh` is unavailable, use `curl`/an HTTP call with `$GITHUB_TOKEN` against the same endpoint" as a named fallback — not lump tool-absence in with API-rejection as if they were the same failure mode.

This is a bigger gap than the `-f`/`-F` nuance the last round fixed: a wrong flag produces a malformed-but-attempted call; a missing binary produces no call at all, and three rounds of review have been spent perfecting the syntax of a command that may never run.

**2. Everything else remains solid**

The empty-list exit, the `(re PR #{number}) {required task title}` title format, the comment-URL requirement (with the description-URL fallback), the parent-issue fallback for both relations, and the reconciled `dev_orchestration.md` Monitor-loop handoff all faithfully translate #339's four steps and read cleanly. The "final check that no before-merging requirement is dropped" phrasing resolves the earlier staleness nit.

Net: close, but the `gh`-availability question needs to be settled before this can merge — otherwise the planner is being handed a recipe that calls for an ingredient that isn't in the kitchen.

**Changes requested.**

---

aunger:
**Author**

Pushed `5d6a0fd` addressing the review.

**Point 1 (`gh` is not available to agents) -- agreed, fixed.** Confirmed in the sandbox: `gh` is not installed (`gh: command not found`), but `curl` is (`/usr/bin/curl`), and the repo's established pattern for hitting the GitHub REST API from agent-adjacent tooling is exactly that, `curl`/`urllib` plus `$GITHUB_TOKEN` (`scripts/ci_monitor.sh`, `scripts/ci_monitor.py`, `scripts/file_test_failure_issues.py`, etc.). So steps 3c and 3d now:

- call the `dependencies/blocked_by` and `sub_issues` REST endpoints with `curl -sX POST -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github+json" ...`, mirroring the header form `ci_monitor.sh` uses, and
- send the issue id as a bare JSON integer in the request body (`-d '{"issue_id": {id}}'`), which is what the API requires and which also dissolves the prior `-f`/`-F` string-vs-integer concern entirely, since a JSON body literal has no such ambiguity.

I also added a parenthetical in each step stating that `gh` is not installed so a future editor does not reintroduce it, and kept the existing parent-issue and plain-text fallbacks unchanged. Tool absence is no longer conflated with API rejection: the primary path now uses a tool that is actually present in the sandbox.

**Point 2 (everything else solid) -- no change needed.** Agreed.

Docs-only change; no build or unit tests apply.

---

aunger:
**Review (Reviewer)**

Re-reviewed at `5d6a0fd`, the round that switched both `gh api` calls to `curl` + `$GITHUB_TOKEN`.

I independently verified the two REST endpoints against current GitHub documentation:

- `POST /repos/{owner}/{repo}/issues/{issue_number}/dependencies/blocked_by` with body `{"issue_id": <integer>}` — confirmed real, GA, and `issue_id` is documented as an integer.
- `POST /repos/{owner}/{repo}/issues/{issue_number}/sub_issues` with body `{"sub_issue_id": <integer>}` — confirmed real, GA, and `sub_issue_id` is documented as an integer.

Both example requests in GitHub's docs use a bare `-d '{"issue_id": 1}'`-style JSON body with curl, exactly matching what steps 3c/3d now specify. Sending the id as a bare JSON integer in the body sidesteps the `-f`/`-F` string-vs-integer problem entirely (a JSON literal has no such ambiguity), so that whole sub-thread of the review is now moot, cleanly. I also confirmed `curl` is present in the sandbox and `gh` is not, so the new primary path is actually reachable, unlike the `gh api` version two rounds ago.

Walking the rest of the diff against issue #339's four steps:

1. Empty-list exit and report to the Orchestrator (line 27): matches.
2. Issue title format `(re PR #{number}) {required task title}` (line 31): matches verbatim.
3. Linking: title carries the PR number, description carries the URL of the calling PR comment with a sensible description-fallback, blocking now has a concrete reachable mechanism with a parent-issue fallback and a plain-text last resort, and sub-issue-of-PR mirrors the same three-tier fallback. All four sub-requirements covered.
4. Report the before-merging list and exit (lines 50-51): matches, and `dev_orchestration.md`'s Monitor-loop text is reconciled (empty-list short-circuit, no manual-vs-automate branch, merge gated on the filed issues).

One smaller observation, not a blocker: steps 3c/3d both phrase the parent-issue fallback as "the parent issue the PR resolves (the issue this PR fixes)." That's the right concept, but the doc never says how the planner is meant to *identify* that issue programmatically, e.g., by parsing a `Fixes #N`/`Closes #N` keyword from the PR description, the way GitHub itself does for auto-closing. A planner could plausibly read the PR description by eye and find it (as I did here, "Fixes #339"), so this probably isn't fatal, but a one-line pointer ("look for a `Fixes #N`/`Closes #N` reference in the PR description") would remove any doubt about how to locate it, especially since this fallback only fires in the comparatively rare case where the PR-number call is rejected.

Net: the mechanism this PR converged on is sound and verified against the actual API, the sandbox tooling assumption is now correct, and the rest of the diff remains a faithful translation of #339's spec. The parent-issue-identification gap above is minor enough that I wouldn't hold the PR on it alone.

LGTM.

---

PR #341, opened 2026-06-08
Dismiss swipe keyguard before camera launch to fix flaky green-feed smoke test

aunger:
Fixes #235.

## Problem

`GalleryButtonVisualE2ETest.test0_smokeGreenFeedVisible` was flaky, passing a little over half the time.
On failure it reported `GREEN coverage in central 60% of screen is 0.0% — expected > 70%`.

## Root cause

`launchPixelCamera()` started the mock camera with a non-secure `STILL_IMAGE_CAMERA` intent.
When the CI emulator's swipe keyguard was still up, the activity launched behind the lock screen, so the screenshot captured the gray (`#E9E8EF`) lock screen instead of the green camera View, yielding 0.0% green coverage.
The original `setUp()` only sent `KEYCODE_WAKEUP`, which turns the display on but does not dismiss the swipe keyguard, and it only ran once, so the screen could re-sleep or re-lock before a later launch.

## Fix

Add a `wakeAndDismissKeyguard()` helper in `E2EFixture` that sends `KEYCODE_WAKEUP` followed by an upward swipe (the same wake + swipe sequence the CI pre-flight smoke check uses in `build.yml`; `wm dismiss-keyguard` did not reliably dismiss the swipe-type lock screen on the API-35 CI emulator).
An upward swipe is preferred over `KEYCODE_MENU`, which can open a foreground activity's options menu and obscure the camera View.
The helper is called on every `launchPixelCamera()` as well as in `setUp()`, since the screen can re-sleep or re-lock between `setUp()` and a later test's launch. When the screen is already unlocked the swipe is a harmless gesture over the foreground activity.

This replaces the temporary mitigation in #337, which only allowed the test to fail without blocking CI rather than fixing the underlying race.

## Verification

`./gradlew :app:compileDebugAndroidTestKotlin` compiles cleanly.

---
_Generated by [Claude Code](https://claude.ai/code/session_011KAycJTfAncDeiLZ1B6gAT)_

PR #352, opened 2026-06-09
Apply wake+swipe keyguard-dismiss fix to all CI test-launch sites

### LINKED ISSUE #344: Apply 341 unlock fix more broadly in CI
- Look for more instances where the fix in #341 may be appropriate. I don’t want to keep adding it one place at a time.

aunger:
Fixes #344.

The fix from #341 (KEYCODE_WAKEUP + upward swipe to dismiss the swipe-type lock screen) is now applied to every CI step that starts a test suite or the emulator.

## Changes

**`build.yml` -- "Wait for emulator service readiness":** Replace the old `input keyevent 82` (KEYCODE_MENU) call with `sleep 1; input swipe 300 1000 300 300; sleep 1`, matching the pre-flight smoke check step and `E2EFixture.wakeAndDismissKeyguard()`.
KEYCODE_MENU could not reliably dismiss the swipe-type lock screen on the API-35 CI emulator (root cause documented in #341).

**`build.yml` -- "Run instrumented tests":** Add ADB variable + wake + swipe before the Gradle invocation.

**`build.yml` -- "Run PixelCameraOverlayE2ETest":** Add wake + swipe before the Gradle invocation.

**`build.yml` -- "Run GalleryButtonVisualE2ETest":** Add wake + swipe before the Gradle invocation.

## Why each site

Each E2E test step can be preceded by 10-30 s of other CI work (mock-camera install, previous test suite, etc.) during which the emulator screen may sleep or re-lock.
The pre-step unlock prevents the first `launchPixelCamera()` in each suite from landing behind the lock screen and capturing the gray (`#E9E8EF`) keyguard instead of the green camera View.

The Kotlin-side fix (`E2EFixture.wakeAndDismissKeyguard()` called in `setUp()` and `launchPixelCamera()`) was already applied in #341.
This PR closes the remaining gap at the CI-shell level.

## Test plan

- [ ] CI passes on this PR (all three E2E suites run after a wake+swipe unlock).
- [ ] Python unit tests: `python3 -m unittest discover -s scripts -p "test_*.py"` -- 208 tests pass (verified locally).

https://claude.ai/code/session_017M4NzZ1oJW4g5jTDMDVCDM

---
_Generated by [Claude Code](https://claude.ai/code/session_017M4NzZ1oJW4g5jTDMDVCDM)_

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -222,10 +222,13 @@ jobs:
           echo "=== $(date -u '+%Y-%m-%d %H:%M:%S UTC') === Services ready (${SVCWAIT}s after boot flag)."
 
           # Wake, unlock, keep on, and disable animations now that services are up.
-          # KEYCODE_WAKEUP (224) is more reliable than KEYCODE_MENU (82) for waking
-          # the display from sleep; KEYCODE_MENU then dismisses the lock screen.
+          # Use the same wake + upward-swipe sequence as the pre-flight step and
+          # E2EFixture.wakeAndDismissKeyguard(): KEYCODE_MENU (82) could not reliably
+          # dismiss the swipe-type lock screen on the API-35 CI emulator (issue #341).
           "$ADB" shell input keyevent 224
-          "$ADB" shell input keyevent 82
+          sleep 1
+          "$ADB" shell input swipe 300 1000 300 300
+          sleep 1
           # Prevent the screen from sleeping while the emulator is "plugged in"
           # (value 7 = stay on for AC/USB/wireless charging combined).
           # Without this the screen can time out between the unlock here and the
@@ -328,6 +331,15 @@ jobs:
         id: run_instrumented
         if: steps.diff_check.outputs.needs_full_build == 'true'
         run: |
+          ADB="$ANDROID_HOME/platform-tools/adb"
+          # Wake and dismiss the swipe keyguard before starting the test suite.
+          # The E2E setup (including mock-camera install) can take 10-30 s, and the
+          # screen may sleep or re-lock in that window. This mirrors
+          # E2EFixture.wakeAndDismissKeyguard() (issue #344 / #341).
+          "$ADB" shell input keyevent 224
+          sleep 1
+          "$ADB" shell input swipe 300 1000 300 300
+          sleep 1
           if ./gradlew connectedDebugAndroidTest; then
             echo "outcome=success" >> "$GITHUB_OUTPUT"
           else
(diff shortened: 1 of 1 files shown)


aunger (commented):
**Review verdict: Changes requested**

The coverage is complete at the CI shell level--all four sites that start a test suite or the emulator now have a wake+swipe unlock. The `ADB` variable addition for "Run instrumented tests" is correct and matches the pattern in the other steps. Comments are clear and accurate.

One issue requires a change before merge.

---

**Inconsistent (and likely insufficient) sleep after the swipe**

The PR description says the new sites "mirror" the pre-flight smoke check and `E2EFixture.wakeAndDismissKeyguard()`. That is not quite true:

- Pre-flight step: `sleep 2` -- with an explicit comment: "sleep 2 lets the swipe animation complete before am start -W fires"
- `E2EFixture.wakeAndDismissKeyguard()`: 300 ms Kotlin sleep (tight loop inside a test process, not between a shell command and a Gradle invocation)
- New additions in this PR: `sleep 1`

The pre-flight step uses 2 seconds deliberately--it was tuned to give the swipe animation time to complete before the `am start -W` fires. The new additions immediately follow the swipe with a `./gradlew` invocation that will itself issue `am start`-style commands at the start of each test suite. The same animation window applies. Using `sleep 1` instead of `sleep 2` may be sufficient, but it is an untested assumption and contradicts the reasoning documented in the existing pre-flight comment.

If the intent is to match the pre-flight step, use `sleep 2` at all four new/changed sites. If `sleep 1` has been shown to work, that reasoning should be documented and the pre-flight step's `sleep 2` should be brought down to match--otherwise future readers will be confused about which value is authoritative.

The "Wait for emulator service readiness" step (also changed in this PR) has a smaller exposure because `stay_on_while_plugged_in 7` is set immediately after, but the swipe animation window concern still applies if any visible activity is being checked afterward.

---

**Nit (not blocking)**

`keyevent 224` (numeric) is used in the changed and new sites; the pre-flight step uses `KEYCODE_WAKEUP` (symbolic). Both produce identical results, but using one form everywhere would prevent any future confusion about whether the two sites do the same thing.

---

aunger:
Both points accepted and addressed in 183bbf2.

**sleep 2:** Agreed -- the pre-flight step's `sleep 2` comment ("lets the swipe animation complete before am start -W fires") applies equally to the new sites, since Gradle also issues `am start`-style commands at the start of each suite. Using `sleep 1` was an untested assumption. All four new/changed sites now use `sleep 2`, and each has a comment citing the same reasoning as the pre-flight step.

**KEYCODE_WAKEUP (nit):** Also changed. All sites now use the symbolic form to match the pre-flight step.

---

aunger:
**Review verdict: LGTM**

The prior review's two requests have been addressed correctly in 183bbf2.

**sleep 2:** All four new/changed sites now use `sleep 2`, consistent with the pre-flight step and its documented rationale. The comments at each new site explicitly cite the same reasoning ("sleep 2 lets the swipe animation complete before Gradle's am start fires (same reasoning as the pre-flight step's post-swipe sleep 2)").

**KEYCODE_WAKEUP (symbolic form):** All live unlock lines use the symbolic form. The only remaining `keyevent 82` or `keyevent 224` text is in the comment at line 25, explaining why the project does not use `reactivecircus/android-emulator-runner@v2`--correct to leave as-is.

**Correctness of the change:** The old "Wait for emulator service readiness" pair (`keyevent 224` + `keyevent 82`) is correctly replaced. The swipe is the right replacement for the MENU-key dismiss on the swipe-type lock screen, which is the root cause documented in #341. The `ADB` variable addition in the "Run instrumented tests" step is correct and matches the pattern used in the other E2E steps.

**Scope:** The PR does exactly what #344 asked--no more, no less.

CI is still running at time of review; nothing in the completed checks raises a concern.

---

PR #27, opened 2026-04-17
Update CI actions to Node 24 runtime (fixes #22)

aunger:
Bumps all GitHub Actions to versions using Node 24 runtime,
replacing Node 20 which is deprecated on GitHub Actions runners.

- actions/checkout v4 → v6
- actions/setup-java v4 → v5
- android-actions/setup-android v3 → v4
- actions/cache v4 → v5
- actions/upload-artifact v4 → v7
- softprops/action-gh-release v2 → v3

https://claude.ai/code/session_01DBb2FA8pXTtAaWGXqr4cM5

PR #362, opened 2026-06-11
Retry first Pixel Camera launch through window-transition teardown on slow emulators

### LINKED ISSUE #233: [PixelCameraOverlayE2ETest] overlayAppearsWhenViewfinderOpens
- See also #186 

---

## Test Failure

| Field | Value |
|-------|-------|
| Method | `overlayAppearsWhenViewfinderOpens` |
| Branch | _unknown_ |
| Commit | `4855c4b` |

### Failure message

java.lang.AssertionError: Overlay should appear within 30 s of launching Pixel Camera viewfinder

### Stack trace

```
java.lang.AssertionError: Overlay should appear within 30 s of launching Pixel Camera viewfinder
	at com.gb4pc.e2e.PixelCameraOverlayE2ETest.overlayAppearsWhenViewfinderOpens(PixelCameraOverlayE2ETest.kt:62)
```

### Links

- _unknown_

---
_Filed automatically by CI on failure of `com.gb4pc.e2e.PixelCameraOverlayE2ETest.overlayAppearsWhenViewfinderOpens`._

aunger:
Fixes #233.

> Note: this PR was re-scoped after a review on the previous (UsageStats-lag) approach.
> The earlier `OverlayServiceLogic` retry change has been dropped because CI logs disproved its premise; see the [investigation comment](https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4678088778).
> This PR now contains an E2E-harness fix only.

## Problem

`PixelCameraOverlayE2ETest.overlayAppearsWhenViewfinderOpens` has failed persistently in CI (26+ recurrences across `main` and many branches): the overlay never becomes active within 30 s of launching Pixel Camera.

## Root cause

Confirmed from CI logcat (run `27109982390`, `main` @ `6b57c39`, job `80005861193`, step "Run PixelCameraOverlayE2ETest"). `overlayAppearsWhenViewfinderOpens` is the **first** test in the suite to launch the camera, and its launch is torn down inside its own OPEN window-transition, ~3 ms after START, before the activity reaches `onResume()`:

```
00:56:14.274  ActivityTaskManager: START ... MockCameraActivity ... result code=0
00:56:14.277  WindowManager: Force removing ActivityRecord{... MockCameraActivity t34}
00:56:14.374  WindowManager: Attempted to add application window with unknown token
              ...MockCameraActivity t-1 f. Aborting.
00:56:44.253  UiDevice: Taking screenshot ..._overlayAppearsWhenViewfinderOpens_failure.png
00:56:45.124  TestRunner: failed: overlayAppearsWhenViewfinderOpens
```

`MockCameraActivity` only calls `openCamera()` from `onResume()`. On this launch the activity never resumes, so:

- there is **no `CameraService::connect`** and **no `CameraDevice.onOpened`** anywhere in the 30 s window (the first `CameraService::connect` is at `00:56:47.463`, in the *next* test);
- `CameraManager.AvailabilityCallback.onCameraUnavailable` never fires;
- `evaluateForeground()` is never entered and the overlay never activates.

This is a launch-transition race in the E2E harness, **not** an `OverlayServiceLogic` timing/UsageStats-lag bug.
Corroboration: the very next test, `overlayAppearsAfterUsageStatsLag`, relaunches the camera and gets a clean `connect` -> `onOpened` (~30 ms) and **passes in ~2.4 s**; every later launch in the run is likewise healthy.
Only the *first* launch is killed before the camera opens.

## Fix

Make `E2EFixture.launchPixelCamera()` resilient to the first-launch teardown: re-issue the `am start` (up to 3 attempts) until `OverlayService.isOverlayActive` becomes true, which is the observable proof the activity reached `onResume()` and opened the camera.

To keep that proof sound, the verify windows are tiered rather than flat:

- A bounded `LAUNCH_BASELINE_MS` (3 s) wait first establishes a known-inactive baseline, so each attempt observes a genuine false -> true transition and a stale `true` left over from a previous test cannot masquerade as confirmation of the new launch.
- The **first** attempt gets `LAUNCH_FIRST_VERIFY_MS` (12 s), inside the 20 s budget `waitForOverlayActive()` already deems legitimate for a slow-but-healthy activation (~9 s camera open + ~1 s UsageStats). This avoids false-positiving a healthy launch and re-issuing `am start` over a mid-resume activity (which would itself race the OPEN transition).
- Follow-up attempts use the shorter `LAUNCH_RETRY_VERIFY_MS` (6 s), since recovery from the teardown failure mode is fast (every observed non-first launch opens the camera in ~30 ms).

The retry is bounded so a genuinely broken launch still fails the caller's own activation assertion rather than hanging.
The worst case for a genuinely broken launch is `3 + 12 + 2 x 6 = 27 s`, under the 30 s assertion, so `launchPixelCamera()` returns and lets the caller's own assertion fail rather than hang.
On a healthy launch it returns as soon as the overlay activates.

No production code is changed and no existing test requirement is loosened.

## Diagnostic logging

`launchPixelCamera()` now emits `Log.i`/`Log.w` lines (tag `GB4PC_E2E`) on each `am start` attempt, on the attempt the overlay activates on, on a retry after a suspected teardown, and on exhaustion.
This is a direct response to the review request for a run whose logcat *shows the teardown race occurring on the first launch and the retry recovering from it*: with this in place, when #233 recurs in CI the logcat will explicitly state which attempt re-issued the launch and which one the overlay activated on, instead of leaving the retry/recovery to be inferred by grepping for the absence of `CameraService::connect`.
This is a test-harness-only change; no production code path is touched.

## Tests and validation status

The fix targets the E2E harness; the affected assertion (`overlayAppearsWhenViewfinderOpens`) runs only in CI on the emulator.
`app:testDebugUnitTest` is unchanged and passes locally; `app:compileDebugAndroidTestKotlin` compiles cleanly.

**The CI evidence does not yet positively confirm the retry path resolves the failure.**
The prior green run (`27331152010`, HEAD `ae71ead`) had its first camera launch succeed immediately on attempt 1 (`onOpened` ~0.9 s after START, no force-removal of the launch's own `ActivityRecord`), so the `repeat(LAUNCH_ATTEMPTS)` retry branches were never exercised, per the [latest review](https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#pullrequestreview-4475731786). The `Force removing ... t36` line in that run is unrelated debris from a previous step's APK reinstall, not the OPEN-transition race this PR targets.

Because #233 is intermittent (it does not reproduce on every run), a single green run cannot by itself validate a race-condition fix: a green result is the expected outcome whether or not the fix works, when the race simply did not trigger.
What this PR can stand on:

- The root-cause logcat evidence (run `27109982390`) that establishes the first-launch teardown as the failure mode.
- The reasoning in the companion-object comment for why the tiered verify windows are correct (no false-positive on a healthy launch; bounded so a broken launch fails the caller's assertion rather than hanging; budget `3 + 12 + 2 x 6 = 27 s` < 30 s).
- The new `GB4PC_E2E` diagnostic logging, which turns the *next* recurrence of #233 in CI into the positive confirmation the reviewer asked for (it will show the teardown race on the first launch and the retry recovering), rather than relying on catching the intermittent race in a single re-run.

The change cannot manufacture the intermittent teardown race on demand, so the conclusive logcat (race + recovery on the first launch) will arrive on the next CI run that happens to trigger it; the diagnostic logging is what makes that run self-evident when it occurs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -381,17 +381,26 @@ jobs:
             echo "outcome=success" >> "$GITHUB_OUTPUT"
           else
             echo "outcome=failure" >> "$GITHUB_OUTPUT"
-            echo "=== LOGCAT (since test start) ==="
-            "$ADB" logcat -d -T "$LOGCAT_SINCE" | \
-              bash scripts/filter_logcat.sh
           fi
+          # Capture the filtered logcat regardless of pass/fail (issues #364/#365).
+          # The GB4PC_E2E diagnostic lines from E2EFixture.launchPixelCamera() are
+          # what make the issue #233 bounded-relaunch path observable, but they are
+          # only useful if they survive a *green* run too -- on a pass the retry
+          # either fired and recovered or never fired, and only the logcat can tell
+          # the two apart. Dumping it only in the failure branch discarded that
+          # evidence on every passing run, so always tee it to a file and upload it.
+          echo "=== LOGCAT (since test start) ==="
+          "$ADB" logcat -d -T "$LOGCAT_SINCE" | \
+            bash scripts/filter_logcat.sh | tee results/e2e-overlay-logcat.txt
 
       - name: Upload overlay E2E markers
         if: always() && steps.diff_check.outputs.needs_full_build == 'true'
         uses: actions/upload-artifact@v7
         with:
           name: testresults-e2e-overlay
-          path: results/e2e-overlay.ndjson
+          path: |
+            results/e2e-overlay.ndjson
+            results/e2e-overlay-logcat.txt
           retention-days: 14
 
       - name: Run GalleryButtonVisualE2ETest
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -52,6 +52,15 @@ class E2EFixture(
 ) {
     private val pcPackage = Constants.PIXEL_CAMERA_PACKAGE
 
+    companion object {
+        // Issue #233: bounded relaunch for the first-launch teardown race in launchPixelCamera().
+        // Up to LAUNCH_ATTEMPTS launches, each given LAUNCH_VERIFY_MS to activate the overlay
+        // before re-issuing. 3 x 5 s = 15 s worst case, comfortably under the 30 s assertion in
+        // overlayAppearsWhenViewfinderOpens (and the 20 s default in waitForOverlayActive).
+        private const val LAUNCH_ATTEMPTS = 3
+        private const val LAUNCH_VERIFY_MS = 5_000L
+    }
+
     /**
      * Run from `@Before`. Performs all preconditions; tests that pass `setUp()` are
      * guaranteed: Pixel Camera installed, OverlayService running and registered, PC
@@ -81,9 +90,33 @@ class E2EFixture(
         // re-lock between setUp() and a later test's launch (and between CI steps), so the
         // dismissal must run on every launch, not only once in setUp().
         wakeAndDismissKeyguard()
-        uiAutomation.executeShellCommand(
-            "am start -a android.media.action.STILL_IMAGE_CAMERA -p $pcPackage"
-        ).close()
+
+        // Issue #233: On the API-35 CI emulator the *first* STILL_IMAGE_CAMERA launch of a
+        // suite is frequently torn down inside its own OPEN window-transition before the
+        // activity reaches onResume(). WindowManager force-removes the just-created
+        // ActivityRecord ("Force removing ActivityRecord{... MockCameraActivity}" / "Attempted
+        // to add application window with unknown token ... Aborting"), so MockCameraActivity
+        // never calls openCamera(), CameraManager.AvailabilityCallback.onCameraUnavailable
+        // never fires, and the overlay never activates -- the test then times out at 30 s.
+        // A re-issued `am start` after the transition has settled reliably brings the activity
+        // up (every non-first launch in CI opens the camera within ~30 ms), so retry the launch
+        // until the overlay activates, which is the observable proof the camera reached
+        // onResume() and opened. The retry is bounded so a genuinely broken launch still fails
+        // the caller's own activation assertion rather than hanging.
+        repeat(LAUNCH_ATTEMPTS) { attempt ->
+            uiAutomation.executeShellCommand(
+                "am start -a android.media.action.STILL_IMAGE_CAMERA -p $pcPackage"
+            ).close()
+            // Give this launch its own short window to reach onResume() and activate the
+            // overlay before deciding whether to re-issue. The sum across attempts stays well
+            // under each caller's activation timeout.
+            if (waitForCondition(LAUNCH_VERIFY_MS) { OverlayService.isOverlayActive }) return
+            if (attempt < LAUNCH_ATTEMPTS - 1) {
+                // The previous launch was torn down before the camera opened. Re-dismiss the
+                // keyguard (the screen can re-sleep between attempts) and try again.
+                wakeAndDismissKeyguard()
+            }
+        }
     }
 
     /**
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -55,8 +55,11 @@ class E2EFixture(
     companion object {
         // Issue #233: bounded relaunch for the first-launch teardown race in launchPixelCamera().
         // Up to LAUNCH_ATTEMPTS launches, each given LAUNCH_VERIFY_MS to activate the overlay
-        // before re-issuing. 3 x 5 s = 15 s worst case, comfortably under the 30 s assertion in
-        // overlayAppearsWhenViewfinderOpens (and the 20 s default in waitForOverlayActive).
+        // before re-issuing. The launch attempts are 3 x 5 s = 15 s worst case. A genuinely
+        // broken launch returns from launchPixelCamera() after that budget (the baseline wait
+        // below only consumes time when a prior overlay is still active, which is itself a
+        // separate failure and not the healthy path), leaving the 30 s assertion in
+        // overlayAppearsWhenViewfinderOpens to fail the test rather than hang.
         private const val LAUNCH_ATTEMPTS = 3
         private const val LAUNCH_VERIFY_MS = 5_000L
     }
@@ -103,6 +106,15 @@ class E2EFixture(
         // until the overlay activates, which is the observable proof the camera reached
         // onResume() and opened. The retry is bounded so a genuinely broken launch still fails
         // the caller's own activation assertion rather than hanging.
+        //
+        // The activation check below treats isOverlayActive becoming true as proof that *this*
+        // launch reached onResume(). That proof is only sound from a known-inactive baseline: if
+        // a previous test's overlay is still active when launchPixelCamera() is called (setUp()'s
+        // waitForOverlayInactive() and the mid-test goHome()/stopPixelCamera() flows do not assert
+        // deactivation), waitForCondition would read the stale true on its first poll and return
+        // before this launch's `am start` had any effect. Wait for a clean inactive baseline first
+        // so each attempt observes a genuine false -> true transition. Bounded so it cannot hang.
+        waitForCondition(LAUNCH_VERIFY_MS) { !OverlayService.isOverlayActive }
         repeat(LAUNCH_ATTEMPTS) { attempt ->
             uiAutomation.executeShellCommand(
                 "am start -a android.media.action.STILL_IMAGE_CAMERA -p $pcPackage"
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -54,14 +54,31 @@ class E2EFixture(
 
     companion object {
         // Issue #233: bounded relaunch for the first-launch teardown race in launchPixelCamera().
-        // Up to LAUNCH_ATTEMPTS launches, each given LAUNCH_VERIFY_MS to activate the overlay
-        // before re-issuing. The launch attempts are 3 x 5 s = 15 s worst case. A genuinely
-        // broken launch returns from launchPixelCamera() after that budget (the baseline wait
-        // below only consumes time when a prior overlay is still active, which is itself a
-        // separate failure and not the healthy path), leaving the 30 s assertion in
-        // overlayAppearsWhenViewfinderOpens to fail the test rather than hang.
+        //
+        // The per-attempt verify windows must not false-positive a *healthy* launch as a failed
+        // one. waitForOverlayActive() documents that on the CI emulator a healthy activation can
+        // take ~9 s (camera open) + ~1 s (UsageStats) and budgets 20 s for it. So the first
+        // attempt is given a window in that range (LAUNCH_FIRST_VERIFY_MS): a slow-but-healthy
+        // first launch activates the overlay inside that window and returns without the loop ever
+        // re-issuing `am start`. Re-issuing on a healthy, mid-resume activity would itself race the
+        // OPEN transition (a second `input swipe` + `am start` over the resuming activity), i.e.
+        // re-create the very failure this fix targets, so the first window deliberately errs long.
+        //
+        // Only the *teardown* failure mode (activity force-removed ~3 ms after START, before
+        // onResume(), so the camera never opens and the overlay can never activate) survives that
+        // first window. Recovery from it is fast: every observed non-first launch in CI opens the
+        // camera within ~30 ms, so the follow-up attempts use the shorter LAUNCH_RETRY_VERIFY_MS.
+        //
+        // Worst-case budget for a genuinely broken launch:
+        //   baseline inactive wait (LAUNCH_BASELINE_MS, only consumed if a prior overlay is still
+        //   active going in -- itself a separate failure, not the healthy path)
+        //   + LAUNCH_FIRST_VERIFY_MS + (LAUNCH_ATTEMPTS - 1) x LAUNCH_RETRY_VERIFY_MS
+        //   = 3 + 12 + 2 x 6 = 27 s, under the 30 s overlayAppearsWhenViewfinderOpens assertion,
+        // so launchPixelCamera() returns and lets the caller's own assertion fail rather than hang.
         private const val LAUNCH_ATTEMPTS = 3
-        private const val LAUNCH_VERIFY_MS = 5_000L
+        private const val LAUNCH_FIRST_VERIFY_MS = 12_000L
+        private const val LAUNCH_RETRY_VERIFY_MS = 6_000L
+        private const val LAUNCH_BASELINE_MS = 3_000L
     }
 
     /**
@@ -114,15 +131,18 @@ class E2EFixture(
         // deactivation), waitForCondition would read the stale true on its first poll and return
         // before this launch's `am start` had any effect. Wait for a clean inactive baseline first
         // so each attempt observes a genuine false -> true transition. Bounded so it cannot hang.
-        waitForCondition(LAUNCH_VERIFY_MS) { !OverlayService.isOverlayActive }
+        waitForCondition(LAUNCH_BASELINE_MS) { !OverlayService.isOverlayActive }
         repeat(LAUNCH_ATTEMPTS) { attempt ->
             uiAutomation.executeShellCommand(
                 "am start -a android.media.action.STILL_IMAGE_CAMERA -p $pcPackage"
             ).close()
-            // Give this launch its own short window to reach onResume() and activate the
-            // overlay before deciding whether to re-issue. The sum across attempts stays well
-            // under each caller's activation timeout.
-            if (waitForCondition(LAUNCH_VERIFY_MS) { OverlayService.isOverlayActive }) return
+            // The first attempt gets the full healthy-activation window so a slow-but-healthy
+            // launch is never mistaken for a failed one and re-issued (which would itself race the
+            // OPEN transition). Only the teardown failure mode survives that window, and recovery
+            // from it is fast, so later attempts use the shorter retry window. See the
+            // companion-object note for the budget rationale.
+            val verifyMs = if (attempt == 0) LAUNCH_FIRST_VERIFY_MS else LAUNCH_RETRY_VERIFY_MS
+            if (waitForCondition(verifyMs) { OverlayService.isOverlayActive }) return
             if (attempt < LAUNCH_ATTEMPTS - 1) {
                 // The previous launch was torn down before the camera opened. Re-dismiss the
                 // keyguard (the screen can re-sleep between attempts) and try again.
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -2,7 +2,6 @@ package com.gb4pc.e2e
 
 import android.app.KeyguardManager
 import android.app.UiAutomation
-import android.util.Log
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
@@ -12,6 +11,7 @@ import android.media.MediaScannerConnection
 import android.os.Build
 import android.os.Environment
 import android.provider.MediaStore
+import android.util.Log
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.uiautomator.UiDevice
 import com.gb4pc.Constants
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -2,6 +2,7 @@ package com.gb4pc.e2e
 
 import android.app.KeyguardManager
 import android.app.UiAutomation
+import android.util.Log
 import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
@@ -53,6 +54,12 @@ class E2EFixture(
     private val pcPackage = Constants.PIXEL_CAMERA_PACKAGE
 
     companion object {
+        // Tag for the diagnostic logcat lines emitted by launchPixelCamera() (issue #233). These
+        // make the bounded-relaunch path observable in a CI run's logcat: when the first-launch
+        // teardown race occurs, the log shows which attempt re-issued `am start` and which one the
+        // overlay finally activated on, instead of leaving the retry/recovery to be inferred.
+        private const val TAG = "GB4PC_E2E"
+
         // Issue #233: bounded relaunch for the first-launch teardown race in launchPixelCamera().
         //
         // The per-attempt verify windows must not false-positive a *healthy* launch as a failed
@@ -133,6 +140,7 @@ class E2EFixture(
         // so each attempt observes a genuine false -> true transition. Bounded so it cannot hang.
         waitForCondition(LAUNCH_BASELINE_MS) { !OverlayService.isOverlayActive }
         repeat(LAUNCH_ATTEMPTS) { attempt ->
+            Log.i(TAG, "launchPixelCamera: am start attempt ${attempt + 1}/$LAUNCH_ATTEMPTS")
             uiAutomation.executeShellCommand(
                 "am start -a android.media.action.STILL_IMAGE_CAMERA -p $pcPackage"
             ).close()
@@ -142,11 +150,25 @@ class E2EFixture(
             // from it is fast, so later attempts use the shorter retry window. See the
             // companion-object note for the budget rationale.
             val verifyMs = if (attempt == 0) LAUNCH_FIRST_VERIFY_MS else LAUNCH_RETRY_VERIFY_MS
-            if (waitForCondition(verifyMs) { OverlayService.isOverlayActive }) return
+            if (waitForCondition(verifyMs) { OverlayService.isOverlayActive }) {
+                Log.i(TAG, "launchPixelCamera: overlay active on attempt ${attempt + 1}")
+                return
+            }
             if (attempt < LAUNCH_ATTEMPTS - 1) {
                 // The previous launch was torn down before the camera opened. Re-dismiss the
                 // keyguard (the screen can re-sleep between attempts) and try again.
+                Log.w(
+                    TAG,
+                    "launchPixelCamera: overlay still inactive after ${verifyMs} ms on attempt " +
+                        "${attempt + 1} (first-launch teardown race); re-issuing am start"
+                )
                 wakeAndDismissKeyguard()
+            } else {
+                Log.w(
+                    TAG,
+                    "launchPixelCamera: overlay still inactive after $LAUNCH_ATTEMPTS attempts; " +
+                        "letting the caller's own assertion fail"
+                )
             }
         }
     }
--- a/scripts/filter_logcat.sh
+++ b/scripts/filter_logcat.sh
@@ -10,8 +10,10 @@
 #      suppressed.
 #   2. Lines are filtered to only those relevant for CI failure diagnosis:
 #      AndroidRuntime, FATAL, "Process crashed", app/service tags
-#      (com.gb4pc, OverlayService, MockCamera, CameraService), and any
-#      logcat error tag (E/<tag>).
+#      (com.gb4pc, OverlayService, MockCamera, CameraService), the E2E
+#      harness diagnostic tag (GB4PC_E2E, emitted by
+#      E2EFixture.launchPixelCamera() to make the issue #233 bounded-relaunch
+#      path observable), and any logcat error tag (E/<tag>).
 #
 # Usage:
 #   adb logcat -d | scripts/filter_logcat.sh
@@ -22,4 +24,4 @@
 #   bitmap_url=data:image/png;base64,ABC= →  bitmap_url=data:image/png;base64,[elided]
 
 sed 's/;base64,[A-Za-z0-9+/=]\{8,\}/;base64,[elided]/g' | \
-  grep -E "AndroidRuntime|FATAL|Process crashed|com\.gb4pc|OverlayService|MockCamera|CameraService|E/\w" || true
+  grep -E "AndroidRuntime|FATAL|Process crashed|com\.gb4pc|OverlayService|MockCamera|CameraService|GB4PC_E2E|E/\w" || true
--- a/scripts/test_filter_logcat.sh
+++ b/scripts/test_filter_logcat.sh
@@ -136,6 +136,7 @@ KEPT_CASES["com.gb4pc"]="05-25 14:08:16.131  1204  1204 D com.gb4pc.app: some me
 KEPT_CASES["OverlayService"]="05-25 14:08:16.131  1204  1204 D OverlayService: some message"
 KEPT_CASES["MockCamera"]="05-25 14:08:16.131  1204  1204 D MockCamera: some message"
 KEPT_CASES["CameraService"]="05-25 14:08:16.131  1204  1204 D CameraService: some message"
+KEPT_CASES["GB4PC_E2E"]="05-25 14:08:16.131  1204  1204 I GB4PC_E2E: launchPixelCamera: am start attempt 1/3"
 KEPT_CASES["E/<tag>"]="05-25 14:08:16.131  1204  1204 E/SomeTag: some error"
 
 for LABEL in "${!KEPT_CASES[@]}"; do


aunger (commented):
Reviewed the diff and traced through the new unit tests against the modified `scheduleActivationRetry()`. The implementation itself is internally consistent: `activationRetryPending` correctly gates re-entrant scheduling during `evaluateForeground()`, `activationRetriesScheduled` is incremented/reset correctly (including on success via `cancelActivationRetry()`), and the bound at `MAX_ACTIVATION_RETRIES` terminates the chain as the new tests verify. No logic bugs found in the diff itself.

My concern is with the root-cause diagnosis, which I think needs more justification before this is likely to actually fix the recurring CI failure.

**The "several seconds of UsageStats lag" claim isn't backed by evidence in this PR.** `E2EFixture.waitForOverlayActive()` (in `app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt`) documents the previously-understood lag profile as "camera open can take ~9 s and UsageStats detection adds up to 1 s more" - i.e. ~1 s of UsageStats lag, which is exactly what the existing single retry at `ACTIVATION_RETRY_MS` (1 s) was designed to cover. This PR asserts the lag is actually "several seconds," which is a 5-10x increase over that prior understanding, but doesn't cite any logcat/DebugLog output from a failing CI run to support it.

**The failure pattern looks more like a permanent miss than a transient multi-second race.** Issue #233 has recurred 26+ times across completely unrelated branches (CodeQL workflow changes, test-result markers, foreground-app logging, etc.) over more than two weeks, always with the identical assertion failure at the 30 s mark. If `ForegroundDetector.getForegroundPackage()` is *never* seeing `GoogleCamera` as a foreground event in CI at all (e.g. the mock camera's activity doesn't generate a `MOVE_TO_FOREGROUND`/`ACTIVITY_RESUMED` event UsageStats reports, or a permission/appops issue on the CI emulator), then this fix changes nothing: `evaluateForeground()` will keep returning early via the `!isPixelCamera` branch on every one of the 15 retries, the overlay never activates, and the E2E test still times out at 30 s - just as it has every time so far.

**PR #325 (merged 3 days ago) added exactly the diagnostic logging needed to distinguish these two cases** (`all FG apps=` in the `ForegroundDetector` summary line). Has anyone pulled the `e2e-test-results` artifact from a recent failed run (e.g. the June 8 run on `main`, commit `6b57c39`) and confirmed that `GoogleCamera` does eventually appear in `allForegroundPackages` within the UsageStats window, just a few seconds later than the camera-unavailable callback? Without that confirmation, I'd want to see this fix validated against a real failure log before merging - otherwise we may be looking at recurrence #27 with the same assertion failure on this branch's own CI run.

Smaller note: the PR description's claim that "16 retries x ACTIVATION_RETRY_MS gives a 15 s window" (here paraphrased as "15 retries x ACTIVATION_RETRY_MS gives a ~15 s activation window" per the `Constants.kt` comment) is roughly right, but worth double-checking against the E2E test's 30 s budget once camera-open time (~9 s per the existing comment) is also accounted for - 9 s open + 15 s retry window = 24 s, leaving only ~6 s of margin in the 30 s test, which is tight if camera-open is ever slower than the documented ~9 s.

---

aunger:
You were right to push on the root cause. I pulled the logcat from a recent failing run (run 27109982390, `main` @ `6b57c39`, the June 8 failure) and traced the `Run PixelCameraOverlayE2ETest` step. The evidence contradicts the "UsageStats lag" diagnosis this PR is built on, so I do not think this change can fix #233 as written.

## What the failing run actually shows

`overlayAppearsWhenViewfinderOpens` is the **first** test to launch the camera in the suite. Its timeline:

```
00:56:12.686  TestRunner: started: overlayAppearsWhenViewfinderOpens
00:56:14.274  ActivityTaskManager: START ... MockCameraActivity ... result code=0
00:56:14.277  WindowManager: Force removing ActivityRecord{8fcca73 ... MockCameraActivity t34}
00:56:14.374  WindowManager: Attempted to add application window with unknown token
              ...ActivityRecord{8fcca73 ... MockCameraActivity t-1 f}. Aborting.
00:56:14.515  UsageStatsService: Unexpected activity event reported! (...MockCameraActivity event : 23)
00:56:14.554  ActivityManager: Start proc 4953:...GoogleCamera for next-top-activity {MockCameraActivity}
              ... (nothing) ...
00:56:44.253  UiDevice: Taking screenshot ..._overlayAppearsWhenViewfinderOpens_failure.png
00:56:45.124  TestRunner: failed: overlayAppearsWhenViewfinderOpens
```

The activity record for the first launch is **force-removed during its OPEN transition, 3 ms after START**, and the subsequent window add is aborted with an unknown token. The activity never reaches `onResume()`.

`MockCameraActivity` only calls `openCamera()` from `onResume()`. So on this launch:

- there is **no `CameraService::connect`** and **no `CameraDevice.onOpened`** anywhere in the 30 s window (I grepped the whole step: the first `CameraService::connect` is at `00:56:47.463`, which belongs to the *next* test);
- therefore `CameraManager.AvailabilityCallback.onCameraUnavailable` never fires;
- therefore `evaluateForeground()` is never entered, and the DT-06a retry loop this PR extends is never scheduled.

The retry budget can only help once `onCameraUnavailable` has fired. Here it never fires, so making the retry bounded-repeating changes nothing for this failure mode.

## Corroboration

The very next test, `overlayAppearsAfterUsageStatsLag`, relaunches the camera at `00:56:45.134`, gets a clean `CameraService::connect` → `CameraDevice.onOpened` at `00:56:47.4-5`, and **passes in ~2.4 s**. Every later launch in the run shows the same healthy `connect` → `onOpened` within ~30 ms. Only the first launch is torn down before the camera opens. This is consistent with #233 recurring on the *first* camera launch regardless of branch, and explains why it is not a timing/lag issue (a 2.4 s pass has enormous margin under the existing single 1 s retry, let alone 30 s).

## Why the first launch is killed

The force-removal at `00:56:14.277` races the launch transition. `launchPixelCamera()` runs `wakeAndDismissKeyguard()` (a `KEYCODE_WAKEUP`, a 300 ms sleep, an `input swipe`, another 300 ms sleep) and then immediately issues `am start STILL_IMAGE_CAMERA`. On the first launch the activity is created and torn down inside the same transition (`t34` → `t-1`), so `onResume()`/`openCamera()` never run. Subsequent launches do not hit this because the task already exists.

## Recommendation

This looks like an E2E-harness reliability bug, not an `OverlayServiceLogic` bug. The fix should make the first camera launch reliably reach `onResume()` (open the camera) before the 30 s wait, e.g. retry/verify the launch in `launchPixelCamera()` until `CameraDevice.onOpened` (or `onCameraUnavailable`) is observed, and/or separate the keyguard-dismiss swipe from the launch so they do not race the OPEN transition. I do not think the `OverlayServiceLogic` retry change in this PR should merge as a fix for #233, since it cannot affect a launch where the camera never opens.

Happy to repurpose this branch toward the harness fix. Flagging first so we do not land a change whose own CI run would be recurrence #27 with the identical assertion.

(Evidence: run 27109982390, job 80005861193, step "Run PixelCameraOverlayE2ETest".)

---

aunger:
Acted on the investigation above: I dropped the `OverlayServiceLogic`/`Constants` retry change (and its modified unit test) and force-pushed a harness-only fix.

`launchPixelCamera()` now re-issues `am start` up to 3 times (5 s verify window each) until `OverlayService.isOverlayActive`, which can only become true once `MockCameraActivity` has reached `onResume()` and opened the camera. This directly counters the first-launch window-transition teardown seen in the logs; it returns immediately on a healthy launch and is bounded (15 s worst case, under the 30 s assertion) so a truly broken launch still fails the caller's own assertion.

Title/description updated to match. Locally: `app:testDebugUnitTest` passes and `app:compileDebugAndroidTestKotlin` compiles; the E2E assertion itself can only be exercised by CI on the emulator, so the real validation is this branch's CI run.

---

aunger (commented):
The investigation comment is good work: tracing the logcat to the force-removed `ActivityRecord` during the OPEN transition, and corroborating with the next test's clean `connect` -> `onOpened`, makes a convincing case that this is a harness launch race rather than an `OverlayServiceLogic` timing bug. Dropping the previous approach once the premise was disproved was the right call.

One correctness concern with the retry implementation itself:

`waitForCondition` checks `condition()` immediately, before its first sleep (`E2EFixture.kt` lines 158-163). That means in `launchPixelCamera()`, if `OverlayService.isOverlayActive` happens to already be `true` at the moment the loop's first `am start` is issued, `waitForCondition(LAUNCH_VERIFY_MS) { OverlayService.isOverlayActive }` returns `true` on its very first poll and `launchPixelCamera()` returns immediately, having issued only the `am start` with no verification that *this* launch actually took effect.

`setUp()` calls `waitForOverlayInactive()` before each test, but that helper does not assert (per its own doc comment: "Tolerates the case where the overlay was never active... does not fail if the timeout expires"). So if a previous test's deactivation is still in flight when the next test's `setUp()` runs, `isOverlayActive` could still read `true` going into `launchPixelCamera()`, and the new retry/verify logic would treat that stale state as proof of success for the new launch. This is a pre-existing risk in the sense that the old code didn't verify launches at all, but the new code's framing ("the overlay activating is the observable proof the camera reached onResume()") is specifically undermined by this stale-state path, since it can short-circuit before the new `am start` has had any chance to take effect.

Worth considering: capture `OverlayService.isOverlayActive` before issuing the first `am start` and, if already `true`, either wait for it to go false first or require a state *transition* (false -> true) rather than just a true reading, so a stale flag can't masquerade as confirmation of the new launch.

Minor/non-blocking: with this change, every `launchPixelCamera()` call site (7 across the suite) now carries a worst-case 15 s retry budget on top of whatever the caller waits afterward (e.g., `test0_smokeGreenFeedVisible`'s own 30 s `waitForGreenCoverage`). On healthy launches this returns immediately, so it's only a worst-case stacking concern, but if the first-launch teardown bug ever recurs on a test that doesn't itself wait on `isOverlayActive`, that test's wall-clock time grows by up to 15 s. Probably fine given the 30 s assertion headroom, just flagging for awareness.

CI's "Build & Test" run for `bb26e0f` was still in progress at review time, so the real-world validation (whether the retry actually clears the recurring `overlayAppearsWhenViewfinderOpens` failure) isn't confirmed yet.

---

aunger:
Good catch on the stale-state short-circuit, that path does undermine the whole premise of the fix ("the overlay activating is the observable proof the camera reached onResume()"), so I have addressed it rather than treated it as pre-existing.

Fixed in `0ba85a9`: `launchPixelCamera()` now waits (bounded, `LAUNCH_VERIFY_MS`) for `OverlayService.isOverlayActive` to be false before entering the retry loop. From that known-inactive baseline, each attempt's `waitForCondition(...) { isOverlayActive }` can only succeed on a genuine false -> true transition, so a stale `true` left over from a previous test (whose deactivation `setUp()`'s `waitForOverlayInactive()` and the mid-test `goHome()`/`stopPixelCamera()` flows don't assert) can no longer masquerade as confirmation of the new launch.

On timing: the baseline wait does not change the worst-case budget for a *broken* launch. It only consumes time when a prior overlay is still active going in, which is itself a separate failure rather than the healthy path; on a healthy start the overlay is already inactive and it returns immediately. The launch-attempt budget remains 3 x 5 s = 15 s, well under the 30 s assertion. I updated the companion-object comment to state this accurately rather than implying a flat 15 s ceiling.

On the minor/non-blocking worst-case-stacking note (7 call sites each carrying up to 15 s): I agree it is only a worst-case concern and is comfortably within the 30 s assertion headroom, so I have left it as-is for this PR.

The real validation remains this branch's CI run of `overlayAppearsWhenViewfinderOpens` on the emulator. Locally `app:testDebugUnitTest` passes and `app:compileDebugAndroidTestKotlin` compiles cleanly.

---

aunger:
Reviewed `0ba85a9`. The re-scoping to a harness-only fix is reasonable, and the investigation that disproved the `OverlayServiceLogic`/UsageStats-lag premise is solid work, the logcat evidence is convincing about the actual failure mode (first-launch teardown before `onResume()`).

However, I think the fix as written has a real risk of not working, and possibly making things worse, due to an internal inconsistency in the timing assumptions.

## Main concern: `LAUNCH_VERIFY_MS = 5_000L` may be too short for a *healthy* launch, causing the retry to fire on a launch that was already succeeding

`waitForOverlayActive()` (unchanged by this PR, in the same file) documents:

> The default timeout is 20 s to accommodate slow CI emulators where camera open can take ~9 s and UsageStats detection adds up to 1 s more.

That's an existing, documented claim about how long a *healthy* activation can legitimately take on this CI emulator. But the new per-attempt verify window is only 5 s. If attempt 1's `am start` actually succeeds and the activity is on a normal (if slow, ~9-10 s) path to `onResume()` -> `openCamera()` -> overlay activation, `waitForCondition(5000) { isOverlayActive }` will time out at 5 s anyway, and the loop will re-issue `wakeAndDismissKeyguard()` (which performs an `input swipe` over the foreground activity) plus a second `am start` on top of an activity that is mid-resume.

That's the same category of problem this PR is trying to fix, just self-inflicted: a second launch racing the first launch's transition. In the worst case this could make the test fail *more* often, not less, on the slower end of the CI emulator's range.

Either:
- `LAUNCH_VERIFY_MS` needs to be long enough to not false-positive on the documented healthy-case latency (e.g., closer to the existing 20 s default used elsewhere in this file), or
- the PR needs to explain why the "~9 s camera open" figure in `waitForOverlayActive`'s doc comment doesn't apply to the first launch covered by this fix.

As written, the 5 s figure isn't justified anywhere in the PR description, the corroborating evidence cited (the *second* test's launch passing in ~2.4 s) is for a non-first launch, which by the PR's own root-cause analysis is not the problematic case.

## Secondary: new baseline wait adds latency to all 7 call sites, not just the first

`waitForCondition(LAUNCH_VERIFY_MS) { !OverlayService.isOverlayActive }` now runs unconditionally at the top of `launchPixelCamera()`, which is called from 7 places, not just `overlayAppearsWhenViewfinderOpens`. The PR's own comment acknowledges that `goHome()`/`stopPixelCamera()` don't assert deactivation, so on any test where the overlay is still deactivating when the next `launchPixelCamera()` runs, this silently adds up to 5 s of wait that didn't exist before. Probably fine given the existing timeout headroom, but it's a behavior change to every call site that the "no existing test requirement is loosened" line doesn't really cover, worth calling out explicitly even if no assertion changes.

## Nit: worst-case figure in the companion-object comment is now wrong

The companion-object comment says "The launch attempts are 3 x 5 s = 15 s worst case" and that `launchPixelCamera()` "returns from launchPixelCamera() after that budget." But the new baseline `waitForCondition(LAUNCH_VERIFY_MS) {...}` added in `0ba85a9` can itself consume up to 5 s before the loop starts, making the true worst case 20 s. The inline comment near the baseline wait gets this right ("does not change the worst-case budget for a *broken* launch" is true only because 20s < 30s assertion, but the *stated* 15 s ceiling in the companion object is no longer accurate as a description of `launchPixelCamera()`'s total worst case). The two comments should agree on the actual ceiling.

## CI status

Build status is currently "pending" on `0ba85a9`. Given this is the third attempt on this issue and the central claim of the PR ("the real validation remains this branch's CI run") is unverified, I'd want to see `overlayAppearsWhenViewfinderOpens` actually pass before merging, especially in light of the timing concern above.

Verdict: changes requested, primarily on the `LAUNCH_VERIFY_MS` vs. documented ~9-10 s healthy-activation time inconsistency.

---

aunger:
Addressed the review in `b9... ` (latest push to this branch). The core concern was correct and load-bearing, so I restructured the verify windows rather than just bumping a constant.

## Main concern: 5 s window false-positiving a healthy launch

You are right that `LAUNCH_VERIFY_MS = 5_000L` was inconsistent with `waitForOverlayActive()`'s own documented ~9 s (camera open) + ~1 s (UsageStats) healthy-activation profile, and that the corroborating ~2.4 s figure was a *non-first* launch and therefore not the case this fix covers. A 5 s window on a slow-but-healthy first launch would time out and re-issue `wakeAndDismissKeyguard()` (an `input swipe`) + `am start` over the mid-resume activity, re-creating the same OPEN-transition race, exactly the self-inflicted failure you described.

Fix: the windows are no longer flat. The **first** attempt now gets `LAUNCH_FIRST_VERIFY_MS = 12_000L`, inside the 20 s budget `waitForOverlayActive()` already deems legitimate for a healthy activation. A slow-but-healthy first launch activates the overlay within that window and returns without the loop ever re-issuing. Only the *teardown* failure mode (activity force-removed ~3 ms after START, never reaches `onResume()`, camera never opens, overlay can never activate) survives that window, and recovery from it is fast (every observed non-first CI launch opens the camera in ~30 ms), so the follow-up attempts use the shorter `LAUNCH_RETRY_VERIFY_MS = 6_000L`. The rationale is in the companion-object comment.

## Secondary: baseline wait on all 7 call sites

Kept the baseline inactive wait (it is what makes "overlay active == this launch reached onResume()" sound from a known-inactive state), but trimmed it to `LAUNCH_BASELINE_MS = 3_000L`. It still only consumes time when a prior overlay is genuinely still active going in, which is a separate failure rather than the healthy path; on a healthy start the overlay is already inactive and it returns immediately. Flagging the cross-call-site behavior change explicitly as you asked.

## Nit: worst-case figure in the companion-object comment

Corrected. The comment now states the true ceiling as `LAUNCH_BASELINE_MS + LAUNCH_FIRST_VERIFY_MS + (LAUNCH_ATTEMPTS - 1) x LAUNCH_RETRY_VERIFY_MS = 3 + 12 + 2 x 6 = 27 s`, which is the real worst case for a genuinely broken launch and stays under the 30 s `overlayAppearsWhenViewfinderOpens` assertion, so `launchPixelCamera()` returns and lets the caller's own assertion fail rather than hang. No stale "15 s" claim remains.

Locally `app:testDebugUnitTest` passes and `app:compileDebugAndroidTestKotlin` compiles cleanly. As before, the real validation is this branch's CI run of `overlayAppearsWhenViewfinderOpens` on the emulator; I agree it should be green before merge.

---

aunger (commented):
I pulled the full job log for this branch's green run (run `27331152010`, job `80743659339`, HEAD `ae71ead`) to check whether it actually exercises the new retry path, since that is the PR's stated validation. It does not, and the run does not contain the failure mode this PR targets at all.

## The first camera launch in this run succeeded immediately, on attempt 1

Searching the `Run PixelCameraOverlayE2ETest` step's logcat for the first `STILL_IMAGE_CAMERA` `START`:

```
06-11 07:37:54.339  ActivityTaskManager: START ... STILL_IMAGE_CAMERA ... MockCameraActivity ... taskId=37 ... type = OPEN
06-11 07:37:54.464  TopTaskTracker: onTaskMovedToFront: taskId=37
06-11 07:37:54.574  ActivityManager: Start proc 5432:.../MockCameraActivity
06-11 07:37:55.260  MockCameraActivity: CameraDevice.onOpened: 1
06-11 07:37:56.026  ActivityTaskManager: Displayed .../MockCameraActivity for user 0: +1s675ms
```

`onOpened` fires ~0.9 s after `START`, with no force-removal of task 37's `ActivityRecord` anywhere in between. `##GB4PC_TEST##` reports `overlayAppearsWhenViewfinderOpens` as `PASS`. This is exactly the "healthy, non-first-launch" pattern the PR describes from `overlayAppearsAfterUsageStatsLag` in the original failing run, just happening on the *first* launch this time.

`waitForCondition(LAUNCH_FIRST_VERIFY_MS) { OverlayService.isOverlayActive }` would have returned `true` on or near its first poll here. The `repeat(LAUNCH_ATTEMPTS)` loop's retry branches (`am start` re-issue, `wakeAndDismissKeyguard()` on failure, `LAUNCH_RETRY_VERIFY_MS` window) were never reached. **The retry logic added by this PR was not exercised by this CI run.**

## The "Force removing ActivityRecord ... t36" event in this run is unrelated

There is a `Force removing ActivityRecord{... MockCameraActivity t36 f}}: app died, no saved state` at `07:37:48.410`, ~6 s before the test's first `am start`. Looking at the surrounding lines, this happens in the middle of the `com.gb4pc` test APK being reinstalled (`Package [com.gb4pc] reported as REPLACED...`, `ActivityManager: Force stopping com.gb4pc.mockgallery ... installPackageLI`, etc.) — it's debris from a previous step's mock-camera task being torn down by the app reinstall, not the OPEN-transition race this PR describes for task 37. It's not part of `overlayAppearsWhenViewfinderOpens`'s own launch.

## Why this matters for merging

Issue #233 has been intermittent across 26+ runs, not 100% reproducing on every run, including some `main` runs that presumably passed in between the failures shown in the issue history. A single green run on a branch whose fix targets a race condition is exactly the kind of result you'd expect whether or not the fix works, if the race simply didn't trigger this time, which is what the logcat shows happened here. The PR description's closing claim ("This branch's latest CI run ... is green ... so the real validation is this branch's CI run") isn't supported. We don't have positive evidence the retry path is even reachable as written, let alone that it resolves the teardown failure mode when it does occur.

This is the third major iteration on this PR (UsageStats-lag retry -> harness retry -> baseline-wait fix -> tiered verify windows), each time citing a fresh CI run as validation. Given the failure is intermittent, I don't think any single green run can validate this without either (a) a run where the logcat actually shows the teardown race occurring on the first launch *and* the retry recovering from it, or (b) several consecutive green runs to build confidence the change at least doesn't regress anything, combined with reasoning about why the tiered windows are correct (which the companion-object comment does reasonably well on its own terms).

I don't see a logic bug in the retry/verify implementation itself (the budget arithmetic, `LAUNCH_ATTEMPTS=3`/`LAUNCH_FIRST_VERIFY_MS=12_000`/`LAUNCH_RETRY_VERIFY_MS=6_000`/`LAUNCH_BASELINE_MS=3_000` sums to 27 s < 30 s as documented, and the false->true baseline transition fix from the earlier review round looks correct). My objection is purely about the validation claim, given how many times this issue has recurred and how many approaches have already been tried and merged/reverted.

Verdict: changes requested. Either get a CI run whose logcat shows the actual teardown race + recovery on the first launch, or temper the PR description's validation claim and consider whether more confirmation (e.g. re-running CI a few times) is warranted before merging given the history of this issue.

---

aunger:
Addressed the changes-requested verdict from [the latest review](https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#pullrequestreview-4475731786). You confirmed there is no logic bug in the retry/verify implementation; the objection was purely about the validation claim, given the green run did not exercise the retry path and #233 is intermittent. You offered two acceptable resolutions: (a) a CI run whose logcat shows the teardown race + recovery on the first launch, or (b) temper the description's validation claim and consider more confirmation before merge.

I cannot manufacture the intermittent teardown race on demand (it does not reproduce on every run), so I have done both halves of what is actually achievable, and added a change that converts resolution (a) from "hope the next re-run happens to trigger it" into "the next run that triggers it is self-evident":

**Resolution (b), description tempered.** The "Tests and validation status" section already states the CI evidence does not positively confirm the retry path, explains why the prior green run's first launch succeeded on attempt 1 (so the retry branches were never reached), and explains why a single green run cannot validate an intermittent race fix. No stale "the real validation is this branch's green run" claim remains.

**Resolution (a), made reachable: diagnostic logging (`e8f3382`).** `launchPixelCamera()` now emits `Log.i`/`Log.w` lines under tag `GB4PC_E2E` on each `am start` attempt, on the attempt the overlay activates on, on a retry after a suspected teardown, and on exhaustion. When #233 next recurs in CI, the logcat will explicitly show which attempt re-issued the launch and which one recovered, rather than requiring it to be inferred by grepping for the absence of `CameraService::connect`. This is a test-harness-only change (uses the existing `android.util.Log` convention already present in this module's rules); no production code is touched.

I attempted to re-run this branch's Build & Test workflow to accumulate consecutive green runs, but the integration token cannot trigger re-runs (`403 Resource not accessible by integration`). A maintainer re-running CI a few times would build the confidence you noted is warranted given the recurrence history; the diagnostic logging above means whichever of those runs trips the race will be conclusive on its own.

Locally `app:testDebugUnitTest` passes and `app:compileDebugAndroidTestKotlin` compiles cleanly on `e8f3382`.

---

aunger:
Reviewed `e8f3382`. This addresses both halves of what the previous review ([`4475731786`](https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#pullrequestreview-4475731786)) asked for, and that review already confirmed there is no logic bug in the retry/verify implementation. I re-traced the implementation and the budget arithmetic against `waitForCondition` and `waitForOverlayActive`, and it holds up.

## Validation claim (the prior blocking concern): resolved

- Resolution (b) is done. The "Tests and validation status" section now states plainly that the CI evidence does not positively confirm the retry path, explains why the prior green run (`ae71ead`) never reached the retry branches (first launch opened the camera on attempt 1), and explains why a single green run can't validate an intermittent race fix. The old "the real validation is this branch's green run" framing is gone.
- Resolution (a) is made reachable. The `GB4PC_E2E` `Log.i`/`Log.w` lines mean the next CI run that actually trips the teardown race will show, in logcat, which attempt re-issued `am start` and which one the overlay activated on, rather than leaving it to be inferred from the absence of `CameraService::connect`. That is the right way to turn an intermittent race into self-evident evidence without being able to force the race on demand.

The integration-token `403` on re-runs is a real constraint, not an evasion; accumulating consecutive green runs is a maintainer action, and I agree it's worth doing given the recurrence history, but it isn't something this PR can perform itself.

## Implementation: re-verified, correct

- Tiered windows: `LAUNCH_BASELINE_MS(3) + LAUNCH_FIRST_VERIFY_MS(12) + (LAUNCH_ATTEMPTS-1)(2) x LAUNCH_RETRY_VERIFY_MS(6) = 27 s < 30 s`. Matches the companion-object comment and the code.
- The baseline `waitForCondition(LAUNCH_BASELINE_MS) { !isOverlayActive }` establishes the known-inactive baseline so each attempt observes a genuine false -> true transition; this correctly closes the stale-`true` short-circuit raised in the earlier round.
- `waitForCondition` polls every 100 ms and re-checks at the deadline, so the 12 s first window comfortably covers the documented ~9 s + ~1 s healthy activation without false-positiving and re-issuing `am start` over a mid-resume activity.
- The logging change is test-harness-only (`android.util.Log`), touches no production path, and the `repeat` loop still terminates and lets the caller's own assertion fail rather than hang.

## Nit (non-blocking)

The new `import android.util.Log` is inserted between `android.app.UiAutomation` and `android.content.BroadcastReceiver`, breaking the otherwise-alphabetical import order in this file (it sorts after `android.content.*`, not before). There's no ktlint/detekt/spotless in the build and `analyze-kotlin` is CodeQL, so this won't fail CI, but it's an avoidable inconsistency in an otherwise tidy file.

## CI

At review time, `build-and-test` and `analyze-kotlin` are still in progress on the latest commit; `strip-bylines` and `detect-changes` are green. I'm not blocking on CI per the review protocol. Note that, by this PR's own (now-correct) reasoning, a green `build-and-test` here neither confirms nor refutes the fix unless its logcat happens to show the race-and-recovery, so the `GB4PC_E2E` lines are what to check if it does go green.

Verdict: LGTM. The blocking validation-claim objection has been resolved exactly as the prior review specified, the implementation was already accepted as correct, and the only remaining item is a cosmetic import-order nit. Confidence-building re-runs are a maintainer follow-up, not a blocker on this diff.

---

aunger:
I checked the latest CI run (run [27346613576](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27346613576), `build-and-test` job 80939422700, HEAD `e8f3382`, completed 2026-06-12 01:25:46 UTC, green) against issues #364 and #365. **It satisfies neither.**

## #365 (GB4PC_E2E lines present on a green run): not satisfied

This run reproduces the exact situation #365 already documented for the prior green run (`27331152010`):

- `overlayAppearsWhenViewfinderOpens` reports `PASS` with `ms:0` via `##GB4PC_TEST##`.
- I grepped this job's full captured log for `GB4PC_E2E`: **zero matches**. None of the `Log.i`/`Log.w` lines this PR added to `launchPixelCamera()` appear anywhere in the output.
- The first `STILL_IMAGE_CAMERA` `ActivityTaskManager: START` in the captured logcat is at `01:23:51.493` (taskId=36). The failure-screenshot markers visible in this job's log (`..._failure.png`) all belong to `GalleryButtonVisualE2ETest` (tests 1a, 2a, 3a, 5a), not `PixelCameraOverlayE2ETest`.
- This is consistent with #365's hypothesis: either `PixelCameraOverlayE2ETest`'s logcat window isn't captured for a passing run, or the device-wide `-d` dump's ring buffer wrapped before the final dump. Either way, #365 asks specifically to pull the `e2e-test-results`/`instrumented-test-results` artifacts (not just the job's console log) and confirm whether the `GB4PC_E2E` lines are present there. I did not have access to those artifacts, only the job log, so this remains unconfirmed by this run.

## #364 (positive confirmation the retry recovers from the teardown race): not satisfied

#364 requires a CI run where the `GB4PC_E2E` logcat either shows (a) the teardown race occurring and the retry recovering, or (b) no race and attempt 1 succeeding outright, confirmed via the diagnostic lines. Since no `GB4PC_E2E` lines appear in this run's captured output at all, this run provides no evidence either way about which path was taken. A `PASS` with `ms:0` is exactly as consistent with "attempt 1 succeeded, diagnostic logging never fired/was never captured" as it is with "the retry fired and recovered, but the recovery evidence was lost." #364 is explicitly not closed by an absence of recurrence alone ("though inconclusive, per the PR's own reasoning").

## Net assessment

This run is another data point in the same category as the one #365 was filed against: green, `PASS`, `ms:0`, no diagnostic evidence either way. It doesn't move either blocking issue forward. Per #365's own suggested next step, what's needed is to pull the actual artifacts (not just the job console log) for a run on this HEAD and check whether `GB4PC_E2E` lines exist there; if they're absent from the artifacts too, that's itself the finding #365 asks to be understood before merge, since it would mean the new logging is either not being emitted, not being captured by the workflow's logcat step for this suite, or lost to ring-buffer wraparound regardless of pass/fail.

I did not look at correctness of the `launchPixelCamera()` retry/verify implementation itself, that was settled in the prior review rounds and is out of scope for this check. This is purely about whether #364/#365's CI-evidence requirements are met by this run.

---

aunger:
Pushed two commits (`9ef7a11`, `2e2bfd0`) that address the root cause of the missing `GB4PC_E2E` evidence flagged in the last check against #364/#365. The diagnostic lines were being emitted by `launchPixelCamera()` but discarded twice over before they could reach the captured output, which is why grepping the job log for `GB4PC_E2E` returned zero matches on both green runs.

## Two independent reasons the lines never appeared

1. **The logcat was only dumped on failure.** `9ef7a11` is unrelated to this; the bigger gap is the workflow itself. In `build.yml`, the `Run PixelCameraOverlayE2ETest` step dumped logcat only inside the `else` (failure) branch. On a *green* run the step wrote `outcome=success` and never dumped logcat at all, so there was nothing for the `GB4PC_E2E` lines to appear in regardless of filtering. This is exactly the "passing run, no diagnostic evidence either way" situation the last comment described.

2. **`filter_logcat.sh` stripped the tag even when logcat *was* dumped.** The CI logcat is piped through `scripts/filter_logcat.sh`, which keeps only an allowlist of tags (`AndroidRuntime`, `OverlayService`, `MockCamera`, `CameraService`, `com.gb4pc`, `E/<tag>`, etc.). `GB4PC_E2E` was not in that allowlist, so even a *failing* run that did dump logcat would have had the `GB4PC_E2E` lines grepped out before they reached the job log.

## What changed

- `9ef7a11` (**Keep GB4PC_E2E diagnostic lines in the filtered CI logcat**): adds `GB4PC_E2E` to the kept-tag set in `scripts/filter_logcat.sh`, updates the header comment, and adds a `KEPT_CASES["GB4PC_E2E"]` case to `scripts/test_filter_logcat.sh`. The shell test suite passes (19/19, including the new case).
- `2e2bfd0` (**Always capture and upload the E2E overlay logcat, not only on failure**): moves the filtered-logcat dump out of the failure-only branch so it runs on every outcome, tees it to `results/e2e-overlay-logcat.txt`, and adds that file to the `testresults-e2e-overlay` upload. This directly answers #365's request to be able to pull the diagnostic lines from the run's *artifacts* (not just the console log) on a green run.

Net effect: on the next CI run on this HEAD, the `GB4PC_E2E` lines will be present both in the `Run PixelCameraOverlayE2ETest` console output and in the `e2e-overlay-logcat.txt` artifact, on a pass as well as a failure. That makes #364's "which attempt re-issued the launch / which one the overlay activated on" determinable from a green run, and satisfies #365's artifact-presence check. Both changes are CI-workflow/shell-script only; no production code and no test requirement is touched.

The build.yml change keeps `set -o pipefail` semantics intact: the step's recorded `outcome` is written before the logcat dump, and `filter_logcat.sh` ends in `|| true`, so an empty grep result cannot fail the step.

---

aunger:
Reviewed `2e2bfd0` (latest two commits: `9ef7a11`, `2e2bfd0`), on top of the prior LGTM (`e8f3382`, https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4680534105).

## What changed since the LGTM

The prior LGTM was contingent on the `GB4PC_E2E` diagnostic lines actually surfacing in CI. The follow-up check (https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4686831155) found they did not, for two independent reasons, and `9ef7a11`/`2e2bfd0` fix both:

- `9ef7a11` adds `GB4PC_E2E` to `scripts/filter_logcat.sh`'s tag allowlist (previously it would have been grepped out even on a failing run) and adds a corresponding case to `scripts/test_filter_logcat.sh`.
- `2e2bfd0` moves the filtered-logcat dump in `build.yml`'s `run_overlay_e2e` step out of the failure-only branch so it runs unconditionally, tees it to `results/e2e-overlay-logcat.txt`, and adds that file to the `testresults-e2e-overlay` artifact upload.

I verified both independently:

- `bash scripts/test_filter_logcat.sh` passes 19/19, including the new `GB4PC_E2E` case.
- `.github/workflows/build.yml` parses as valid YAML.
- The `Gate on test failures` step reads `steps.run_overlay_e2e.outputs.outcome`, which is written by the `if/else` block *before* the new unconditional logcat-dump line, so the pass/fail verdict is unaffected by this change. The new pipeline ends in `filter_logcat.sh`'s `|| true` and `tee` to an already-`mkdir -p`'d directory, so under `set -o pipefail` it shouldn't spuriously fail the step (and thus shouldn't spuriously flip `outcome` to failure) even on an empty/odd logcat dump.
- `actions/upload-artifact@v7` defaults to `if-no-files-found: warn`, so even in an edge case where `e2e-overlay-logcat.txt` wasn't produced, the upload step wouldn't fail the job.

This directly closes the gap #365 identified (file is now produced and uploaded on every run, pass or fail, with the tag preserved) and gives #364 what it needs going forward (the next CI run, green or red, will have `GB4PC_E2E` lines available in `e2e-overlay-logcat.txt` to confirm whether the retry path was exercised and whether it recovered).

## Outstanding (non-blocking, carried over)

The import-order nit from the prior review (https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4680534105) is still present: `import android.util.Log` in `E2EFixture.kt` sorts after `android.content.*` alphabetically but is placed before it. Still cosmetic, still doesn't fail CI (no ktlint/detekt/spotless enforced), still not blocking.

## CI

`build-and-test` was "pending" on `2e2bfd0` at review time. Per the review protocol I'm not blocking on it, but given issues #364/#365 are specifically about what the *next* CI run's logcat artifact shows, that run (once complete) is the first real chance to act on #364/#365 using this PR's own changes.

## Verdict

LGTM. The two new commits are narrow, CI-config/shell-script-only changes that correctly fix the artifact-visibility gap #365 found, without touching the retry/verify logic that was already reviewed and accepted across four prior rounds. The implementation concerns from earlier rounds remain resolved; nothing in `9ef7a11`/`2e2bfd0` reopens them.

---

aunger:
Reviewed the current HEAD `2e2bfd0` (the two new commits `9ef7a11` and `2e2bfd0` on top of the previously-accepted `e8f3382`).

## Scope of this round

The retry/verify logic in `launchPixelCamera()` and the tiered-window budget were settled and accepted in the prior rounds (the false->true baseline transition, `LAUNCH_BASELINE_MS(3) + LAUNCH_FIRST_VERIFY_MS(12) + 2 x LAUNCH_RETRY_VERIFY_MS(6) = 27 s < 30 s`). Nothing in these two commits reopens that, so I focused on the new CI-config/shell-script changes, which are exactly the right fix for the "GB4PC_E2E lines never reached the captured output" gap.

I re-verified both halves independently:

- `9ef7a11` adds `GB4PC_E2E` to the allowlist in `scripts/filter_logcat.sh` and a matching `KEPT_CASES["GB4PC_E2E"]` case in `scripts/test_filter_logcat.sh`. `bash scripts/test_filter_logcat.sh` passes 19/19 including the new case. Without this, the tag would have been grepped out even on a failing run.
- `2e2bfd0` moves the filtered-logcat dump in `build.yml`'s `run_overlay_e2e` step out of the failure-only `else` branch so it runs unconditionally, tees it to `results/e2e-overlay-logcat.txt`, and adds that file to the `testresults-e2e-overlay` upload. This is what makes the diagnostic lines available from a green run's artifact, closing #365's request and giving #364 the evidence channel it needs.

On the step's exit-code safety (the thing that would matter most here, since the dump now runs after `outcome` is recorded): the `outcome=success`/`failure` output is written inside the `if/else` *before* the new dump, so the gate verdict can't be flipped by the dump. The final pipeline is `adb logcat | filter_logcat.sh | tee results/e2e-overlay-logcat.txt`; `filter_logcat.sh` ends in `|| true` and `tee` writes to the `mkdir -p results` directory created earlier in the step, so under `set -o pipefail` an empty or odd logcat dump won't spuriously fail the step. `upload-artifact@v7` defaults to `if-no-files-found: warn`, so a missing file won't fail the upload either. All correct.

## Outstanding nit (still unresolved, still non-blocking)

`import android.util.Log` in `E2EFixture.kt` (line 5) is placed between `android.app.UiAutomation` and `android.content.BroadcastReceiver`, but `android.util.*` sorts *after* `android.content.*` and `android.media.*`. This same import-order nit was raised in the two prior review rounds and is still present. It's cosmetic, there's no ktlint/detekt/spotless in the build so it won't fail CI, but it's an avoidable inconsistency in an otherwise alphabetically-ordered import block and trivial to fix while the branch is open. Not blocking.

## CI

Green on `2e2bfd0`: `build-and-test`, `analyze-kotlin`, `analyze-actions`, and CodeQL all succeeded. Per this PR's own (correct) reasoning, a green `build-and-test` neither confirms nor refutes the retry path unless its `e2e-overlay-logcat.txt` artifact happens to show the race-and-recovery; that artifact is now produced on every run, which is the point of these commits. The intermittent-race confidence-building re-runs remain a maintainer follow-up, not a blocker on this diff.

## Verdict

LGTM. The artifact-visibility gap that #364/#365 flagged is correctly closed by two narrow, CI-config/shell-only commits; the previously-accepted retry/verify logic is untouched; CI is green. The only open item is the cosmetic import-order nit, which I'd encourage fixing before merge but which does not warrant blocking.

---

aunger:
Reviewer check on HEAD `2e2bfd0`, focused on the #364/#365 evidence requirements (the prior rounds already settled the retry/verify logic).

## Evidence: the artifact channel is now working

Run [27391057782](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27391057782) (`build-and-test` 80948419402, HEAD `2e2bfd0`, green) is the first run produced after `9ef7a11`/`2e2bfd0`. I pulled the `testresults-e2e-overlay` artifact and `e2e-overlay-logcat.txt` now contains the `GB4PC_E2E` lines that were absent on the two prior green runs. Full breakdown in https://github.com/aunger/gallery-button-for-pixel-camera/issues/364#issuecomment-4687113585.

- #365 (lines present in the artifact on a green run): **satisfied**. Six `GB4PC_E2E` lines present, one `am start` + one `overlay active` per `launchPixelCamera()` call.
- #364 (confirm the retry actually fixes #233): **partially**. The lines show `overlayAppearsWhenViewfinderOpens` launching the camera and activating the overlay on attempt 1 with no teardown race, which is #364's scenario (b). They do **not** confirm scenario (a) (race + retry recovery), because the intermittent race did not trip on this run, so the retry branches were never exercised. #364 stays open as a watch item, as its own text anticipates.

So this run is the proof the diagnostic channel works as designed, but it is not the conclusive race+recovery run, and by this PR's own (correct) reasoning a green run without the race neither confirms nor refutes the fix.

## Blocking nit: import-order, now in its fourth round

`import android.util.Log` is at line 5 of `E2EFixture.kt`, placed between `android.app.UiAutomation` (line 4) and `android.content.BroadcastReceiver` (line 6). `android.util.*` sorts after `android.content.*`, `android.media.*`, `android.os.*`, and `android.provider.*`, so it belongs at the bottom of the `android.*` block, immediately before the `androidx.*` imports. This has been flagged as a non-blocking nit three times (https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4680534105, https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4686856503, https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4687086742) and is still unfixed. It is a one-line move that the branch has been open long enough to absorb. I'd like it cleaned up before merge rather than carried indefinitely; please move it.

## Verdict

Changes requested, on the import-order nit only. The retry/verify logic and the CI-config commits are correct and the evidence channel is proven working; the substantive #364 confirmation (race + recovery) is a maintainer/CI follow-up that this diff cannot force. Two tracking issues (#364, #365) are marked "Blocks PR #362"; #365's artifact-presence requirement is now met by this run, #364's race-recovery requirement is not yet met and remains the gating follow-up.

---

aunger:
Addressed the blocking import-order nit from [the latest review](https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4687116581) in `7d94d20`.

`import android.util.Log` was inserted between `android.app.UiAutomation` and `android.content.BroadcastReceiver` by the diagnostic-logging commit. `android.util.*` sorts after `android.content.*`, `android.media.*`, `android.os.*`, and `android.provider.*`, so it has been moved to the bottom of the `android.*` block, immediately before the `androidx.*` imports. That is the only change in this commit.

Locally `app:compileDebugAndroidTestKotlin` compiles and `app:testDebugUnitTest` passes.

On the substantive items: per the prior rounds, the retry/verify logic and CI-config commits are accepted and unchanged, and #365's artifact-presence requirement is met. #364's race-and-recovery confirmation depends on a CI run that actually trips the intermittent first-launch teardown, which this diff cannot force; the diagnostic channel that makes such a run self-evident is now in place and proven working.

---

aunger:
Reviewed HEAD `7d94d20` (the import-order move on top of the previously-accepted `2e2bfd0`).

## Blocking nit from last round: resolved

`import android.util.Log` is now at line 14 of `E2EFixture.kt`, at the bottom of the `android.*` block immediately before the `androidx.*` imports, where `android.util.*` sorts. This was the only blocking item in my prior review ([#issuecomment-4687116581](https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4687116581)), and `7d94d20` is a one-line import move with no other change, so it doesn't reopen anything.

## Re-verified, still correct

- The retry/verify logic in `launchPixelCamera()` is unchanged from the accepted rounds: known-inactive baseline via `waitForCondition(LAUNCH_BASELINE_MS) { !isOverlayActive }`, then tiered windows `LAUNCH_BASELINE_MS(3) + LAUNCH_FIRST_VERIFY_MS(12) + (LAUNCH_ATTEMPTS-1)(2) x LAUNCH_RETRY_VERIFY_MS(6) = 27 s < 30 s`. Code and companion-object comment agree.
- `bash scripts/test_filter_logcat.sh` passes 19/19, including the `GB4PC_E2E` case from `9ef7a11`.
- The `build.yml` change writes `outcome=success`/`failure` inside the `if/else` *before* the now-unconditional logcat dump, so the `Gate on test failures` verdict can't be flipped by the dump; the pipeline ends in `filter_logcat.sh`'s `|| true` and `tee` to the `mkdir -p results` dir, safe under `set -o pipefail`, and `upload-artifact@v7` defaults to `if-no-files-found: warn`.

## Evidence status (carried, non-blocking on this diff)

- #365 (GB4PC_E2E lines present in the artifact on a green run): satisfied, per the artifact pulled on `2e2bfd0` ([#issuecomment-4687116581](https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4687116581)).
- #364 (race + retry recovery confirmed): not yet met; it depends on a CI run that actually trips the intermittent first-launch teardown, which this diff cannot force. The diagnostic channel that makes such a run self-evident is now in place and proven working. This is a maintainer/CI follow-up, not a defect in this diff.

## CI

`build-and-test` is pending on `7d94d20` at review time. Not blocking per the review protocol; per this PR's own (correct) reasoning a green run neither confirms nor refutes the race fix unless `e2e-overlay-logcat.txt` happens to show race-and-recovery.

Verdict: LGTM. Every blocking item from the prior rounds is resolved, the implementation and CI-config changes are correct and re-verified, and the only remaining #364 confirmation is a maintainer follow-up this diff cannot perform.

---

aunger:
Reviewed HEAD `7d94d20`, re-verifying the full diff from a clean checkout rather than just the latest commit.

## Blocking nit from the prior rounds: resolved

`import android.util.Log` is now at line 14 of `E2EFixture.kt`, at the bottom of the `android.*` block immediately before the `androidx.*` imports, which is where `android.util.*` sorts (after `android.app.*`, `android.content.*`, `android.media.*`, `android.os.*`, `android.provider.*`). The repeatedly-flagged import-order inconsistency is gone, and `7d94d20` is a pure one-line move with no other change.

## Implementation: re-verified against the checked-out source, correct

- `waitForCondition(timeoutMs)` checks the condition immediately, sleeps 100 ms, and re-checks once more at the deadline (lines 209-217). So the baseline `waitForCondition(LAUNCH_BASELINE_MS) { !OverlayService.isOverlayActive }` does establish a known-inactive starting point, and each attempt's `waitForCondition(verifyMs) { OverlayService.isOverlayActive }` can only succeed on a genuine false -> true transition. The stale-`true` short-circuit raised in the early rounds is correctly closed.
- Tiered windows match the companion-object comment: `LAUNCH_BASELINE_MS(3) + LAUNCH_FIRST_VERIFY_MS(12) + (LAUNCH_ATTEMPTS-1)(2) x LAUNCH_RETRY_VERIFY_MS(6) = 27 s < 30 s` assertion. The first 12 s window covers the documented ~9 s + ~1 s healthy activation, so a slow-but-healthy first launch returns without re-issuing `am start` over a mid-resume activity. The `repeat` loop terminates and lets the caller's own assertion fail rather than hang.
- The change is test-harness-only. The diff touches exactly four files (`E2EFixture.kt`, `build.yml`, `scripts/filter_logcat.sh`, `scripts/test_filter_logcat.sh`); no production code path is modified, and `app:testDebugUnitTest` is unaffected.
- CI-config commits are correct: `GB4PC_E2E` added to the `filter_logcat.sh` allowlist with a matching `test_filter_logcat.sh` case (19/19), and the `build.yml` logcat dump is moved out of the failure-only branch, tee'd to `results/e2e-overlay-logcat.txt`, and uploaded. `outcome=success`/`failure` is written inside the `if/else` before the unconditional dump, so the gate verdict can't be flipped by it; the pipeline ends in `filter_logcat.sh`'s `|| true` and `upload-artifact@v7` defaults to `if-no-files-found: warn`, so neither can spuriously fail the step or upload.

## Evidence status (carried, not blocking this diff)

- #365 (GB4PC_E2E lines present in the artifact on a green run): satisfied, per the `2e2bfd0` artifact pull.
- #364 (positive race + retry-recovery confirmation): not yet met. Run `27391057782` on this HEAD's predecessor showed `overlayAppearsWhenViewfinderOpens` activating the overlay on attempt 1 with no teardown race (scenario (b)), so the retry branches were not exercised. That confirmation depends on a CI run that actually trips the intermittent first-launch teardown, which this diff cannot force. The diagnostic channel that makes such a run self-evident is in place and proven working. This is a maintainer/CI follow-up tracked by #364, not a defect in this diff.

## CI

Green on `7d94d20`: `build-and-test`, `analyze-kotlin`, `analyze-actions`, and CodeQL all succeeded; `mergeable_state` is `clean`. Per this PR's own (correct) reasoning a green `build-and-test` neither confirms nor refutes the race fix unless `e2e-overlay-logcat.txt` happens to show race-and-recovery; that artifact is now produced on every run.

## Verdict

LGTM. Every blocking item from the prior five rounds is resolved, the retry/verify logic and CI-config changes are correct and re-verified, the import-order nit is fixed, and CI is green. The only outstanding #364 item (the conclusive race + recovery logcat) is a maintainer/CI follow-up this diff cannot perform, and it is already tracked as a blocking issue against this PR.

---

PR #370, opened 2026-06-13
Fix Pixel Camera missing-card test for mock-camera CI environment

aunger:
## Summary

Fixes #303.

`MainSettingsScreenTest#mainScreen_showsPixelCameraMissingCard_whenNotInstalled` asserted that the "Pixel Camera is not installed" card is always displayed, based on the comment "On an emulator Pixel Camera is never present." That assumption is no longer true: `e2e-mock-camera` installs a stub APK under Pixel Camera's applicationId (`com.google.android.GoogleCamera`) before the instrumented test suite runs (see the "CI pre-flight" and "Run E2E emulator setup" steps in `.github/workflows/build.yml`). As a result, `PermissionHelper.isPixelCameraInstalled()` returns `true` on CI, `MainSettingsScreen` correctly hides the missing-camera card, and the old assertion that it must be displayed always failed.

## Changes

- Renamed and rewrote the test as `mainScreen_showsPixelCameraMissingCard_matchingInstalledState`: it now reads `PermissionHelper.isPixelCameraInstalled()` at runtime and asserts the card is shown only when Pixel Camera (or its CI stub) is *not* installed, and hidden (`assertDoesNotExist()`) when it *is* installed. This makes the test correct both on CI (mock camera installed) and on a real device without Pixel Camera.
- Removed the test from `.github/allowed-test-failures.txt` (issue #303 entry), since it should now pass.

## Test plan

- [x] `./gradlew assembleDebug assembleDebugAndroidTest testDebugUnitTest` passes locally.
- [ ] CI instrumented test run confirms `MainSettingsScreenTest#mainScreen_showsPixelCameraMissingCard_matchingInstalledState` passes (mock camera installed -> card hidden).

PR #371, opened 2026-06-13
Fix picker instrumented test for default photo-app filter

### LINKED ISSUE #305: [MainActivityRedirectTest] pickerScreen_loadsApps_andShowsSettingsApp
## Test Failure

| Field | Value |
|-------|-------|
| Suite | Standard Instrumented Tests |
| Class | `com.gb4pc.ui.MainActivityRedirectTest` |
| Method | `pickerScreen_loadsApps_andShowsSettingsApp` |
| Branch | `fix/issue-265-wire-instrumented-tests-ci` |
| Commit | `320d833` |
| Detected at | 2026-05-31 04:39 UTC |

### Stack trace

```
androidx.compose.ui.test.ComposeTimeoutException: Condition still not satisfied after 10000 ms
at androidx.compose.ui.test.AndroidComposeUiTestEnvironment$AndroidComposeUiTestImpl.waitUntil(ComposeUiTest.android.kt:441)
at androidx.compose.ui.test.junit4.AndroidComposeTestRule.waitUntil(AndroidComposeTestRule.android.kt:306)
at com.gb4pc.ui.picker.PickerActivityTest.pickerScreen_loadsApps_andShowsSettingsApp(PickerActivityTest.kt:36)
```

### Links

- [CI run](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/26703267361)
- [Test artifact: instrumented-test-results](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/26703267361#artifacts)
- [PR](https://github.com/aunger/gallery-button-for-pixel-camera/pull/297)

---
_Filed automatically by CI on failure of `com.gb4pc.ui.MainActivityRedirectTest.pickerScreen_loadsApps_andShowsSettingsApp`._

aunger:
Fixes #305.

`PickerActivityTest.pickerScreen_loadsApps_andShowsSettingsApp` waited for "Settings" to appear in the picker's app list and then timed out after 10 s.

The picker now defaults to a "photo-related apps" filter (UI-09), introduced after this test was written. "Settings" never matches that filter, so it never appears unless the user taps "Show all apps". The "Show all apps" button is shown whenever `showAll` is false; `showAll` auto-flips to true only when `AppListFilter.buildPhotoRelatedPackages` returns an empty candidate set, in which case the full app list (including "Settings") is shown directly without the button.

The test now:
1. Waits for either "Settings" or the "Show all apps" toggle to appear (covering the async app-list query from #8).
2. If "Show all apps" is present (the photo-related candidate set was non-empty), clicks it and waits for "Settings" to appear in the full list.
3. Asserts "Settings" is displayed.

This covers both cases: an emulator where the photo-related candidate set is empty (picker auto-shows all apps) and one where it is non-empty (test switches to the full list via "Show all apps").

### DIFF
--- a/.github/allowed-test-failures.txt
+++ b/.github/allowed-test-failures.txt
@@ -21,7 +21,6 @@
 # removed from this list as it is fixed.
 
 com.gb4pc.ui.MainActivityRedirectTest#setupScreen_showsGrantButton                              # issue #304
-com.gb4pc.ui.MainActivityRedirectTest#pickerScreen_loadsApps_andShowsSettingsApp                # issue #305
 
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test1a_overlayShowsBlueAtConfiguredPosition            # issue #229
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test2a_emptyGalleryNoGreenAfterTap                     # issue #230
--- a/app/src/androidTest/java/com/gb4pc/ui/picker/PickerActivityTest.kt
+++ b/app/src/androidTest/java/com/gb4pc/ui/picker/PickerActivityTest.kt
@@ -1,8 +1,10 @@
 package com.gb4pc.ui.picker
 
 import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.hasText
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
 import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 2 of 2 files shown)


aunger (commented):
**Review: Changes requested**

The test fix itself is correct: the new wait condition (`hasText("Settings") or hasText("Show all apps")`), the conditional click on "Show all apps", and the second wait correctly mirror `PickerScreen`'s `showAll` state machine in `PickerActivity.kt` (lines 84-109). Both branches (photo-related set empty -> `showAll` auto-true -> "Settings" shown directly; photo-related set non-empty -> "Show all apps" rendered -> click switches to full list -> "Settings" shown) are handled.

Blocking issue: `.github/allowed-test-failures.txt` still lists

```
com.gb4pc.ui.MainActivityRedirectTest#pickerScreen_loadsApps_andShowsSettingsApp                # issue #305
```

PR #370, the precedent for this exact pattern (fixing a test referenced by issue #303), removed its corresponding allowlist entry as part of the fix. The allowlist's own header says "remove entries as the tests are fixed." This PR claims to fix the test for #305 but leaves CI tolerating its failure, so a regression in this very fix wouldn't fail CI. Please remove that line.

Separate nit on the same line: the allowlist entry's class is `com.gb4pc.ui.MainActivityRedirectTest`, but the test actually being changed lives in `com.gb4pc.ui.picker.PickerActivityTest` (per this diff's file path). That mismatch predates this PR, but since you're touching this entry anyway to remove it, it's worth confirming the allowlist matcher actually keys off method name only (so the mismatched class didn't silently make this entry a no-op) - if it does match by full qualified name, this entry may never have been suppressing the right test in the first place, which would be useful context either for this PR or a follow-up.

Smaller point on the PR description: it states "Show all apps" "only appears when the filtered list is non-empty." Looking at `PickerActivity.kt` lines 130 and 143, both the "Showing photo-related apps" hint and the "Show all apps" button are gated purely on `!showAll`, not on `filteredApps` being non-empty. `showAll` only auto-flips to `true` when `photoRelatedPackages` (the raw candidate set from `AppListFilter.buildPhotoRelatedPackages`) is empty, regardless of what `filteredApps` ends up containing. This doesn't break the test's branching (which correctly keys off "Show all apps" presence rather than list emptiness), but the description's stated rationale doesn't match the code as written.

---

aunger:
Thanks for the review. Addressed:

1. **Blocking: allowlist entry removed.** Dropped the `com.gb4pc.ui.MainActivityRedirectTest#pickerScreen_loadsApps_andShowsSettingsApp` line from `.github/allowed-test-failures.txt` now that the test is fixed.

2. **Class-name mismatch nit.** Confirmed `check_allowed_failures.py` matches on `class_name` (the JUnit `testsuite name`, i.e. the actual fully-qualified class from the XML) and `class_name#method_name`. The removed entry used `com.gb4pc.ui.MainActivityRedirectTest`, but the test actually lives in `com.gb4pc.ui.picker.PickerActivityTest`, so neither `is_allowed` branch (`failure.class_name in allowlist` or `failure.qualified in allowlist`) could ever have matched. The entry was a no-op from the start; it never suppressed this test's failure (and CI failing on this test is exactly how issue #305 got filed). No further action needed here since the line is gone, but flagging in case it's useful for auditing other allowlist entries for the same class-path drift.

3. **PR description rationale.** Updated the description: the "Show all apps" button (and the filtered-list hint) are gated purely on `!showAll`, not on `filteredApps` being non-empty. `showAll` only auto-flips to `true` when `AppListFilter.buildPhotoRelatedPackages` returns an empty candidate set. The test's branching logic was already correct since it keys off "Show all apps" presence, not list emptiness - only the description's wording was off.

---

PR #372, opened 2026-06-13
Fix setup screen test for non-Grant first-step button labels

### LINKED ISSUE #304: [MainActivityRedirectTest] setupScreen_showsGrantButton
## Test Failure

| Field | Value |
|-------|-------|
| Suite | Standard Instrumented Tests |
| Class | `com.gb4pc.ui.MainActivityRedirectTest` |
| Method | `setupScreen_showsGrantButton` |
| Branch | `fix/issue-265-wire-instrumented-tests-ci` |
| Commit | `320d833` |
| Detected at | 2026-05-31 04:39 UTC |

### Stack trace

```
java.lang.AssertionError: Assert failed: The component is not displayed!
at androidx.compose.ui.test.AssertionsKt.assertIsDisplayed(Assertions.kt:34)
at com.gb4pc.ui.SetupActivityTest.setupScreen_showsGrantButton(SetupActivityTest.kt:46)
```

### Links

- [CI run](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/26703267361)
- [Test artifact: instrumented-test-results](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/26703267361#artifacts)
- [PR](https://github.com/aunger/gallery-button-for-pixel-camera/pull/297)

---
_Filed automatically by CI on failure of `com.gb4pc.ui.MainActivityRedirectTest.setupScreen_showsGrantButton`._

aunger:
## Summary

Fixes #304.

`SetupActivityTest.setupScreen_showsGrantButton` asserted that a node with text containing "Grant" is displayed on the setup screen. On API 33+ emulators, the setup flow's first step is `NOTIFICATION`, whose button label is "Allow Notifications" (it does not contain "Grant"). Similarly, the `BATTERY` step's button reads "Exclude from Battery Optimization". When either of these is the first displayed step, the substring assertion fails with:

```
java.lang.AssertionError: Assert failed: The component is not displayed!
```

## Fix

Renamed the test to `setupScreen_showsActionButton` and changed it to check whether any of the four possible step-button labels (`setup_notification_button`, `setup_usage_access_button`, `setup_overlay_button`, `setup_battery_button`) is currently displayed. This makes the test pass regardless of which step is shown first, while still asserting that the setup screen renders its primary action button.

Also removed the now-resolved `com.gb4pc.ui.MainActivityRedirectTest#setupScreen_showsGrantButton` entry from `.github/allowed-test-failures.txt` (the entry's class name was a mislabeling from the auto-filed issue; the actual failing test is `com.gb4pc.ui.SetupActivityTest`, per the reported stack trace).

## Test plan

- [x] `./gradlew compileDebugKotlin compileDebugAndroidTestKotlin` succeeds
- [x] `./gradlew :app:testDebugUnitTest` passes
- [x] Instrumented test `com.gb4pc.ui.SetupActivityTest#setupScreen_showsActionButton` passes on CI (requires emulator)

### DIFF
--- a/.github/allowed-test-failures.txt
+++ b/.github/allowed-test-failures.txt
@@ -20,8 +20,6 @@
 # issue #309 removed it.  Each already has a tracking issue and must be
 # removed from this list as it is fixed.
 
-com.gb4pc.ui.MainActivityRedirectTest#setupScreen_showsGrantButton                              # issue #304
-
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test1a_overlayShowsBlueAtConfiguredPosition            # issue #229
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test2a_emptyGalleryNoGreenAfterTap                     # issue #230
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test3a_populatedGalleryShowsGreenAfterTap              # issue #231
--- a/app/src/androidTest/java/com/gb4pc/ui/SetupActivityTest.kt
+++ b/app/src/androidTest/java/com/gb4pc/ui/SetupActivityTest.kt
@@ -2,13 +2,16 @@ package com.gb4pc.ui
 
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.junit4.createAndroidComposeRule
+import androidx.compose.ui.test.onAllNodesWithText
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
+import com.gb4pc.R
 import com.gb4pc.data.PrefsManager
 import com.gb4pc.ui.setup.SetupActivity
 import org.junit.After
+import org.junit.Assert.assertTrue
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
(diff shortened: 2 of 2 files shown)


aunger (commented):
Reviewed the diff against the issue #304 stack trace and the SetupActivity/SetupState source.

Findings:

- Root cause analysis checks out. The stack trace in every auto-filed comment is `SetupActivityTest.setupScreen_showsGrantButton(SetupActivityTest.kt:46)`, and the old test asserted on the substring "Grant". `setup_notification_button` ("Allow Notifications") and `setup_battery_button` ("Exclude from Battery Optimization") indeed don't contain "Grant", so when NOTIFICATION or BATTERY is the first undone step, the old assertion fails. That matches the failure.
- The four string resources referenced (`setup_notification_button`, `setup_usage_access_button`, `setup_overlay_button`, `setup_battery_button`) match `strings.xml` exactly, and none is a substring of another, so `onAllNodesWithText` won't produce false positives across steps.
- `onAllNodesWithText(...).fetchSemanticsNodes().isNotEmpty()` matches the existing pattern already used in `PickerActivityTest.kt`, so it's consistent with precedent in this codebase rather than a new idiom.
- The `assertIsDisplayed` and `onNodeWithText` imports remain used elsewhere in the file (lines 38, 43, 79), so nothing is left dangling.
- The `allowed-test-failures.txt` removal is precisely scoped: only the `MainActivityRedirectTest#setupScreen_showsGrantButton` line is removed, and the PR's claim that this was a mislabeled class name (the actual failing class per every stack trace is `SetupActivityTest`) is correct; the other unrelated entries (#229-#232) are untouched.

Minor nit (non-blocking): the new test doesn't call `composeRule.waitForIdle()` before checking `fetchSemanticsNodes()`. Given how much CI flakiness this repo has chased in #233/#249/etc., it might be worth a defensive `waitForIdle()` call before the snapshot check, though `createAndroidComposeRule` does auto-sync before idling resources settle, so this is likely a non-issue in practice.

Overall this is a clean, well-scoped fix that directly addresses the reported failure with a test design that should be robust regardless of which setup step is shown first. LGTM.

---

PR #375, opened 2026-06-13
Remove 'state=open' filter from test-failure-archive query

aunger:
The intended `test-failure` tag process is for issues to be closed as normal, by a PR. The tag will remain in place when the issue is closed. If the same test fails soon thereafter (within a configurable number of archive days), CI will reopen the issue. After that archive duration, a scheduled Action will replace the tag with `test-failure-archive`, which marks the issue as no longer revivable.

That process means that the scheduled archiver process must replace the tags even on issues that are already closed.

### DIFF
--- a/scripts/archive_stale_test_failures.py
+++ b/scripts/archive_stale_test_failures.py
@@ -67,7 +67,7 @@ def fetch_stale_issues(repo: str, token: str, cutoff: datetime.datetime) -> list
     page = 1
     while True:
         batch = gh_api(
-            f"repos/{repo}/issues?state=open&labels=test-failure&per_page=100&page={page}",
+            f"repos/{repo}/issues?labels=test-failure&per_page=100&page={page}",
             token=token,
         )
         if not batch:
--- a/scripts/archive_stale_test_failures.py
+++ b/scripts/archive_stale_test_failures.py
@@ -1,10 +1,14 @@
 #!/usr/bin/env python3
 """Archive stale test-failure GitHub issues.
 
-Finds all open issues labelled ``test-failure`` whose ``updated_at``
-timestamp is older than STALE_DAYS (default 21).  For each such issue
-the ``test-failure`` label is removed and ``test-failure-archive`` is
-added, so it falls out of the Phase 4 dedup search scope.
+Finds all issues (open or closed) labelled ``test-failure`` whose
+``updated_at`` timestamp is older than STALE_DAYS (default 21).  For
+each such issue the ``test-failure`` label is removed and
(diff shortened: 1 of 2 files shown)


aunger:
Review

This is a one-line change to `scripts/archive_stale_test_failures.py` that drops `state=open` from the issue-list query, so `fetch_stale_issues` now considers closed issues too.

Concerns:

1. **No PR description.** There's nothing explaining why this filter is being removed. Reading the surrounding code, I can construct a plausible rationale (see below), but the Author should state the actual motivation so reviewers and future readers don't have to reverse-engineer it.

2. **Stale docstrings.** Both the module docstring ("Finds all open issues labelled `test-failure`...") and `fetch_stale_issues`'s own docstring ("Return all open test-failure issues last updated before *cutoff*") still say "open", but the function no longer filters by state. After this change those docstrings are inaccurate and should be updated to describe the new (any-state) behavior.

3. **No test coverage for the behavior change.** None of the tests in `scripts/test_archive_stale_test_failures.py` exercise a closed issue. If the intent is that stale closed issues with the `test-failure` label now get swept up and relabeled to `test-failure-archive`, a test should demonstrate that (e.g., a closed, stale, `test-failure`-labeled issue is returned by `fetch_stale_issues` and archived by `main`).

For context on point 1: `file_test_failure_issues.py`'s `find_existing_issue` dedups on `label:test-failure` without an `is:open`/`is:closed` filter, and can reopen a closed issue carrying that label. So a closed issue that still has `test-failure` would remain in dedup scope and could get reopened later, undermining the "falls out of the Phase 4 dedup search scope" goal described in this script's own docstring. If that's the motivating bug, this change makes sense, but it should say so, and the docstrings/tests should reflect it.

Decision: Changes requested (docstrings out of sync with new behavior, no test coverage, and no PR description explaining intent).

---

aunger:
I updated the PR description.

aunger:
Review (follow-up)

The follow-up commit (b39a310) addresses all three points from the previous review:

1. PR description now states the motivation directly.
2. Both docstrings (module-level and `fetch_stale_issues`) accurately describe the new open-or-closed behavior, and explain the `find_existing_issue` dedup rationale that motivates the change.
3. New tests (`test_stale_closed_issue_returned`, `test_query_does_not_filter_by_state`, `test_archives_stale_closed_issue`) cover the closed-issue behavior. Ran the full suite locally: all 24 tests pass.

No further issues found in this diff. The query change, docstring updates, and test additions are consistent with each other and with `archive_issue`'s label-swap logic, which is unaffected and still correct regardless of issue state.

Decision: LGTM.

---

PR #379, opened 2026-06-14
Strip Claude bylines whose link omits the /session_ suffix

### LINKED ISSUE #377: Automated byline removal failed
See PR 375, which has several comments ending thus:

> Generated by [Claude Code](https://claude.ai/code)

Please fix the workflow.

aunger:
Fixes #377.

PR #375's review comments ended with `_Generated by [Claude Code](https://claude.ai/code)_`, a Markdown-link byline footer whose URL has no `/session_<token>` path. The `strip-session-bylines` workflow's regex only matched that footer when the URL included a `/session_<token>` suffix, so this byline form passed through untouched.

This change makes the `/session_<token>` suffix optional in the Markdown-link footer alternative of `_BYLINE_RE`, so both forms (`https://claude.ai/code` and `https://claude.ai/code/session_<token>`) are stripped from issue/PR bodies, comments, and review bodies.

The bare-URL alternative (`\nhttps://claude.ai/code/session_<token>`) is left requiring the session suffix, since making it optional risked truncating unrelated URLs like `https://claude.ai/code/other_thing` into a dangling `/other_thing`.

Added two new tests covering the no-session-suffix Markdown-link footer, with and without the `---` separator.


aunger (commented):
🤖 Reviewer

Reviewed the regex change and tests on the branch.

- The fix is correctly scoped: making `/session_<token>` optional only in the Markdown-link footer alternative, while leaving the bare-URL alternative untouched, is the right call. I verified the alternative (making the bare-URL suffix optional too) would corrupt unrelated URLs like `https://claude.ai/code/other_thing` into a dangling `/other_thing`, exactly as the PR description warns.
- Verified against the actual byline text from PR #375's review comments (`...issue state.\n\n---\n_Generated by [Claude Code](https://claude.ai/code)_`); the updated regex strips it correctly, leaving a clean trailing newline.
- Ran the full test module on the branch: all 41 tests pass, including the two new ones for the no-suffix footer (with and without the `---` separator).
- The comment-only edit to `test_unchanged_for_markdown_link_without_session_prefix` doesn't change its input/assertion, so this isn't a loosened test requiring declaration; the new comment is also more accurate (that text has no leading newline, so it was never going to match regardless of session suffix).
- Docstring and inline comments accurately describe the new optional-suffix behavior and cite issue #377.
- Single, well-scoped commit; no scope creep.

No issues found. Decision: LGTM.

PR #383, opened 2026-06-14
Apply verified label when before-merging list is empty

### LINKED ISSUE #382: Add verified label
Instruct the verification planner agent to apply the `verified` label to both PR and issue, if its before-merge requirement list is empty.

aunger:
Fixes #382

Instructs the Verification Planner agent to apply the `verified` label to both the PR and the issue it resolves when its before-merging list is empty, signaling that no outstanding unautomated verification steps or out-of-repo requirements remain.

### DIFF
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -24,7 +24,7 @@ You file one tracking issue per item (see **What to do**) so that each requireme
 1. Read the issue description, PR description, and all comments on both.
    Look for both kinds of outstanding requirement described under **Role**: unautomated verification steps, and changes outside the repo (such as an issue that needs to be filed).
    Assemble the *before merging* list, labeling each item as either an unautomated verification step or a change outside the repo, and noting for each item the URL of the specific PR comment that called for it.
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

This is a minimal, well-scoped change that does exactly what issue #382 asks: a single line added to step 2 of `agents/verification_planning.md` instructing the Verification Planner to apply the `verified` label to both the PR and its parent issue when the before-merging list is empty.

Notes:
- The `verified` label already exists in this repo (green, id `LA_kwDORy4CL88AAAACnU6ocA`), so no setup is needed for the instruction to be actionable.
- "the issue it resolves" is consistent with the existing fallback language in step 3d ("the parent issue the PR resolves"), so the Planner has a well-established way to identify which issue to label.
- Applying a label is a routine operation the Planner can do via the GitHub API/MCP tools without needing the detailed curl/fallback instructions that steps 3c/3d spell out for the less-common issue-dependency and sub-issue operations, so the terseness here is fine.
- Docs-only change, no automated tests needed or expected.

LGTM.

PR #384, opened 2026-06-14
Move ci_monitor.py into scripts/ci_monitor/ and split off its usage docs

### LINKED ISSUE #381: Separate ci_monitor usage instructions from orchestrator
- Move `scripts/ci_monitor.py` into `scripts/ci_monitor/`.
- Extract the usage instructions for `ci_monitor.py` from `dev_orchestration.md`
    - form `scripts/ci_monitor/README.md`
    - replace the instructions with a reference to the new file
    - This excludes any Orchestrator-specific instruction, which must remain in `dev_orchestration.md`
- Create CLAUDE.md in the new folder, with only the line `@README.md`, preceded by a notice that this file simply @ includes README.

aunger:
Fixes #381.

- Moves `scripts/ci_monitor.py` to `scripts/ci_monitor/ci_monitor.py`.
- Extracts the usage instructions for `ci_monitor.py` (CLI syntax, per-test outcome filters, outcome vocabulary) out of `agents/dev_orchestration.md` and into the new `scripts/ci_monitor/README.md`. `dev_orchestration.md` now links to the README for that content and retains only the Orchestrator-specific routing notes (the Monitor loop algorithm, the 30-minute escalation rule, and the rule against starting a new Author round on informational lines).
- Adds `scripts/ci_monitor/CLAUDE.md`, which simply `@`-includes `README.md`, preceded by a notice explaining that.
- Updates `scripts/test_ci_monitor.py` to locate the relocated module and script at their new path.

All 128 existing `test_ci_monitor.py` cases still pass after the move.

### DIFF
--- a/scripts/ci_monitor/README.md
+++ b/scripts/ci_monitor/README.md
@@ -1,0 +1,78 @@
+# CI Monitor
+
+`ci_monitor.py` polls a pull request's CI and streams a terminal outcome plus
+per-test signals. Each stdout line is the interface: terminal lines
+(`Clear`/`Blocked`/`Infra`) end the loop, while informational lines
+(`in_progress` heartbeat, per-step deltas, per-test `FAIL`/`SKIP`/`PASS`) keep
+it alive.
+
+For how the Orchestrator uses this script as part of the development cycle
+(the Monitor loop, routing decisions, and the Verification Planner dispatch),
+see [`agents/dev_orchestration.md`](../../agents/dev_orchestration.md).
+
+## Running the monitor
+
+Run it from the repo root, passing the PR number via `--pr`:
+
+```bash
+python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER> [filter flags]
+```
+
+`OWNER`/`REPO` default to this repo at the top of the script, and it reads
+`$GITHUB_TOKEN` from the environment (required). The script catches transient
+REST/parse failures per call so they cannot kill the resilient poll loop.
+
+## Per-test outcome filters
+
+By default the monitor reports **all FAIL markers**, **all SKIP markers**, and
+**no PASS markers**. Independent filter flags narrow or expand which per-test
+outcomes are streamed:
+
+| Flag | Effect |
+|---|---|
+| `--include-fail [PATTERN]` | Report FAIL markers (default); optionally restrict to those whose `name` matches PATTERN. |
+| `--no-include-fail` | Suppress all FAIL markers. |
+| `--include-skip [PATTERN]` | Report SKIP markers (default); optionally restrict to those whose `name` matches PATTERN. |
+| `--no-include-skip` | Suppress all SKIP markers. |
+| `--include-pass [PATTERN]` | Report PASS markers (not the default); optionally restrict to those whose `name` matches PATTERN. |
+| `--no-include-pass` | Suppress all PASS markers (explicit form of the default). |
+
+Each `--include-*` flag takes an **optional regex** matched against the
+marker's `name` field. Supplied without a pattern it includes *all* markers of
+that outcome. The three outcomes keep their distinct labels in output:
+`--include-pass` never relabels a SKIP as PASS.
+
+**Task-relevance validation.** To verify that a task-relevant test actually
+ran and passed (rather than being silently skipped), supply `--include-pass`
+with a regex matching the test(s) of interest:
+
+```bash
+python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER> --include-pass 'MyFeatureTest'
+```
+
+This emits a `PASS` line when the matching test passes and a `SKIP` line if it
+was skipped (which would be a false-validation trap: the code path was never
+exercised). With no pattern (`--include-pass ''`), every passing test is
+reported.
+
+## Outcome vocabulary
+
+| Line emitted          | Meaning                                                                                               |
+|-----------------------|-------------------------------------------------------------------------------------------------------|
+| `PR#N: Clear ...`     | All CI checks passed and `mergeable_state` is `clean` or `unstable`; PR may be merged.               |
+| `PR#N: Blocked ...`   | A check failed (`failure`/`action_required`) or `mergeable_state` is `behind`/`dirty`; new Author round needed. |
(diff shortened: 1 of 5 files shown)


aunger (commented):
🤖 Reviewer

Checked this against the three requirements in #381:

1. `scripts/ci_monitor.py` -> `scripts/ci_monitor/ci_monitor.py`: done, executable bit preserved, `scripts/test_ci_monitor.py` updated to the new `sys.path` and `_MONITOR_PATH`. Ran the suite locally on this branch: 128 passed, 0 failed, matching the PR description.
2. Usage docs extracted to `scripts/ci_monitor/README.md`, with `dev_orchestration.md` left holding only Orchestrator-specific routing notes (the `timeout_ms` rule, the "informational lines don't start a new Author round" rule, and "don't subscribe to PR events"). The "Monitor reads results at step granularity from two polled REST signals..." bullet was moved entirely into the README rather than kept in both places. That's a defensible call (it's an implementation detail of the script, not an Orchestrator rule), but flagging it in case the intent was for it to stay as Orchestrator-facing context too.
3. `scripts/ci_monitor/CLAUDE.md` matches the spec exactly: an HTML-comment notice followed by `@README.md`.

I grepped the whole repo for leftover `scripts/ci_monitor.py` references (workflows, docs, scripts) and found none outside the new `ci_monitor/ci_monitor.py` path itself, so the rename is complete.

Two minor notes, neither blocking:

- `scripts/ci_monitor.sh` (a pre-existing, untouched-by-this-PR bash implementation) still has `bash scripts/ci_monitor.sh <PR_NUMBER>` usage docs in its header and duplicates the old outcome vocabulary verbatim. It's out of scope for #381 since this PR doesn't touch it, but since #381's goal was de-duplicating this documentation, it might be worth a follow-up issue if `ci_monitor.sh` is still meant to be a live alternative to the Python script.
- The four `—` (em dash) characters in the relocated outcome-vocabulary table in `README.md` violate `.claude/rules/prose-style.md` ("avoid... em-dash characters"), but they're carried over verbatim from the pre-existing table in `dev_orchestration.md`, so this is a pre-existing issue, not something newly introduced here.

`mergeable_state` currently shows `blocked` with 0 checks reported yet (CI hasn't started/reported), which I understand the Monitor loop will resolve after this review.

This is a clean, faithful extraction/move. LGTM.

scripts/ci_monitor/README.md
```diff
@@ -0,0 +1,78 @@
...
+was skipped (which would be a false-validation trap: the code path was never
+exercised). With no pattern (`--include-pass ''`), every passing test is
+reported.
+
+## Outcome vocabulary
+
+| Line emitted          | Meaning                                                                                               |
+|-----------------------|-------------------------------------------------------------------------------------------------------|
+| `PR#N: Clear ...`     | All CI checks passed and `mergeable_state` is `clean` or `unstable`; PR may be merged.               |
+| `PR#N: Blocked ...`   | A check failed (`failure`/`action_required`) or `mergeable_state` is `behind`/`dirty`; new Author round needed. |
```
aunger: This is an example of Orchestrator-specific detail, and shouldn't be in this file

scripts/ci_monitor/README.md
aunger, line 1: Follow word wrap rules. Break lines at sentence breaks.

aunger (commented):
See comments in this review

scripts/ci_monitor/README.md
aunger: 🤖 Author

Agreed, fixed. Removed "; new Author round needed" from the `Blocked` row. That routing decision already lives in `dev_orchestration.md` ("if Monitor emits a Blocked line -> goto newAuthor"), so the README now just describes what `Blocked` means without prescribing the Orchestrator's response.

scripts/ci_monitor/README.md
aunger: 🤖 Author

Fixed. Rewrapped all prose paragraphs in `scripts/ci_monitor/README.md` to break at sentence boundaries instead of mid-sentence column wrapping, per `.claude/rules/prose-style.md`. While in there, I also replaced the four remaining em-dash characters in the outcome-vocabulary table (carried over from `dev_orchestration.md`) with semicolons, since they're now newly part of this file and easy to clean up alongside the wrap fix.

aunger (commented):
🤖 Reviewer

Re-reviewed after the latest commit (314e805), which addresses both review-thread comments from the prior round.

Checked against #381's three requirements again:

1. `scripts/ci_monitor.py` -> `scripts/ci_monitor/ci_monitor.py`: confirmed, executable bit preserved (`-rwxr-xr-x`), and a repo-wide grep for `scripts/ci_monitor.py` / `scripts.ci_monitor` finds no leftover references outside the new path.
2. Usage docs live in `scripts/ci_monitor/README.md`; `dev_orchestration.md` retains only the Orchestrator-specific routing notes. The fix in 314e805 (removing "; new Author round needed" from the Blocked row, since that routing decision is already covered by "if Monitor emits a Blocked line -> goto newAuthor") is correct and reads well.
3. `scripts/ci_monitor/CLAUDE.md` matches the spec: HTML-comment notice + `@README.md`.

Ran the test suite directly (pytest isn't installed in this sandbox, so `python3 scripts/test_ci_monitor.py`): 128 passed, 0 failed, matching the PR description.

The word-wrap rewrap and em-dash-to-semicolon fixes in `scripts/ci_monitor/README.md` look correct per `.claude/rules/prose-style.md`.

One new nit, not caught in the prior round: the "Orchestrator-specific notes" bullet list this PR adds to `agents/dev_orchestration.md` sits right next to a pre-existing line that still has an em dash:

> The 30-minute escalation threshold is enforced by `timeout_ms: 1800000` on the Monitor call — no elapsed-time tracking needed.

That line is unchanged by this PR (confirmed via `git diff 8ce774d 314e805 -- agents/dev_orchestration.md`), so it's the same "pre-existing issue, not newly introduced here" category as the README em dashes the prior review flagged, and non-blocking for the same reason. Flagging it now mainly because the Author already did the cleanup for the analogous case in the README, so the inconsistency is now more visible side by side. Not asking for a change in this PR, just noting it; could be picked up incidentally in a future edit to that file, or in a small follow-up if anyone wants to do a prose-style sweep of `agents/`.

This is a clean, faithful extraction/move with the prior round's feedback addressed correctly. LGTM.

PR #388, opened 2026-06-14
Add FLAG_LAYOUT_IN_SCREEN to fix overlay vertical position offset

aunger:
Fixes #229.

## Root cause

`OverlayManager.createLayoutParams()` computes `xPx`/`yPx` via `calculateOverlayXPx`/`calculateOverlayYPx`, which treat `(0, 0)` as the top-left of the full physical screen (matching `displayWidth`/`displayHeight` from `currentWindowMetrics.bounds`).

However, the `TYPE_APPLICATION_OVERLAY` window was only setting `FLAG_LAYOUT_NO_LIMITS`, not `FLAG_LAYOUT_IN_SCREEN`. Without `FLAG_LAYOUT_IN_SCREEN`, a window's `Gravity.TOP or Gravity.START` origin for `x`/`y` is offset below the status bar rather than the true screen origin. The result: the overlay consistently rendered about 128 px lower than the configured `yPercent` on a 1080x2400 display.

This is the cause of the long-standing `test1a_overlayShowsBlueAtConfiguredPosition` E2E failure (issue #229), which has reported the exact same offset (`BLUE centroid at (216.0, 1784.0)` vs expected `(216.0, 1656.0)`, a 128 px gap) on every CI run since 2026-05-25.

## Fix

Add `FLAG_LAYOUT_IN_SCREEN` alongside `FLAG_LAYOUT_NO_LIMITS` in both branches of `createLayoutParams()` (focusable and non-focusable overlay), so the overlay's `x`/`y` are positioned relative to the true screen origin, matching the coordinates `calculateOverlayXPx`/`calculateOverlayYPx` compute and what `test1a_overlayShowsBlueAtConfiguredPosition` expects.

## Tests

- Added a Robolectric regression test (`overlay window uses FLAG_LAYOUT_IN_SCREEN so position is relative to screen origin`) asserting both `FLAG_LAYOUT_IN_SCREEN` and `FLAG_LAYOUT_NO_LIMITS` are set on the overlay's `WindowManager.LayoutParams`.
- All existing unit tests pass (`./gradlew :app:testDebugUnitTest`).
- The E2E suite (including `test1a_overlayShowsBlueAtConfiguredPosition`) runs as part of CI's `build-and-test` job.

PR #389, opened 2026-06-14
Stop allowing test2a_emptyGalleryNoGreenAfterTap to fail (depends on #388)

aunger:
## Problem

`GalleryButtonVisualE2ETest#test2a_emptyGalleryNoGreenAfterTap` has been failing consistently, reporting GREEN coverage of about 87% after tapping the overlay, when it should drop below 10% once the gallery opens over the empty roll.

CI screenshots from a recent run show the before-tap and after-tap screenshots are identical: the green MockCameraActivity screen with the overlay icon unchanged. The gallery never opens, meaning the tap had no effect.

## Root cause

Issue #229 established that the overlay's actual rendered position is about 128px lower than the position computed from `PrefsManager.getOverlayPosition()` (via `calculateOverlayYPx`). `E2EFixture.tapOverlay()` taps the *computed* position, so its tap lands on empty camera-app space above the icon and never hits the overlay's clickable view. The gallery is never launched, so the green camera feed stays on screen (issue #230, and likely the same cause behind #231 and #232).

## Fix

PR #388 fixes that root cause directly: it adds `FLAG_LAYOUT_IN_SCREEN` to the overlay's window flags, so the overlay's *rendered* position matches the *computed* position that `tapOverlay()` already uses. Once #388 merges, `tapOverlay()`'s existing computed-position tap lands on the overlay correctly, and `test2a_emptyGalleryNoGreenAfterTap` should pass without any change to `tapOverlay()` itself.

This PR therefore:

1. Removes `test2a_emptyGalleryNoGreenAfterTap` from `.github/allowed-test-failures.txt` now that #388 fixes the underlying cause.
2. Documents `tapOverlay()`'s dependency on #229/#388's fix in its KDoc, so a future regression in the overlay's window flags is easier to connect back to this assumption.

An earlier version of this PR independently reworked `tapOverlay()` to locate the overlay by screenshotting the screen and tapping the centroid of its icon's color mask, as a way to sidestep the computed/rendered position mismatch without depending on #388. Per review discussion, that added meaningful complexity on top of a fix (#388) that closes the gap at its source, so it has been reverted in favor of this smaller, #388-dependent change.

Doesn't Fix (but really tried) #230

## Test plan

- [x] `./gradlew :app:compileDebugAndroidTestKotlin -q` succeeds.
- [x] `./gradlew :app:testDebugUnitTest -q` passes.
- [x] `python3 -m unittest scripts.test_check_allowed_failures -v` passes after the allowlist edit.
- [ ] This PR should merge after #388. CI instrumented E2E run confirms `test2a_emptyGalleryNoGreenAfterTap` now passes once #388's fix is present.

### DIFF
--- a/.github/allowed-test-failures.txt
+++ b/.github/allowed-test-failures.txt
@@ -21,6 +21,5 @@
 # removed from this list as it is fixed.
 
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test1a_overlayShowsBlueAtConfiguredPosition            # issue #229
-com.gb4pc.e2e.GalleryButtonVisualE2ETest#test2a_emptyGalleryNoGreenAfterTap                     # issue #230
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test3a_populatedGalleryShowsGreenAfterTap              # issue #231
 com.gb4pc.e2e.GalleryButtonVisualE2ETest#test5a_secureCameraLockedPopulatedGalleryShowsGreen    # issue #232
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -349,9 +349,10 @@ class E2EFixture(
      *
      * Issue #229 established that the overlay's rendered position does not match the
      * coordinates computed from [PrefsManager.getOverlayPosition]: the overlay renders about
-     * 128px lower than [calculateOverlayYPx] predicts. Tapping the *computed* position
-     * therefore misses the overlay's clickable view entirely--the tap lands on empty camera-app
-     * space above the icon and silently fails to launch the gallery (issues #230, #231, #232).
+     * 128px lower than `OverlayManager.calculateOverlayYPx` predicts. Tapping the *computed*
+     * position therefore misses the overlay's clickable view entirely--the tap lands on empty
+     * camera-app space above the icon and silently fails to launch the gallery (issues #230,
+     * #231, #232).
      *
      * To tap reliably regardless of that rendering discrepancy, this locates the overlay by its
      * actual on-screen pixels: it takes a fresh screenshot and taps the centroid of the
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -345,24 +345,40 @@ class E2EFixture(
     }
 
     /**
-     * Taps the overlay button at its configured screen position.
+     * Taps the overlay button at its actual rendered position.
      *
-     * Reads the active [com.gb4pc.data.OverlayPosition] from [PrefsManager], computes pixel
-     * coordinates from [android.view.WindowMetrics] (API 30+) or [android.util.DisplayMetrics]
-     * (API 26–29), then dispatches a tap via [UiDevice].
+     * Issue #229 established that the overlay's rendered position does not match the
+     * coordinates computed from [PrefsManager.getOverlayPosition]: the overlay renders about
+     * 128px lower than [calculateOverlayYPx] predicts. Tapping the *computed* position
+     * therefore misses the overlay's clickable view entirely--the tap lands on empty camera-app
+     * space above the icon and silently fails to launch the gallery (issues #230, #231, #232).
      *
-     * If the overlay is not rendered (e.g. blocked in secure-camera mode) this call taps empty
-     * screen space and has no visible effect — the plan requires this silent behaviour for
-     * Tests 4a/5a's baseline-failure scenario.
+     * To tap reliably regardless of that rendering discrepancy, this locates the overlay by its
+     * actual on-screen pixels: it takes a fresh screenshot and taps the centroid of the
+     * YELLOW union BLUE mask (the gallery icon's background and foreground colors, as used by
+     * test1c_overlayOuterIsSquircle).
+     *
+     * If the overlay is not rendered (e.g. blocked in secure-camera mode), the mask is empty and
+     * this falls back to the position computed from [PrefsManager.getOverlayPosition], tapping
+     * empty screen space and having no visible effect, which Tests 4a/5a's baseline-failure
+     * scenario requires.
      */
     fun tapOverlay() {
         val (displayWidth, displayHeight) = displaySize()
 
-        val aspectRatio = AspectRatioUtil.quantize(displayWidth, displayHeight)
-        val pos = PrefsManager(context).getOverlayPosition(aspectRatio)
+        val screen = Screenshot.captureScreen()
+        val iconMask = ColorMatch.union(
+            ColorMatch.mask(screen, Rgb.BLUE),
+            ColorMatch.mask(screen, Rgb.YELLOW)
+        )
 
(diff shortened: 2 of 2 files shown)


aunger:
🤖 Reviewer

Code itself looks solid: `tapOverlay()`'s screenshot-based centroid lookup reuses the exact `ColorMatch.union(mask(BLUE), mask(YELLOW))` pattern already validated by `test1c_overlayOuterIsSquircle`, the fallback to the computed position when the mask is empty correctly preserves the no-op behavior tests 4a/5a depend on, the allowlist edit is correct (verified `python3 -m unittest scripts.test_check_allowed_failures -v` still passes), and `./gradlew :app:compileDebugAndroidTestKotlin -q` compiles cleanly. Prose style (double-hyphens, no smart quotes/em-dashes) is followed.

One significant concern that needs to be addressed before merge:

**Overlap with PR #388 (issue #229), which is open concurrently and fixes the same root cause.**

PR #388 adds `FLAG_LAYOUT_IN_SCREEN` to `OverlayManager.createLayoutParams()`, which it documents as the actual root cause of the ~128px rendering offset that this PR's description also cites as the reason `tapOverlay()` misses the icon. If #388 is correct, once it merges the overlay's *rendered* position will match `calculateOverlayYPx`/`getOverlayPosition()`, and the *original* (pre-this-PR) `tapOverlay()` implementation, tapping the computed position, would also have started hitting the icon and fixing #230. That would make this PR's screenshot-based icon-locating logic unnecessary complexity layered on top of a fix that's no longer needed for its stated purpose (though it would still function harmlessly, just always taking the "mask found" branch).

Neither PR's description acknowledges the other. This PR's "Root cause" section presents the 128px gap as the reason `tapOverlay()` needs a screenshot-based rewrite, without noting that a sibling PR proposes closing that same gap at the source. Whichever of #388/#389 merges second should at minimum:
- Acknowledge the other PR/issue and explain why this approach is still warranted (e.g. if there's a reason the visual-lookup approach is preferable even after #229's flag fix is in, that should be stated explicitly), or
- Be revisited/simplified in light of the other fix landing.

This isn't a request to expand the scope of this PR's diff, just to get the ordering/overlap discussed and recorded before either merges, so we don't end up with two independent "fixes" for the same underlying offset and confusing history.

Minor nit (non-blocking): the new KDoc comment references `[calculateOverlayYPx]`, which is an `internal` function in `com.gb4pc.overlay.OverlayManager` with no import in this file. The cross-reference won't resolve/link (there's no dokka step so it won't fail the build, just won't be a useful link). Consider spelling it out as plain text or fully qualifying it.

**Verdict: Changes requested** (not for the code quality itself, but to get the #388/#389 overlap discussed and recorded before merge).

aunger:
🤖 Author

Agreed, and you're right: once #388's `FLAG_LAYOUT_IN_SCREEN` fix lands, the overlay's rendered position will match `calculateOverlayYPx`/`getOverlayPosition()`, so the original computed-position `tapOverlay()` would land on the icon correctly on its own. The screenshot/color-mask rewrite was solving a problem #388 already solves at the source, so I've reverted it.

This PR now only:
- Reverts `tapOverlay()` to its original computed-position implementation, with a KDoc note that it depends on #229/#388's `FLAG_LAYOUT_IN_SCREEN` fix for the rendered position to match.
- Keeps the `allowed-test-failures.txt` removal for `test2a_emptyGalleryNoGreenAfterTap`.

Also fixed the KDoc nit (spelled out `OverlayManager.calculateOverlayYPx` as plain text rather than an unresolvable cross-reference) before the revert.

Updated the PR description to reflect this and to note the dependency on #388 merging first. `./gradlew :app:compileDebugAndroidTestKotlin -q` and `./gradlew :app:testDebugUnitTest -q` both still pass.

aunger (commented):
🤖 Reviewer

The revert addresses the prior concern fully. Verified:

- `tapOverlay()`'s function body is now byte-for-byte identical to the pre-PR implementation on main (confirmed via `git diff` against the merge-base); only the KDoc changed.
- The new KDoc paragraph documents the dependency on #229/#388's `FLAG_LAYOUT_IN_SCREEN` fix, explaining why this PR's allowlist removal is correct once #388 lands.
- The `[calculateOverlayYPx]` cross-reference nit is fixed: it's now spelled out as plain text (`OverlayManager.calculateOverlayYPx`) rather than an unresolvable link.
- The new `[PrefsManager.getOverlayPosition]` cross-reference resolves correctly: `getOverlayPosition` is a public member of `PrefsManager` (confirmed in `PrefsManager.kt`), and this file already uses the same `[PrefsManager.x]` cross-reference style elsewhere.
- Prose-style fixes (en-dash to hyphen, em-dash to double-hyphen) are correctly applied.
- `python3 -m unittest scripts.test_check_allowed_failures -v` passes (22 tests, OK).
- The allowlist diff is minimal and correct: only `test2a_emptyGalleryNoGreenAfterTap` is removed, with the other two issue-tracked entries untouched.
- The PR title/body now explicitly document the dependency on #388, and `mergeable_state: blocked` is consistent with that (not something this PR needs to resolve on its own).

Diff is now minimal (+6/-3, 2 files) and well-scoped. No further concerns.

**Verdict: LGTM**

aunger:
🤖 Author

CI on this branch (after merging in main, which includes #388's `FLAG_LAYOUT_IN_SCREEN` fix) has run and the result contradicts this PR's premise: [run 27490570874](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27490570874), `build-and-test` failed.

From `e2e-test-results/TEST-com.gb4pc.e2e.GalleryButtonVisualE2ETest.xml`:

- `test1a_overlayShowsBlueAtConfiguredPosition`: **PASS** -- so #388's `FLAG_LAYOUT_IN_SCREEN` fix does correct the position mismatch from #229 (good sign for #388 on its own).
- `test2a_emptyGalleryNoGreenAfterTap`: **still FAILS**, GREEN coverage 88.391045% after tap (was 87.320335% before #388 -- same failure mode, gallery never opens).

The `2a-green-mask`/failure screenshot shows the overlay icon rendered right where it should be (~20%, ~69% of the display, matching `Constants.DEFAULT_X_PERCENT`/`DEFAULT_Y_PERCENT`), and `tapOverlay()` taps that exact computed center point. So the *position* match that #388 fixes looks correct now -- but the tap still doesn't open the gallery; the green camera feed (with overlay icon still visible) is unchanged after the tap.

This means #230 has (at least) two independent causes:
1. The position mismatch from #229, now fixed by #388.
2. Something else preventing `UiDevice.click(x, y)` from reaching `OverlayManager`'s `TYPE_APPLICATION_OVERLAY` window / triggering `handleTap()` -- even when the tap coordinates land in the dead center of the rendered icon.

Given this, removing `test2a_emptyGalleryNoGreenAfterTap` from `.github/allowed-test-failures.txt` in this PR is premature -- it would turn the gate red on a real, currently-reproducing failure. I'm leaving the allowlist removal out of this PR for now and will investigate cause (2) -- likely in `OverlayManager.createLayoutParams()`'s window-flag combination (`FLAG_NOT_FOCUSABLE`/`FLAG_NOT_TOUCH_MODAL` and touch routing for `TYPE_APPLICATION_OVERLAY`) or in how `handleTap()`/`OverlayService` wires up the click -- as a follow-up, likely under a new issue split off from #230 since #388 already closes the #229-shaped part of it.

Will update this PR (or open a new one) once I've confirmed the second cause and have a fix that actually turns `test2a` green in CI.

PR #400, opened 2026-06-15
Grant READ_MEDIA_IMAGES to com.gb4pc (not com.gb4pc.test) to fix captureOnePhoto() timeout

### LINKED ISSUE #231: [GalleryButtonVisualE2ETest] test3a_populatedGalleryShowsGreenAfterTap
## Test Failure

| Field | Value |
|-------|-------|
| Method | `test3a_populatedGalleryShowsGreenAfterTap` |
| Branch | _unknown_ |
| Commit | `4855c4b` |

### Failure message

java.lang.AssertionError: captureOnePhoto(): new image row did not appear in MediaStore within 15 s (even after MediaScannerConnection.scanFile fallback)

### Stack trace

```
java.lang.AssertionError: captureOnePhoto(): new image row did not appear in MediaStore within 15 s (even after MediaScannerConnection.scanFile fallback)
	at com.gb4pc.e2e.E2EFixture.captureOnePhoto(E2EFixture.kt:225)
	at com.gb4pc.e2e.GalleryButtonVisualE2ETest.test3a_populatedGalleryShowsGreenAfterTap(GalleryButtonVisualE2ETest.kt:253)
```

### Links

- _unknown_

---
_Filed automatically by CI on failure of `com.gb4pc.e2e.GalleryButtonVisualE2ETest.test3a_populatedGalleryShowsGreenAfterTap`._

aunger:
Fixes #231. Addresses #232 (partial — only the reported `captureOnePhoto()` timeout is fixed; `test5a` remains a red-light test for the separate #156 secure-camera overlay regression, so **#232 stays open** and is intentionally not auto-closed by this PR).

## Root cause

Both issues report the same failure in `GalleryButtonVisualE2ETest`:

```
java.lang.AssertionError: captureOnePhoto(): new image row did not appear in MediaStore within 15 s (even after MediaScannerConnection.scanFile fallback)
```

`E2EFixture.captureOnePhoto()` sends `ACTION_SHUTTER` to the mock camera (`e2e-mock-camera`, `applicationId = "com.google.android.GoogleCamera"`), which writes a synthetic GREEN JPEG to MediaStore via its own `ContentResolver` (confirmed in CI logcat: `MockCameraActivity: saveJpegToMediaStore: saved content://media/external/images/media/...`). `captureOnePhoto()` then polls `countMediaStoreImages()`, which queries `MediaStore.Images.Media.EXTERNAL_CONTENT_URI`.

Under scoped storage (API 29+), an app without `READ_MEDIA_IMAGES`/`READ_EXTERNAL_STORAGE` can only see MediaStore rows it created itself. The row written by `com.google.android.GoogleCamera` was invisible to the query, so the count never increments and `captureOnePhoto()` times out after 15 s, regardless of code changes elsewhere.

### First attempt (didn't fully fix it)

An earlier commit on this branch granted `READ_MEDIA_IMAGES` to `com.gb4pc.test` (the instrumentation test package) via `appops set`, and switched `E2EFixture`'s MediaStore queries to use `InstrumentationRegistry.getInstrumentation().context` (the `com.gb4pc.test` context), mirroring the existing `e2e-mock-gallery`/`com.gb4pc.mockgallery` precedent.

CI still failed `test3a`/`test5a` with the same timeout. CI logcat showed why: `am instrument` runs the instrumented test code inside the **app-under-test's process and UID** (`com.gb4pc`), not a separate `com.gb4pc.test` process. No `com.gb4pc.test` process is ever started. So the `ContentResolver.query()` call is always made as `com.gb4pc`, regardless of which `Context` object (`context` vs `testContext`) issues it. Granting `READ_MEDIA_IMAGES` to `com.gb4pc.test` had no effect.

## Fix (this PR)

- Add `app/src/debug/AndroidManifest.xml` declaring `READ_MEDIA_IMAGES` (API 33+) and `READ_EXTERNAL_STORAGE` (API 29-32, via `maxSdkVersion`) for **debug builds only**, so production/release builds never declare these permissions.
- Remove `app/src/androidTest/AndroidManifest.xml` (declared the permissions on the wrong package, `com.gb4pc.test`).
- In `connectedE2EAndroidTest`'s `doLast` block, grant `READ_MEDIA_IMAGES` to `com.gb4pc` (not `com.gb4pc.test`) via `pm grant` (the `appops set` approach was a no-op for this runtime permission — see PR comments).
- Revert `E2EFixture`'s `countMediaStoreImages()`/`clearCameraRoll()` to use `context` (`com.gb4pc`, the app-under-test context) instead of `testContext`, matching the package that now holds the permission and that the instrumented test code actually runs as.
- Remove `test3a_populatedGalleryShowsGreenAfterTap` (#231) from `.github/allowed-test-failures.txt` now that the shared `captureOnePhoto()` blocker is fixed. `test5a_secureCameraLockedPopulatedGalleryShowsGreen` remains allowlisted referencing #156 (its residual failure is the secure-camera overlay regression, not `captureOnePhoto()`).

## Tests

- `./gradlew :app:assembleDebug :app:compileDebugAndroidTestKotlin` succeeds; the merged debug manifest (package `com.gb4pc`) includes the new permissions, and the androidTest manifest (package `com.gb4pc.test`) no longer declares them.
- `./gradlew testDebugUnitTest` passes.
- The instrumented `test3a` (now un-allowlisted) is exercised by CI's instrumented E2E suite and passes.

## Before merging

- [x] `test5a_secureCameraLockedPopulatedGalleryShowsGreen` has a doc comment describing it as an intentional red-light test for a separate secure-camera overlay regression (#156). CI confirmed it now fails on the green-coverage assertion (a different failure point than `captureOnePhoto()`); that is #156's scope, not #232's reported timeout, so test5a stays allowlisted under #156 and #232 stays open. See follow-up #406 (cross-package `clearCameraRoll()` cleanup).


aunger (commented):
🤖 Reviewer

LGTM.

I traced the root cause through `E2EFixture.kt`: `countMediaStoreImages()` and `clearCameraRoll()` both query `MediaStore.Images.Media.EXTERNAL_CONTENT_URI` from `com.gb4pc.test`, and under scoped storage (API 29+) a process without `READ_MEDIA_IMAGES`/`READ_EXTERNAL_STORAGE` can't see rows written by other packages, here `e2e-mock-camera` (`com.google.android.GoogleCamera`). That fully explains the `captureOnePhoto()` timeout in both #231 and #232.

The fix mirrors the existing `e2e-mock-gallery`/`com.gb4pc.mockgallery` precedent exactly: new manifest declaring the permission, plus `appops set com.gb4pc.test READ_MEDIA_IMAGES allow` in `connectedE2EAndroidTest`'s `doLast`, placed after `e2eTestApk` is installed (so the package exists when appops targets it, same ordering as the mockgallery grant).

A few things I checked and confirm are fine:
- No manifest merge conflict: `app/src/main/AndroidManifest.xml` doesn't declare `READ_MEDIA_IMAGES`/`READ_EXTERNAL_STORAGE`, so the new `androidTest` manifest merges cleanly.
- `READ_EXTERNAL_STORAGE` with `maxSdkVersion="32"` is a no-op on CI's API 35 image, but the doc comment already says as much ("harmless to request"), so this isn't a hidden inconsistency.
- The "Before merging" checklist item about `test5a`'s doc comment checks out: the existing doc comment at `GalleryButtonVisualE2ETest.kt:380-385` does correctly describe it as a red-light test for the separate secure-camera regression (#156), distinct from the `captureOnePhoto()` timeout this PR fixes.
- Allowlist removal for `test3a`/`test5a` matches the stated rationale.

Small, focused, well-justified diff. No concerns.

aunger:
🤖 Author

CI on the first commit (081e8b9) still failed `test3a`/`test5a` with the same `captureOnePhoto()` MediaStore timeout (now at `E2EFixture.kt:357`, the post-`scanFile` fallback), even though the mock camera's logcat confirmed it wrote the row (`saveJpegToMediaStore: saved content://media/external/images/media/1000000020`).

Root cause of the remaining failure: `081e8b9` correctly granted `READ_MEDIA_IMAGES` to `com.gb4pc.test` via `appops set` and declared it in a new `app/src/androidTest/AndroidManifest.xml`. But `E2EFixture`'s `context` field is `instrumentation.targetContext`, i.e. the context of `com.gb4pc` (the app under test), not `com.gb4pc.test`. `countMediaStoreImages()` and `clearCameraRoll()` ran their `ContentResolver` queries through that `context`, so they executed as `com.gb4pc` -- a package that never declared or held `READ_MEDIA_IMAGES`. The appops grant on `com.gb4pc.test` had no effect on those queries.

Fix (744a602): added a `testContext` (`InstrumentationRegistry.getInstrumentation().context`, i.e. `com.gb4pc.test`) and switched `countMediaStoreImages()` and `clearCameraRoll()` to use it for their `ContentResolver` calls, matching the package the permission grant actually targets.

`./gradlew :app:compileDebugAndroidTestKotlin` and `./gradlew testDebugUnitTest` both pass locally. Pushed to this branch; CI will re-run.

aunger (commented):
🤖 Reviewer

Reviewed both commits (081e8b9, 744a602) plus the manifest addition and build.gradle.kts change.

The root-cause analysis is correct and well-supported: under scoped storage (API 29+), a `ContentResolver` query only returns MediaStore rows the querying package itself owns, so `com.gb4pc.test` needed `READ_MEDIA_IMAGES` (declared in the new androidTest manifest, granted via `appops set` mirroring the existing `com.gb4pc.mockgallery` precedent in build.gradle.kts).

The second commit's fix is correctly scoped: `testContext` (the instrumentation context, i.e. `com.gb4pc.test`) is only substituted for the two `contentResolver` calls in `clearCameraRoll()` and `countMediaStoreImages()`. All other `context` usages (broadcast receiver registration, keyguard, window manager, package manager) correctly remain on the app-under-test context, since those need to act as `com.gb4pc`. I checked for any other `context.contentResolver` call sites that might've been missed; there are none.

The "before merging" checklist item about test5a's doc comment is already satisfied: `GalleryButtonVisualE2ETest.kt` lines 334-348 already document test5a as an intentional red-light test tied to issue #156 with explicit "Do NOT skip" language, so no further action is needed there.

The in-code comments explaining *why* `testContext` exists and why the appops grant alone wasn't sufficient are genuinely useful for future readers; the author's mid-PR comment transparently walks through the first commit's miss and the second commit's fix, which is exactly the kind of debugging trail that should live in PR history.

The `.github/allowed-test-failures.txt` removal of the two entries matches the issues being fixed (#231, #232) and leaves the file's remaining structure intact.

No issues found. LGTM.

aunger:
🤖 Author

CI on 744a602 still failed `test3a`/`test5a` with the same `captureOnePhoto()` timeout. CI logcat for that run shows why: `am instrument com.gb4pc.test/androidx.test.runner.AndroidJUnitRunner` runs the instrumented test code inside the **app-under-test's process and UID** (`com.gb4pc`, e.g. PID 5343 / u0a212 in the logs). No `com.gb4pc.test` process is ever started. So `ContentResolver.query()` from `E2EFixture` is always made as `com.gb4pc`, regardless of whether the code calls `context` or `testContext` -- both `Context` objects share the same process/UID/binder identity. The `appops set com.gb4pc.test READ_MEDIA_IMAGES allow` grant from 081e8b9 was therefore a no-op.

Pushed 09f1a93, which:
- Grants `READ_MEDIA_IMAGES` to `com.gb4pc` (not `com.gb4pc.test`) via `appops set` in `connectedE2EAndroidTest`.
- Declares `READ_MEDIA_IMAGES`/`READ_EXTERNAL_STORAGE` in a new debug-only manifest (`app/src/debug/AndroidManifest.xml`) so the permission exists for the appops grant to act on, without adding it to release builds.
- Removes `app/src/androidTest/AndroidManifest.xml` (declared the permissions on the wrong package).
- Reverts `E2EFixture`'s `testContext` change back to `context`, since `context` (`com.gb4pc`) is the package that now holds the permission and that the test code actually runs as.

`./gradlew :app:assembleDebug :app:compileDebugAndroidTestKotlin` and `./gradlew testDebugUnitTest` both pass locally; the merged debug manifest (package `com.gb4pc`) carries the new permissions. CI will re-run on this push.

aunger:
🤖 Reviewer

I traced through the three commits and the final diff against the current code.

**The fix is sound and internally consistent:**

- `app/src/debug/AndroidManifest.xml` correctly declares `READ_MEDIA_IMAGES` and `READ_EXTERNAL_STORAGE` (`maxSdkVersion=32`), debug-only, so release builds are unaffected.
- `app/src/androidTest/AndroidManifest.xml` is removed, and a search of the tree confirms no remaining references to a `testContext`/`com.gb4pc.test`-context split in `E2EFixture.kt`. `countMediaStoreImages()` and `clearCameraRoll()` both use `context` (`com.gb4pc`) consistently, as the diff claims.
- The new `appops set com.gb4pc READ_MEDIA_IMAGES allow` line in `connectedE2EAndroidTest` runs after the app APK install (line 223), so the manifest-declared permission exists on the device before the grant is applied. Correct ordering.
- This mirrors the existing precedent in `e2e-mock-gallery/src/main/AndroidManifest.xml`, which already declares `READ_MEDIA_IMAGES` for the same "appops grant a declared permission" pattern.
- The allowlist removals match #231/#232 exactly.

The escalating commit history (081e8b9 -> 744a602 -> 09f1a93) and the author's comments do a good job of documenting *why* the first two attempts failed; that context will be useful if this regresses.

**One nit, not a blocker for this PR's stated scope:**

`E2EFixture.clearCameraRoll()`'s doc comment (and its implementation) assumes deleting all rows from `MediaStore.Images.Media.EXTERNAL_CONTENT_URI` "should not occur" to throw `RecoverableSecurityException` on API 33+, reasoning that "the test suite controls all insertions." But the entire premise of this PR is that `e2e-mock-camera` (a different package/UID) inserts the row that `captureOnePhoto()` observes. `READ_MEDIA_IMAGES` is read-only; it doesn't grant `com.gb4pc` delete rights over another app's MediaStore rows. So if `clearCameraRoll()` ever runs while a mock-camera-inserted row is still present on API 33+, that assumption could be violated and the delete could throw. This was a latent assumption before this PR too, so I'm not asking for it to be fixed here, just flagging that this PR is the first change that makes "another package's rows visible to `com.gb4pc`" concretely true, which is exactly the situation that comment says shouldn't arise.

CI (build-and-test, analyze-kotlin) was still in progress as of this review; I didn't block on it.

Overall: LGTM. The reasoning is well-documented, the diff matches the described changes, and I didn't find any inconsistency between the manifest, gradle, and Kotlin changes.

aunger:
🤖 Author

CI on 09f1a93 still failed `test3a`/`test5a` with the same `captureOnePhoto()` timeout (`E2EFixture.kt:357`, the post-`scanFile` fallback), even though the mock camera's logcat again confirmed it wrote the row (`saveJpegToMediaStore: saved content://media/external/images/media/1000000021`).

Root cause of this remaining failure: `READ_MEDIA_IMAGES` is a normal dangerous/runtime permission, not an AppOps-gated special permission like `SYSTEM_ALERT_WINDOW` or `GET_USAGE_STATS`. `appops set com.gb4pc READ_MEDIA_IMAGES allow` only adjusts the AppOps mode for that op; it does not flip the PackageManager-level permission grant state that `ContentResolver`'s permission check (`PermissionChecker`) consults. So even with the permission declared in `app/src/debug/AndroidManifest.xml` (from 09f1a93), `com.gb4pc`'s `EXTERNAL_CONTENT_URI` query was still treated as ungranted and only returned rows owned by `com.gb4pc` itself.

Pushed 1f92af6, which:
- Replaces `appops set com.gb4pc READ_MEDIA_IMAGES allow` with `pm grant com.gb4pc android.permission.READ_MEDIA_IMAGES` in `connectedE2EAndroidTest` -- the same mechanism already used for `CAMERA` (both in this task and in `build.yml`).
- Makes the same `appops set` -> `pm grant` change for `com.gb4pc.mockgallery`'s `READ_MEDIA_IMAGES` grant, which has the identical latent bug (not currently exercised by a passing assertion, since test2a only checks the empty-gallery case, but `LastPhotoActivity` does the same cross-package `EXTERNAL_CONTENT_URI` query).
- Updates the now-inaccurate "granted via appops" comments in `app/src/debug/AndroidManifest.xml` and `e2e-mock-gallery/src/main/AndroidManifest.xml`.

`./gradlew :app:assembleDebug :app:compileDebugAndroidTestKotlin testDebugUnitTest` passes locally. CI will re-run on this push.

aunger (commented):
🤖 Reviewer

I re-traced the final commit (1f92af6) against the rest of the diff and the repo state.

**The fix is sound:**

- `pm grant com.gb4pc android.permission.READ_MEDIA_IMAGES` and `pm grant com.gb4pc.mockgallery android.permission.READ_MEDIA_IMAGES` replace the prior `appops set ... allow` calls, matching the existing `pm grant com.google.android.GoogleCamera android.permission.CAMERA` precedent (both in `app/build.gradle.kts` and `build.yml`). The author's diagnosis (appops mode vs. PackageManager grant state consulted by `PermissionChecker`) is consistent with how `ContentResolver` permission checks actually work.
- `app/src/debug/AndroidManifest.xml` declares both `READ_MEDIA_IMAGES` and `READ_EXTERNAL_STORAGE` (maxSdkVersion 32), and I confirmed the merged debug manifest for `com.gb4pc` carries both, while the `com.gb4pc.test` androidTest manifest does not declare either, confirming the earlier removal of `app/src/androidTest/AndroidManifest.xml` is still correct.
- The comment updates in `app/src/debug/AndroidManifest.xml` and `e2e-mock-gallery/src/main/AndroidManifest.xml` ("granted via appops" -> "granted via `pm grant`") keep the docs in sync with the new mechanism.
- `./gradlew :app:assembleDebug :app:compileDebugAndroidTestKotlin testDebugUnitTest` passes locally.
- The allowlist removals in `.github/allowed-test-failures.txt` still correctly match #231/#232.

**One nit (not a blocker):** the `com.gb4pc.mockgallery` `pm grant` change is, by the author's own comment, fixing a "latent bug... not currently exercised by a passing assertion." That's a reasonable opportunistic fix given it's the same one-line mechanism change as the `com.gb4pc` grant right above it, but it's worth flagging that it's untested by this PR's own test plan.

CI for 1f92af6 is still pending as of this review; I'm not blocking on it, but given this is the fourth attempt at the same root cause, I'd want to actually see `test3a`/`test5a` go green in this run before merge, not just "no regression in the rest of the suite."

The previous review round (before 1f92af6) also flagged a non-blocking observation about `clearCameraRoll()`'s assumption that `com.gb4pc` only ever sees its own MediaStore rows on API 33+, an assumption this PR is the first to make concretely false. That's still unresolved but still not a blocker for this PR's scope.

Overall: LGTM, conditional on CI actually showing `test3a`/`test5a` passing on this commit.

aunger:
🤖 Author

CI on 1f92af6 fixed `captureOnePhoto()`, but introduced two new failures (`test4a`, `test5a`) and changed `test3a`'s failure point. Pushed two fixes (6a34af7, c6ba30d) addressing both.

**test4a/test5a — `clearCameraRoll()` failed its post-condition (`E2EFixture.kt:276`)**

This is exactly the latent issue the Reviewer flagged as a non-blocking nit on the previous review. `ContentResolver.delete(EXTERNAL_CONTENT_URI, null, null)` only ever removes rows owned by the calling package (`com.gb4pc`). The post-delete query was unfiltered, and now that `com.gb4pc` holds a working `READ_MEDIA_IMAGES` grant (from 1f92af6), that unfiltered query also returns rows owned by other packages -- specifically the mock camera's GREEN photo (owned by `com.google.android.GoogleCamera`) captured during `test3a`. `clearCameraRoll()` removed everything it could, but then asserted the unfiltered count was 0 and failed with `expected:<0> but was:<1>`.

Fix (6a34af7): filter both the post-delete query and its expectation to `OWNER_PACKAGE_NAME = com.gb4pc`, matching what `delete()` actually affects. Updated the doc comment accordingly (its old premise -- that cross-package rows "should not occur" -- is exactly what this PR makes concretely true, as the Reviewer noted).

**test3a -- GREEN coverage after tap was 0.0% (`GalleryButtonVisualE2ETest.kt:269`)**

Not a permissions issue. The CI logcat confirms the capture pipeline now works end to end: `saveJpegToMediaStore: saved content://media/external/images/media/1000000020`, followed by `Overlay tapped: locked=false, gallery=com.gb4pc.mockgallery` at 05:19:40.395. But `ActivityTaskManager: Displayed com.gb4pc.mockgallery/.LastPhotoActivity` doesn't fire until 05:19:41.822 -- 1.4 s later. `test3a`'s `fixture.pause(1000)` after the tap takes the screenshot ~400 ms before the gallery activity finishes its cold start (process spawn + MediaStore query + JPEG decode), so it captures whatever was on screen during the transition -- measuring 0% GREEN.

`test2a` has the identical 1.4 s gap but never noticed, because its assertion (`coverage < 10%`) passes trivially on an unrendered screen. `test0` hit this same class of cold-start timing issue for the mock camera and already has a fix for it: `waitForGreenCoverage()`.

Fix (c6ba30d): replace `test3a`'s fixed `pause(1000)` with `fixture.waitForGreenCoverage(minCoverage = 0.40f, timeoutMs = 15_000L)`, following `test0`'s precedent. Its return value is discarded -- it only gates the wait -- and the existing assertion still re-measures full-screen coverage against the original 40% threshold on a fresh screenshot.

**Note on test4a/test5a after this fix**

Both are documented red-light tests for issue #156. Fixing `clearCameraRoll()` lets them proceed past the precondition failure to their actual assertions, which are expected to still fail for the documented #156 reason (secure-camera overlay not rendered/tappable) -- not pass. No assertion requirements were loosened.

`./gradlew :app:compileDebugAndroidTestKotlin :app:assembleDebug :app:testDebugUnitTest` passes locally. Pushed to this branch; CI will re-run.

aunger (commented):
🤖 Reviewer

Reviewing the two commits pushed after the previous LGTM (6a34af7, c6ba30d), which address the test4a/test5a and test3a failures the Author found on 1f92af6.

**6a34af7 (`clearCameraRoll()` fix): correct.**

This resolves exactly the latent issue flagged as a non-blocking nit in the previous review. The post-delete query and assertion are now both filtered to `OWNER_PACKAGE_NAME = ${context.packageName}` (= `com.gb4pc`), matching what `ContentResolver.delete()` actually affects. The updated doc comment accurately reflects the new behavior and explains why the old unfiltered assertion broke once cross-package rows became visible. No issue here.

**c6ba30d (`waitForGreenCoverage` in test3a): correct, follows existing precedent.**

Replacing the fixed `pause(1000)` with `waitForGreenCoverage(minCoverage = 0.40f, timeoutMs = 15_000L)` mirrors `test0`'s handling of the same cold-start timing class. The comment correctly documents that the return value is discarded and the subsequent full-screen `coverageFraction` assertion is the real check. I confirmed `waitForGreenCoverage` measures coverage over a central 60% region while the final assertion measures the full screen — different regions, but that's consistent with how `test0` already uses this helper as a polling gate distinct from the final assertion, so this isn't a new inconsistency.

**One nit, not a blocker:**

`MediaStore.Images.Media.OWNER_PACKAGE_NAME` (`MediaColumns.OWNER_PACKAGE_NAME`) was added in API 29, but `app/build.gradle.kts` sets `minSdk = 26`. CI only runs the API-35 emulator, so this won't surface there, but if `clearCameraRoll()` is ever run against an API 26-28 device/emulator, a query selection referencing a column that doesn't exist on that MediaStore schema could throw at runtime rather than just returning 0 rows. Worth a one-line comment or a `Build.VERSION.SDK_INT` guard at some point, but I agree this doesn't need to hold up this PR given CI's actual target.

**Status:** CI on c6ba30d (the latest commit) was still pending as of this review; I'm not blocking on it per our process, but the Orchestrator should confirm test3a/test4a/test5a/captureOnePhoto() all pass before merge, since this PR's stated goal is specifically to un-allowlist test3a and test5a.

Both new commits are sound, well-reasoned, and consistent with the rest of the PR's documentation style. LGTM.

aunger:
🤖 Reviewer

Re-reviewing after the two follow-up commits (6a34af7, c6ba30d) pushed since my previous LGTM on 09f1a93.

**6a34af7 (clearCameraRoll fix for test4a/test5a) -- introduces a new latent issue for test4a**

The fix itself is locally correct: `delete()` only removes rows `com.gb4pc` owns, so checking the post-condition against `OWNER_PACKAGE_NAME = com.gb4pc` matches what `delete()` actually affects. No argument there.

But this changes `clearCameraRoll()`'s effective behavior: after this PR, a GREEN photo inserted by `com.google.android.GoogleCamera` (test3a's capture) can no longer be removed by any later `clearCameraRoll()` call, because `delete()` never had rights to remove it and the assertion no longer notices it's still there.

`test3a` runs before `test4a` and `test5a` (alphabetical `FixMethodOrder`), and leaves exactly such a row behind. `test4a`'s doc comment (lines ~290-308) explicitly reasons:

> once the regression is fixed (overlay renders correctly in secure-camera mode), the tap succeeds, the gallery opens, and the empty-state screen is black -> coverage < 10% -> assertion PASSES

That reasoning is now false: once `tapOverlay()` succeeds, the gallery is no longer showing an "empty roll" -- it contains test3a's leftover GREEN photo (owned by GoogleCamera, un-deletable by `clearCameraRoll()`). So `test4a` would show GREEN content and `coverage >= 0.10f`, making the assertion **fail** even after #156 is fixed -- for a reason unrelated to #156, and contradicting the test's own documented rationale.

This is currently masked because `test4a` already fails for the #156 reason (overlay blocked, camera feed stays green), so the CI result doesn't change today. But the PR is shipping a doc comment for `test4a` that is no longer accurate about what "PASS" means, and a latent test-correctness bug that will surface as a confusing failure whenever #156 is eventually fixed. Given this PR is the one that makes "another package's MediaStore row survives `clearCameraRoll()`" concretely true (as I noted on the previous round), I think this consequence belongs in this PR's scope rather than being deferred -- at minimum the `test4a` doc comment needs a note about this caveat, or `clearCameraRoll()`/`setUp()` needs a way to clear cross-package rows the test suite itself created (e.g. via `pm clear` or a shell `content delete` as root, run between tests).

**c6ba30d (waitForGreenCoverage in test3a) -- looks correct**

Matches the `test0` precedent, return value is correctly discarded, and the final assertion still re-measures against the original 40% threshold on a fresh screenshot. No issues found here.

**Other**

CI (build-and-test) was still in progress as of this review; I didn't block on it.

Overall: Changes requested -- not for c6ba30d, but 6a34af7 needs the test4a cross-contamination issue addressed (at minimum documented, ideally fixed) before merge.

aunger:
🤖 Author

Pushed 2dca1c8, addressing the Reviewer's changes-requested review on 6a34af7 and the resulting CI failure on c6ba30d.

**Reviewer's requested change (test4a doc comment)**

Added a "Caveat introduced by PR #400" section to `test4a_secureCameraLockedEmptyGalleryNoGreen`'s doc comment (`GalleryButtonVisualE2ETest.kt:308-323`). It documents that `clearCameraRoll()` can only delete rows owned by `com.gb4pc`, so `test3a`'s GREEN photo (owned by `com.google.android.GoogleCamera`) survives into `test4a` and every later test. Today this is harmless because `test4a` fails for the #156 reason regardless of camera-roll contents, but once #156 is fixed this leftover row will make `test4a`'s empty-gallery assumption false and its assertion fail for an unrelated reason. The comment flags this explicitly so a future #156 fix isn't blocked by a confusing failure, and notes that a real fix would need a way to clear cross-package MediaStore rows (e.g. `pm clear` / shell `content delete` as root) — out of scope for #231/#232.

I went with documenting (the "at minimum" option from the review) rather than implementing the cross-package cleanup now, since that cleanup is a separate piece of work with its own design questions (root shell access in CI, whether to scope it to rows created during the test run, etc.) that deserves its own PR rather than being folded into this permissions fix.

**CI gate failure on c6ba30d (test5a)**

CI on c6ba30d (the commit this review was against) showed `test3a` and `test4a` no longer failing, but `com.gb4pc.e2e.GalleryButtonVisualE2ETest#test5a_secureCameraLockedPopulatedGalleryShowsGreen` as the one remaining blocking failure, now failing at `GalleryButtonVisualE2ETest.kt:385` area — the test body's own GREEN-coverage `fail()` call (the "expected baseline failure for the secure-camera overlay regression (issue #156)" message), not `captureOnePhoto()`/`clearCameraRoll()`.

This is exactly the scenario this PR's "Before merging" checklist anticipated: *"If CI shows it still fails on the green-coverage assertion (a different failure point than `captureOnePhoto()`) after this fix, that's a pre-existing issue distinct from #232's reported `captureOnePhoto()` timeout and will need its own follow-up."* The `captureOnePhoto()`/`clearCameraRoll()` preconditions that #231/#232 actually reported are now fixed — `test5a` reaches its real, documented red-light assertion for #156 instead of timing out earlier.

2dca1c8 re-adds `test5a` to `.github/allowed-test-failures.txt` with a comment explaining why (it's #156's intentional red-light marker, not #231/#232's scope) and instructing it be removed only when #156 is fixed. `test3a` (#231) remains un-allowlisted and is expected to pass.

**Verification**

`./gradlew :app:compileDebugAndroidTestKotlin :app:testDebugUnitTest` passes locally. I confirmed the updated allowlist parses correctly via `check_allowed_failures.py`'s loader and yields exactly `com.gb4pc.e2e.GalleryButtonVisualE2ETest#test5a_secureCameraLockedPopulatedGalleryShowsGreen`, matching the CI gate's reported failure. CI will re-run on this push.

**Out of scope for #231/#232**

- The cross-package `clearCameraRoll()` cleanup needed for `test4a`/`test5a` to be meaningfully assertable once #156 lands — needs its own PR/design (root shell `content delete` or `pm clear` between tests).
- `test5a`'s GREEN-coverage assertion failure itself — tracked by #156, a separate and larger fix to the secure-camera overlay rendering path.
- The pre-existing nit about `MediaStore.Images.Media.OWNER_PACKAGE_NAME` requiring API 29+ while `minSdk = 26` (flagged in a prior review round) — not exercised by CI's API 35 target, deferred as before.

aunger (commented):
🤖 Reviewer

Re-reviewing after 2dca1c8, which addresses my previous changes-requested review on 6a34af7.

**test4a doc comment caveat (the requested change) -- addressed**

The new "Caveat introduced by PR #400" section (`GalleryButtonVisualE2ETest.kt:308-323`) accurately documents the cross-package contamination I flagged: `test3a`'s GREEN photo (owned by `com.google.android.GoogleCamera`) survives `clearCameraRoll()` and will make `test4a`'s empty-gallery assumption false once #156 is fixed. It correctly identifies this as a future, not current, problem (today `test4a` fails for the #156 reason regardless), and correctly scopes the real fix (cross-package `pm clear`/`content delete`) as out of scope for #231/#232. I went in expecting to push back on "document vs. fix," but the documentation is specific enough (names the exact row, the exact test ordering dependency, and what to check first) that a future reader hitting this won't be stuck re-deriving it. No further action needed here.

**test5a re-allowlist for #156 -- consistent with the PR's own checklist**

`.github/allowed-test-failures.txt` now allowlists `test5a` under issue #156 (not #231/#232), with a comment explaining the change in failure point (timeout -> documented GREEN-coverage red-light). This matches the "Before merging" caveat in the PR description almost verbatim, and `test3a` (#231) is correctly left off the allowlist as expected-to-pass. The rationale comment is clear about when to remove the entry (when #156 lands).

**Nit: prose-style inconsistency in the new caveat block**

`GalleryButtonVisualE2ETest.kt:308-323` (the new text) mixes dash styles: line 308 uses a real em-dash ("...#231/#232) — re-check..."), while lines 318 and 321 use double-hyphens ("...empty -- it will show..."). The repo's `.claude/rules/prose-style.md` asks for double-hyphens instead of em-dash characters in prose. The surrounding pre-existing doc comments in this file already use em-dashes throughout, so this isn't a new problem this PR created in isolation, but the new block itself is internally inconsistent between the two styles. Not a blocker, just a nit since I'm asked not to skip them.

**Everything else from prior rounds**

I re-checked the manifest/gradle/Kotlin changes from 09f1a93 and the `pm grant` fix from 1f92af6 again against the current tree -- still consistent with what was described, no regressions from the latest commit.

CI (commit 2dca1c8) is still pending as of this review; I'm not blocking on it, but per the PR's own description, `test3a` is now expected to pass and only `test5a` (under its new #156 allowlist entry) should report as a tolerated failure.

Overall: LGTM. The requested change was addressed thoughtfully (the documentation actually engages with the specific scenario rather than a generic disclaimer), and the test5a allowlist change is well-justified and matches the PR's stated scope.

aunger (commented):
🤖 Reviewer

Re-reviewing after 2dca1c8, which addresses my previous changes-requested review.

**test4a doc comment (2dca1c8) -- addresses the cross-contamination issue**

The new "Caveat introduced by PR #400" section (`GalleryButtonVisualE2ETest.kt:308-323`) accurately describes the situation: `clearCameraRoll()` can only delete rows owned by `com.gb4pc`, `test3a`'s GREEN photo (owned by `com.google.android.GoogleCamera`) survives into every later test, and this is currently harmless only because `test4a`/`test5a` fail for the #156 reason regardless. The comment correctly flags that this will need attention once #156 is fixed, and that a real fix is out of scope here. This was the "at minimum" documentation option my prior review offered, and it's been done thoroughly. I verified the doc comment's claims against `clearCameraRoll()`'s implementation and they hold up.

**Allowlist re-attribution (2dca1c8) -- correct**

`test3a` (#231) is fully removed from `.github/allowed-test-failures.txt`. `test5a` remains listed but is now attributed to #156 with a clear explanation distinguishing it from #232's `captureOnePhoto()` timeout (which is fixed). This matches the PR's own "Before merging" contingency, which anticipated exactly this outcome.

**Verification**

- `./gradlew :app:compileDebugAndroidTestKotlin :app:testDebugUnitTest` passes locally on this branch.
- `clearCameraRoll()`'s `OWNER_PACKAGE_NAME = com.gb4pc` filter (6a34af7) is consistent with the `pm grant` mechanism (1f92af6) and the new debug manifest -- I didn't find any remaining inconsistency across `app/build.gradle.kts`, `app/src/debug/AndroidManifest.xml`, `e2e-mock-gallery/src/main/AndroidManifest.xml`, and `E2EFixture.kt`.
- CI's `build-and-test` job was still in progress as of this review (CodeQL, analyze-kotlin, strip-bylines already passed); not blocking on it per the review guidelines.

**Nit (non-blocking)**

The deferred `OWNER_PACKAGE_NAME` requires API 29+ / minSdk=26 question from an earlier round is still unresolved but, as previously agreed, not exercised by CI's API 35 target and reasonably deferred.

Overall: LGTM. The escalating fix history is well documented, the final diff is internally consistent, and the previously requested change (test4a caveat) has been made.

aunger:
Maintainer direction on scope: **keep #232 open.**

CI is green on `2dca1c8`, but only because `test5a` is allowlisted. To be precise about what this PR closes:

- **#231 / test3a** - genuinely fixed and remains un-allowlisted (passing for real).
- **#232 / test5a** - the reported `captureOnePhoto()` timeout is fixed, but test5a still fails on its green-coverage assertion (`GalleryButtonVisualE2ETest.kt:385`), which is the separate #156 secure-camera overlay regression. test5a is re-allowlisted referencing #156, so #232 is only **partially** addressed and should stay open until test5a passes un-allowlisted.

Before merge: the description's `Fixes #232` should be downgraded to a non-closing reference (e.g. `Addresses #232`) so merging does not auto-close #232. `Fixes #231` is correct and can remain.

Issue #401, opened 2026-06-15
ci_monitor.py: step "success" lines and missing per-test markers make Blocked outcomes untraceable


aunger:
## Summary

`scripts/ci_monitor/ci_monitor.py` repeatedly produced misleading or unhelpful output across PR #398 and PR #400, in two related ways:

1. It emitted `step "Run GalleryButtonVisualE2ETest" -> success` (and other step lines as `success`) on runs where that test step's contained, un-allowlisted tests were genuinely FAILING, while the same poll's terminal line was `Blocked`.
2. On several `Blocked` runs, the Monitor emitted no `step` lines and no `FAIL`/`SKIP`/`PASS` per-test markers at all (even with `--include-pass` supplied), so it never surfaced which test failed.

Both symptoms repeatedly misled the orchestration into believing CI was further along, or believing a specific suite had passed, than was actually the case.

### Evidence

- **PR #398, Run C** (commit b7cbe6c; test2a genuinely FAILING, confirmed independently from real GitHub CI logs at 87.32% GREEN coverage, an assertion failure):
  ```
  step "Build and run unit tests" -> success
  step "Run PixelCameraOverlayE2ETest" -> success
  step "Run GalleryButtonVisualE2ETest" -> success
  Blocked
  ```
- **PR #400, Run F** (commit 744a602; test3a/test5a still genuinely FAILING with a `captureOnePhoto()` MediaStore timeout, confirmed independently from real GitHub CI logs):
  ```
  step "Build and run unit tests" -> success
  step "Run PixelCameraOverlayE2ETest" -> success
  step "Run GalleryButtonVisualE2ETest" -> success
  Blocked
  ```

In both cases every reported step is `success`, including `Run GalleryButtonVisualE2ETest`, the step that contains the actually-failing, un-allowlisted test, yet the Monitor's own terminal line on the same run is `Blocked`.

### Root cause (confirmed from code)

`scripts/ci_monitor/ci_monitor.py`, `parse_steps()` (lines 49-77), reads each completed step's `conclusion` from `/actions/runs/{id}/jobs` and emits `step "<name>" -> <conclusion>` for the named test steps (`Build and run unit tests`, any step whose name contains `E2ETest`) or for any step whose conclusion is itself `failure`/`cancelled`/`timed_out`/`action_required`.

[...]

PR #408, opened 2026-06-15
Drain lagging step/FAIL signals before a CI Monitor Blocked/Infra terminal

### LINKED ISSUE #402: CI Monitor reliability bug: step "success" lines and missing per-test markers contradict the Blocked terminal
## Summary

`scripts/ci_monitor/ci_monitor.py` repeatedly produced output that contradicted itself and hid the actual failing test, during the orchestration on PR #398 and PR #400. Two distinct but likely-related symptoms were observed across multiple independent Monitor invocations:

1. **False-positive step conclusions.** The Monitor emitted `step "Run GalleryButtonVisualE2ETest" -> success` (and other named steps as `success`) on runs whose terminal outcome was `Blocked`, even though that step's contained, non-allowlisted tests were genuinely FAILING (independently confirmed from the real GitHub Actions logs).
2. **Missing per-test markers on failing runs.** On other runs that went straight to `Blocked`, the Monitor emitted no `step` lines at all and no `FAIL`/`SKIP`/`PASS` markers, even with `--include-fail` (default) and `--include-pass` (explicit), so the failing test was never identified.

Together, these mean: `step ... -> success` lines cannot be trusted as evidence that the contained tests passed, and a `Blocked` terminal does not reliably surface which test failed via the Monitor's informational signals.

### PR #398

- **Run A** (commit with `test2a` still in `.github/allowed-test-failures.txt`): `in_progress`; `step "Build and run unit tests" -> success`; `in_progress`; `step "Run PixelCameraOverlayE2ETest" -> success`; `Clear (mergeable_state=clean)`. (Expected/healthy shape.)
- **Run B** (predecessor of b7cbe6c, after `test2a` was removed from the allowlist; `test2a` genuinely FAILING): only `in_progress` heartbeats then `Blocked`. No `step` lines, no `FAIL`/`SKIP` markers. A second capture on the same completed run also emitted only `Blocked` with no per-test markers.
- **Run C** (commit b7cbe6c, "diagnostics" round; `test2a` still genuinely FAILING — later confirmed at 87.32% GREEN coverage / an assertion failure by reading the real GitHub CI logs): `step "Build and run unit tests" -> success`; `step "Run PixelCameraOverlayE2ETest" -> success`; `step "Run GalleryButtonVisualE2ETest" -> success`; `Blocked`. Every named step reported `success`, including `GalleryButtonVisualE2ETest`, which contains the un-allowlisted, actually-failing `test2a`, yet the terminal was `Blocked`.
- **Run D** (commit b2270e3, the real fix; genuinely green): `in_progress` x3 then `Clear (mergeable_state=clean)`. No `step` lines (none of the three named steps happened to complete during the polled window before `Clear`).

### PR #400 (branch `fix/issue-231-232-unallowlist`; `test3a`/`test5a` removed from the allowlist)

- **Run E** (commit 081e8b9; `test3a`/`test5a` genuinely failing on a `captureOnePhoto()` MediaStore timeout): `in_progress` x2 then `Blocked`. No `step` lines, no markers. A capture re-run on the same run: `Blocked` only.
- **Run F** (commit 744a602; later confirmed via the real GitHub CI logs that `test3a`/`test5a` were STILL failing the same `captureOnePhoto()` timeout): `step "Build and run unit tests" -> success`; `step "Run PixelCameraOverlayE2ETest" -> success`; `step "Run GalleryButtonVisualE2ETest" -> success`; `Blocked`. Same pattern as Run C: every named step `success` including the one containing the failing tests, yet `Blocked`.
- **Run G** (commit 09f1a93): three separate Monitor invocations (including dedicated capture re-runs) each produced `in_progress` x3 then `Blocked`, with **no** `step` lines and **no** `FAIL`/`SKIP`/`PASS` markers, even with `--include-pass` supplied. For this commit it was independently verified with local git that the branch was fully up to date with `main` (`git rev-list --count origin/fix/issue-231-232-unallowlist..origin/main` = 0, `git merge-base --is-ancestor origin/main origin/fix/issue-231-232-unallowlist` = true, `git merge-tree` reported 0 conflicts), so `mergeable_state` was not `behind`/`dirty`. Per the README's own definition of `Blocked` ("a check failed (failure/action_required) OR mergeable_state is behind/dirty"), this `Blocked` must reflect a genuinely failed check — yet the Monitor surfaced no failing step and no per-test marker identifying which test failed.

## Relevant code

- `scripts/ci_monitor/ci_monitor.py`, `parse_check_result()` (around lines 157-176): determines `Clear`/`in_progress`/`Infra`/`Blocked`/`all_passed` from `/repos/{owner}/{repo}/commits/{sha}/check-runs`.
- `scripts/ci_monitor/ci_monitor.py`, `parse_run_id()` (around lines 179-184) and `parse_steps()` (around lines 49-77): determine the single `run_id` from `/actions/runs?head_sha={sha}&event=pull_request&per_page=5` (first non-cancelled run) and then read step conclusions from `/actions/runs/{run_id}/jobs`.
- `scripts/ci_monitor/ci_monitor.py`, main loop (around lines 443-544): for each poll, `check_json`/`result` (terminal signal) is fetched first; then `runs_json`/`jobs_json`/`artifacts_json` (informational `step`/`FAIL`/`SKIP`/`PASS` signals) are fetched and emitted via `emit_block`; finally, if `result in ("Blocked", "Infra")`, the terminal line is printed and the loop **breaks immediately** — there is no further poll.
- `.github/workflows/build.yml`: the `Run PixelCameraOverlayE2ETest` and `Run GalleryButtonVisualE2ETest` steps (around lines 359-434) deliberately do **not** use `continue-on-error`. Instead each wraps `./gradlew connectedE2EAndroidTest ...` in `if ... ; then echo "outcome=success" >> "$GITHUB_OUTPUT"; else echo "outcome=failure" >> "$GITHUB_OUTPUT"; fi` — so the **shell step itself always exits 0** regardless of whether the contained test passed. The actual pass/fail verdict is deferred to the final `Gate on test failures` step (around lines 524-536), which is the step that actually fails the job/check-run when a non-allowlisted test fails (per `.github/workflows/CLAUDE.md`'s description of the red-to-green allowlist design from issue #309).

### Confirmed from code

- **Symptom 1's "success" is technically accurate but misleading.** Because `Run GalleryButtonVisualE2ETest` (and `Run PixelCameraOverlayE2ETest`) never propagate a failing exit code — they capture `outcome=failure` into a step output instead — the GitHub Actions API genuinely reports `conclusion: success` for these steps even when the test they ran failed. `parse_steps()` faithfully relays this `success` conclusion. The actual failure is only visible in the separate `Gate on test failures` step's conclusion, which `parse_steps()` would also emit (its `concl` would be `failure`, matching the `concl in ("failure", ...)` branch) — but in Runs C and F, no such line was ever observed before `Blocked`.
- **The loop breaks immediately on `Blocked`/`Infra`/`Clear`** (lines ~535-538 and ~512-523), with no subsequent poll. Any informational signal (step conclusion or `testresults-*` artifact) that has not yet become visible via its respective endpoint at the moment `result` resolves to a terminal value will never be emitted, because there is no later iteration to catch up.

### Best-supported hypothesis (not fully confirmed from code alone)

The single most parsimonious explanation unifying both symptoms is **cross-endpoint consistency lag combined with the immediate `break`**:

- `parse_check_result()` reads `/commits/{sha}/check-runs`, while `parse_steps()`/artifact fetching read `/actions/runs/{run_id}/jobs` and `/actions/runs/{run_id}/artifacts` for a `run_id` chosen independently via `/actions/runs?head_sha={sha}...`. These are different API surfaces and may not update in lockstep.
- On the poll where `check-runs` first reflects the job's overall `conclusion: failure` (-> `Blocked`, triggering the `break`), the `/actions/runs/{run_id}/jobs` view may still show the final `Gate on test failures` step as not yet `status: completed` (and/or the `testresults-e2e-gallery` artifact as not yet visible via `/actions/runs/{run_id}/artifacts`). `parse_steps()` skips any step with `status != "completed"` (line ~63), so the failing gate step is silently dropped on that poll — and there is no later poll to pick it up, because the loop has already broken.
- This explains Run C/F (the three earlier-completing named steps were already captured as `success` on prior polls; the final failing gate step's `failure` conclusion never becomes visible before the terminal `Blocked` ends the loop) and Runs B/E/G (if `check-runs` flips straight from `in_progress` to `failure` between polls without any named test step yet showing as `completed` in the jobs view, no `step` lines appear at all, and the `testresults-*` artifacts likewise may not yet be listed).
- This hypothesis is **not independently confirmed** here (it would require correlating raw, timestamped API responses across the lagging poll and the terminal poll for the same run); it is presented as the best-supported explanation given the code paths and the observed symptom pattern, in case it points investigators in the right direction.

## Impact

- Orchestration sessions relying on the Monitor's `step "..." -> success` lines as evidence that a named E2E test step's contained tests passed are misled: `success` here only means the wrapper shell script exited 0, not that the tests passed.
- Orchestration sessions relying on the Monitor to identify *which* test caused a `Blocked` terminal get no information at all in the majority of observed failing runs (B, E, and all three captures of G), forcing a fallback to manually reading the raw GitHub Actions logs — defeating the purpose of the per-test marker signal described in the README.
- Run G in particular demonstrates this is not limited to merge-state blocks (`behind`/`dirty`): a genuine check failure on a cleanly mergeable, up-to-date branch produced `Blocked` with zero diagnostic informational lines across three separate Monitor invocations.

aunger:
Fixes #402.

## Problem

`scripts/ci_monitor/ci_monitor.py`'s poll loop broke immediately on a `Blocked`/`Infra` terminal, derived from `/commits/{sha}/check-runs`, without giving `/actions/runs/{id}/jobs` (per-step conclusions) or `/actions/runs/{id}/artifacts` (`testresults-*` per-test markers) a chance to catch up. Both endpoints can lag behind `check-runs`. This produced two contradictory symptoms observed across PR #398 and PR #400:

1. Every named E2E test step reported `step "..." -> success` (because the wrapper shell script always exits 0 by design), with `Blocked` as the terminal and no `failure` line from the actual gating step (`Gate on test failures`).
2. Some `Blocked` terminals emitted zero `step`/`FAIL`/`SKIP`/`PASS` lines at all, even with `--include-pass`, because neither the jobs view nor the artifacts list had caught up before the loop broke.

## Fix

- Added `drain_then_print(sha, terminal_line)`: before printing a `Blocked`/`Infra` terminal line, re-poll the step/artifact signals (`poll_signals`), retrying up to `DRAIN_MAX_ATTEMPTS` (3) times with a `DRAIN_DELAY_SECONDS` (5s) pause between attempts, stopping early as soon as an attempt emits something new. Any step delta or `FAIL`/`SKIP`/`PASS` marker that was still lagging on the poll that produced the terminal result now gets a chance to be emitted before the loop ends, even if the lag spans more than one poll.
- Refactored the per-poll step/artifact signal fetching into a shared `poll_signals(sha)` helper, used both by the main loop and by the new drain step. `poll_signals` returns whether it emitted anything; `drain_then_print` uses this to decide whether to keep retrying.
- If every drain attempt comes up empty, the Monitor prints `drain poll found no new diagnostic signals` immediately before the `Blocked`/`Infra` terminal line, so a terminal with no diagnostics is explicitly flagged rather than looking identical to a clean (if undiagnosed) terminal.
- Updated `scripts/ci_monitor/README.md`:
  - Documented that the named E2E test steps (`Run *E2ETest`) always conclude `success` by design (the wrapper records `outcome=success|failure` to `$GITHUB_OUTPUT` but always exits 0), and that the real pass/fail verdict comes from the `Gate on test failures` step and the per-test markers. This folds in the README documentation gap noted in the duplicate issue #401.
  - Documented the new bounded-retry drain behavior, the new diagnostic line, and that a lag outliving the bounded retry (issue #402's Run G, which spanned multiple Monitor process invocations) still surfaces as this flagged, undiagnosed terminal.

### Response to review

- **2 (unexplained `DRAIN_DELAY_SECONDS = 5`)**: the constant is unchanged, but it's no longer a single roll of the dice — `DRAIN_MAX_ATTEMPTS = 3` means up to 15s of cumulative lag is now covered, and a comment in the code explains the reasoning (Runs B/C/E/F/T needed only one attempt; this gives headroom for slightly longer lags without unbounded delay).
- **3 (`poll_signals`'s return value unused)**: now consumed by `drain_then_print` to decide whether to keep retrying and whether to print the new "found no new diagnostic signals" line.
- **4 (no fallback diagnostic when the drain comes up empty)**: addressed by the new `drain poll found no new diagnostic signals` line.
- **1 (Run G's multi-process lag)**: not fully solved by a bounded in-process retry (Run G's lag spanned multiple separate Monitor invocations over minutes, which no single-process retry can cover). The new diagnostic line makes this case explicit and distinguishable from a coincidental empty terminal, so the Orchestrator/user knows to fall back to reading the raw GitHub Actions logs rather than treating a bare `Blocked` as fully diagnosed. Fully closing Run G's gap would need a different mechanism (e.g., the Orchestrator re-running the Monitor after a longer delay, or a persistent poll outside the process lifetime); flagged here as a known residual limitation rather than addressed.

## Tests

- Updated every existing `main()`-driven test case that ends in `Blocked`/`Infra` (groups i, j, m, p, q, s/s2/s3/s4) to account for the bounded drain retries (each attempt finds nothing new here, so all 3 attempts run) and the new `drain poll found no new diagnostic signals` line.
- Group (t): reproduces the Run B/E/G shape from the issue, a `Blocked` terminal where `check-runs` flips to `failure` on poll 1 while `jobs`/`artifacts` have not yet caught up; the first drain attempt surfaces the `Gate on test failures` step failure and the per-test `FAIL` marker before the `Blocked` line.
- Added group (u): reproduces a two-poll lag, where drain attempt 1 still finds nothing new and only attempt 2 surfaces the step failure and FAIL marker, confirming the bounded retry recovers lags longer than one poll and that the "found no new diagnostic signals" line is correctly suppressed when a later attempt does find something.
- `GITHUB_TOKEN=fake python3 scripts/test_ci_monitor.py` (run locally; not wired into the build CI) → 147 passed, 0 failed.


aunger (commented):
🤖 Reviewer

Ran `GITHUB_TOKEN=fake python3 scripts/test_ci_monitor.py` locally: 134 passed, 0 failed, confirming the PR description's test claim.

The refactor itself is clean: `poll_signals(sha)` is a sensible extraction shared by the main loop and the new `drain_then_print`, and the README updates correctly fold in both the always-`success` E2E wrapper caveat and the new drain-before-terminal behavior. The new test group (t) is a good reproduction of the Run C/F/B/E "single-poll lag" shape, and the bookkeeping updates to groups i/j/m/p/q/s/s2/s3/s4 for the extra drain-poll requests look consistent.

A few concerns:

1. **The fix doesn't address the strongest evidence in the issue (Run G).** Run G reported `Blocked` with zero `step`/`FAIL`/`SKIP`/`PASS` lines across **three separate Monitor process invocations**, including dedicated capture re-runs, i.e., the lag persisted across multiple fresh processes over what was presumably minutes, not one extra poll's worth of time. A single `DRAIN_DELAY_SECONDS = 5` re-poll within the same process addresses a one-poll lag (matching Runs B/C/E/F's shape, and the new test group (t)), but there's nothing in the PR explaining why this is expected to also resolve Run G's shape, where the lag apparently outlived multiple independent process lifetimes. If the root cause for Run G is actually different (e.g., `parse_run_id` picking a run whose jobs/artifacts never populate, rather than a short-lived lag), this PR doesn't touch that path and Run G could still reproduce after this fix.

2. **`DRAIN_DELAY_SECONDS = 5` is an unexplained, seemingly arbitrary constant.** No measurement, citation, or reasoning is given for why 5 seconds is enough for `/actions/runs/{id}/jobs` and `/actions/runs/{id}/artifacts` to catch up with `/commits/{sha}/check-runs`. If it's too short, the drain poll just reproduces the same "nothing new" result and the issue persists with an extra 5-second delay tacked on.

3. **`poll_signals`'s return value (`emitted[0]`) is unused.** `drain_then_print` calls `poll_signals(sha)` but discards the boolean. Either use it (e.g., to note explicitly when the drain poll found nothing new, so a `Blocked` terminal with zero diagnostics can be distinguished from "genuinely nothing new to report" vs. "still lagging after the drain"), or remove the return value as dead code.

4. **No fallback diagnostic when the drain poll comes up empty.** If `Blocked` still emits zero step/FAIL lines after the drain (the Run G shape), the Monitor silently prints bare `Blocked`, exactly the symptom-2 failure from the issue, just delayed by 5 seconds. A line like "drain poll found no new diagnostic signals" would at least flag that distinction to the Orchestrator/user instead of looking identical to a clean (if undiagnosed) `Blocked`.

These aren't blocking nits about code correctness (the implemented single-poll-lag fix is correct and well-tested for that case), but the PR title and description position this as fixing issue #402 broadly, and #402's headline evidence (Run G) isn't clearly covered.

aunger (commented):
🤖 Reviewer

Ran `GITHUB_TOKEN=fake python3 scripts/test_ci_monitor.py` locally: 147 passed, 0 failed, matching the PR description's updated test claim.

This round's commit (0170f1a) addresses all four points from the prior review:

1. The new "drain poll found no new diagnostic signals" line and the honest framing in the PR description and README (Run G's multi-process lag is a "known residual limitation," not claimed as fixed) is the right call. It doesn't overclaim, and gives the Orchestrator/user a clear signal to fall back to raw logs when this line appears.
2. `DRAIN_DELAY_SECONDS = 5` now has a comment explaining the reasoning (3 attempts x 5s = 15s, covering the observed single-poll-lag runs), with an explicit acknowledgment that a longer lag still falls through to point 1's flagged line. Good enough given the constant is still a guess, but now a documented one.
3. `poll_signals`'s return value is now consumed by `drain_then_print` to decide whether to keep retrying and whether to print the "found no new diagnostic signals" line. Dead code concern resolved.
4. The new test groups (t) and (u) are good: (t) reproduces the one-poll-lag shape from Runs B/C/E/F, and (u) specifically reproduces a two-poll lag and confirms the bounded retry recovers it and correctly suppresses the "found nothing new" line when a later attempt succeeds.

The refactor (`poll_signals` shared between the main loop and `drain_then_print`) is clean, and the bookkeeping updates across the other test groups (i/j/m/p/q/s/s2/s3/s4) for the extra drain-poll request counts look correct and consistent with the new code path.

One nit: `agents/dev_orchestration.md` (lines 149-150 and 174) documents which Monitor output lines the Orchestrator relays and how (`step`/`FAIL`/`SKIP`/`PASS` as informational deltas; `Blocked`/`Infra`/`Clear` as terminals). This PR adds a new line type, `drain poll found no new diagnostic signals`, and updates `scripts/ci_monitor/README.md`'s outcome vocabulary table for it, but doesn't mention it in `dev_orchestration.md`. Since the new line is informational and immediately precedes a terminal line that's already handled by the existing rules, this is low-stakes, but worth a one-line addition to keep the Orchestrator's documented line-handling rules in sync with the README's outcome vocabulary, the way the rest of that table is kept in sync.

That's not blocking; the core fix is correct, well-tested, and the PR's self-assessment of its own limitations is accurate.

aunger (commented):
🤖 Reviewer

Ran `GITHUB_TOKEN=fake python3 scripts/test_ci_monitor.py` locally: 147 passed, 0 failed, matching the PR description's claim.

CI on the head commit is green: `build-and-test` completed successfully, CodeQL passed, no failures.

All four points from the first review round are addressed. The second review round's nit (document the new `drain poll found no new diagnostic signals` line type in `dev_orchestration.md`) is also addressed by the third commit, which adds the `undiagnosedTerminal` routing block and the out-of-process 5-minute recheck pattern.

The refactor is clean: `poll_signals` is a well-scoped extraction used consistently by both the main loop and `drain_then_print`, and the `seen_steps`/`seen_arts`/`seen_fails` sets flow through correctly (they are the same sets in both callers, so the drain cannot re-emit something the main-loop poll already emitted, and an artifact seen by the drain is marked seen so it is not re-downloaded on the next poll -- which in this code path does not exist, but the invariant is still upheld).

One issue worth naming, though it does not change my verdict:

**`drain_then_print` stops retrying after the first attempt that emits anything.** If the gate step becomes visible on drain attempt 1 but the `testresults-*` artifact only resolves on drain attempt 2 (two endpoints lagging by different amounts), the FAIL marker is silently dropped for this process's lifetime. The `seen_arts` set would not be populated, so a fresh Orchestrator recheck would re-fetch the artifact and surface the FAIL, but within this process the information is lost. In practice, the observed failure cases (Runs B/C/E/F/G) show steps and artifacts lagging together, and the test group (u) explicitly reproduces a two-attempt lag for both signals simultaneously, so this partial-lag case is not evidenced from the issue. The conservative fix would be to continue retrying after a fruitful attempt (up to `DRAIN_MAX_ATTEMPTS`) rather than breaking immediately, but that changes the semantics the test groups verify and is not required by anything in the issue. Noting it as a known gap, not requesting a change here.

The commit organization is consistent with `code_edit.md`'s guidance: the initial fix commit bundles the behavior change, README updates, and test updates together (all part of one concern); the second commit bundles the bounded-retry additions, the `poll_signals` return-value use, and the empty-drain diagnostic line with corresponding test updates; the third commit is purely an Orchestrator-side procedural change in `dev_orchestration.md` with no code changes. That's three well-bounded concerns in three commits, none of which are pure refactors mixed with behavior changes.

No prose-style violations introduced: all new documentation is ASCII, using straight quotes and hyphens/parentheses rather than smart quotes or em dashes in the new paragraphs. The `->` arrows in `dev_orchestration.md` are pre-existing routing pseudocode symbols, exempt per the style rule.

LGTM.

PR #420, opened 2026-06-17
Break poll loop on Clear with no check runs

### LINKED ISSUE #415: ci_monitor.py poll loop appears to repeatedly emit the same Clear terminal line instead of exiting
## Summary

While running the CI Monitor for PR #413 (`python3 scripts/ci_monitor/ci_monitor.py --pr 413`, via the `Monitor` tool, task id `bpx6walvc`, `timeout_ms: 1800000`), the script repeatedly emitted the same terminal line, `PR#413: Clear`, as separate notification events, instead of emitting it once and exiting.

## Evidence

- The working tree was on branch `fix/issue-410-run-g-multi-process-lag` at commit `9fa222d`, which includes the #408 drain-fix commits to `ci_monitor.py` (`62192b7`/`776dfd0`, +157/-46 lines vs `origin/main`). So this was observed against the #408-fixed version of `ci_monitor.py`, not a stale pre-fix checkout.
- At least 4 separate `PR#413: Clear` notification events were delivered for the same monitor task before the Orchestrator called `TaskStop` to end it.

## Possible cause

Per issue #402's description of the main loop (around lines 443-544 at the time): "if `result in ('Blocked', 'Infra')`, the terminal line is printed and the loop **breaks immediately**". This phrasing implies the loop does not break for other terminal states such as `Clear`, so the script keeps polling and re-printing `Clear` on every subsequent poll, up to its own timeout or the `Monitor` tool's `timeout_ms`.

## Impact

- Each repeated `Clear` line arrives as a fresh task-notification event in the orchestration session, even though the Orchestrator already acted on the first one.
- The Orchestrator must remember to manually stop the monitor task after the first `Clear`, or it keeps emitting events (and consuming the poll budget) until the 30-minute timeout.

## Suggested follow-up

Consider whether the main loop should also break (or at least suppress repeat emissions of an unchanged terminal line) once a `Clear` result is reached, consistent with how `Blocked`/`Infra` already break the loop immediately.

## Note

This is a behavioral observation from an Orchestrator session, not a diagnosis; filing per `agents/dev_orchestration.md`'s guidance to report infrastructure gaps rather than work around them silently.

aunger:
Fixes #415

## What changed

`parse_check_result` returns `'Clear'` when `total_count == 0` (no CI check runs are registered for the commit). Before this fix, the main poll loop handled that result in the `elif result is not None:` catch-all, which printed `PR#N: Clear` but did not `break`. The loop then slept 30 seconds and repeated, emitting a fresh `Clear` line on every subsequent poll until the 30-minute Monitor timeout.

The fix adds an explicit `elif result == 'Clear':` branch before the catch-all. It prints the terminal line and breaks immediately. No drain is needed: with no check runs there are no lagging step/FAIL signals to surface.

The existing `Clear` path via `all_passed` (where `parse_check_result` returns `'all_passed'` and `mergeable_state` is `clean` or `unstable`) already had its own `break` and is unaffected.

## Test coverage

New test (v) in `scripts/test_ci_monitor.py`: a single poll with `total_count=0` check-runs emits exactly one `Clear` terminal line, consumes exactly the three expected mocked requests (pulls, check-runs, runs), and exits with rc 0. All 150 tests pass (147 existing + 3 new).

### DIFF
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -601,6 +601,12 @@ def main(argv):
             # terminal line; see drain_then_print and DRAIN_DELAY_SECONDS.
             drain_then_print(sha, "PR#%s: %s" % (pr, result))
             break
+        elif result == "Clear":
+            # No check runs registered (total_count == 0) — PR is already clear.
+            # Break immediately; no drain needed (there are no failing signals).
+            print("PR#%s: Clear" % pr)
+            sys.stdout.flush()
+            break
         elif result is not None:
             print("PR#%s: %s" % (pr, result))
             sys.stdout.flush()
--- a/scripts/test_ci_monitor.py
+++ b/scripts/test_ci_monitor.py
@@ -26,6 +26,7 @@ Covers:
   (s) #260 CLI flags: _parse_outcome_filters and main() pass filter flags through
   (t) #402 Gap E: drain_then_print surfaces a step/FAIL that lags one poll behind Blocked
   (u) #402 Gap E (review): drain_then_print's bounded retry recovers a two-poll lag
+  (v) #415: Clear from parse_check_result (no checks) breaks the loop exactly once
 
 No network calls required; no GITHUB_TOKEN needed.
 Always exits 0 on success, non-zero on failure.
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

## Verdict: LGTM

The fix is correct and the test coverage is solid. A few notes follow.

### The fix itself

`parse_check_result` has always returned `'Clear'` when `total_count == 0`, but the main loop had no explicit handler for that token, so it fell through to `elif result is not None:` -- which printed the line but did not `break`. The new `elif result == 'Clear':` branch is placed before the catch-all and breaks immediately. No drain is warranted here (there are no check runs, so there are no lagging signals to surface), and the comment explains that clearly.

The placement in the if/elif chain is correct: it comes after the `all_passed` branch (which produces its own `Clear` printout via a different code path) and before the generic catch-all.

### On the issue comment ("might have already solved it")

I checked: the base commit (`a8fc540`, the #408 drain fix) did not add any handling for the `total_count == 0` `'Clear'` result. The `elif result is not None:` catch-all remained unchanged. The issue was live at the time of the PR branch, and this fix is still needed.

### Test (v)

The test correctly sets up a three-entry deque, asserts exactly one `Clear` line is printed, asserts the deque is fully consumed (proving the loop exits after the first poll), and asserts `rc == 0`. The deque entry count matches the actual call order: `fetch_pr_with_retry` -> `_request(check-runs)` -> `poll_signals` -> `_request(runs)`.

One nit in the test comment (no behavior impact): the inline annotation on `RUNS_EMPTY_V` says "before check-runs result is evaluated", but `poll_signals` is called *after* `check_json` is fetched and `result` is computed (lines 550-560 of `ci_monitor.py`). The deque count (3) and the test assertions are both correct; only that one phrase in the comment is misleading.

Issue #406, opened 2026-06-15
(re PR #400) Design cross-package clearCameraRoll() cleanup so test4a/test5a are meaningfully assertable once #156 is fixed


aunger:
Per the author's comment in https://github.com/aunger/gallery-button-for-pixel-camera/pull/400#issuecomment-4704988104 (under "Out of scope for #231/#232"):

> The cross-package `clearCameraRoll()` cleanup needed for `test4a`/`test5a` to be meaningfully assertable once #156 lands — needs its own PR/design (root shell `content delete` or `pm clear` between tests).

## Background

PR #400 grants `com.gb4pc` `READ_MEDIA_IMAGES` via `pm grant`, which fixes the `captureOnePhoto()` MediaStore-visibility timeout (#231/#232). A side effect: `ContentResolver.delete(EXTERNAL_CONTENT_URI, ...)` in `clearCameraRoll()` only removes rows owned by `com.gb4pc` itself, so MediaStore rows written by other packages (e.g. `com.google.android.GoogleCamera`'s GREEN photo captured during `test3a`) survive `clearCameraRoll()` and persist into later tests (`test4a`, `test5a`, ...).

This is documented as a "Caveat introduced by PR #400" in `GalleryButtonVisualE2ETest.kt` (around lines 308-323, added in commit 2dca1c8). Today it's harmless because `test4a`/`test5a` fail for the unrelated #156 reason regardless. But once #156 (secure-camera overlay regression) is fixed, this leftover cross-package row will make `test4a`'s "empty gallery" assumption false, causing its assertion to fail for a reason unrelated to #156.

## What's needed

A design and implementation for clearing cross-package MediaStore rows created during the E2E test run, for example:
- Root shell `content delete` between tests, or
- `pm clear` of relevant packages between tests, or
- Some other mechanism scoped to rows created during the test run.

This needs its own design discussion (root shell access in CI, scoping considerations) and its own PR, per the author's comment. It should be resolved before (or in tandem with) #156, so that fixing #156 doesn't immediately introduce a new, confusing `test4a`/`test5a` failure.

## Related

- #156 (secure-camera overlay regression — test5a's current red-light reason)
- #231, #232 (fixed/partially fixed by PR #400, which introduced this caveat)

---

[...]

PR #424, opened 2026-06-17
Add non-blocking follow-on track to Verification Planner

### LINKED ISSUE #407: Verification Planner cannot file non-blocking follow-up issues (every filed issue is treated as a merge blocker)
## Problem

The Verification Planner role (`agents/verification_planning.md`) has no way to file a follow-up issue that does not block the PR. By construction, every issue it files is a merge blocker:

- **Role (lines 5-7):** it only hunts for "outstanding requirements that must be handled before merging."
- **The before-merging list (lines 16-20):** "the PR may be merged only once every item on it is resolved," and each filed issue "blocks the PR until done."
- **Step 3c (lines 34-42):** every new issue is recorded as `blocked_by` against the PR, with fallbacks all the way down to a plain-text "Blocks PR #..." line in the issue body.
- **Step 2 (line 27):** the only non-blocking outcome is the empty-list case, where the Planner files nothing and applies the `verified` label.

So the two categories it recognizes (unautomated verification steps; changes outside the repo) are both blocking buckets. Pure follow-on work has nowhere to go: the Planner must either mis-stamp it as a blocker, or drop it to avoid that (losing it). Both are bad.

## Live example

On PR #400, the Planner filed #406 (cross-package `clearCameraRoll()` cleanup). That work is explicitly out of scope for #231/#232 and is deferred to its own PR (see https://github.com/aunger/gallery-button-for-pixel-camera/pull/400#issuecomment-4704988104), so it should not block #400. But the Planner stamped it "Blocks PR #400" because that is the only category it has. The blocking framing had to be removed from #406 by hand.

## Proposed fix

Add a parallel, non-blocking track to `agents/verification_planning.md`:

1. Have the Planner assemble two lists: a *before-merging (blocking)* list and a *follow-on (non-blocking)* list, classifying each finding into one. Deferred / "out of scope, needs its own PR" items (like #406) go in the follow-on list.
2. File a tracking issue for follow-on items too (so nothing is lost), using the same title convention and originating-comment link, but skip the `blocked_by` dependency call and state in the body that it does not block the PR. A sub-issue link is optional.
3. Make the merge gate and the `verified` label depend only on the *blocking* list being empty, even when follow-on issues were filed.
4. Report both lists to the Orchestrator, clearly separated.

## Related

- #400 (PR where this surfaced)
- #406 (the follow-on issue that was mis-stamped as blocking)

aunger:
Fixes #407

Previously the Verification Planner had a single before-merging (blocking) list.
Every finding was either stamped as a merge blocker or silently dropped--there was no third option.
This caused #406 (a deferred cleanup item) to be mis-stamped "Blocks PR #400" on PR #400, requiring a manual correction.

## Changes

`agents/verification_planning.md` gains a parallel follow-on (non-blocking) track:

- **Step 1** now asks the Planner to assemble two lists: the existing *before merging* list plus a new *follow-on* list for work that is explicitly deferred or out of scope.
- **New step 4** instructs the Planner to file a tracking issue for each follow-on item using the same title convention, but to omit the `blocked_by` dependency call and to state in the issue body that it does not block the PR.
- **Step 2** (empty blocking list = `verified`) now explicitly notes that a non-empty follow-on list does not prevent applying the `verified` label.
- **Step 5** (was step 4) instructs the Planner to report both lists to the Orchestrator, making clear that only the blocking list gates the merge.
- The former "The before-merging list" section is replaced by "Two lists, not one," which introduces the two tracks and states that only the blocking list controls the merge gate.

### DIFF
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -13,18 +13,26 @@ Two kinds of outstanding requirement are your responsibility:
 2. **Changes outside the repo**: requirements that are not satisfied by any change to a file in the repo, such as an issue that needs to be filed, a setting that must be changed in an external system, or a manual operational step.
    These are easy to lose because the review process is centered on file changes; surfacing them is explicitly part of your job.
 
-## The before-merging list
+## Two lists, not one
 
-Assemble a single *before merging* list that names every outstanding requirement you find, of either kind above.
-This list is the deliverable the merge decision depends on: the PR may be merged only once every item on it is resolved.
-You file one tracking issue per item (see **What to do**) so that each requirement is followed up and blocks the PR until done.
+Classify each finding into exactly one of two tracks:
+
+- **Before-merging (blocking)**: requirements from the two kinds above that are in scope for this PR and must be resolved before it can be merged.
+- **Follow-on (non-blocking)**: work that is explicitly deferred, out of scope for this PR, or otherwise not a condition of merging (for example, cleanup in another package that the PR explicitly deferred to a follow-up).
+
+Both lists are deliverables.
+File a tracking issue for every item in either list so nothing is lost.
+Only the before-merging list controls the merge gate.
 
 ## What to do
 
 1. Read the issue description, PR description, and all comments on both.
    Look for both kinds of outstanding requirement described under **Role**: unautomated verification steps, and changes outside the repo (such as an issue that needs to be filed).
-   Assemble the *before merging* list, labeling each item as either an unautomated verification step or a change outside the repo, and noting for each item the URL of the specific PR comment that called for it.
+   Assemble two lists:
+   - the *before merging* list, labeling each item as either an unautomated verification step or a change outside the repo, and noting for each item the URL of the specific PR comment that called for it; and
+   - the *follow-on* list, noting for each item the URL of the source comment or description and a brief reason it is not a merge blocker (e.g., "explicitly deferred in PR comment," "out of scope for this PR").
 2. If the *before merging* list is empty, apply the `verified` label to both the PR and the issue it resolves, report this success to the Orchestrator (no unautomated steps or outside-the-repo requirements were identified; the PR may be merged) and exit.
+   A non-empty follow-on list does not prevent you from applying the `verified` label; apply it anyway and also report the follow-on list.
 3. Otherwise, open one GitHub issue for each item on the *before merging* list.
    Do NOT communicate with the user, and do NOT ask whether to test manually or to automate.
    For each item:
@@ -47,8 +55,17 @@ You file one tracking issue per item (see **What to do**) so that each requireme
       Send the id as a bare JSON integer in the request body (not a quoted string), as the API requires.
       If the PR number is not accepted for a sub-issue relationship, fall back to making the new issue a sub-issue of the **parent issue** the PR resolves instead, by sending the same request to `.../issues/{parent issue number}/sub_issues`.
       If neither call succeeds, skip this link without failing.
-4. Report the *before merging* list to the Orchestrator and exit.
-   The PR may be merged once every one of these filed issues is resolved.
+4. Open one GitHub issue for each item on the *follow-on* list.
+   For each item:
+   a. Title the issue `(re PR #{number}) {task title}`, using the same convention as blocking issues.
+   b. In the issue description, include a URL to the source comment or description, and state clearly that this issue does **not** block PR #{number} (for example, "This is a follow-on item and does not block merging PR #{number}.").
+   c. Do **not** call the `blocked_by` dependency endpoint for follow-on issues.
+      Do **not** add any "Blocks PR #..." line to the issue body.
+   d. Optionally make the new issue a sub-issue of the PR (same API call as step 3d), as a convenient organizational link.
+      Skip this without failing if it does not succeed.
+5. Report both lists to the Orchestrator and exit.
+   The PR may be merged once every item on the *before merging* list is resolved.
+   Follow-on issues do not gate the merge.
 
 ## Boundaries
 
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -58,14 +58,10 @@ Only the before-merging list controls the merge gate.
       If neither call succeeds, skip this link without failing.
 4. Open one GitHub issue for each item on the *follow-on* list.
    For each item:
-   a. Title the issue `(re PR #{number}) {task title}`, using the same convention as blocking issues.
+   a. Title the issue simply `{task title}`, without referencing the current PR.
    b. In the issue description, include a URL to the source comment or description, and state clearly that this issue does **not** block PR #{number} (for example, "This is a follow-on item and does not block merging PR #{number}.").
    c. Do **not** call the `blocked_by` dependency endpoint for follow-on issues.
       Do **not** add any "Blocks PR #..." line to the issue body.
-   d. Optionally make the new issue a sub-issue of the PR (same API call as step 3d), as a convenient organizational link.
-      Skip this without failing if it does not succeed.
-      (Unlike blocking issues, follow-on issues do not fall back to the parent issue if the PR sub-issue call fails.
-      The organizational link is a convenience only; losing it for a non-blocking item carries no risk to the merge gate.)
 5. Report both lists to the Orchestrator and exit.
    The PR may be merged once every item on the *before merging* list is resolved.
    Follow-on issues do not gate the merge.
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -14,7 +14,7 @@ Two kinds of outstanding requirement are your responsibility:
 2. **Changes outside the repo**: requirements that are not satisfied by any change to a file in the repo, such as an issue that needs to be filed, a setting that must be changed in an external system, or a manual operational step.
    These are easy to lose because the review process is centered on file changes; surfacing them is explicitly part of your job.
 
-## Two lists, not one
+## Two lists
 
 Classify each finding into exactly one of two tracks:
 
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

The change correctly addresses all four points in issue #407: two lists are introduced, follow-on issues are filed without a `blocked_by` call, the merge gate and `verified` label depend only on the blocking list, and step 5 reports both lists to the Orchestrator separately.
Two defects need to be fixed before merge.

**Defect 1 -- Role section not updated (incorrect documentation)**

The "Role" section (lines 5-7) still reads:

> You scan the linked issue and PR for outstanding requirements that must be handled before merging.

After this PR, the Planner also scans for follow-on items that are explicitly *not* requirements before merging.
The role description is now inaccurate: it describes only half the job.
It should be updated to reflect that the Planner assembles two lists--one blocking, one non-blocking.

**Defect 2 -- Step 4d silently drops the parent-issue sub-issue fallback**

Step 3d (for blocking issues) has a two-level fallback:
1. Try to make the new issue a sub-issue of the PR.
2. If that fails, fall back to the parent issue.
3. If neither succeeds, skip without failing.

Step 4d (for follow-on issues) says only:

> Skip this without failing if it does not succeed.

There is no fallback to the parent issue.
If this asymmetry is deliberate (follow-on items warrant a weaker link), say so.
If it is an oversight, add the same parent-issue fallback as in step 3d.
As written, a Planner following the instructions will silently drop the organizational link for follow-on issues whenever the PR sub-issue call fails, with no explanation of why the behavior differs from the blocking case.

aunger:
🤖 Reviewer

The PR correctly implements the core of issue #407: the "Two lists, not one" section, steps 1/2/3/4/5, and the merge-gate logic are all sound.
However, neither defect called out in the previous review has been addressed.
Both must be fixed before merge.

**Defect 1 -- Role section still inaccurate (unresolved from prior review)**

Lines 5-7 still read:

> You scan the linked issue and PR for outstanding requirements that must be handled before merging, and you file a tracking GitHub issue for each one so none is lost.

After this PR the Planner does two distinct jobs: (1) find blocking requirements and file them as blocking issues, and (2) find follow-on items and file them as non-blocking issues.
The role description still says nothing about the second job, so a Planner reading only the Role section will think it is looking only for merge blockers--exactly the bug this PR fixes.
The role description must be updated.

**Defect 2 -- Step 4d sub-issue fallback still missing and unexplained (unresolved from prior review)**

Step 3d (blocking issues) has a two-level fallback for the sub-issue call: try the PR, then fall back to the parent issue, then skip without failing.
Step 4d (follow-on issues) still says only:

> Skip this without failing if it does not succeed.

The prior review asked whether the asymmetry is deliberate or an oversight.
No answer was given and the text was not changed.
As written, a Planner following the instructions will silently drop the organizational link for follow-on issues whenever the PR sub-issue call fails, with no explanation of why the behavior differs from the blocking case.
Either add the same parent-issue fallback as step 3d, or add a sentence explaining why follow-on issues intentionally do not get that fallback.

aunger:
🤖 Reviewer

The second commit resolves both defects from the prior review.
The Role section now correctly describes both jobs, and step 4d explains why follow-on issues intentionally omit the parent-issue fallback.
All four requirements from issue #407 are addressed.

One new defect must be fixed before merge.

**Defect -- em-dash characters in the Role section (prose-style violation)**

Line 6 of the updated file contains two U+2014 em-dash characters:

> You are the final check that no requirement--blocking or follow-on--is lost.

(The dashes shown above are how that sentence should read; the actual file contains `—`, not `--`.)

The project prose-style rule (`.claude/rules/prose-style.md`) says:

> Replace dashes with commas, semicolons, or parentheses when grammatically possible, and double-hyphens otherwise.

Parentheses are grammatically natural here.
Replace the line with one of:

> You are the final check that no requirement (blocking or follow-on) is lost.

or, keeping the aside structure with double-hyphens:

> You are the final check that no requirement--blocking or follow-on--is lost.

No other issues found.

aunger:
Fixed the em-dash prose-style violation from the last review.

Line 6 of `agents/verification_planning.md` previously contained two U+2014 em-dash characters:

> You are the final check that no requirement—blocking or follow-on—is lost.

Replaced with parentheses per the prose-style rule:

> You are the final check that no requirement (blocking or follow-on) is lost.

Commit: b4b0e31

aunger:
🤖 Reviewer

All four requirements from issue #407 are implemented correctly, and every defect raised in prior review cycles has been resolved.

**Issue #407 checklist**

1. Two lists assembled and classified -- the "Two lists, not one" section and step 1 both cover this.
2. Follow-on issues filed without `blocked_by`, with explicit non-blocking statement -- step 4 handles this correctly.
3. Merge gate and `verified` label depend only on the blocking list -- step 2 explicitly states a non-empty follow-on list does not prevent applying `verified`.
4. Both lists reported to the Orchestrator -- step 5 reports both and makes clear only the blocking list gates the merge.

**Prior defects, all resolved**

- Role section (lines 6-8): now correctly describes both jobs (assemble two lists, file blocking and non-blocking issues).
- Step 4d asymmetry: the parenthetical "(Unlike blocking issues, follow-on issues do not fall back to the parent issue if the PR sub-issue call fails. The organizational link is a convenience only; losing it for a non-blocking item carries no risk to the merge gate.)" explains the deliberate difference clearly.
- Em-dash prose-style violation: confirmed absent. Line 6 reads "You are the final check that no requirement (blocking or follow-on) is lost." -- parentheses, no em-dashes.

**CI**: all check runs completed successfully (build-and-test, CodeQL, file-issues all green).

No issues found.

agents/verification_planning.md
```diff
@@ -47,8 +56,19 @@ You file one tracking issue per item (see **What to do**) so that each requireme
       Send the id as a bare JSON integer in the request body (not a quoted string), as the API requires.
       If the PR number is not accepted for a sub-issue relationship, fall back to making the new issue a sub-issue of the **parent issue** the PR resolves instead, by sending the same request to `.../issues/{parent issue number}/sub_issues`.
       If neither call succeeds, skip this link without failing.
-4. Report the *before merging* list to the Orchestrator and exit.
-   The PR may be merged once every one of these filed issues is resolved.
+4. Open one GitHub issue for each item on the *follow-on* list.
+   For each item:
+   a. Title the issue `(re PR #{number}) {task title}`, using the same convention as blocking issues.
```
aunger:
```suggestion
   a. Title the issue simply `{task title}`, without referencing the current PR.
```

agents/verification_planning.md
```diff
@@ -47,8 +56,19 @@ You file one tracking issue per item (see **What to do**) so that each requireme
...
+   b. In the issue description, include a URL to the source comment or description, and state clearly that this issue does **not** block PR #{number} (for example, "This is a follow-on item and does not block merging PR #{number}.").
+   c. Do **not** call the `blocked_by` dependency endpoint for follow-on issues.
+      Do **not** add any "Blocks PR #..." line to the issue body.
+   d. Optionally make the new issue a sub-issue of the PR (same API call as step 3d), as a convenient organizational link.
```
aunger: Remove this (d) four-line suggestion to make it a sub issue

agents/verification_planning.md
```diff
@@ -13,18 +14,26 @@ Two kinds of outstanding requirement are your responsibility:
 2. **Changes outside the repo**: requirements that are not satisfied by any change to a file in the repo, such as an issue that needs to be filed, a setting that must be changed in an external system, or a manual operational step.
    These are easy to lose because the review process is centered on file changes; surfacing them is explicitly part of your job.
 
-## The before-merging list
+## Two lists
```
aunger, line 17: I edited this, so remember to pull. (Nobody needs convincing this is two lists instead of one. Once you change the instructions, the old ones are forgotten.)

aunger:
🤖 Author

Addressed the two unresolved review comments (commit 49cb2c1):

- **Step 4a**: follow-on issue titles are now plain `{task title}` with no PR reference. The PR context is already conveyed in the issue body.
- **Step 4d removed**: the optional sub-issue suggestion for follow-on issues is dropped entirely.

Also pulled the owner's "Simplify" commit (739b57d) that renamed the section heading from "Two lists, not one" to "Two lists" before making these edits.

aunger (commented):
🤖 Reviewer

All four requirements from issue #407 are implemented correctly in the current state of the file (commit 49cb2c1), and every defect raised across all prior review cycles has been resolved.

**Issue #407 checklist**

1. Two lists assembled and classified -- the "Two lists" section and step 1 both cover this correctly.
2. Follow-on issues filed without `blocked_by`, with explicit non-blocking statement in the body -- step 4 handles this; step 4c is unambiguous.
3. Merge gate and `verified` label depend only on the blocking list -- step 2 explicitly says a non-empty follow-on list does not prevent applying `verified`, and step 5 makes clear only the blocking list gates the merge.
4. Both lists reported to the Orchestrator, clearly separated -- step 5 covers this.

**Changes since the last clean-pass review (commit b4b0e31)**

Two further changes were made after the 05:38 approval: the owner renamed the section heading ("Two lists, not one" -> "Two lists"), and the Author addressed inline review comments by (a) dropping the `(re PR #{number})` prefix from follow-on issue titles and (b) removing step 4d (the optional sub-issue link for follow-on issues) entirely. Both changes are correct and improve the document.

**Open review thread on line 17**

The owner's comment ("I edited this, so remember to pull") on line 17 is purely informational--the Author pulled as directed (commit 49cb2c1 explicitly says so). No code change is required; I will resolve the thread.

**Prose style**: confirmed clean. No em-dashes, no curly quotes, no ellipsis characters.

**CI**: all check runs completed successfully per the CI summary comment.

No issues found.

PR #425, opened 2026-06-17
Forward DebugLog to logcat under the GB4PC tag

### LINKED ISSUE #399: App logger (DebugLog) should reach the OS logging facility (logcat)
## Problem

The reasonable expectation is for application logging to go to the OS logging facility (logcat).
The app's primary logger, `DebugLog` (`app/src/main/java/com/gb4pc/util/DebugLog.kt`), does not meet that expectation: `DebugLog.log(...)` writes only to an in-memory circular buffer and never emits to logcat.

Today `DebugLog.log(...)` is effectively the production logger, called from roughly 86 sites across production code (`OverlayManager`, `ForegroundDetector`, `OverlayService`, `OverlayServiceLogic`, `ServiceChecker`, `BootReceiver`, `PickerActivity`, `SecureViewerActivity`, `MediaDeletionManager`).
None of those messages appear in logcat.
Direct `android.util.Log` calls exist only in E2E/test scaffolding (`E2EFixture`, `ScreenshotTestRule`, `FailureScreenshotRule`, `MockCameraActivity`) and, recently, in three spots in `OverlayManager`.
So the same class now logs a few lines to logcat and most lines only to memory, which is an inconsistent split.

## Why this matters (recent evidence)

PR #398 (issues #397 / #230) burned multiple debugging rounds because overlay diagnostics added through `DebugLog` were invisible in CI logcat.
The investigation only moved forward once the `dispatchTouchEvent`/`handleTap` diagnostics were rerouted to `android.util.Log`.
Any logcat-based debugging (CI log capture, `adb logcat`, device bug reports) is currently blind to everything logged through `DebugLog`.

## History (for context)

- `DebugLog` was introduced in the original data layer (commit `e8fecb4`, "Add data layer with TDD: models, prefs, debug log, permissions"; spec tags DA-03, UI-10) as an in-memory buffer feeding an in-app viewer.
- The in-app viewer side: #14 / PR #17 (make the debug log text selectable and copyable) and #83 / PR #85 (move the log viewer to its own screen).
- A thread-safety fix for the buffer: #9 (`DebugLog.listener` invoked outside the synchronized block).
- Logging-content changes layered on top of the in-memory model: #65 / PR #71 (add more logging), #90 / PR #95 (remove trace-level logging), #324 / PR #325 (log foreground apps).
- Separately, E2E diagnostics were sent straight to logcat under the `GB4PC_E2E` tag (for example #364, #365, around PR #362), precisely because logcat is what CI can read.
  That is exactly the capability the in-app logger lacks.

## Suggested options (pick one)

1. Delegate, always: `DebugLog.log(...)` also forwards each entry to `android.util.Log` (a single tag such as `GB4PC`, with a sensible level). Lowest risk, keeps the in-app viewer, and makes every existing call site visible in logcat with no call-site changes.
2. Delegate, when configured: same as option 1 but gated behind a setting or build flag (for example debug builds only, or a "mirror logs to logcat" toggle next to the in-app viewer). Useful if mirroring everything raises privacy concerns.
3. Replace: drop the bespoke logger and standardize on `android.util.Log` (optionally Timber over it) as the single logging path. The in-app viewer would then need a logcat-backed source or be retired.

Recommendation: option 1, or option 2 if privacy gating is wanted.
Either is the smallest change that closes the expectation gap while preserving the on-device viewer.

## Brief argument against using logcat

The in-memory buffer exists so users (and the maintainer) can read recent logs on-device through the in-app viewer without a computer or adb, and logcat is a shared, fixed-size system ring buffer the app does not control, so its messages can be throttled or dropped and are wiped on reboot.
Mirroring (options 1 or 2) keeps the in-app viewer's guarantees while still satisfying the logcat expectation.

aunger:
Fixes #399

## What changed

`DebugLog.log()` now calls `android.util.Log.d("GB4PC", message)` before writing to the in-memory buffer.
Every message logged through `DebugLog` -- across all 86+ call sites in production code -- is now visible in logcat, `adb logcat`, CI log captures, and device bug reports.

The in-memory circular buffer, the on-device viewer, and the `listener` callback are all unchanged.
This is Option 1 from the issue: always delegate, lowest risk, no call-site changes required.

## Why this approach

The issue noted that PR #398's debugging round was blocked because `DebugLog` messages were invisible in CI logcat.
Mirroring to logcat (rather than replacing the in-memory logger) keeps the on-device viewer's guarantees while closing the visibility gap.

## Tests

Added `DebugLogLogcatTest` (Robolectric) covering:
- Each `log()` call produces a `DEBUG` entry under the `GB4PC` tag in logcat.
- Multiple entries arrive in order.
- Forwarding still occurs when the in-memory buffer is at capacity (eviction does not suppress the logcat write).
- The `LOGCAT_TAG` constant equals `"GB4PC"`.

All existing `DebugLogTest` tests continue to pass without modification.

### DIFF
--- a/app/src/main/java/com/gb4pc/util/DebugLog.kt
+++ b/app/src/main/java/com/gb4pc/util/DebugLog.kt
@@ -1,13 +1,20 @@
 package com.gb4pc.util
 
+import android.util.Log
 import com.gb4pc.Constants
 import java.util.LinkedList
 
 /**
  * In-memory circular buffer debug log (DA-03, UI-10).
  * Thread-safe. Entries are lost when the process is killed.
+ *
+ * Each call to [log] also forwards the message to [android.util.Log] under the tag [LOGCAT_TAG]
+ * so that all application log messages are visible in logcat and CI log captures (Issue #399).
  */
 object DebugLog {
+
+    /** logcat tag used for all forwarded messages. */
+    const val LOGCAT_TAG = "GB4PC"
 
     data class Entry(val timestamp: Long, val message: String)
 
@@ -18,6 +25,7 @@ object DebugLog {
     var listener: (() -> Unit)? = null
 
     fun log(message: String) {
+        Log.d(LOGCAT_TAG, message)
         val snapshot: (() -> Unit)?
         synchronized(lock) {
             buffer.addLast(Entry(System.currentTimeMillis(), message))
--- a/app/src/main/java/com/gb4pc/util/DebugLog.kt
+++ b/app/src/main/java/com/gb4pc/util/DebugLog.kt
@@ -25,9 +25,9 @@ object DebugLog {
     var listener: (() -> Unit)? = null
 
     fun log(message: String) {
-        Log.d(LOGCAT_TAG, message)
         val snapshot: (() -> Unit)?
         synchronized(lock) {
+            Log.d(LOGCAT_TAG, message)
             buffer.addLast(Entry(System.currentTimeMillis(), message))
             while (buffer.size > Constants.DEBUG_LOG_BUFFER_SIZE) buffer.removeFirst()
             snapshot = listener
--- a/app/src/test/java/com/gb4pc/util/DebugLogLogcatTest.kt
+++ b/app/src/test/java/com/gb4pc/util/DebugLogLogcatTest.kt
@@ -1,0 +1,70 @@
+package com.gb4pc.util
+
+import android.util.Log
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.shadows.ShadowLog
+
+/**
+ * Verifies that DebugLog.log() forwards each message to android.util.Log (Issue #399).
+ *
+ * Robolectric is required so ShadowLog captures Log.d() calls made by DebugLog.
+ */
+@RunWith(RobolectricTestRunner::class)
+class DebugLogLogcatTest {
+
+    @Before
+    fun setUp() {
+        ShadowLog.reset()
+        DebugLog.clear()
+    }
+
+    @Test
+    fun `log forwards message to logcat under the GB4PC tag`() {
+        DebugLog.log("hello logcat")
+
+        val logged = ShadowLog.getLogsForTag(DebugLog.LOGCAT_TAG)
+        assertEquals("Expected exactly one logcat entry for tag GB4PC", 1, logged.size)
+        assertEquals("hello logcat", logged[0].msg)
+        assertEquals(Log.DEBUG, logged[0].type)
+    }
+
+    @Test
+    fun `log forwards every message to logcat in order`() {
+        DebugLog.log("first")
+        DebugLog.log("second")
+        DebugLog.log("third")
+
+        val logged = ShadowLog.getLogsForTag(DebugLog.LOGCAT_TAG)
+        assertEquals(3, logged.size)
+        assertEquals("first",  logged[0].msg)
+        assertEquals("second", logged[1].msg)
+        assertEquals("third",  logged[2].msg)
+    }
+
+    @Test
+    fun `logcat tag constant equals GB4PC`() {
+        assertEquals("GB4PC", DebugLog.LOGCAT_TAG)
+    }
+
+    @Test
+    fun `log forwards message to logcat even when buffer is at capacity`() {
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

The approach is correct and the tests are solid. Two issues require changes before merge.

---

## Issue 1 (substantive): `Log.d` is placed outside the synchronized block

In the new `log()` implementation the logcat write precedes the `synchronized(lock)` block:

```kotlin
fun log(message: String) {
    Log.d(LOGCAT_TAG, message)       // outside lock
    val snapshot: (() -> Unit)?
    synchronized(lock) {
        buffer.addLast(...)
        ...
    }
    snapshot?.invoke()
}
```

This introduces two problems:

**a) Behavioral regression under concurrent access.**
If two threads call `log()` concurrently, their `Log.d` calls can be interleaved in a different order than their `buffer.addLast()` calls. The in-memory buffer and logcat can then disagree on message order. Before this PR, the buffer was the single authoritative record; after this PR, logcat is supposed to be a faithful mirror, but the ordering guarantee is broken under concurrent writers.

**b) Silent message loss if `Log.d` throws.**
`android.util.Log` does not document throwing, but it is a native call that can be interrupted or suppressed by the system. If it throws, execution never reaches `synchronized(lock)`, so the message is lost from the in-memory buffer. Pre-PR, `log()` always wrote to the buffer (barring JVM-level errors). This PR introduces a regression path that did not exist before.

Both problems are fixed by moving `Log.d` inside the synchronized block, immediately before `buffer.addLast()`:

```kotlin
fun log(message: String) {
    val snapshot: (() -> Unit)?
    synchronized(lock) {
        Log.d(LOGCAT_TAG, message)
        buffer.addLast(Entry(System.currentTimeMillis(), message))
        while (buffer.size > Constants.DEBUG_LOG_BUFFER_SIZE) buffer.removeFirst()
        snapshot = listener
    }
    snapshot?.invoke()
}
```

Holding the lock during `Log.d` is fine: `android.util.Log` does not call back into `DebugLog`, so there is no deadlock risk.

---

## Issue 2 (nit): unrelated file in the diff

`.claude/rules/github-mcp-authenticate.md` appears in this PR's diff. That file is already present on `main` and has nothing to do with issue #399. It should not be part of this PR. Please rebase or otherwise remove it from the diff before merge.

---

## What is good

- Option 1 (always delegate) is the right call per the issue.
- `LOGCAT_TAG = "GB4PC"` as a named constant is clean.
- The four Robolectric tests cover the cases that matter: single-message forwarding, ordering, the constant value, and at-capacity eviction. Existing `DebugLogTest` is untouched.
- No call-site changes required--all 86+ production sites benefit automatically.

aunger:
🤖 Author

Both issues addressed.

**Issue 1 (Log.d outside synchronized block):** Agreed -- the reviewer's analysis is correct on both counts (ordering under concurrent writers and the silent-loss regression path). `Log.d` is now inside the `synchronized(lock)` block, immediately before `buffer.addLast()`. All four `DebugLogLogcatTest` tests and all eight existing `DebugLogTest` tests still pass.

**Issue 2 (unrelated file in diff):** The `.claude/rules/github-mcp-authenticate.md` commits were on this branch because the branch was created before those commits landed on main via a separate PR. Fixed by rebasing onto the current `origin/main` and cherry-picking only the DebugLog commit. The diff now contains exactly two files: `DebugLog.kt` and `DebugLogLogcatTest.kt`.

aunger:
🤖 Reviewer

Both issues from the previous round are resolved. This is good to merge.

**Issue 1 (Log.d placement):** `Log.d(LOGCAT_TAG, message)` is now inside the `synchronized(lock)` block, immediately before `buffer.addLast()`. The concurrent ordering problem and the silent-loss regression path are both eliminated. Holding the lock during `Log.d` is safe; `android.util.Log` does not call back into `DebugLog`.

**Issue 2 (unrelated files):** Confirmed via `git diff origin/main..feature/issue-399-debuglog-logcat --name-only`--the branch diverges from `origin/main` by exactly two files: `DebugLog.kt` and `DebugLogLogcatTest.kt`. The unrelated file is gone.

**Implementation:** Clean. `LOGCAT_TAG = "GB4PC"` as a named constant, KDoc updated, no call-site changes required, in-memory buffer and on-device viewer fully preserved. Satisfies Option 1 exactly as the owner elected.

**Tests:** All four `DebugLogLogcatTest` cases are correct: single-message forwarding with tag and level, ordering across multiple messages, constant value assertion, and logcat forwarding under buffer-full eviction. Existing `DebugLogTest` is untouched.

CI checks (`build-and-test`, `analyze-kotlin`) were still in progress at review time.

PR #427, opened 2026-06-17
Add silentVanish recovery path when Monitor task disappears without notification

### LINKED ISSUE #411: Orchestrator session stalled for ~2 hours after a Monitor task vanished without a terminal notification
## Summary

While orchestrating issue #402 / PR #408, the CI Monitor tool (`Monitor` launching `python3 scripts/ci_monitor/ci_monitor.py --pr 408`) started normally but never delivered a terminal result, and the session stalled for roughly two hours afterward with no waking event.

## Timeline

- 11:17:38 UTC: Monitor launched with `timeout_ms: 1800000` (30 minutes), task id `birx5v7wr`. Startup notification received: "monitor PID 28464 ... if TaskStop is unavailable, send SIGTERM to this PID to stop me".
- No further notification (in_progress, Clear, Blocked, Infra, or timeout) was ever delivered.
- 13:26:13 UTC: the user's next message woke the session, roughly 2 hours and 9 minutes later.
- On investigation: `ps -p 28464` showed the process was no longer running, and `TaskOutput` for task id `birx5v7wr` returned "No task found with ID: birx5v7wr".

## Expected behavior

Per `agents/dev_orchestration.md`, Monitor events (in_progress heartbeats, and the terminal Clear/Blocked/Infra/timeout lines) are supposed to keep the orchestration session alive, with a hard 30-minute timeout enforced by `timeout_ms`. Here, neither the 30-minute timeout notification nor any other terminal signal arrived, and the task disappeared from tracking entirely.

## Note

This gap was observed independently of the `ci_monitor.py` reliability issues tracked in #402; it is in the session/task-notification infrastructure around the Monitor tool, not in `ci_monitor.py` itself. (The PR #408 fix for #402 was separately verified directly via the GitHub API, since the user authorized direct log access for this session.)

Per `agents/dev_orchestration.md`'s guidance ("If a session ever stalls with no such event for an extended period, file a bug describing the gap rather than adding a sleep loop to work around it"), filing this as a tracking issue rather than adding a workaround.

aunger:
Fixes #411

## What happened

When the CI Monitor task's process exited without the task-notification infrastructure delivering any terminal line (including the 30-minute timeout notification), the Orchestrator session stalled indefinitely -- no waking event ever arrived. The Monitor process died (PID gone), and `TaskOutput` returned "No task found with ID: <id>", meaning the task record was dropped entirely. The session waited ~2 hours until the user sent a message.

## What this PR does

Adds a `silentVanish` recovery branch to the Monitor loop in `agents/dev_orchestration.md`.

When a user message wakes the session while the Monitor is nominally still pending:

1. The Orchestrator checks the Monitor task's status via `TaskOutput` (using the task ID from the Monitor invocation).
2. If `TaskOutput` returns "No task found with ID: <id>", the task record was dropped -- a silent vanish. The Orchestrator informs the user and re-launches the Monitor immediately, then resumes normal routing (including the `undiagnosedTerminal` check if warranted).
3. If the Monitor task is still registered, the user message is not proof of a vanish -- the Orchestrator continues waiting.
4. If the fresh re-launch also vanishes silently (a second user message arrives before any terminal line), the Orchestrator escalates to the user.

Also updates the "No sleep loops needed" section: instead of advising to file a bug (which was the right advice when no protocol existed), it now points to the `silentVanish` recovery path for Monitor-task stalls, and reserves the file-a-bug advice for stalls with no known recoverable cause.

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -194,6 +194,10 @@ silentVanish:
   // undiagnosedTerminal if warranted).
   // If the fresh invocation also vanishes silently (a second user message
   // arrives before any terminal line), escalate to the user; stop.
+  // Note: the user message that triggered this branch is treated as a wake-up event only.
+  // Its content (if any) is set aside; the Orchestrator's narrow scope during CI monitoring
+  // means user questions or instructions cannot be addressed mid-monitor. The user is informed
+  // of CI status (see branches below), which is the appropriate response in this context.
   if TaskOutput for the Monitor's task ID returns "No task found with ID: <id>":
     Inform the user that the Monitor task vanished without a terminal notification and is being re-launched.
     Re-launch the Monitor tool call (same command as the original, fresh invocation).
@@ -235,7 +239,7 @@ Orchestrators do not need `sleep`-based keep-alive loops while waiting on sub-ag
 Task-notification events (sub-agent completion) and Monitor events (CI status) keep the session alive on their own.
 Dispatch and wait; do not add artificial delays.
 
-If a session ever stalls with no such event for an extended period, use the `silentVanish` recovery path (see the Monitor loop above) rather than adding a sleep loop.
+If the session stalls (no Monitor or sub-agent event arrives) and a user message later wakes it, apply the `silentVanish` recovery path (see the Monitor loop above) if the Monitor was pending.
 If the stall cannot be explained by a known recoverable cause (e.g., a Monitor task vanish), file a bug describing the gap.
 
 ## When to abort
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -152,6 +152,7 @@ Monitor output lines are relayed to the user verbatim; this is user-facing statu
     if Monitor emits a Blocked line  → goto newAuthor
     if Monitor emits an Infra line   → escalate to user; stop
     if Monitor times out (30 min)    → escalate to user; stop
+    if a user message wakes the session before Monitor delivers any terminal line → goto silentVanish
     if Monitor emits a Clear line:
       // Step: Surface outstanding before-merging requirements
       //   (unautomated verification tests and changes outside the repo, such as an issue that needs to be filed)
@@ -178,6 +179,29 @@ undiagnosedTerminal:
     → treat the re-run's outcome as authoritative; resume the routing above from "Act only on the terminal lines..." using the re-run's lines (still without re-applying the undiagnosedTerminal check)
   else (the re-run repeats `drain poll found no new diagnostic signals` followed by the same Blocked/Infra terminal):
     → proceed with the original terminal's routing (Blocked → newAuthor; Infra → escalate to user, stop) without a further re-run
+
+silentVanish:
+  // Issue #411: the Monitor task can silently vanish -- the process exits
+  // without the task-notification infrastructure delivering any terminal line
+  // (not even a timeout notification). This leaves the session stuck until
+  // the user sends a message. When a user message wakes the session while
+  // a Monitor invocation is still nominally pending, check whether the task
+  // is still alive using TaskOutput (passing the Monitor's task ID).
+  // If TaskOutput returns "No task found with ID: <id>", the task record has
+  // been dropped -- a silent vanish. Re-launch the Monitor once immediately
+  // and resume the routing above from "Act only on the terminal lines..."
+  // with this fresh invocation (applying all normal checks, including
+  // undiagnosedTerminal if warranted).
+  // If the fresh invocation also vanishes silently (a second user message
+  // arrives before any terminal line), escalate to the user; stop.
+  if TaskOutput for the Monitor's task ID returns "No task found with ID: <id>":
+    Inform the user that the Monitor task vanished without a terminal notification and is being re-launched.
+    Re-launch the Monitor tool call (same command as the original, fresh invocation).
+    Resume the routing above from "Act only on the terminal lines..." with the fresh invocation.
+    if this fresh invocation also produces no terminal line before the next user message:
+      → escalate to user; stop
+  else (the Monitor task is still registered -- a user message is not proof of a vanish):
+    Continue waiting for the Monitor's terminal line; do not re-launch.
 ```
 
 ### Monitor script
@@ -211,7 +235,8 @@ Orchestrators do not need `sleep`-based keep-alive loops while waiting on sub-ag
 Task-notification events (sub-agent completion) and Monitor events (CI status) keep the session alive on their own.
 Dispatch and wait; do not add artificial delays.
 
-If a session ever stalls with no such event for an extended period, file a bug describing the gap rather than adding a sleep loop to work around it.
+If a session ever stalls with no such event for an extended period, use the `silentVanish` recovery path (see the Monitor loop above) rather than adding a sleep loop.
+If the stall cannot be explained by a known recoverable cause (e.g., a Monitor task vanish), file a bug describing the gap.
 
 ## When to abort
 
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -144,7 +144,7 @@ Monitor output lines are relayed to the user verbatim; this is user-facing statu
 ```
   if Reviewer requested changes → goto newAuthor
   if Reviewer gave LGTM:
-    Orchestrator launches a Monitor tool call running `python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER>` from the repo root (run_in_background: true, timeout_ms: 1800000). Record the task ID returned by the Monitor tool call for use in silentVanish recovery.
+    Orchestrator launches a Monitor tool call running `python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER>` from the repo root (run_in_background: true, timeout_ms: 1800000). Record the task ID returned by the Monitor tool call for use in silentVanish recovery, and clear the silentVanish re-launch flag (this original launch is not a re-launch).
     Each stdout line arrives as a task-notification event; relay each line to the user verbatim.
     Act only on the terminal lines Clear, Blocked, or Infra. Relay in_progress lines to the user as brief status updates (the script suppresses these unless no other output has been emitted for over 120 seconds).
     Relay `step "..." -> ...` and `FAIL [...] ...` lines to the user as informational test-result deltas; they do NOT end the loop or start a new Author round.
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Reviewer

The `silentVanish` recovery path is well-conceived and the structure is sound. Two issues need to be addressed before merge.

---

**Issue 1 (blocker): The Orchestrator is never told to record the Monitor task ID at launch time**

The `silentVanish` block checks `TaskOutput` "for the Monitor's task ID." But the Monitor-launch pseudocode reads:

```
Orchestrator launches a Monitor tool call running `python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER>` from the repo root (run_in_background: true, timeout_ms: 1800000)
Each stdout line arrives as a task-notification event; relay each line to the user verbatim.
```

There is no instruction to capture or record the task ID returned by the Monitor tool call. If the Orchestrator does not record it, the `TaskOutput` check in `silentVanish` cannot be executed -- the Orchestrator has no ID to query. The launch step needs an explicit instruction such as "Record the task ID returned by the Monitor tool call for use in silentVanish recovery."

---

**Issue 2 (blocker): The `silentVanish` else branch leaves the user message unaddressed**

The else branch says:

```
  else (the Monitor task is still registered -- a user message is not proof of a vanish):
    Continue waiting for the Monitor's terminal line; do not re-launch.
```

The user message already woke the session. After determining the task is still alive, the Orchestrator needs to know what to do with the user's message. Should it relay it to a sub-agent? Ignore it? Reply to the user that CI is still running? The say-nothing rule and relay rules govern what can be said to sub-agents, but there is no guidance here for the Orchestrator's response (if any) to the user. The else branch should at minimum say something like "Relay to the user that the Monitor is still running; then continue waiting."

---

No other concerns. The double-vanish escalation, the `TaskOutput` "No task found" string match, the positioning of the `silentVanish` goto relative to the timeout and other terminals, and the "No sleep loops needed" section update are all correct.

aunger:
🤖 Reviewer

Both blockers from the prior review round are resolved: the Monitor launch line now instructs the Orchestrator to record the task ID, and the else branch now tells the Orchestrator to relay status to the user before continuing to wait. Those are clean fixes.

One new blocker, and one nit.

---

**Issue 1 (blocker): "No sleep loops needed" update misstates how silentVanish is triggered**

The updated text reads:

> If a session ever stalls with no such event for an extended period, use the `silentVanish` recovery path (see the Monitor loop above) rather than adding a sleep loop.

`silentVanish` cannot be proactively invoked. It is triggered by a user message waking the stalled session. An Orchestrator that is stalled (no user message has arrived) has no way to reach for `silentVanish` on its own -- the stall is the stall, and the only escape is the user sending a message. The text as written implies the Orchestrator can actively "use" `silentVanish` as a remedy during the stall, which is false. A future Orchestrator reading this could spin looking for a trigger that never comes, or worse, invent one.

The sentence should be rewritten to make clear that `silentVanish` applies when the user's message is itself the wake-up event -- something like:

> If the session stalls (no Monitor or sub-agent event arrives) and a user message later wakes it, apply the `silentVanish` recovery path if the Monitor was pending.

The immediately following sentence ("If the stall cannot be explained by a known recoverable cause...") is correct and should stay.

---

**Issue 2 (nit): silentVanish discards the user message's content without saying so**

The `silentVanish` branch treats the incoming user message purely as a trigger event and does not address what to do with its content. Both the vanish branch and the else branch tell the Orchestrator how to respond about CI status, but neither branch says what to do if the user's message contained an instruction or question. This may be intentional -- the Orchestrator's narrow scope and say-nothing rule may be intended to swallow the message -- but if so, a brief comment in the pseudocode stating that would prevent future confusion. If it is not intentional, an explicit routing step is needed.

Not requiring resolution of the nit before merge. The blocker must be fixed.

aunger:
🤖 Programmer

Both issues from the latest review round are addressed in commit 16878d8.

**Blocker fixed**: The "No sleep loops needed" sentence no longer implies the Orchestrator can proactively invoke `silentVanish` during a stall. It now reads:

> If the session stalls (no Monitor or sub-agent event arrives) and a user message later wakes it, apply the `silentVanish` recovery path (see the Monitor loop above) if the Monitor was pending.

This makes clear that `silentVanish` is triggered by the user message waking the session, not something reachable while the stall is ongoing.

**Nit addressed**: Added a comment at the top of the `silentVanish` decision block stating explicitly that the triggering user message's content is set aside — the Orchestrator's narrow CI-monitoring scope means mid-monitor instructions cannot be addressed, and CI status (relayed in the branches below) is the appropriate response.

aunger (commented):
🤖 Reviewer

Both blockers from the prior round are correctly resolved. The "No sleep loops" sentence now accurately describes `silentVanish` as triggered by a user message waking the session, not something reachable proactively during the stall. The `silentVanish` block now contains an explicit note explaining that the triggering user message's content is set aside, with the reasoning stated. The task-ID recording instruction at launch is in place. The else branch tells the Orchestrator to relay CI status and continue waiting.

One blocker remains before merge.

---

**Issue 1 (blocker): Three prose `--` dashes have surrounding spaces, violating the prose style rule**

`.claude/rules/prose-style.md` states:

> Do not surround dashes or double-hyphens with spaces
> Omit the spaces--like this.

The `silentVanish` block introduces three instances of ` -- ` (space, double-hyphen, space) in prose comment lines:

1. `// Issue #411: the Monitor task can silently vanish -- the process exits`
2. `// been dropped -- a silent vanish.`
3. `else (the Monitor task is still registered -- a user message is not proof of a vanish):`

Each should have the surrounding spaces removed so the double-hyphen runs directly into the adjacent words, e.g.:

- `can silently vanish--the process exits`
- `been dropped--a silent vanish.`
- `still registered--a user message is not proof of a vanish`

No other concerns. The logic, the TaskOutput string match, the double-vanish escalation path, and the "No sleep loops needed" update are all correct.

aunger:
🤖 Programmer

Blocker from the latest review round is resolved in commit 2dee14a.

**Blocker fixed**: The three ` -- ` (space-dash-dash-space) instances in the `silentVanish` block now run without surrounding spaces, per `.claude/rules/prose-style.md`:

- `can silently vanish--the process exits`
- `been dropped--a silent vanish.`
- `still registered--a user message is not proof of a vanish`

No other changes were made.

aunger:
🤖 Reviewer

All blockers from prior rounds are resolved. The task-ID recording instruction is in place, the else branch tells the Orchestrator what to say to the user, the "No sleep loops needed" sentence correctly frames `silentVanish` as triggered by a user-message wake-up (not proactively reachable during a stall), the user-message content note is present, and the three double-hyphen spacing violations are fixed.

One blocker remains.

---

**Issue 1 (blocker): The double-vanish escalation is unreachable as written**

The re-launch step reads:

```
  if TaskOutput for the Monitor's task ID returns "No task found with ID: <id>":
    Inform the user that the Monitor task vanished without a terminal notification and is being re-launched.
    Re-launch the Monitor tool call (same command as the original, fresh invocation).
    Resume the routing above from "Act only on the terminal lines..." with the fresh invocation.
    if this fresh invocation also produces no terminal line before the next user message:
      → escalate to user; stop
```

The instruction `Resume the routing above from "Act only on the terminal lines..."` sends the Orchestrator back into the full routing loop. That loop includes `if a user message wakes the session before Monitor delivers any terminal line → goto silentVanish`. So if the fresh invocation also vanishes silently and a second user message arrives, the routing loop intercepts it and re-enters `silentVanish` before the nested escalation instruction is ever reached. The nested `if this fresh invocation also produces no terminal line before the next user message: → escalate to user; stop` is dead code: control flow never reaches it.

On that second entry to `silentVanish`, if the re-launched task has also vanished, `TaskOutput` returns "No task found" again, and the Orchestrator would inform the user and re-launch a third time--not escalate. The escalation is never triggered.

The fix is to make the double-vanish check part of `silentVanish` itself rather than a nested instruction after `Resume`. One way: add a "is this a re-launch?" flag (e.g., note at the top of `silentVanish` that the Orchestrator tracks whether the current Monitor invocation is a silentVanish re-launch), and add a branch at the top of `silentVanish`:

```
  if this entry is for a silentVanish re-launch (not the original Monitor invocation):
    → escalate to user; stop
```

Another equivalent approach: add a goto label `silentVanishRelaunch` after the re-launch, and use that for the second pass, where the only path is escalation. The specific mechanism is up to the Author; the requirement is that a second user-message-before-terminal event during the re-launched Monitor triggers escalation rather than another re-launch.

---

No other concerns. The logic for the first-time vanish case, the else branch, the prose style, and the "No sleep loops" update are all correct.

aunger:
🤖 Programmer

The double-vanish escalation blocker is resolved in commit 534c2f9.

You were right that the nested `if this fresh invocation also produces no terminal line before the next user message: → escalate` was dead code. The routing loop routes every user-message wake-up through `goto silentVanish`, so a second vanish re-enters `silentVanish` from the top and the nested instruction was never reached.

The fix introduces a silentVanish re-launch flag that survives across `goto silentVanish` re-entries:

- The original Monitor launch clears the flag (this original launch is not a re-launch).
- On a confirmed vanish (`TaskOutput` returns "No task found with ID: <id>"), the flag decides what to do: if it is already set, the re-launched Monitor has vanished too, so escalate and stop; otherwise re-launch and set the flag.
- The dead nested `if` after `Resume` is removed.

One refinement beyond the literal suggestion: I put the re-launch-flag check inside the confirmed-vanish branch rather than at the very top of `silentVanish`. Putting it at the top would escalate on any user message during a healthy re-launched Monitor (the task is still alive, not vanished). Gating escalation on both the flag and a confirmed vanish means a still-alive re-launched task with an incoming user message falls through to the else branch (relay CI-still-running status, keep waiting), which preserves the single-recovery-attempt guarantee you asked for while not escalating on healthy re-launches.

No spaced double-hyphens were introduced (verified). The diff is one file, +17/-6.

aunger (commented):
🤖 Reviewer

LGTM. All blockers from prior rounds are resolved, and the double-vanish escalation fix in 534c2f9 is correct.

I traced the full state machine against the flag lifecycle:

- Original launch records the task ID and clears the re-launch flag. (line 147)
- Routing entry `if a user message wakes the session before Monitor delivers any terminal line → goto silentVanish` is correctly positioned after the timeout/Infra/Blocked terminal checks. (line 155)
- First vanish: flag is clear, so the escalation check is skipped; re-launch, set flag, resume. Second vanish: the re-launch flag is set, so the confirmed-vanish branch escalates and stops. The escalation is now reachable, since `goto silentVanish` re-enters the block from the top rather than relying on the previously-dead nested instruction.
- Gating escalation on both the flag AND a confirmed vanish (rather than at the very top of `silentVanish`) is the right call: a still-alive re-launched task that receives an incidental user message falls through to the else branch and keeps waiting, instead of escalating spuriously.
- The else branch now relays CI-still-running status to the user, addressing the waking message. The user-message-content note and the "No sleep loops needed" reframing (silentVanish as a user-message-triggered recovery, not proactively reachable during a stall) are both correct.
- Prose style: the three spaced ` -- ` instances are gone; the remaining em-dashes and `→` arrows in the file are pre-existing and outside this diff.

One non-blocking nit: the comment block states the re-launch flag should also be cleared "whenever a Monitor invocation delivers any terminal line," but no executable step in the pseudocode does that, only the original-launch line clears it. In practice this is harmless, since a terminal line exits the Monitor loop to its routing (newAuthor / escalate / Clear path) and the stale flag is never re-consulted within the same CI cycle. Worth an imperative clear-on-terminal step for clarity if you touch this again, but it is not a correctness defect and does not block merge.

CI is docs-only (the summary comment notes no full build was needed); not blocking on it.

Issue #410, opened 2026-06-15
(re PR #408) Close Run G multi-process CI Monitor lag (residual limitation of #402 fix)


aunger:
Follow-up from PR #408 (https://github.com/aunger/gallery-button-for-pixel-camera/pull/408), which fixes #402.

## Requirement

PR #408 fixes the single-poll-lag shape of issue #402 (Runs B/C/E/F: a `Blocked` terminal contradicted by `step ... -> success` lines, or with no step/FAIL markers, because `/actions/runs/{id}/jobs` and `/actions/runs/{id}/artifacts` lag one poll behind `/commits/{sha}/check-runs`). It does this via a bounded in-process drain (`drain_then_print`, up to `DRAIN_MAX_ATTEMPTS=3` retries at `DRAIN_DELAY_SECONDS=5` apart, i.e. up to 15s of cumulative lag).

However, issue #402's Run G demonstrated a lag that spanned three separate Monitor process invocations over what was presumably minutes, not one extra poll's worth of time within a single process. PR #408 explicitly does not claim to fix this case: it adds a `drain poll found no new diagnostic signals` line so this shape is flagged rather than looking like a coincidental empty terminal, but the underlying gap (a genuine check failure on a cleanly mergeable, up-to-date branch producing `Blocked` with zero diagnostic informational lines across multiple Monitor invocations) remains open.

Per the PR description: "Fully closing Run G's gap would need a different mechanism (e.g., the Orchestrator re-running the Monitor after a longer delay, or a persistent poll outside the process lifetime); flagged here as a known residual limitation rather than addressed."

## Source

PR #408 description ("Response to review", point 1):
https://github.com/aunger/gallery-button-for-pixel-camera/pull/408

## Suggested follow-up

Design and implement a mechanism that can survive across Monitor process invocations to recover Run G's multi-minute, multi-process lag, such as the Orchestrator re-running the Monitor after a longer delay when it sees the new `drain poll found no new diagnostic signals` line followed by a `Blocked`/`Infra` terminal, or a persistent out-of-process poll.

Blocks PR #408.

aunger:
Owner says make this minimal, or skip it.

PR #428, opened 2026-06-17
Continue draining all attempts so two lagging endpoints both surface

### LINKED ISSUE #419: CI Monitor: handle two endpoints lagging by different amounts?
This ticket is to decide: Should we implement this?

Owner’s note: Would this obviate the out-of-process retry that #408 coded into `dev-orchestration.md`? If yes, then I think this is worth doing. Otherwise, I need to hear a recommendation.

---
_Originally posted by 🤖 Reviewer in https://github.com/aunger/gallery-button-for-pixel-camera/pull/408#pullrequestreview-4507562425_

drain_then_print stops retrying after the first attempt that emits anything. If the gate step becomes visible on drain attempt 1 but the testresults-* artifact only resolves on drain attempt 2 (two endpoints lagging by different amounts), the FAIL marker is silently dropped for this process's lifetime. The seen_arts set would not be populated, so a fresh Orchestrator recheck would re-fetch the artifact and surface the FAIL, but within this process the information is lost. In practice, the observed failure cases (Runs B/C/E/F/G) show steps and artifacts lagging together, and the test group (u) explicitly reproduces a two-attempt lag for both signals simultaneously, so this partial-lag case is not evidenced from the issue. The conservative fix would be to continue retrying after a fruitful attempt (up to DRAIN_MAX_ATTEMPTS) rather than breaking immediately, but that changes the semantics the test groups verify and is not required by anything in the issue. Noting it as a known gap, not requesting a change here.

aunger:
Fixes #419

## Decision

The issue asks whether to implement the partial-lag fix, and whether doing so would obviate the out-of-process retry that #408 coded into `dev_orchestration.md` (the `undiagnosedTerminal` recheck).

**Recommendation: implement the fix. It does NOT obviate the out-of-process retry.** The two mechanisms cover disjoint lag regimes:

- **Partial lag (this issue):** both signals resolve within the bounded in-process drain window (up to `DRAIN_MAX_ATTEMPTS` x `DRAIN_DELAY_SECONDS`, ~15s), but on *different* drain attempts. This is fully fixable in-process and is what this PR does.
- **`undiagnosedTerminal` (#408 / #410, Run G):** a multi-minute lag that outlives the entire Monitor process. The drain finds nothing, prints `drain poll found no new diagnostic signals`, and exits; only a fresh, out-of-process Monitor launch minutes later recovers it. No amount of in-process retry reaches signals that are not available during the process's lifetime.

So the fix closes a real correctness gap on its own merits and the out-of-process recheck remains necessary for the longer lag regime.

## The bug

`drain_then_print` broke out of the bounded drain loop on the first attempt that emitted anything. When the failing gate step (from `/actions/runs/{id}/jobs`) and the `testresults-` artifact (from `/actions/runs/{id}/artifacts`) settle on different drain attempts, e.g. the step on attempt 1 and the artifact on attempt 2, the early break dropped the later-arriving `FAIL` marker for this Monitor process's lifetime. `seen_arts` was never populated for the dropped artifact, so an out-of-process recheck would still re-fetch it, but the information was lost within the process.

## The fix

Run every drain attempt (still bounded by `DRAIN_MAX_ATTEMPTS`) instead of stopping at the first fruitful one, so both endpoints get the full drain window to catch up. The bound and the `drain poll found no new diagnostic signals` flag are unchanged.

## Tests

- New group `(w)` reproduces the partial lag (step surfaces on drain attempt 1, artifact `FAIL` only on attempt 2) and asserts both signals are surfaced before the terminal `Blocked` line. Verified it fails against the old break-early code with `lagging FAIL marker was dropped`.
- Groups `(t)` and `(u)` updated to supply the now-always-run remaining drain attempts; their assertions about which signals surface are unchanged.
- Full suite: 157 passed, 0 failed (`python3 scripts/test_ci_monitor.py`).

## Merge note

Merged latest `main` (which had advanced past this branch's base) to resolve the conflict the owner flagged. Both this branch and `main` had independently added a test group lettered `(v)` (`main`'s was the #415 no-check-runs Clear case). Kept `main`'s `(v)` and relettered this PR's group to `(w)`, renaming its `_V`/`_v` identifiers to `_W`/`_w` to avoid the variable-name collision. No assertions changed.


aunger (commented):
Reviewed against issue #419, the PR description, and the implementation on `feature/issue-419-ci-monitor-two-endpoint-lag`.
I ran the full test suite from the worktree and confirmed 154 passed, 0 failed.

## Decision framing

The issue owner asked two questions:

1. Should we implement the fix?
2. Would it obviate the `undiagnosedTerminal` out-of-process retry from #408?

The PR answers both correctly.
The fix is worth doing: it closes a real in-process correctness gap where a `FAIL` marker is silently dropped for the lifetime of the process when the two lagging endpoints settle on different drain attempts.
The out-of-process retry remains necessary and is not made redundant: it handles a disjoint lag regime (multi-minute, outlives the entire Monitor process), while this fix handles a sub-`DRAIN_MAX_ATTEMPTS * DRAIN_DELAY_SECONDS` (~15s) lag that is fully recoverable in-process.
The PR description makes this partitioning explicit and correct.

## Code change

The change is a one-line removal of `break` from the drain loop in `drain_then_print`.
The resulting loop runs every attempt (still bounded by `DRAIN_MAX_ATTEMPTS`), sets `drained = True` on any fruitful attempt, and continues.
The "no new signals" sentinel (`drain poll found no new diagnostic signals`) still prints only when `drained` remains `False` after all attempts--correct, because the sentinel is only meaningful when every attempt was empty.
The logic is right.

## Comments and documentation

The module-level `DRAIN_MAX_ATTEMPTS` block comment, the `drain_then_print` docstring, and the README bullet are all updated to describe the new behavior accurately and to cross-reference issue #419.
No stale "stopping early" or "stops as soon as a poll emits" language survives in the updated text (remaining occurrences explain the old behavior as the rationale for the fix, which is appropriate).

## Tests

Group `(v)` is a clean reproduction of the partial-lag case: step settles on drain attempt 1 (emits), artifact settles on drain attempt 2 (emits), attempt 3 finds nothing new.
The 15-request mock count is correctly derived (5 + 3 + 4 + 3).
The assertions cover: gate step surfaced, FAIL surfaced, Blocked emitted exactly once, both diagnostics precede the terminal, no spurious "no new diagnostic signals" line, and all mocked requests consumed.
Groups `(t)` and `(u)` are updated to supply the now-always-run remaining drain attempts (adding 3 requests each), and their pass/fail assertions are unchanged--confirming no regression in previously passing cases.

## One minor note (not a blocker)

New code comments in the test file use em-dash characters where the project prose-style rule (`/.claude/rules/prose-style.md`) asks for double-hyphens or rephrasing.
Examples: `# drain attempt 1 -- caught up`, `# drain attempt 2 -- everything already seen`.
This is a pre-existing pattern throughout `test_ci_monitor.py` (13 occurrences on `main` before this PR), so this PR is consistent with the established file style rather than introducing a new violation.
Worth noting for a future cleanup pass but not a reason to hold this PR.

## Verdict

The fix is correct, minimal, well-explained, and fully tested.
**Recommend merging.**

---

aunger:
Note the merge conflict

aunger:
🤖 Reviewer

The fix is correct and the reasoning is sound.
Removing the `break` from `drain_then_print` so every drain attempt runs (up to `DRAIN_MAX_ATTEMPTS`) directly addresses the partial-lag case described in issue #419.
The `drained` flag logic is preserved correctly: it becomes `True` if any attempt emits something, so the "no new diagnostic signals" message is still suppressed when at least one drain attempt was fruitful.
The decision that this fix does NOT obviate the out-of-process `undiagnosedTerminal` recheck is also correct--the two mechanisms cover disjoint lag regimes as described.

The new test group `(w)` faithfully reproduces the partial-lag scenario (step surfaces on attempt 1, artifact FAIL on attempt 2) and its failure message `"lagging FAIL marker was dropped"` makes a regression obvious.
Groups `(t)` and `(u)` are updated correctly to supply the now-always-consumed remaining drain attempts without changing their assertions.
All 157 tests pass.
CI checks all pass (build-and-test success, CodeQL success, strip-bylines success).

One issue requires a fix before merge:

**Prose style violation -- em-dashes in new comments**

The `.claude/rules/prose-style.md` rule says: "Replace dashes with commas, semicolons, or parentheses when grammatically possible, and double-hyphens otherwise."

Seven inline comments introduced by this PR in `scripts/test_ci_monitor.py` use em-dashes (`—`) where double-hyphens (`--`) are required:

```
# drain attempt 1 — caught up: step + FAIL surface
# drain attempt 2 — everything already seen, nothing new
# drain attempt 3 — everything already seen, nothing new
# drain attempt 3 — everything already seen, nothing new
# drain attempt 1 — STEP caught up, ARTIFACT still lagging
# drain attempt 2 — ARTIFACT now caught up
# drain attempt 3 — everything already seen, nothing new
```

Each `—` should be replaced with `--`.
The existing em-dashes elsewhere in the file predate this PR and are not this PR's responsibility to fix, but the seven newly introduced ones are.

Decision: **Changes requested**

aunger:
Addressed the requested change in a544069.

Replaced the em-dash characters with double-hyphens (`--`) in the seven drain-attempt comments this PR introduced in groups `(t)` and `(u)`, per `.claude/rules/prose-style.md`. (Group `(w)`'s comments already used `--` from the merge relettering, so no change was needed there.) The pre-existing em-dashes elsewhere in the file are left untouched as out of scope.

Full suite still passes: 157 passed, 0 failed (`python3 scripts/test_ci_monitor.py`).

aunger:
🤖 Reviewer

The previous change request (em-dashes in drain-attempt comments) has been addressed correctly.
All seven newly introduced `# drain attempt N — ...` lines in groups `(t)`, `(u)`, and `(w)` now use `--`.
The surviving em-dashes in the file are all pre-existing and not this PR's responsibility.

The core fix, documentation, and tests are sound:

- Removing the `break` from `drain_then_print` so every attempt runs (still bounded by `DRAIN_MAX_ATTEMPTS`) is the minimal correct fix for the partial-lag case.
- The `drained` flag logic is intact: it is set on any fruitful attempt, so the "no new diagnostic signals" sentinel still fires only when every attempt was empty.
- The decision that this fix does NOT obviate the out-of-process `undiagnosedTerminal` recheck is correct. The two mechanisms cover disjoint lag regimes and the PR description explains this clearly.
- Group `(w)` faithfully reproduces the partial-lag scenario (step on attempt 1, artifact FAIL on attempt 2, attempt 3 empty). The request count arithmetic (5 + 3 + 4 + 3 = 15) is correct. All assertions are appropriate.
- Groups `(t)` and `(u)` are updated correctly to supply the now-always-consumed remaining drain attempts; their block comments and check message strings are updated to describe the new 15-request totals; their pass/fail assertions are unchanged.
- The `DRAIN_MAX_ATTEMPTS` block comment, the `drain_then_print` docstring, and the README bullet all accurately describe the new behavior and cross-reference issue #419. No stale "stopping early" or "stops as soon as a poll emits" language survives in prose describing the current behavior.
- CI checks on the latest commit show no failures.

Decision: **LGTM**

PR #434, opened 2026-06-17
test: semgrep findings smoke test (do not merge)

aunger:
Temporary PR to verify semgrep CI catches security violations that pre-commit does not.

Contains `subprocess.run(..., shell=True)` with user-controlled input and `eval()` -- both expected semgrep findings.

Do not merge.


aunger:
This was a CI test, never intended to be merged.

PR #438, opened 2026-06-18
Make Verification Planner idempotent: skip duplicate issue filing

### LINKED ISSUE #429: Verification planner shouldn't file duplicate issues
The Verification Planner agent's issue filing should be idempotent. Sometimes it runs more than once.

aunger:
Fixes #429

The Verification Planner could file duplicate tracking issues when dispatched more than once for the same PR (for example, after an Orchestrator re-dispatch).

This change adds a comment-based deduplication record. After filing issues, the planner posts a structured comment (marked with `<!-- gb4pc-verification-plan -->`) on the PR that lists every filed issue number in checkbox format, split into before-merging and follow-on sections. On a re-run, the planner reads this comment first, extracts the already-filed issue numbers, fetches each issue to get its title and internal id, and skips filing any item whose expected title (per step 4a or 5a) matches a prior issue's title. The blocking-dependency and sub-issue linking steps (4c and 4d) are still attempted for already-covered items, using the internal id fetched from the prior comment. Step 4c treats a 409 or 422 response as success so duplicate-link attempts do not trigger the plain-text fallback.

### DIFF
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -38,6 +38,7 @@ Only the before-merging list controls the merge gate.
    Search the PR's comments for one that contains the exact HTML marker `<!-- gb4pc-verification-plan -->`.
    If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue URL of the form `https://github.com/{owner}/{repo}/issues/{n}`).
    Treat those issues as already filed and do not create duplicates for the corresponding items.
+   Record the comment's id (the numeric id returned by the comments API, not its URL) for use in step 6.
    If the comment does not exist, proceed with filing all items normally.
 4. If the before-merging list is not empty (i.e., step 2 did not exit), open one GitHub issue for each item on the *before merging* list that is not already covered by a prior-run comment (step 3).
    Do NOT communicate with the user, and do NOT ask whether to test manually or to automate.
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -39,6 +39,8 @@ Only the before-merging list controls the merge gate.
    If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue URL of the form `https://github.com/{owner}/{repo}/issues/{n}`).
    Treat those issues as already filed and do not create duplicates for the corresponding items.
    Record the comment's id (the numeric id returned by the comments API, not its URL) for use in step 6.
+   For each parsed issue URL, fetch the issue (`GET https://api.github.com/repos/{owner}/{repo}/issues/{n}`) and record its title.
+   In steps 4 and 5, an item is "already covered" if its expected title matches the title of a prior issue.
    If the comment does not exist, proceed with filing all items normally.
 4. If the before-merging list is not empty (i.e., step 2 did not exit), open one GitHub issue for each item on the *before merging* list that is not already covered by a prior-run comment (step 3).
    Do NOT communicate with the user, and do NOT ask whether to test manually or to automate.
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -40,7 +40,7 @@ Only the before-merging list controls the merge gate.
    a. Title the issue `(re PR #{number}) {required task title}`, where `{number}` is the current PR number and `{required task title}` is a short title for the outstanding requirement.
    b. **Before filing**: search GitHub for an open or closed issue whose title exactly matches the title from step 3a.
       Use the GitHub search API: `curl -sG -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github+json" https://api.github.com/search/issues --data-urlencode "q=repo:{owner}/{repo} in:title \"{exact title}\" is:issue"`.
-      If any result has a title that exactly matches (case-insensitively), the issue was already filed in a prior run--skip filing and use that existing issue's id and number for steps 3c and 3d.
+      If any result has a title that exactly matches (case-insensitively), the issue was already filed in a prior run--skip filing (step 3c is moot for an existing issue); proceed to steps 3d and 3e using the existing issue's id and number.
       Only create a new issue when no exact title match exists.
    c. In the issue description, include a URL to the particular PR comment that called for this requirement.
       (If the requirement came from the PR or issue description itself rather than a comment, link to that description instead.)
@@ -64,8 +64,8 @@ Only the before-merging list controls the merge gate.
    For each item:
    a. Title the issue simply `{task title}`, without referencing the current PR.
    b. **Before filing**: search GitHub for an open or closed issue whose title exactly matches the title from step 4a.
-      Use the same search approach as step 3b.
-      If an exact title match exists, skip filing and use the existing issue for tracking; do not file a duplicate.
+      Use the same search approach as step 3b, with `q=repo:{owner}/{repo} in:title "{exact title}" is:issue` as the query string.
+      If an exact title match exists, record the existing issue's number for inclusion in the step-5 report; do not re-file or re-write its description.
    c. In the issue description, include a URL to the source comment or description, and state clearly that this issue does **not** block PR #{number} (for example, "This is a follow-on item and does not block merging PR #{number}.").
    d. Do **not** call the `blocked_by` dependency endpoint for follow-on issues.
       Do **not** add any "Blocks PR #..." line to the issue body.
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -39,7 +39,7 @@ Only the before-merging list controls the merge gate.
    If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue URL of the form `https://github.com/{owner}/{repo}/issues/{n}`).
    Treat those issues as already filed and do not create duplicates for the corresponding items.
    If the comment does not exist, proceed with filing all items normally.
-4. Otherwise, open one GitHub issue for each item on the *before merging* list that is not already covered by a prior-run comment (step 3).
+4. If the before-merging list is not empty (i.e., step 2 did not exit), open one GitHub issue for each item on the *before merging* list that is not already covered by a prior-run comment (step 3).
    Do NOT communicate with the user, and do NOT ask whether to test manually or to automate.
    For each item:
    a. Title the issue `(re PR #{number}) {required task title}`, where `{number}` is the current PR number and `{required task title}` is a short title for the outstanding requirement.
@@ -51,6 +51,7 @@ Only the before-merging list controls the merge gate.
    b. In the issue description, include a URL to the source comment or description, and state clearly that this issue does **not** block PR #{number} (for example, "This is a follow-on item and does not block merging PR #{number}.").
 6. Post (or replace) the verification-plan comment on the PR.
    The comment must contain the exact HTML marker `<!-- gb4pc-verification-plan -->` on its own line so future runs can find it.
+   The rebuilt comment must list all issues--those parsed from the prior-run comment in step 3 and any newly filed in steps 4-5--so that future runs can find the complete record and will not re-file already-existing issues.
    Format the comment as follows (use the actual issue URLs):
 
    ```markdown
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -5,7 +5,7 @@
 You are a Verification Planner.
 You are the final check that no requirement (blocking or follow-on) is lost.
 You scan the linked issue and PR and assemble two lists: (1) outstanding requirements that must be handled before merging, and (2) follow-on work that is explicitly deferred or out of scope for this PR.
-You file a tracking GitHub issue for every item in either list and post a verification-plan comment on the PR that records which issues are blocking and which are follow-on.
+You file a tracking GitHub issue for every item in either list, mark blocking items as merge blockers, and post a verification-plan comment on the PR that records which issues were filed.
 You do not communicate with the user and you do not implement anything.
 
 Two kinds of outstanding requirement are your responsibility:
@@ -36,44 +36,65 @@ Only the before-merging list controls the merge gate.
    A non-empty follow-on list does not prevent you from applying the `verified` label; apply it anyway and also report the follow-on list.
 3. **Before filing any issues**, check whether the PR already has a verification-plan comment from a prior run.
    Search the PR's comments for one that contains the exact HTML marker `<!-- gb4pc-verification-plan -->`.
-   If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue URL of the form `https://github.com/{owner}/{repo}/issues/{n}`).
+   If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue number of the form `#{issue number}`).
    Treat those issues as already filed and do not create duplicates for the corresponding items.
    Record the comment's id (the numeric id returned by the comments API, not its URL) for use in step 6.
-   For each parsed issue URL, fetch the issue (`GET https://api.github.com/repos/{owner}/{repo}/issues/{n}`) and record its title.
-   In steps 4 and 5, an item is "already covered" if its expected title matches the title of a prior issue.
+   For each parsed issue ID n, fetch the issue (`GET https://api.github.com/repos/{owner}/{repo}/issues/{n}`) and record its title and internal id (the `id` field, not the issue number).
+   In steps 4 and 5, an item is "already covered" if the title that would be assigned to it by step 4a (for before-merging items) or step 5a (for follow-on items) matches the title of a prior issue.
    If the comment does not exist, proceed with filing all items normally.
 4. If the before-merging list is not empty (i.e., step 2 did not exit), open one GitHub issue for each item on the *before merging* list that is not already covered by a prior-run comment (step 3).
    Do NOT communicate with the user, and do NOT ask whether to test manually or to automate.
    For each item:
    a. Title the issue `(re PR #{number}) {required task title}`, where `{number}` is the current PR number and `{required task title}` is a short title for the outstanding requirement.
+      (Skip this sub-step for already-covered items; use the existing issue's internal id recorded in step 3 for steps 4c and 4d instead.)
    b. In the issue description, include a URL to the particular PR comment that called for this requirement.
       (If the requirement came from the PR or issue description itself rather than a comment, link to that description instead.)
+      (Skip this sub-step for already-covered items.)
+   c. Record that the new issue **blocks** the PR, using GitHub's issue-dependencies feature.
+      GitHub exposes this through the REST API.
+      The `gh` CLI is not installed in the sandbox, so call the REST endpoint with `curl` and `$GITHUB_TOKEN`, using the auth headers (`-H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github+json"`).
+      To mark the PR as blocked by the new issue:
+      `curl -sX POST -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github+json" https://api.github.com/repos/{owner}/{repo}/issues/{number}/dependencies/blocked_by -d '{"issue_id": {issue's internal id}}'`,
+      where `{number}` is the current PR number and `{issue's internal id}` is the issue's internal id (the `id` field, not its number; obtain it from the issue-creation response for new issues, or from the fetch in step 3 for already-covered items).
+      Send the id as a bare JSON integer in the request body (not a quoted string), as the API requires.
+      If the PR number is not accepted for an issue-dependency relationship, fall back to blocking the **parent issue** the PR resolves (the issue this PR fixes) instead, by sending the same request to `.../issues/{parent issue number}/dependencies/blocked_by`.
+      Only if neither call succeeds for a reason other than the dependency already existing (treat a 409 or 422 response as success rather than a failure that triggers the fallback), state the blocking relationship in plain text in the new issue's description (for example, "Blocks PR #{number}") so it is not lost, and skip the formal link without failing.
+   d. Make the new issue a **sub-issue** of the PR, using GitHub's sub-issues feature.
+      GitHub exposes this through the REST API; call it with `curl` and `$GITHUB_TOKEN` as in step 4c (the `gh` CLI is not installed):
+      `curl -sX POST -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github+json" https://api.github.com/repos/{owner}/{repo}/issues/{number}/sub_issues -d '{"sub_issue_id": {issue's internal id}}'`,
+      where `{number}` is the current PR number and `{issue's internal id}` is the issue's internal id (not its number).
+      Send the id as a bare JSON integer in the request body (not a quoted string), as the API requires.
+      If the PR number is not accepted for a sub-issue relationship, fall back to making the new issue a sub-issue of the **parent issue** the PR resolves instead, by sending the same request to `.../issues/{parent issue number}/sub_issues`.
+      If neither call succeeds, skip this link without failing.
 5. Open one GitHub issue for each item on the *follow-on* list that is not already covered by a prior-run comment (step 3).
    For each item:
    a. Title the issue simply `{task title}`, without referencing the current PR.
    b. In the issue description, include a URL to the source comment or description, and state clearly that this issue does **not** block PR #{number} (for example, "This is a follow-on item and does not block merging PR #{number}.").
+   c. Do **not** call the `blocked_by` dependency endpoint for follow-on issues.
+      Do **not** add any "Blocks PR #..." line to the issue body.
 6. Post (or replace) the verification-plan comment on the PR.
    The comment must contain the exact HTML marker `<!-- gb4pc-verification-plan -->` on its own line so future runs can find it.
    The rebuilt comment must list all issues--those parsed from the prior-run comment in step 3 and any newly filed in steps 4-5--so that future runs can find the complete record and will not re-file already-existing issues.
-   Format the comment as follows (use the actual issue URLs):
+   When rebuilding the comment, preserve the checked (`- [x]`) or unchecked (`- [ ]`) state of each item from the prior-run comment for issues that already existed; newly filed issues start as unchecked.
+   Format the comment as follows (use actual issue numbers):
 
    ```markdown
    <!-- gb4pc-verification-plan -->
    ### Verification plan
 
    **Before merging** (must resolve before PR #{number} can be merged):
-   - [ ] {title of item 1} -- {issue URL}
-   - [ ] {title of item 2} -- {issue URL}
+   - [ ] #{issue number of item 1}
+   - [ ] #{issue number of item 2}
 
    **Follow-on** (does not block merging):
-   - [ ] {title of item A} -- {issue URL}
+   - [ ] #{issue number of item A}
    ```
 
    If either section is empty, write "None." in place of the list.
    If a prior-run comment already exists (step 3), replace it rather than posting a second comment.
    To replace it, use the GitHub REST API to edit the existing comment body:
    `curl -sX PATCH -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github+json" https://api.github.com/repos/{owner}/{repo}/issues/comments/{comment_id} -d '{"body": "..."}'`,
-   where `{comment_id}` is the id of the existing verification-plan comment.
+   where `{comment_id}` is the id of the existing verification-plan comment, and `{body}` is the markdown that will completely replace the existing body.
 7. Report both lists to the Orchestrator and exit.
    The PR may be merged once every item on the *before merging* list is resolved.
    Follow-on issues do not gate the merge.
@@ -85,4 +106,4 @@ Only the before-merging list controls the merge gate.
 - Do not modify source files.
 - Do not commit or push anything.
 - Limit your reading to the issue, PR, and project test infrastructure references.
-- The only repository-changing actions you take are filing the tracking issues described above and posting the verification-plan comment on the PR.
+- The only repository-changing actions you take are filing the tracking issues described above, linking them to the PR, and posting the verification-plan comment on the PR.
--- a/agents/verification_planning.md
+++ b/agents/verification_planning.md
@@ -5,7 +5,7 @@
 You are a Verification Planner.
 You are the final check that no requirement (blocking or follow-on) is lost.
 You scan the linked issue and PR and assemble two lists: (1) outstanding requirements that must be handled before merging, and (2) follow-on work that is explicitly deferred or out of scope for this PR.
-You file a tracking GitHub issue for every item in either list, marking blocking items as merge blockers and follow-on items as non-blocking.
+You file a tracking GitHub issue for every item in either list and post a verification-plan comment on the PR that records which issues are blocking and which are follow-on.
 You do not communicate with the user and you do not implement anything.
 
 Two kinds of outstanding requirement are your responsibility:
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

The deduplication logic addresses the right problem, and the search-then-skip structure is a reasonable approach.
However, there are three correctness problems in the new text that need to be fixed before merge.

---

**Problem 1 -- Step 3b omits step 3e from the "reuse existing" instruction (bug)**

Step 3b says:

> skip filing and use that existing issue's id and number for steps 3c and 3d.

Step 3c is "write the description" and step 3d is "record the blocking dependency."
Step 3e ("make the new issue a sub-issue") is not mentioned.
On a re-run the sub-issue link from the prior run may or may not already exist, so the agent should attempt step 3e as well.
As written, the agent is implicitly told to skip 3e on re-runs, which leaves the sub-issue relationship absent if it happened to fail on the first run.

Fix: change "for steps 3c and 3d" to "for steps 3c, 3d, and 3e."

Wait--see also Problem 2 below, which changes what this phrase should say anyway.

---

**Problem 2 -- Step 3b tells the agent to execute step 3c against an already-existing issue (impossible)**

Step 3c says "In the issue description, include a URL to the particular PR comment..."
That instruction assumes you are creating a new issue and writing its body.
But when the issue already exists (the dedup path), executing step 3c means editing an existing issue's description, which is not what is intended and is not what the step instructs.
The phrase "use that existing issue's id and number for steps 3c and 3d" is self-contradictory: step 3c is a write-body action that only makes sense at creation time.

The intent is clearly: skip creation and description-writing (step 3c is moot on re-runs), and still attempt the linking steps (3d and 3e) in case they did not complete on the prior run.

Fix: Rewrite the skip clause to something like: "skip filing the issue (step 3c is moot for an existing issue); proceed to steps 3d and 3e using the existing issue's id and number."

---

**Problem 3 -- Step 4b says "use the existing issue for tracking" without saying how**

For follow-on items the dedup says:

> skip filing and use the existing issue for tracking; do not file a duplicate.

Unlike the before-merging path, there are no linking steps after this point for follow-on items (4c and 4d only apply to new issues).
The instruction gives the agent nothing concrete to do with the existing issue.
In particular, the agent is not told to include the existing issue number in its step-5 report to the Orchestrator, which is the only way the Orchestrator learns the issue exists.

Fix: Spell out what "use the existing issue for tracking" means; for example: "Record the existing issue's number for inclusion in the step-5 report; do not re-file or re-write its description."

---

**Nit -- Step 4b says "use the same search approach as step 3b" without clarifying the query string**

Step 3b gives the curl command in full.
Step 4b says only "use the same search approach."
A follow-on issue title does not include the `(re PR #{number})` prefix, so the `q=` value differs.
Referring back to 3b is fine for the curl boilerplate, but omitting the concrete `q=` example for step 4b leaves unnecessary ambiguity.
This is a nit, not a blocker on its own, but fixing problems 1-3 is a natural opportunity to also add the concrete query form.

---

CI: CodeQL passed, build-and-test passed (skipped as expected for a docs-only change), strip-bylines passed.

aunger (commented):
🤖 Author

All three correctness problems and the nit are addressed in the follow-up commit (e21f897).

**Problem 1+2**: Rewrote the step 3b skip clause to: "skip filing (step 3c is moot for an existing issue); proceed to steps 3d and 3e using the existing issue's id and number." This removes the self-contradiction (step 3c no longer appears in the re-run path) and adds 3e so the sub-issue link is attempted even on re-runs.

**Problem 3**: The step 4b skip clause now reads: "record the existing issue's number for inclusion in the step-5 report; do not re-file or re-write its description." The agent has a concrete action instead of a vague directive.

**Nit**: Step 4b now includes the concrete query string `q=repo:{owner}/{repo} in:title "{exact title}" is:issue` so the agent does not have to infer the format from step 3b.

aunger:
🤖 Reviewer

The previous review's three correctness problems and the nit were all addressed correctly in the follow-up commit.
The step 3b skip clause now says "step 3c is moot; proceed to steps 3d and 3e," step 4b now gives a concrete action and a concrete query string, and CI is all green.
However there is one new correctness gap that the previous review did not catch, and one residual concern.

---

**Problem 1 -- Step 4b does not tell the agent which follow-on steps are skipped on a re-run (bug)**

Step 3b was fixed in the prior round to say explicitly: "step 3c is moot for an existing issue; proceed to steps 3d and 3e."
Step 4b was not given the same treatment.

As written, step 4b says:

> "If an exact title match exists, record the existing issue's number for inclusion in the step-5 report; do not re-file or re-write its description."

The phrase "do not re-write its description" implicitly covers 4c (which writes the description), but only by implication.
An agent that reads "do not re-write its description" and then arrives at step 4c ("In the issue description, include a URL...") sees two instructions pulling in opposite directions without a clear resolution--the same tension that existed in the before-merging path before it was fixed.

The before-merging path resolved this by using the word "moot."
Step 4b should do the same: replace "do not re-file or re-write its description" with "step 4c is moot for an existing issue."
That matches the phrasing already established in step 3b and removes any ambiguity about which steps apply on a re-run.

---

**Problem 2 -- Step 3d's plaintext fallback is not idempotent on re-runs (bug)**

Step 3d says: "Only if neither call succeeds, state the blocking relationship in plain text in the new issue's description."

On a re-run where the blocking link was already successfully created in the prior run, the dependency API may return a 409 or 422 because the link already exists.
If that happens, the agent would attempt to edit the existing issue's description to add "Blocks PR #...," which contradicts the intent established in step 3b ("step 3c is moot for an existing issue") and is not idempotent.

Step 3e does not have this problem because its fallback is simply "skip this link without failing"--it never writes to the issue body.

The fix for step 3d is to qualify the plaintext fallback.
For example: "Only if neither call succeeds for a reason other than the dependency already existing (for example, treat a 409 or 422 response as success), state the blocking relationship in plain text..."
Alternatively, note that a 409/422 on this endpoint should be treated as success rather than a failure that triggers the fallback.

---

aunger:
Author response (commit 1c658c3).

Both problems from the second review are addressed:

**Problem 1 (step 4b)**: Replaced "do not re-file or re-write its description" with "step 4c is moot for an existing issue." This matches the "moot" phrasing established in step 3b and removes any ambiguity about which steps apply on a re-run.

**Problem 2 (step 3d)**: Qualified the plaintext fallback to exclude the case where the dependency already exists. The step now reads: "Only if neither call succeeds for a reason other than the dependency already existing (treat a 409 or 422 response as success rather than a failure that triggers the fallback)..." This makes step 3d idempotent on re-runs--a duplicate-link response no longer causes the agent to write to the issue body.

aunger (commented):
🤖 Reviewer

All three correctness problems and the nit from the first review, and both correctness problems from the second review, are correctly addressed in the current HEAD (1c658c3).

Going through each point in the current file:

**Step 3b**: The deduplication check is present with the full curl command.
The skip clause reads "skip filing (step 3c is moot for an existing issue); proceed to steps 3d and 3e using the existing issue's id and number."
This is correct: description-writing is excluded, linking steps are included, and 3e is not silently dropped.

**Step 3d plaintext fallback**: Now reads "Only if neither call succeeds for a reason other than the dependency already existing (treat a 409 or 422 response as success rather than a failure that triggers the fallback)..."
A duplicate-link response on re-run no longer causes the agent to write to the issue body.
The step is idempotent.

**Step 4b**: The deduplication check is present with the concrete query string (`q=repo:{owner}/{repo} in:title "{exact title}" is:issue`).
The skip clause reads "step 4c is moot for an existing issue," matching the phrasing established in 3b.
The agent has a concrete action (record for the step-5 report) and no ambiguity about which steps to skip.

**Step 4d**: The "do not call blocked_by" prohibition is a blanket restriction, not a conditional fallback, so it does not need the same idempotency treatment as 3d.
It is correct as written.

**Step numbering**: The re-lettering from b/c/d to b/c/d/e in step 3, and b/c/d in step 4, is consistent throughout.
The cross-reference in 3e ("as in step 3d") is correct.

CI is all green (CodeQL, Semgrep, build-and-test, file-issues all pass; dev-build correctly skipped for a docs-only change).

**Verdict: LGTM.**

agents/verification_planning.md
```diff
@@ -38,19 +38,23 @@ Only the before-merging list controls the merge gate.
    Do NOT communicate with the user, and do NOT ask whether to test manually or to automate.
    For each item:
    a. Title the issue `(re PR #{number}) {required task title}`, where `{number}` is the current PR number and `{required task title}` is a short title for the outstanding requirement.
-   b. In the issue description, include a URL to the particular PR comment that called for this requirement.
+   b. **Before filing**: search GitHub for an open or closed issue whose title exactly matches the title from step 3a.
```
aunger: The title generated in 3a isn’t deterministic. 

Also, the links (blocked by, child of) really don’t work at this point. 

I propose that the Verification Planner should add a comment to the PR with links and titles of the created tasks (and check for such a comment to deduplicate before creating them).

aunger: 🤖 Author

Both points accepted. Implemented in commit 37ab116.

**Title non-determinism**: Agreed--an AI-generated title is not a reliable deduplication key across runs. The new approach abandons title-based search entirely.

**Links (blocked_by and sub-issues)**: Also removed. Those API calls did not work reliably in practice, so they were dead weight.

**New approach**: The planner now posts a structured PR comment (marked with the HTML anchor `<!-- gb4pc-verification-plan -->`) after filing issues. The comment lists every filed issue URL in a checkbox format, split into before-merging and follow-on sections. On a re-run, the planner reads this comment first, extracts the already-filed issue URLs, and skips those items--no title matching required. If a comment already exists, it is replaced (PATCH) rather than a duplicate being posted.

This is fully idempotent: the comment is the single source of truth, the key is the stable issue URL rather than a generated title, and replacing the comment on a re-run is a no-op if nothing changed.

aunger:
🤖 Reviewer

The pivot from title-based search to a comment-based deduplication record is the right call: it eliminates non-deterministic title matching as a failure mode, removes the broken dependency/sub-issue API calls, and gives the planner a single stable source of truth. The structure is clean.

Two correctness problems remain in the current HEAD (37ab116).

---

**Problem 1 -- "Otherwise" in step 4 has a broken antecedent (ambiguity)**

Steps are now numbered 2, 3, 4, 5, 6, 7. The early-exit condition is in step 2. Step 3 is the new idempotency check (no branching; always runs when step 2 did not exit). Step 4 opens with:

> "Otherwise, open one GitHub issue for each item on the *before merging* list..."

"Otherwise" made sense in the old file where step 2 was immediately followed by step 3 ("Otherwise, open..."). Now there is an intervening step with no else clause, and an agent reading step 4 after completing step 3 cannot tell what "Otherwise" is the opposite of. Is it the else-branch of step 3's "if the comment does not exist" clause? Is it still the else-branch of step 2? The correct reading is still step 2's else-branch, but the word no longer signals that clearly.

Replace "Otherwise" with something self-contained such as "If the before-merging list is not empty (i.e., step 2 did not exit)," so the conditional is explicit.

---

**Problem 2 -- Step 6 does not say the rebuilt comment must include previously filed issues (correctness gap)**

Step 3 parses existing issue URLs out of the prior-run comment and marks those items as "already filed." Steps 4 and 5 file only the remaining (unfiled) items. Step 6 then says: "Post (or replace) the verification-plan comment" with a format that uses "the actual issue URLs."

On a partial re-run (some items already filed, some new), the comment is rebuilt from scratch (PATCH replaces the body). "The actual issue URLs" is ambiguous: it could mean only the issues filed in the current run (steps 4-5), which would erase the prior entries and break the next run's deduplication. The intent is obviously to include all issues--prior and new--but that intent is not stated.

Add a sentence to step 6 making this explicit, for example: "The rebuilt comment must list all issues--those parsed from the prior-run comment in step 3 and any newly filed in steps 4-5--so that future runs can find the complete record."

Without this, an agent that files one new issue after finding two prior ones could post a comment with only one URL in it, causing the next run to re-file the other two.

---

CI: Semgrep, CodeQL, build-and-test, file-issues, strip-bylines all pass. dev-build correctly skipped for a docs-only change.

aunger:
🤖 Author

Both problems from the latest review are addressed in commit 470af41.

**Problem 1 ("Otherwise" broken antecedent)**: Replaced "Otherwise, open one GitHub issue..." with "If the before-merging list is not empty (i.e., step 2 did not exit), open one GitHub issue..." so the conditional is explicit and self-contained. An agent completing step 3 now has an unambiguous signal for when step 4 applies, without having to trace back through an intervening step.

**Problem 2 (rebuilt comment must include prior-run issues)**: Added a sentence to step 6 stating: "The rebuilt comment must list all issues--those parsed from the prior-run comment in step 3 and any newly filed in steps 4-5--so that future runs can find the complete record and will not re-file already-existing issues." This closes the gap where an agent filing one new issue after finding two prior ones could have posted a comment with only the new URL, causing the next run to re-file the other two.

aunger (commented):
Reviewer

Both fixes from the previous review round are correctly applied at HEAD (470af41).

Problem 1 ("Otherwise" broken antecedent): Step 4 now reads "If the before-merging list is not empty (i.e., step 2 did not exit)..." -- correct and unambiguous.

Problem 2 (rebuilt comment must include all prior-run issues): Step 6 now contains the explicit sentence requiring the rebuilt comment to list all issues from step 3 and steps 4-5. Correct.

However, there is one new correctness gap that none of the prior rounds caught.

---

**Problem -- Step 6 references `{comment_id}` that step 3 never tells the agent to record (bug)**

Step 6 says:

> "If a prior-run comment already exists (step 3), replace it rather than posting a second comment. To replace it, use the GitHub REST API to edit the existing comment body: `curl ... https://api.github.com/repos/{owner}/{repo}/issues/comments/{comment_id} ...`, where `{comment_id}` is the id of the existing verification-plan comment."

Step 3 says only: search for the comment, parse its issue URLs, treat those issues as already filed.

Step 3 never says to record the comment's id.

An agent executing the steps in order arrives at step 6 needing `{comment_id}`, but was given no instruction to capture it in step 3. The agent would have to infer from context that it should have saved the id, or re-issue the search. Neither is specified.

The fix is straightforward: add one sentence to step 3 telling the agent to record the comment's id when it is found. For example, after "Treat those issues as already filed and do not create duplicates for the corresponding items," add: "Record the comment's id (the numeric id returned by the comments API, not its URL) for use in step 6."

---

CI: Semgrep, CodeQL, build-and-test, file-issues all pass. dev-build correctly skipped for a docs-only change.

aunger:
Author response (commit 46e8d69).

The problem from the latest review is addressed.

**Problem (step 3 never records the comment id)**: Added a sentence to step 3 immediately after "Treat those issues as already filed and do not create duplicates for the corresponding items." The new sentence reads: "Record the comment's id (the numeric id returned by the comments API, not its URL) for use in step 6." An agent executing steps in order now has an explicit instruction to capture the value that step 6's PATCH curl command needs as `{comment_id}`.

aunger:
🤖 Reviewer

The pivot to a comment-based deduplication record is sound, and all previously requested fixes are correctly applied at HEAD (46e8d69). CI passes (Semgrep, CodeQL, build-and-test, file-issues all green; dev-build correctly skipped for a docs-only change).

One correctness gap remains.

---

**Problem -- Step 3 does not tell the agent how to match prior issue URLs back to assembled items (correctness gap)**

Step 3 says: parse the prior-run comment to extract issue URLs, then "Treat those issues as already filed and do not create duplicates for the corresponding items."

Step 4 then says to open an issue for each before-merging item "that is not already covered by a prior-run comment (step 3)."
Step 5 says the same for follow-on items.

But "already covered" requires matching a parsed URL to an assembled item. The agent has two things in hand: a list of assembled items (from step 1, which are descriptions/titles, not URLs) and a list of prior issue URLs (from step 3). There is no way to determine which item "corresponds to" which URL without fetching each prior issue and comparing its title to the assembled item titles.

Step 3 gives no instruction to do that fetch. An agent completing step 3 holds a set of URLs and no titles. When it arrives at step 4 and tries to decide whether item X is "already covered," it has no basis to match without an unspecified round-trip to the GitHub API.

The title-search approach used in the earlier revision of this PR handled this naturally: the agent searched per-item before filing, so the match was intrinsic to the loop. The new approach inverts the order (bulk-parse first, then filter), which requires an explicit reconciliation step that is currently absent.

The fix is to add one instruction to step 3 telling the agent to fetch each parsed issue URL (a GET to the GitHub issues API) and record the title of each prior issue, so that in steps 4 and 5 it can compare item titles against prior-issue titles to decide what is "already covered." For example, after "Record the comment's id ... for use in step 6," add: "For each parsed issue URL, fetch the issue (GET `https://api.github.com/repos/{owner}/{repo}/issues/{n}`) and record its title. In steps 4 and 5, an item is 'already covered' if its expected title matches the title of a prior issue."

Without this, the matching logic is left entirely to agent inference, and an agent could plausibly treat "already covered" as covering all items (re-filing nothing on a re-run even when the list has grown) or could re-file everything (having no way to match URLs to items without the titles).

---

aunger:
🤖 Author

The correctness gap from the latest review is addressed in commit 1937f14.

**Problem (step 3 never tells the agent to fetch prior issue titles)**: After "Record the comment's id (the numeric id returned by the comments API, not its URL) for use in step 6," added two sentences: one directing the agent to GET each parsed issue URL and record the issue's title, and one defining what "already covered" means in steps 4 and 5 (a title match against the set of prior-issue titles). An agent arriving at step 4 or 5 now has an explicit, unambiguous basis for the comparison without any unspecified API round-trip.

aunger:
🤖 Reviewer

The comment-based deduplication structure is sound and all problems from the previous round are addressed. CI passes across the board (Semgrep, CodeQL, build-and-test, file-issues all green; dev-build correctly skipped). One correctness gap remains.

---

**Problem -- "Expected title" in step 3 is undefined (bug)**

The two sentences added in commit 1937f14 read:

> "For each parsed issue URL, fetch the issue (`GET https://api.github.com/repos/{owner}/{repo}/issues/{n}`) and record its title."
> "In steps 4 and 5, an item is 'already covered' if its expected title matches the title of a prior issue."

The instruction to fetch and record titles is correct. The matching definition, however, uses the phrase "its expected title" without saying what that means. Steps 4a and 5a define the title formats (before-merging: `(re PR #{number}) {required task title}`; follow-on: `{task title}`), but they appear after step 3. An agent reading step 3 in isolation has no definition to work from.

The failure mode is real: an agent could plausibly interpret "expected title" as just the bare task description (without the `(re PR #{number})` prefix), match nothing, and re-file everything. Or it could match only on partial strings. The previous title-search approach avoided this because the per-item search query was explicit. The current approach requires the agent to construct the expected title and compare--but "expected" is never grounded to a concrete format.

The fix is to replace "its expected title" with an explicit reference to the format in steps 4a/5a. For example: "In steps 4 and 5, an item is 'already covered' if the title that would be assigned to it by step 4a (for before-merging items) or step 5a (for follow-on items) matches the title of a prior issue."

---

**Nit -- PR description is stale**

The PR body says "The existing issue's id is then used for the normal blocking-dependency and sub-issue links." The current implementation has no dependency or sub-issue linking; those steps were removed in the pivot to the comment-based scheme. The PR description should be updated to reflect what the implementation actually does.

---

agents/verification_planning.md
aunger: > Links (blocked_by and sub-issues): Also removed. Those API calls did not work reliably in practice, so they were dead weight.

Restore these. I’d rather fix it, but fixing and deleting are both beyond scope. 

I just pointed out that they don’t work because they can’t be used for the planner’s discovery.

agents/verification_planning.md
```diff
@@ -34,35 +34,47 @@ Only the before-merging list controls the merge gate.
...
+
+   **Follow-on** (does not block merging):
+   - [ ] {title of item A} -- {issue URL}
+   ```
+
+   If either section is empty, write "None." in place of the list.
+   If a prior-run comment already exists (step 3), replace it rather than posting a second comment.
+   To replace it, use the GitHub REST API to edit the existing comment body:
+   `curl -sX PATCH -H "Authorization: Bearer $GITHUB_TOKEN" -H "Accept: application/vnd.github+json" https://api.github.com/repos/{owner}/{repo}/issues/comments/{comment_id} -d '{"body": "..."}'`,
+   where `{comment_id}` is the id of the existing verification-plan comment.
```
aunger:
```suggestion
   where `{comment_id}` is the id of the existing verification-plan comment, and {body} is the markdown that will completely replace the existing body.
```

Correct me or push back if I’m wrong or you disagree.

agents/verification_planning.md
aunger, line 1: Make a note to preserve and recreate the checked status of existing items.

aunger, line 1: Should existing items that weren’t completed but no longer apply be kept as strikethrough without a checkbox? Or is that needless complication?

agents/verification_planning.md
```diff
@@ -34,35 +34,47 @@ Only the before-merging list controls the merge gate.
...
    A non-empty follow-on list does not prevent you from applying the `verified` label; apply it anyway and also report the follow-on list.
-3. Otherwise, open one GitHub issue for each item on the *before merging* list.
+3. **Before filing any issues**, check whether the PR already has a verification-plan comment from a prior run.
+   Search the PR's comments for one that contains the exact HTML marker `<!-- gb4pc-verification-plan -->`.
+   If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue URL of the form `https://github.com/{owner}/{repo}/issues/{n}`).
+   Treat those issues as already filed and do not create duplicates for the corresponding items.
+   Record the comment's id (the numeric id returned by the comments API, not its URL) for use in step 6.
+   For each parsed issue URL, fetch the issue (`GET https://api.github.com/repos/{owner}/{repo}/issues/{n}`) and record its title.
```
aunger:
```suggestion
   For each parsed issue ID n, fetch the issue (`GET https://api.github.com/repos/{owner}/{repo}/issues/{n}`).
```

agents/verification_planning.md
```diff
@@ -34,35 +34,47 @@ Only the before-merging list controls the merge gate.
...
+   In steps 4 and 5, an item is "already covered" if its expected title matches the title of a prior issue.
```
aunger:
```suggestion
   In steps 4 and 5, an item is "already covered" if it is perfectly described by a prior issue.
```

agents/verification_planning.md
```diff
@@ -34,35 +34,47 @@ Only the before-merging list controls the merge gate.
    - the *follow-on* list, noting for each item the URL of the source comment or description and a brief reason it is not a merge blocker (e.g., "explicitly deferred in PR comment," "out of scope for this PR").
 2. If the *before merging* list is empty, apply the `verified` label to both the PR and the issue it resolves, report this success to the Orchestrator (no unautomated steps or outside-the-repo requirements were identified; the PR may be merged) and exit.
    A non-empty follow-on list does not prevent you from applying the `verified` label; apply it anyway and also report the follow-on list.
-3. Otherwise, open one GitHub issue for each item on the *before merging* list.
+3. **Before filing any issues**, check whether the PR already has a verification-plan comment from a prior run.
+   Search the PR's comments for one that contains the exact HTML marker `<!-- gb4pc-verification-plan -->`.
+   If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue URL of the form `https://github.com/{owner}/{repo}/issues/{n}`).
```
aunger:
```suggestion
   If such a comment exists, parse it to extract the list of already-filed issues (each line with a `- [ ]` or `- [x]` checkbox carries an issue number of the form `#{issue number}`).
```

agents/verification_planning.md
```diff
@@ -34,35 +34,47 @@ Only the before-merging list controls the merge gate.
...
+   ```markdown
+   <!-- gb4pc-verification-plan -->
+   ### Verification plan
+
+   **Before merging** (must resolve before PR #{number} can be merged):
+   - [ ] {title of item 1} -- {issue URL}
+   - [ ] {title of item 2} -- {issue URL}
+
+   **Follow-on** (does not block merging):
+   - [ ] {title of item A} -- {issue URL}
```
aunger:
```suggestion
   - [ ] #{issue number of item A}
```

agents/verification_planning.md
```diff
@@ -34,35 +34,47 @@ Only the before-merging list controls the merge gate.
...
+   The rebuilt comment must list all issues--those parsed from the prior-run comment in step 3 and any newly filed in steps 4-5--so that future runs can find the complete record and will not re-file already-existing issues.
+   Format the comment as follows (use the actual issue URLs):
+
+   ```markdown
+   <!-- gb4pc-verification-plan -->
+   ### Verification plan
+
+   **Before merging** (must resolve before PR #{number} can be merged):
+   - [ ] {title of item 1} -- {issue URL}
+   - [ ] {title of item 2} -- {issue URL}
```
aunger:
```suggestion
   - [ ] #{issue number of item 2}
```

agents/verification_planning.md
```diff
@@ -34,35 +34,47 @@ Only the before-merging list controls the merge gate.
...
+   The comment must contain the exact HTML marker `<!-- gb4pc-verification-plan -->` on its own line so future runs can find it.
+   The rebuilt comment must list all issues--those parsed from the prior-run comment in step 3 and any newly filed in steps 4-5--so that future runs can find the complete record and will not re-file already-existing issues.
+   Format the comment as follows (use the actual issue URLs):
+
+   ```markdown
+   <!-- gb4pc-verification-plan -->
+   ### Verification plan
+
+   **Before merging** (must resolve before PR #{number} can be merged):
+   - [ ] {title of item 1} -- {issue URL}
```
aunger:
```suggestion
   - [ ] #{issue number of item 1}
```

aunger (commented):
Owner review

aunger:
🤖 Author

All open review feedback addressed in commit 20cae41.

**Reviewer bug (step 3, "expected title" undefined)**: Replaced "its expected title matches the title of a prior issue" with "the title that would be assigned to it by step 4a (for before-merging items) or step 5a (for follow-on items) matches the title of a prior issue." The matching key is now explicitly grounded to the concrete title formats already stated in steps 4a and 5a.

**Owner request (restore blocking-dependency and sub-issue links)**: Steps 4c and 4d are restored. The plaintext fallback in step 4c is idempotent: a 409 or 422 response (dependency already exists) is treated as success rather than triggering the fallback. For already-covered items, steps 4a and 4b are skipped but 4c and 4d still run, using the internal id recorded in step 3.

**Owner suggestion (line 39, parse issue numbers not URLs)**: The comment format now carries `#{issue number}` per checkbox line instead of full issue URLs. Step 3 now says "carries an issue number of the form `#{issue number}`."

**Owner suggestion (line 42, fetch by issue ID)**: Changed "For each parsed issue URL" to "For each parsed issue ID n." Step 3 also now records the internal `id` field alongside the title, so steps 4c and 4d have it on re-runs.

**Owner suggestion (line 43, "already covered" definition)**: Declined the suggestion to replace with "it is perfectly described by a prior issue"--that phrasing is vaguer, not clearer, and reintroduces the ambiguity the reviewer identified. The explicit step-4a/5a reference is kept instead.

**Owner suggestion (line 76, body description in PATCH curl)**: Added ", and `{body}` is the markdown that will completely replace the existing body" after the `{comment_id}` description.

**Owner suggestion (preserve checked status)**: Added a sentence to step 6: "When rebuilding the comment, preserve the checked (`- [x]`) or unchecked (`- [ ]`) state of each item from the prior-run comment for issues that already existed; newly filed issues start as unchecked."

**Owner question (strikethrough for removed items)**: Declining this as needless complication. An item that no longer applies is simply not included in the rebuilt comment; a missing entry in the new comment is equally clear and avoids the complexity of tracking which prior items are "removed vs. resolved." If this is wanted, a follow-on issue is the right place for it.

**Owner suggestions (lines 65-69, checkbox format)**: The comment template now uses `- [ ] #{issue number of item N}` instead of `- [ ] {title} -- {issue URL}`. This is consistent with the parsing change in step 3.

**PR description**: Updated to accurately describe the final implementation.

aunger (commented):
🤖 Reviewer

The comment-based deduplication structure is sound, all problems from the previous rounds are addressed, and CI passes across the board (Semgrep, CodeQL, build-and-test, file-issues, strip-bylines all green; dev-build and CodeQL language jobs correctly skipped for a docs-only change).

One correctness gap remains at HEAD (20cae41).

---

**Problem -- Step 4's outer loop and step 4a's parenthetical contradict each other, making the 4c/4d retry unreachable for already-covered items (bug)**

Step 4 opens with:

> "If the before-merging list is not empty (i.e., step 2 did not exit), open one GitHub issue for each item on the *before merging* list that is not already covered by a prior-run comment (step 3)."

The phrase "that is not already covered" scopes the entire loop to non-covered items only. An agent reading this sets aside every already-covered item before entering the loop.

Step 4a then says:

> "(Skip this sub-step for already-covered items; use the existing issue's internal id recorded in step 3 for steps 4c and 4d instead.)"

This parenthetical assumes already-covered items are inside the loop--it says to skip 4a for them but still run 4c and 4d. But those items were already excluded by the step 4 header. An agent following the instructions top-to-bottom would never reach 4a for an already-covered item, and would therefore never attempt the blocking-dependency (4c) or sub-issue (4d) link retries for it.

This matters in practice. If the first run successfully filed an issue but the 4c curl call failed (network timeout, rate limit, etc.), a re-run would correctly identify the issue as already-covered, exclude it from the loop per the step 4 header, and never retry 4c. The blocking-dependency gap from the first run would persist permanently.

The before-merging path is the only place this applies; step 5 has no sub-steps beyond filing, so its outer "not already covered" filter is correct and consistent.

The fix is to change the step 4 loop scope so it is explicit that already-covered items still enter the loop for sub-steps 4c and 4d. For example, replace "open one GitHub issue for each item on the *before merging* list that is not already covered by a prior-run comment (step 3)" with something like: "For each item on the *before merging* list, do the following. Skip sub-steps 4a and 4b for items already covered by a prior-run comment (step 3), but still execute sub-steps 4c and 4d for those items using the internal id recorded in step 3."

That removes the outer filter that makes already-covered items invisible to the loop, while preserving the intent that 4a and 4b are only for new issues.

---

aunger:
🤖 Author

The correctness bug from the latest review is addressed in commit 1a59c40.

**Problem (step 4 outer loop excludes already-covered items from 4c/4d)**: The step 4 header previously said "open one GitHub issue for each item on the before merging list that is not already covered by a prior-run comment (step 3)." The "that is not already covered" filter made already-covered items invisible to the loop, so an agent following instructions top-to-bottom would never reach sub-steps 4c or 4d for them--meaning a failed blocking-dependency or sub-issue link from a prior run could never be retried.

The fix replaces the outer filter with an explicit loop-and-skip structure: "for each item on the before merging list, do the following. Skip sub-steps 4a and 4b for items already covered by a prior-run comment (step 3), but still execute sub-steps 4c and 4d for those items using the internal id recorded in step 3." Already-covered items now enter the loop, skip only the filing sub-steps, and still reach the linking sub-steps. The redundant "(Skip this sub-step for already-covered items.)" parentheticals in 4a and 4b are cleaned up accordingly (4a retains a short note; 4b's is removed since the outer instruction now covers both).

aunger (commented):
🤖 Reviewer

The comment-based deduplication design is sound, and every correctness gap identified in prior rounds has been correctly closed. Here is my per-step assessment of the final text (1a59c40).

**Step 3 (parse prior-run comment)**

- Marker search, issue-number extraction, comment-id recording, per-issue GET to capture title and internal id, and the "already covered" definition (grounded explicitly to the title formats in steps 4a/5a) are all present and consistent.
- The `- [ ]` / `- [x]` parsing format matches the template in step 6. No ambiguity.

**Step 4 (before-merging loop)**

- The outer loop now reads "for each item on the before merging list, do the following. Skip sub-steps 4a and 4b for items already covered..., but still execute sub-steps 4c and 4d for those items using the internal id recorded in step 3." This correctly fixes the prior bug where the outer "not already covered" filter made already-covered items invisible to the loop, preventing 4c/4d retries.
- Step 4a retains the "(Skip this sub-step for already-covered items.)" parenthetical; step 4b's redundant note was removed; both are consistent with the outer instruction.
- Step 4c's 409/422 success treatment is present and makes the plaintext fallback idempotent on re-runs.
- Step 4d's fallback is "skip without failing," not "write to issue body," so a duplicate-link response on re-runs causes a harmless silent skip rather than an incorrect body write. No idempotency problem here.
- Cross-reference in 4d correctly says "as in step 4c" (old "step 3c" reference is gone).

**Step 5 (follow-on list)**

- Simple outer filter "not already covered by a prior-run comment (step 3)" is correct: follow-on items have no linking sub-steps, so there is no retry concern that would require items to enter the loop.

**Step 6 (post or replace comment)**

- Explicit sentence requiring the rebuilt comment to list all issues (from step 3 and steps 4-5) is present. This closes the prior gap where a partial re-run could have overwritten prior entries.
- Preserved checked/unchecked state instruction is present.
- Both `{comment_id}` and `{body}` are defined in the PATCH curl description. `{comment_id}` is grounded to step 3's recording instruction.
- The `<!-- gb4pc-verification-plan -->` marker appears consistently in step 3 (search) and step 6 (emit).

**Role sentence and Boundaries bullet** are updated to accurately reflect the new actions (posting the verification-plan comment, linking).

**CI**: Semgrep, CodeQL, build-and-test, file-issues, strip-bylines all pass. dev-build and language-analysis jobs correctly skipped for a docs-only change.

No correctness gaps found.

**Verdict: LGTM.**

PR #439, opened 2026-06-18
Add imperative clear-on-terminal step for silentVanish re-launch flag

### LINKED ISSUE #430: Re-launch flag nit in Orchestrator CI Monitor instructions
_Originally posted by 🤖 Reviewer in https://github.com/aunger/gallery-button-for-pixel-camera/pull/427#pullrequestreview-4513084938_

---

One non-blocking nit: the comment block states the re-launch flag should also be cleared "whenever a Monitor invocation delivers any terminal line," but no executable step in the pseudocode does that, only the original-launch line clears it. In practice this is harmless, since a terminal line exits the Monitor loop to its routing (newAuthor / escalate / Clear path) and the stale flag is never re-consulted within the same CI cycle. Worth an imperative clear-on-terminal step for clarity if you touch this again, but it is not a correctness defect and does not block merge.

aunger:
The `silentVanish` comment block in `agents/dev_orchestration.md` stated the re-launch flag should be cleared "whenever a Monitor invocation delivers any terminal line," but no executable pseudocode step actually did so.
Only the original-launch line cleared the flag.

This adds explicit "clear the silentVanish re-launch flag" steps to each terminal branch (Blocked, Infra, and Clear) so the pseudocode matches its own stated invariant.

Fixes #430

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -149,11 +149,12 @@ Monitor output lines are relayed to the user verbatim; this is user-facing statu
     Act only on the terminal lines Clear, Blocked, or Infra. Relay in_progress lines to the user as brief status updates (the script suppresses these unless no other output has been emitted for over 120 seconds).
     Relay `step "..." -> ...` and `FAIL [...] ...` lines to the user as informational test-result deltas; they do NOT end the loop or start a new Author round.
     if Monitor emits `drain poll found no new diagnostic signals` immediately followed by a Blocked or Infra line → goto undiagnosedTerminal
-    if Monitor emits a Blocked line  → goto newAuthor
-    if Monitor emits an Infra line   → escalate to user; stop
+    if Monitor emits a Blocked line  → clear the silentVanish re-launch flag; goto newAuthor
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

LGTM.

The three added "clear the silentVanish re-launch flag" steps (Blocked, Infra, and Clear branches) directly close the gap the issue identified: the comment block promised the flag would be cleared "whenever a Monitor invocation delivers any terminal line," but the pseudocode only cleared it on the original launch.
After this change the pseudocode matches that invariant exactly.

A few things I checked explicitly:

- The timeout branch intentionally has no clear.
  The comment says "whenever a Monitor invocation delivers any terminal line"; a timeout is an event, not a delivered line, so the omission is correct.
- The undiagnosedTerminal re-run path routes back to "Act only on the terminal lines..." and therefore hits the same Blocked/Infra/Clear branches with their new clears.
  No separate clear is needed there.
- The silentVanish comment block (lines 200-205) still accurately describes the flag semantics after this change; no prose update is needed.

The change is minimal, correct, and directly addresses the issue.

PR #440, opened 2026-06-18
Remove verified label automatically on new PR push

### LINKED ISSUE #431: GitHub Action: remove `verified` label on new push to PR
## Problem

The `verified` label applied by the Verification Planner becomes stale as soon as a new commit is pushed to a PR.
A PR can carry the label even after changes have been made that have not been re-verified, giving a false impression of readiness.

## Proposed fix

Add a GitHub Actions workflow that triggers on `pull_request` (types: `synchronize`, `reopened`) and removes the `verified` label from the PR if it is present.

This ensures the label only ever reflects the most recent verification pass.

aunger:
The `verified` label applied by the Verification Planner becomes stale as soon as a new commit is pushed to a PR. This adds a GitHub Actions workflow that removes it on `synchronize` and `reopened` events so the label only ever reflects the most recent verification pass.

Fixes #431

## How it works

The new `.github/workflows/remove-verified-on-push.yml` workflow triggers on `pull_request` types `synchronize` and `reopened`. It calls the GitHub Labels REST API to check whether `verified` is present on the PR and, if so, issues a `DELETE` to remove it. When the label is absent the step logs that and exits cleanly.

## Test plan

- [ ] Push a commit to a PR that carries the `verified` label and confirm the workflow run removes the label.
- [ ] Push a commit to a PR that does not carry the `verified` label and confirm the workflow run completes without error.

### DIFF
--- a/.github/workflows/remove-verified-on-push.yml
+++ b/.github/workflows/remove-verified-on-push.yml
@@ -11,7 +11,7 @@ on:
 
 permissions:
   issues: write
-  pull-requests: read
+  pull-requests: write
 
 jobs:
   remove-verified-label:
--- a/.github/workflows/remove-verified-on-push.yml
+++ b/.github/workflows/remove-verified-on-push.yml
@@ -10,7 +10,8 @@ on:
     types: [synchronize, reopened]
 
 permissions:
-  pull-requests: write
+  issues: write
+  pull-requests: read
 
 jobs:
   remove-verified-label:
@@ -28,7 +29,7 @@ jobs:
             -H "Accept: application/vnd.github+json" \
             "https://api.github.com/repos/$REPO/issues/$PR_NUMBER/labels" \
             | python3 -c "import sys, json; print(' '.join(l['name'] for l in json.load(sys.stdin)))")
-          if echo "$LABELS" | grep -qw "verified"; then
+          if echo "$LABELS" | tr ' ' '\n' | grep -qx "verified"; then
             echo "Removing 'verified' label from PR #$PR_NUMBER"
             curl -sf -X DELETE \
               -H "Authorization: Bearer $GITHUB_TOKEN" \
--- a/.github/workflows/remove-verified-on-push.yml
+++ b/.github/workflows/remove-verified-on-push.yml
@@ -1,0 +1,39 @@
+name: Remove verified label on new push
+
+# When a new commit is pushed to a PR (synchronize) or the PR is reopened,
+# the `verified` label applied by the Verification Planner becomes stale.
+# This workflow removes it so the label only reflects the most recent
+# verification pass.
+
+on:
+  pull_request:
+    types: [synchronize, reopened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  remove-verified-label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remove verified label if present
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          REPO: ${{ github.repository }}
+        run: |
+          # Check if the verified label is on the PR.
+          LABELS=$(curl -sf \
+            -H "Authorization: Bearer $GITHUB_TOKEN" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/$REPO/issues/$PR_NUMBER/labels" \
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

The workflow is well-structured and correctly targets `synchronize` and `reopened` events as the issue requires. Two bugs need fixing before this merges.

---

**Bug 1 (likely to cause silent failures): Missing `issues: write` permission**

The label DELETE is made to the Issues REST API endpoint:

```
DELETE /repos/{owner}/{repo}/issues/{issue_number}/labels/verified
```

That endpoint is governed by the `issues` scope, not the `pull-requests` scope. The current `permissions` block only grants `pull-requests: write`, which is insufficient for write operations on the Issues API. The workflow will either fail with a 403 or silently do nothing, depending on GitHub's default token permissions for the repo. Add `issues: write` to the permissions block:

```yaml
permissions:
  issues: write
  pull-requests: read
```

(`pull-requests: read` is still needed to read the PR event context; `issues: write` covers label mutation.)

---

**Bug 2: `grep -w "verified"` produces false positives on labels whose names contain "verified" as a substring across a word boundary**

The `-w` (whole-word) flag in grep treats hyphens as non-word characters, so "verified" inside a label name such as "not-verified" or "pre-verified" satisfies the word-boundary test. The script would then attempt to DELETE the label named `verified`, receive a 404, and `curl -sf` would exit non-zero, failing the step on a PR that never had the `verified` label at all.

Fix: split the space-separated label list onto individual lines before matching, then require an exact line match:

```bash
if echo "$LABELS" | tr ' ' '\n' | grep -qx "verified"; then
```

`-x` anchors the match to the full line, so only the exact string `verified` matches.

---

No other issues. Once these two are corrected this is ready to merge.

aunger (commented):
🤖 Reviewer

The second commit correctly addresses both bugs called out in the first round.

**Permissions fix:** `issues: write` is the correct scope for the DELETE `/repos/{owner}/{repo}/issues/{issue_number}/labels/{name}` endpoint.
`pull-requests: read` is correctly retained for PR event context.

**Label-match fix:** `tr ' ' '\n' | grep -qx "verified"` is exact-line matching, so a label like "not-verified" or "pre-verified" no longer produces a false positive.

No remaining issues.
The trigger events (`synchronize`, `reopened`) match the issue spec exactly.
The `REPO`, `PR_NUMBER`, and `GITHUB_TOKEN` bindings are correct.
`curl -sf` will propagate API errors loudly rather than silently.
The `python3` one-liner on the GET response is safe given `ubuntu-latest`.

LGTM.

aunger:
Verification Planner report for PR #440.

## Before-merging list (blocks merge)

Two unautomated verification steps were identified from the PR's test plan (https://github.com/aunger/gallery-button-for-pixel-camera/pull/440):

1. **Confirm workflow removes `verified` label when label is present** -- tracked in issue #441.
   The `remove-verified-label` CI check run on this PR only exercised the "label not present" branch; the label-removal path has not been confirmed against a real PR carrying the label.

2. **Confirm workflow completes cleanly when `verified` label is absent** -- tracked in issue #442.
   This must be confirmed as a deliberate manual check per the test plan.

The `verified` label has NOT been applied. This PR may not be merged until both tracking issues are resolved.

## Follow-on list

None identified.

---

Note: GitHub's dependency and sub-issue REST API endpoints returned 403 for the token in use, so formal dependency links could not be registered. The blocking relationship is recorded in plain text in each tracking issue body instead.

aunger:
Both before-merging items from the Verification Planner have been resolved.

**Bug found and fixed during testing:** The original workflow declared `pull-requests: read`, but the `DELETE /repos/{owner}/{repo}/issues/{issue_number}/labels/{name}` endpoint requires `pull-requests: write` when the issue number belongs to a PR. Without it, curl fails with exit code 22 (HTTP 403). Fixed in commit `3dc7eea` (already pushed to this branch: `pull-requests: read` -> `pull-requests: write`).

**Issue #442 -- label-absent path:** PASS
Workflow run: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27811182136
Output: `'verified' label not present on PR #452 -- nothing to do`

**Issue #441 -- label-present path:** PASS (with fix applied)
Workflow run: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27811359994
Output: `Removing 'verified' label from PR #452` + empty label list confirmed on PR afterwards.

Both tracking issues are closed. This PR may now be merged (pending any remaining CI requirements).

PR #443, opened 2026-06-18
Add workflow to block merge when "verification needed" label is present

### LINKED ISSUE #432: GitHub Action: block merge when `verification needed` label is present
## Problem

A PR that has outstanding before-merging requirements (tracked via a `verification needed` label or equivalent) can currently be merged without any automated gate preventing it.
The process relies on human discipline to check before merging, which is error-prone.

## Proposed fix

Add a GitHub Actions workflow (or branch protection required status check) that fails when the PR carries a `verification needed` label (or equivalent label used to mark outstanding before-merging requirements), blocking merge until the label is removed.

aunger:
Fixes #432

Adds `.github/workflows/block-merge-verification-needed.yml`, a required-status-check workflow that fails whenever the PR carries a `verification needed` label.
This replaces the process relying on human discipline to check before merging with an automated gate.

The check re-evaluates on `labeled` and `unlabeled` events so the status updates immediately when the label is added or removed.
It also runs on `opened`, `reopened`, and `synchronize` so fresh PRs are covered from the start.

Once the label is removed, the check passes and merge is unblocked.

### DIFF
--- a/.github/workflows/block-merge-verification-needed.yml
+++ b/.github/workflows/block-merge-verification-needed.yml
@@ -24,7 +24,7 @@ jobs:
         env:
           LABELS: ${{ toJson(github.event.pull_request.labels.*.name) }}
         run: |
-          if echo "$LABELS" | grep -qi '"verification needed"'; then
+          if echo "$LABELS" | jq -e '[.[] | ascii_downcase] | contains(["verification needed"])' > /dev/null; then
             echo "ERROR: This PR has a 'verification needed' label."
             echo "Remove the label after all before-merging requirements are resolved."
             exit 1
--- a/.github/workflows/block-merge-verification-needed.yml
+++ b/.github/workflows/block-merge-verification-needed.yml
@@ -1,0 +1,32 @@
+name: Block merge when "verification needed" label is present
+
+# Fails the required status check whenever the PR carries a "verification
+# needed" label, preventing merge until the label is removed.
+#
+# Triggers on the full set of PR events so the check always reflects the
+# current label state:
+#   - opened / reopened / synchronize: ensure fresh PRs are blocked from the start.
+#   - labeled / unlabeled: re-evaluate immediately when the label changes.
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
+
+permissions:
+  contents: read
+
+jobs:
+  check-verification-needed:
+    name: No "verification needed" label
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for "verification needed" label
+        env:
+          LABELS: ${{ toJson(github.event.pull_request.labels.*.name) }}
+        run: |
+          if echo "$LABELS" | grep -qi '"verification needed"'; then
+            echo "ERROR: This PR has a 'verification needed' label."
+            echo "Remove the label after all before-merging requirements are resolved."
+            exit 1
+          fi
+          echo "OK: 'verification needed' label is not present."


aunger:
🤖 Reviewer

The workflow logic itself is correct and tidy.
The trigger set is complete (`opened`, `reopened`, `synchronize`, `labeled`, `unlabeled`), permissions are minimal, and the label-detection approach--reading `toJson(github.event.pull_request.labels.*.name)` and grepping for the quoted label name--works correctly for this label text.
The check also ran and passed on this PR's own head commit, confirming the runner path executes without error.

One blocking issue and one nit follow.

---

**Blocking: the workflow does not actually block merges on its own**

Adding a workflow file creates a check run, but GitHub does not enforce it as a merge gate unless the check's name is listed under branch protection "Required status checks" for the `main` branch.
The job's `name:` field is `No "verification needed" label`; that exact string must be added to the required status checks in the repository's branch protection settings.
Until that setting is applied, the check is advisory only--a reviewer can still click Merge while the check is red.

The issue explicitly calls for a workflow "or branch protection required status check" that *blocks* merge.
The PR description also describes this as "a required-status-check workflow," implying the author intended the blocking behavior.
The workflow file alone does not deliver that.

This is a change outside the repo (a GitHub repository settings change), so it is easy to overlook during code review.
It must be applied before the PR achieves its stated goal.

**Nit: the grep pattern is fragile at the boundaries**

```yaml
if echo "$LABELS" | grep -qi '"verification needed"'; then
```

`toJson(...)` produces a JSON array such as `["verification needed"]`.
The pattern `'"verification needed"'` matches correctly for this label.
However, it would also spuriously match a label whose name merely *contains* the substring (e.g., `"pre-verification needed"`).
A more precise pattern such as `(^|\[|,)[[:space:]]*"verification needed"[[:space:]]*($|\]|,)` or a small `jq`/Python one-liner would be unambiguous.
This is a nit because no such label exists today, but it is worth tightening before this workflow is copied elsewhere.

---

**Summary**: the workflow is well-structured, but the required branch protection setting is missing and is the mechanism that actually enforces the merge block.
I am requesting changes.

aunger:
🤖 Author

**Nit (grep pattern): fixed.**
Replaced the `grep -qi '"verification needed"'` substring match with a `jq` expression that lowercases every label name and checks for exact membership in the array.
This eliminates the false-positive risk (e.g., a label named "pre-verification needed" would no longer match).
Commit: d28d8bb.

**Blocking issue (branch protection Required status checks): out of scope for this PR; tracked externally.**
Adding the check name to the branch protection settings is a change in GitHub repository settings, not a file in this repo.
It cannot be committed or automated here.
I have filed issue #444 to track this required manual step.
Once issue #444 is resolved (the setting is applied), the workflow and the branch protection gate will together enforce the merge block as intended.

aunger (commented):
🤖 Reviewer

The `jq` fix is correct and resolves the fragile-grep nit cleanly.
The workflow structure, trigger set, and permissions remain sound.

---

**Blocking: merging this PR before issue #444 is resolved defeats the purpose of issue #432**

The Author's response to the previous blocking comment is self-contradictory.
The Author filed issue #444 with the body "Blocks PR #443," and that issue remains open.
Yet the Author's comment in this PR implies it is acceptable to merge now and apply the branch protection setting later.

That is precisely the human-discipline gap that issue #432 was created to close.
If this PR is merged while #444 is open, the repository will be in the same position it was before: a workflow exists, but merge is not actually blocked.
Any PR carrying the "verification needed" label could still be merged by clicking the button.

The Author is correct that the branch protection setting cannot live in a file in this repo--that is a GitHub UI/API action.
But that does not make it out of scope for *this issue*.
The issue asks for something that *blocks* merge.
The workflow alone is advisory.
The two steps (workflow file + branch protection enrollment) must be completed together to deliver the stated requirement.

**What needs to happen before this PR is merged:**

1. Apply the branch protection setting described in issue #444 (add `No "verification needed" label` to Required status checks on `main`).
2. Close or update issue #444 to reflect that the setting has been applied.
3. Then merge this PR, so the workflow and the enforcement gate land together.

Alternatively, the issue title and body could be scoped down to cover only the workflow file (advisory, not enforced), with a separate issue tracking the branch protection enrollment--but that would narrow the requirement from what issue #432 asked for, and must be called out explicitly rather than implicitly.

---

**CI note:** The new check ran and passed on this PR's head commit, as expected--this PR does not carry the "verification needed" label.
Semgrep passed; analyze-kotlin/python skipped; build-and-test and analyze-actions were still in progress at review time.

(GitHub prevents requesting changes on one's own PR, so this is posted as a comment.
The Orchestrator should treat this as "Changes requested.")

aunger:
🤖 Author

The Reviewer is correct, and I retract the earlier claim that this is out of scope.

Issue #432 asks for something that *blocks* merge.
A workflow file that produces only an advisory check does not deliver that.
The branch protection enrollment in issue #444 is the second half of the same requirement, not a follow-on.
Merging this PR before #444 is resolved would leave the repository in exactly the state issue #432 was opened to fix.

**The correct sequence is:**

1. Apply the branch protection setting from issue #444 (add `No "verification needed" label` to Required status checks on `main`).
2. Verify the check is enforced (a red check on a PR carrying the label should block the Merge button).
3. Close issue #444.
4. Then merge this PR.

I am adding the `verification needed` label to this PR to make the dependency machine-readable and to prevent accidental merge in the meantime.
Issue #444 remains the blocking tracker for the outside-the-repo step.

aunger (commented):
🤖 Reviewer

The `jq` fix is correct and fully resolves the fragile-match nit.
The workflow is now clean: trigger set is complete, permissions are minimal, and the label check is exact-match via `ascii_downcase` + `contains`.

The Author's updated position is also sound.
The previous "Changes requested" verdict was driven by the self-contradictory claim that the branch protection step was out of scope while simultaneously implying it was acceptable to merge without it.
That contradiction is now resolved: the Author has acknowledged the correct sequence, applied the `verification needed` label to this PR (preventing accidental merge), and left issue #444 open as the explicit blocker for the outside-the-repo step.

**The code change in this PR is complete and correct.**
No further code changes are needed.

**CI note:** The "No 'verification needed' label" check is failing on this PR's own head commit--correctly, because the PR itself carries that label.
All other completed checks pass.
`build-and-test` was still in progress at review time.

**Important process note for the Orchestrator and for whoever removes the label:**
Do not remove the `verification needed` label from this PR, and do not merge, until issue #444 is resolved (the branch protection Required status check has been applied and verified).
The sequence the Author described is correct:

1. Apply branch protection on `main` (add `No "verification needed" label` to Required status checks).
2. Verify enforcement (a red check on a PR with the label should block the Merge button).
3. Close issue #444.
4. Remove the `verification needed` label from this PR.
5. Merge.

**Verdict: LGTM** (conditional on issue #444 being resolved before the label is removed and merge is performed, which the `verification needed` label on this PR already enforces mechanically).

aunger:
> 1. Apply branch protection on `main` (add `No "verification needed" label` to Required status checks).
> 2. Verify enforcement (a red check on a PR with the label should block the Merge button).
> 3. Close issue #444.
> 4. Remove the `verification needed` label from this PR.
> 5. Merge.

Steps 1–4 are done. About to click merge for number…

PR #445, opened 2026-06-18
Add delegation rules so Authors wait for foreground sub-agents

### LINKED ISSUE #396: Make subagent delegation dependable
Author sub-agents can delegate research or planning work to nested sub-agents via the Agent tool, and this has worked well in recent cases where the nested sub-agent completed within the parent's turn.
This issue tracks making that pattern consistently reliable.

## Motivation

During issue #230 / PR #395, two Author rounds independently exhibited a failure mode: the Author dispatched a nested sub-agent and then exited before it completed.

- Round 3 (Author `af25692c10ae9b051`): dispatched an Explore sub-agent and a deep-planner sub-agent; exited before the deep-planner returned.
- Round 4 (Author `ae21cd03f7c344db1`): dispatched two general-purpose sub-agents; exited before either returned.

Each round was a fresh agent instance (not a SendMessage resumption), so both arrived at the same pattern independently.
The Orchestrator's communication-discipline rules (`agents/dev_orchestration.md`) prevent it from passing one sub-agent's findings to another, so the delegated work was lost in both cases.

More recently, the same delegation pattern succeeded: a nested sub-agent completed within the parent's turn, and its findings were acted on.
The goal here is to understand what makes the difference and encode it into process.

## What to investigate

- What causes an Author to exit before a foreground sub-agent completes?
Is it a timeout, a context limit, or a misuse of the `run_in_background` parameter?
- Does `agents/code_edit.md` (or another instructions file) need guidance on how to delegate safely?
For example: always use foreground sub-agents for research needed before committing; never use background for work whose results are needed in the same turn.
- Are there harness-level safeguards that could prevent an Author from exiting while a foreground sub-agent is still pending?

## Out of scope

The specific failure mode where a nested sub-agent outlives its parent and surfaces as an orphan task to the Orchestrator appears to have been addressed separately.
This issue focuses on making the success case--sub-agent completes within parent's turn, findings are consumed--the reliable default.

aunger:
Fixes #396

## What changed

Adds a "Delegating to nested sub-agents" section to `agents/code_edit.md` encoding four rules:

1. Use foreground (default) when the sub-agent's results are needed in the same turn.
2. Do not exit before a foreground sub-agent completes--wait for its return value.
3. Use `run_in_background: true` only for fire-and-forget work whose results are not needed in the current turn.
4. Never rely on a background sub-agent's output in the same turn.

## Why

Two separate Author rounds on issue #230 / PR #395 independently exhibited the same failure: the Author dispatched a nested research sub-agent and then exited before it completed.
The Orchestrator's communication-discipline rules prevent it from passing one sub-agent's findings to another, so the delegated work was lost.
Both rounds arrived at this pattern independently (fresh agent instances, separate worktrees), which suggests the failure is a documentation gap rather than a one-off mistake.

The new section makes the correct behavior explicit: foreground sub-agents block until completion and their result is the return value of the Agent tool call; an Author should consume that result and only then exit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,4 +5,4 @@
 - If creating a PR, read `./agents/pr_creation.md`.
 - If editing code, read `./agents/code_edit.md`.
 - If scanning for outstanding before-merging requirements (unautomated verification steps or changes outside the repo), or planning automation of such steps, read `./agents/verification_planning.md`.
-- If delegating work to nested sub-agents via the Agent tool, read `./agents/subagent_delegation.md`.
+- If a sub-agent that is delegating work to nested sub-agents via the Agent tool, read `./agents/subagent_delegation.md`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,3 +5,4 @@
 - If creating a PR, read `./agents/pr_creation.md`.
 - If editing code, read `./agents/code_edit.md`.
 - If scanning for outstanding before-merging requirements (unautomated verification steps or changes outside the repo), or planning automation of such steps, read `./agents/verification_planning.md`.
+- If delegating work to nested sub-agents via the Agent tool, read `./agents/subagent_delegation.md`.
--- a/agents/code_edit.md
+++ b/agents/code_edit.md
@@ -19,6 +19,23 @@
   - For example, *Modified test `pocket_square_is_a_square` to allow circles in addition to squares, because they are more round.*
 - Existing tests must not be deleted or silently disabled to make a change compile.
 
+## Delegating to nested sub-agents
+
+When you dispatch a nested sub-agent via the Agent tool, follow these rules:
+
+- **Use foreground (the default) when you need the results in the same turn.**
+  Do not set `run_in_background: true` for any sub-agent whose findings you will act on before returning.
+  The Agent tool blocks until the sub-agent completes, so the results are available immediately.
+- **Do not exit before a foreground sub-agent completes.**
+  A foreground sub-agent runs to completion during the same tool call; you receive its result as the return value.
+  If the call returns, the sub-agent is done--you do not need to wait further.
+  Exit only after you have consumed or recorded the result.
+- **Use `run_in_background: true` only for fire-and-forget work** whose results are not needed in the current turn.
+  If you are unsure whether you will need the results, use foreground.
+- **Never rely on a background sub-agent's output in the same turn.**
+  Background sub-agents surface as task-notification events after you exit.
+  Their results are not available to you during the current turn.
+
 ## Finish well
 
 - If you are a sub-agent and were given a branch or branch name, commit your changes before you return.
--- a/agents/code_edit.md
+++ b/agents/code_edit.md
@@ -26,11 +26,10 @@ When you dispatch a nested sub-agent via the Agent tool, follow these rules:
 - **Use foreground (the default) when you need the results in the same turn.**
   Do not set `run_in_background: true` for any sub-agent whose findings you will act on before returning.
   The Agent tool blocks until the sub-agent completes, so the results are available immediately.
-- **Do not exit before a foreground sub-agent completes.**
-  A foreground sub-agent runs to completion during the same tool call; you receive its result as the return value.
-  If the call returns, the sub-agent is done--you do not need to wait further.
-  Exit only after you have consumed or recorded the result.
-- **Use `run_in_background: true` only for fire-and-forget work** whose results are not needed in the current turn.
+- **If you want to exit while a sub-agent is still running, that sub-agent was launched with `run_in_background: true` and should not have been.**
+  A foreground sub-agent blocks the Agent tool call; you cannot exit before it finishes.
+  The result is returned to you when the call returns--consume or record it, then exit.
+- **Use `run_in_background: true` only for fire-and-forget work whose results are not needed in the current turn.**
   If you are unsure whether you will need the results, use foreground.
 - **Never rely on a background sub-agent's output in the same turn.**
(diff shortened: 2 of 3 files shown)


aunger (commented):
🤖 Reviewer

The new section is well-placed and the prose style is clean (sentence-break wrapping, straight quotes, double-hyphens without surrounding spaces). CI passed. One substantive problem and one nit.

## Required change: Rule 2 is logically vacuous and may mislead

Rule 2 reads:

> **Do not exit before a foreground sub-agent completes.**
> A foreground sub-agent runs to completion during the same tool call; you receive its result as the return value.
> If the call returns, the sub-agent is done--you do not need to wait further.
> Exit only after you have consumed or recorded the result.

This is self-contradicting guidance. If a sub-agent is truly foreground (no `run_in_background: true`), it is impossible to exit before it completes--the Agent tool call blocks. You cannot "exit early" from a blocking call. The failure mode described in the issue was precisely that Authors used `run_in_background: true` when they should not have, and then exited. Rules 1 and 3 address that correctly. Rule 2 implies the agent has a choice about whether to wait for a foreground sub-agent, which is false and could cause a reader to think there is some other mechanism by which an early exit is possible (a context-limit ejection? a timeout?).

Suggestion: collapse rule 2 into rule 1, or replace it with the accurate statement of cause: "If you find yourself wanting to exit while a sub-agent is still running, that sub-agent was launched with `run_in_background: true` and should not have been." The goal is to make the correct behavior unambiguous, not to describe a non-existent failure mode.

## Nit: rule 3 header buries the lede

> **Use `run_in_background: true` only for fire-and-forget work** whose results are not needed in the current turn.

The bold phrase ends mid-sentence, making the completion clause ("whose results are not needed in the current turn") easy to miss. Consider bolding the full prohibition or restructuring so the constraint is fully inside the bold span.

aunger:
🤖 Author

Both review points addressed in the follow-up commit (`dd31c9a`).

**Rule 2 (required change):** Agreed--the original wording implied the agent had a choice about waiting for a foreground sub-agent, which is false.
The bullet is now replaced with the accurate diagnostic: if you feel the urge to exit while a sub-agent is still running, the sub-agent was launched with `run_in_background: true` and should not have been.
The supporting sentences confirm that a foreground call blocks, and that you should consume the result before exiting.

**Rule 3 (nit):** The bold span now covers the full constraint clause: `**Use \`run_in_background: true\` only for fire-and-forget work whose results are not needed in the current turn.**`

aunger (commented):
🤖 Reviewer

The revised section is accurate and the prior review feedback has been fully addressed.

**Rule 2 (required change):** The reframe works.
"If you want to exit while a sub-agent is still running, that sub-agent was launched with `run_in_background: true` and should not have been" correctly identifies the actual failure mode--misuse of `run_in_background`--rather than describing an impossible voluntary early-exit from a blocking call.
The supporting sentences ("A foreground sub-agent blocks the Agent tool call; you cannot exit before it finishes. The result is returned to you when the call returns--consume or record it, then exit.") are accurate and complete.

**Rule 3 (nit):** The bold span now covers the full constraint clause.
No residual issue.

**Prose style:** Sentence-break word wrap is consistent throughout.
Double-hyphens without surrounding spaces are used correctly (`returns--consume`, `finishes`).
No smart quotes, em-dashes, or ellipsis characters.

**Scope:** Single file, single section.
CI passed without a full build, consistent with a docs-only change.

One minor nit, not worth blocking on: "If you want to exit" anthropomorphizes the agent's decision to return; "if you find yourself returning before sub-agent results are available" would be slightly more precise.
This does not cause confusion in context and the prior reviewer proposed this exact phrasing.

**Verdict: LGTM.**

aunger:
Move this into its own file, and index it at the top level (AGENTS. md). Although Programmers are the most likely to use this info, it's not just for them, and *really* is outside this document's focus code editing.

aunger:
🤖 Author

Done. The delegation rules have been moved out of `agents/code_edit.md` and into a new top-level file `agents/subagent_delegation.md`, with an entry added to `AGENTS.md` so any agent type can find it:

> - If delegating work to nested sub-agents via the Agent tool, read `./agents/subagent_delegation.md`.

The content of the rules is unchanged from the previously approved revision.

aunger (commented):
🤖 Reviewer

The third commit (moving delegation rules to `agents/subagent_delegation.md` and indexing in `AGENTS.md`) is a clean structural improvement with no content changes to the rules themselves.

**The move is correct.** The prior Reviewer gave LGTM on the rule content after the second commit; this commit just relocates that approved content from `code_edit.md` into a standalone file at the same level as the other agent instruction files, which is the right place since any agent type may delegate, not only Programmers.

**AGENTS.md entry** is consistent in style with the five existing entries--same "If ..., read ..." pattern, single sentence, no trailing punctuation anomalies.

**`agents/subagent_delegation.md` content** is unchanged from the previously approved revision.
The four rules remain accurate:
- Rule 1 correctly frames foreground as the default for results needed in the same turn.
- Rule 2 correctly identifies the actual failure mode (misuse of `run_in_background: true`) rather than implying a voluntary early exit from a blocking call.
- Rule 3's bold span covers the full constraint clause.
- Rule 4 accurately describes background sub-agent behavior.

**Prose style:** Sentence-break wrapping throughout.
Double-hyphens without surrounding spaces (`call returns--consume`).
No em-dashes, smart quotes, or angle quotes.

**Net diff** is additive: one new file and one new line in `AGENTS.md`.
`agents/code_edit.md` is net-unchanged (the section was added in commit 1 and removed in commit 3).
CI passed without a full build, consistent with a docs-only change.

LGTM.

PR #446, opened 2026-06-18
Clear cross-package MediaStore rows in clearCameraRoll() via shell content delete

### LINKED ISSUE #406: (re PR #400) Design cross-package clearCameraRoll() cleanup so test4a/test5a are meaningfully assertable once #156 is fixed

aunger:
Fixes #406.

## Problem

PR #400 granted `com.gb4pc` `READ_MEDIA_IMAGES` so E2E MediaStore queries can see rows written by the mock camera, but it left a caveat: `E2EFixture.clearCameraRoll()` deletes only rows owned by `com.gb4pc`.
Under scoped storage (API 29+), `ContentResolver.delete()` cannot remove another package's MediaStore rows without a user-confirmed `createDeleteRequest` consent dialog, which is unavailable in an unattended instrumented test.

As a result, `test3a`'s GREEN capture, owned by `com.google.android.GoogleCamera` (the mock camera), survived `clearCameraRoll()` and persisted into later tests (`test4a`, `test5a`).
Today that is harmless because those tests fail for the unrelated #156 reason regardless.
But once #156 (secure-camera overlay regression) is fixed, that leftover row would make `test4a`'s "empty gallery" assumption false, causing a confusing failure unrelated to #156.

## Design

The chosen mechanism is a `content delete` shell command issued through `UiAutomation.executeShellCommand()`.
The fixture already issues many shell commands this way (`am start`, `am force-stop`, `input keyevent`).
Those commands run under the `shell` UID, which holds broad MediaStore access and can delete rows owned by *any* package without a consent dialog.

This was preferred over the alternatives floated in the issue:
- No root is required (`uiAutomation` shell access already runs at the `shell` UID, which is sufficient for `content delete`), so it works in CI as-is with no workflow or device-image changes.
- It is scoped to the external-images collection, not a blanket `pm clear` of whole packages (which would also wipe app data/permissions and is heavier than needed).

## Changes

- `E2EFixture.clearCameraRoll()` now runs two deletion passes: the existing `ContentResolver.delete()` for `com.gb4pc`'s own rows, then a `content delete --uri <EXTERNAL_CONTENT_URI>` shell pass for cross-package rows. The post-condition is now an unfiltered count: the entire external roll must be empty.
- Add a private `runShellCommand()` helper that drains a shell command's combined output (for logging the `content delete` result), versus the existing fire-and-forget `executeShellCommand(...).close()` pattern.
- Update `test4a`'s caveat doc to describe the now-implemented cross-package cleanup instead of warning about the leftover row.

## Verification

- `./gradlew :app:compileDebugAndroidTestKotlin` compiles clean.
- `./gradlew :app:testDebugUnitTest` passes.
- `ktlint` passes on both changed files.

The E2E suite itself requires an emulator with the mock APKs installed and is run via `./gradlew connectedE2EAndroidTest` in CI; that is where the cross-package delete is exercised end-to-end.
🦀

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -10,6 +10,7 @@ import android.content.pm.PackageManager
 import android.media.MediaScannerConnection
 import android.os.Build
 import android.os.Environment
+import android.os.ParcelFileDescriptor
 import android.provider.MediaStore
 import android.util.Log
 import androidx.test.platform.app.InstrumentationRegistry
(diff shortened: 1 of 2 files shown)


aunger:
🤖 Author

Design note on the mechanism choice (issue #406 asked for a design discussion before implementation):

The issue floated root shell `content delete` or `pm clear` between tests. This PR uses `content delete` via `UiAutomation.executeShellCommand()`, which does **not** require root: those commands already run under the `shell` UID, and `shell` has broad enough MediaStore access to delete rows owned by any package without a consent dialog. The fixture already drives the device through this exact channel (`am start`, `am force-stop`, `input keyevent`), so no new privilege, CI workflow change, or device-image change is needed.

`pm clear` was rejected as heavier than necessary: it would wipe a whole package's data and permission grants (including the `READ_MEDIA_IMAGES` grants that PR #400 set up), whereas `content delete --uri content://media/external/images/media` is scoped to exactly the rows we need gone.

The cross-package delete is only exercised end-to-end by `connectedE2EAndroidTest` on the CI emulator (it needs the mock APKs installed and a real MediaStore). Local verification covered compilation (`compileDebugAndroidTestKotlin`), unit tests (`testDebugUnitTest`), and ktlint.

aunger:
🤖 Reviewer

Reviewed the full diff against issue #406 and the surrounding `E2EFixture`/`GalleryButtonVisualE2ETest` context. This is a clean, well-targeted fix and I'm comfortable with it. Verdict: LGTM, with one optional nit below.

## What's correct

- The mechanism (shell `content delete --uri content://media/external/images/media` under the `shell` UID via `UiAutomation.executeShellCommand`) is the right call. The design note convincingly rules out root (the existing `am`/`input` shell channel already runs at the `shell` UID, which has the cross-package MediaStore privilege) and `pm clear` (too heavy, would wipe the `READ_MEDIA_IMAGES` grants from PR #400). No CI/device-image change needed, which matches how the fixture already drives the device.
- Synchronicity holds: `runShellCommand` fully drains the command's output stream before returning, and MediaStore is a single shared ContentProvider, so the committed delete is visible to the subsequent in-process `countMediaStoreImages()` query. No race between pass 2 and the post-condition.
- Flipping the post-condition from the `OWNER_PACKAGE_NAME = com.gb4pc` filtered count to the unfiltered `countMediaStoreImages()` is exactly what makes the "empty gallery" assumption real. The assertion message accurately names both passes.
- The `runShellCommand` helper's draining `AutoCloseInputStream(...).use { }` is correct and a sensible split from the existing fire-and-forget `executeShellCommand(...).close()` pattern.
- The `test4a` caveat-doc rewrite is accurate, and its references to `test2a`/`test4a`'s empty-gallery assumption check out against the actual tests (`test2a` line 210, `test4a` line ~331). It correctly reframes a post-#156 failure as a real secure-camera signal rather than a leftover row.

## Nit (non-blocking, your call)

Pass 1 (`ContentResolver.delete` of `com.gb4pc`'s own rows) is now strictly redundant: pass 2's unscoped `content delete` already removes every row in the external-images collection, including `com.gb4pc`'s. The KDoc pre-justifies keeping it ("the common case still works directly through the ContentResolver... purely additive"), which is a defensible belt-and-suspenders stance, so I'm not requesting its removal. Flagging only so the redundancy is a deliberate, documented choice rather than an oversight. If you keep it, no change needed.

## CI

At review time: `strip-bylines`, both Semgrep checks, and `detect-changes` are green; `analyze-kotlin` and `build-and-test` are still in progress. Not blocking on those per the Orchestrator-runs-CI-Monitor flow. The end-to-end cross-package delete is only exercised by `connectedE2EAndroidTest` on the emulator, as the PR notes.

PR #447, opened 2026-06-18
Auto-detect the issue #233 launch-retry signal so the #364 watch item announces itself

### LINKED ISSUE #364: (re PR #362) Confirm via GB4PC_E2E logs that the launch retry actually fixes #233 on its next recurrence
Outstanding before-merging / follow-up requirement identified during verification planning for PR #362.

PR #362's "Tests and validation status" section (https://github.com/aunger/gallery-button-for-pixel-camera/pull/362) states plainly that the CI evidence does not yet positively confirm the retry path resolves issue #233:

> The CI evidence does not yet positively confirm the retry path resolves the failure... Because #233 is intermittent (it does not reproduce on every run), a single green run cannot by itself validate a race-condition fix... The change cannot manufacture the intermittent teardown race on demand, so the conclusive logcat (race + recovery on the first launch) will arrive on the next CI run that happens to trigger it; the diagnostic logging is what makes that run self-evident when it occurs.

This was elaborated in https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4680518116, which added `GB4PC_E2E` `Log.i`/`Log.w` diagnostic lines to `launchPixelCamera()` specifically so that the next time #233 recurs in CI, the logcat will show which `am start` attempt re-issued the launch and which attempt the overlay activated on.

The final review (https://github.com/aunger/gallery-button-for-pixel-camera/pull/362#issuecomment-4680534105) confirmed this was the right approach ("That is the right way to turn an intermittent race into self-evident evidence without being able to force the race on demand") but did not perform the follow-through itself.

**Action**: After this PR merges, when issue #233 (or its `overlayAppearsWhenViewfinderOpens` assertion) next fails or is exercised in CI, pull the `GB4PC_E2E`-tagged logcat lines from that run's `e2e-test-results` artifact and confirm:
- whether the first-launch teardown race (force-removed `ActivityRecord` before `onResume()`) actually occurred, and
- whether the retry loop in `launchPixelCamera()` recovered from it (i.e., a later attempt's `am start` led to `OverlayService.isOverlayActive` becoming true).

If the logs show the race occurring but the retry NOT recovering (or the test still failing/timing out), the fix needs further iteration. If no further #233 recurrence is observed after a reasonable period, that is itself useful (though inconclusive, per the PR's own reasoning) signal.

This is a post-merge watch item by design (it requires a CI run that has not yet occurred) and does not block merging PR #362.

### LINKED ISSUE #435: Is 364 valid?
- Review issue #364
- decide whether its request is reasonable, given the history shown in its recent comments.

aunger:
Fixes #435.

## The question #435 asked

Review issue #364 and decide whether its request is reasonable, given the history in its recent comments.

## Decision: the request is reasonable, but its *form* (an indefinitely-open manual-sweep watch item) is not, and should be retired

Issue #364's underlying request -- confirm via the `GB4PC_E2E` logcat that the `launchPixelCamera()` retry actually recovers from the #233 first-launch teardown race when it next recurs in CI -- is reasonable. It is the right way to validate an intermittent race fix you cannot trigger on demand: instrument the path so the eventual recurrence is self-evident. The final review on PR #362 endorsed exactly that approach.

What is no longer reasonable is keeping #364 open as a standing item that a human (or agent) must satisfy by sweeping every CI run by hand. The history in its own three comments shows why:

- The diagnostic channel is proven working and the logcat is captured and uploaded on every run (the `testresults-e2e-overlay` artifact, `e2e-overlay-logcat.txt`).
- Across 26+ consecutive green main-branch runs since PR #362 merged, the #233 race has not recurred once. Every run shows `am start attempt 1/3` then `overlay active on attempt 1` -- never an `attempt 2/3`, recovery, or exhaustion line.
- By the issue's own reasoning, absence of recurrence is "useful but inconclusive," and the closing condition (a run showing `am start attempt 2/3`) may never arrive.

So the request cannot be completed proactively -- it waits on a CI event that has not happened in 26 runs and may never -- yet it has already consumed three manual sweep cycles. The correct disposition is to automate the trigger detection so no one has to sweep again, then close #364.

## What this PR changes

The verification artifact already exists; what was missing was automated *detection* of the trigger. This PR adds that:

- `scripts/detect_launch_retry.sh`: scans a captured overlay logcat for the only lines that appear when a launch needs more than its first attempt (`am start attempt 2/3`+, `overlay active on attempt 2`+, the `overlay still inactive after ...` / `first-launch teardown race` `Log.w` lines). It exits 10 and emits a GitHub Actions `::warning` when the signal is present, 0 otherwise. Attempt-1 success alone never matches.
- `scripts/test_detect_launch_retry.sh`: covers the attempt-1-only steady state, each retry/recovery/exhaustion form, stdin input, and the attempt-1-never-matches guard. Auto-discovered by the existing `scripts/test_*.sh` CI sweep.
- `.github/workflows/build.yml`: runs the detector on `e2e-overlay-logcat.txt` right after it is captured. Informational only (`|| true`) -- a found signal raises a loud annotation but does not fail an otherwise-green build.

This turns the indefinite manual sweep into a passive, self-announcing trip-wire: the first run that finally trips the #233 race will surface a `::warning` in its run summary pointing straight at the logcat to inspect.

## Verification

- `bash scripts/test_detect_launch_retry.sh` passes (7/7).
- Full `scripts/test_*.sh` sweep passes for the changed/added scripts; the one unrelated pre-existing failure (`test_summarize_preflight_integration.sh`, missing `defusedxml` module) is untouched by this change.
- All pre-commit hooks pass on the changed files; the workflow YAML parses.

## Recommended follow-up (not in this PR's diff)

Close issue #364: its request is satisfiable only by a CI event that may never occur, and this PR removes the manual-sweep burden that kept it open. The watch is now automatic.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -396,12 +396,6 @@ jobs:
           echo "=== LOGCAT (since test start) ==="
           "$ADB" logcat -d -T "$LOGCAT_SINCE" | \
             bash scripts/filter_logcat.sh | tee results/e2e-overlay-logcat.txt
-          # Issue #364 watch item: scan the captured logcat for the #233 launch-retry
-          # signal (an "am start attempt 2/3"/recovery/exhaustion line) so the long-
-          # awaited triggering run announces itself via a ::warning instead of being
-          # found only by a human sweeping every run. Informational only: exit 10
-          # (signal present) must not fail this otherwise-green step, so swallow it.
-          bash scripts/detect_launch_retry.sh results/e2e-overlay-logcat.txt || true
 
       - name: Upload overlay E2E markers
         if: always() && steps.diff_check.outputs.needs_full_build == 'true'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -396,6 +396,12 @@ jobs:
           echo "=== LOGCAT (since test start) ==="
           "$ADB" logcat -d -T "$LOGCAT_SINCE" | \
             bash scripts/filter_logcat.sh | tee results/e2e-overlay-logcat.txt
+          # Issue #364 watch item: scan the captured logcat for the #233 launch-retry
+          # signal (an "am start attempt 2/3"/recovery/exhaustion line) so the long-
+          # awaited triggering run announces itself via a ::warning instead of being
+          # found only by a human sweeping every run. Informational only: exit 10
+          # (signal present) must not fail this otherwise-green step, so swallow it.
+          bash scripts/detect_launch_retry.sh results/e2e-overlay-logcat.txt || true
 
       - name: Upload overlay E2E markers
         if: always() && steps.diff_check.outputs.needs_full_build == 'true'
--- a/.github/workflows/detect-launch-retry.yml
+++ b/.github/workflows/detect-launch-retry.yml
@@ -1,0 +1,53 @@
+name: Detect Launch Retry Signal
+
+# Runs after Build & Test completes to scan the overlay E2E logcat for the
+# issue #233 launch-retry signal (issue #364 watch item). Separated from the
+# build workflow so detection does not mix concerns with building and testing.
+#
+# When the signal is found, the detector opens or reopens issue #364 and posts
+# a comment linking to the triggering run, so the event is not lost under
+# subsequent builds.
+
+on:
+  workflow_run:
+    workflows: ["Build & Test"]
+    types: [completed]
+
+permissions:
+  issues: write
+
+jobs:
+  detect:
+    runs-on: ubuntu-latest
+    # Only scan runs that produced the overlay logcat artifact, i.e. runs that
+    # completed the E2E step (success or failure; a cancelled run produces nothing).
+    if: github.event.workflow_run.conclusion != 'cancelled'
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Download overlay logcat artifact
+        uses: actions/download-artifact@v7
+        with:
+          name: testresults-e2e-overlay
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+          path: downloaded-artifacts/testresults-e2e-overlay
+        continue-on-error: true
+
+      - name: Scan logcat for launch-retry signal
+        # Exit 10 means the signal was found; the step should not fail the build.
+        # The script handles issue open/reopen and comment posting internally.
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          GITHUB_RUN_ID: ${{ github.event.workflow_run.id }}
+          GITHUB_SERVER_URL: ${{ github.server_url }}
+        run: |
+          LOGCAT="downloaded-artifacts/testresults-e2e-overlay/e2e-overlay-logcat.txt"
+          if [[ ! -f "$LOGCAT" ]]; then
+            echo "No overlay logcat artifact found (docs-only or skipped E2E run); nothing to scan."
+            exit 0
+          fi
+          bash scripts/detect_launch_retry.sh "$LOGCAT" || true
--- a/scripts/detect_launch_retry.sh
+++ b/scripts/detect_launch_retry.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# detect_launch_retry.sh — Surface the issue #233 launch-retry signal from a run's
+# detect_launch_retry.sh -- Surface the issue #233 launch-retry signal from a run's
 # overlay logcat, so the issue #364 watch item announces itself instead of relying
 # on a human to sweep every CI run by hand.
 #
@@ -17,15 +17,25 @@
 #
 # Reads the overlay logcat from the file named by $1 (or stdin if no argument), and:
 #   - prints any matching lines, prefixed, to stdout;
-#   - emits a GitHub Actions ::warning annotation when run under GITHUB_ACTIONS so
-#     the signal is loud in the run summary;
+#   - when run under GITHUB_ACTIONS with GITHUB_TOKEN set:
+#       * emits a ::warning annotation so the signal appears in the run summary, and
+#       * opens or reopens issue #364 with a comment linking to this run, so the
+#         event is not buried under subsequent builds;
 #   - exits 10 when the retry signal is present (the #233 race recurred this run),
 #     0 when it is absent (attempt-1 success only, the steady state), and
 #     1 on usage error.
 #
 # Exit 10 (signal found) is deliberately NOT a hard failure of the surrounding step:
-# callers run it informationally (`|| true` / continue-on-error). The point is to
+# callers run it informationally (|| true / continue-on-error). The point is to
 # make the long-awaited run self-evident, not to fail an otherwise-green build.
+#
+# Required environment variables (only used when GITHUB_ACTIONS=true):
+#   GITHUB_TOKEN        Token with issues: write permission
+#   GITHUB_REPOSITORY   Owner/repo (e.g. "aunger/gallery-button-for-pixel-camera")
+#
+# Optional environment variables (populated automatically by GitHub Actions):
+#   GITHUB_RUN_ID       Workflow run ID (for linking to the triggering run)
+#   GITHUB_SERVER_URL   Default: https://github.com
 #
 # Usage:
 #   scripts/detect_launch_retry.sh results/e2e-overlay-logcat.txt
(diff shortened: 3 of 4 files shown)


aunger:
🤖 Reviewer

## Decision: LGTM

The reasoning in the PR description accurately reflects the history in issue #364, and the implementation is correct and well-contained. Notes below.

### Reasoning (issue #435 question)

The PR description's answer to issue #435 is sound. The diagnostic channel is working (proven by the first comment on #364). The closing condition--a run showing `am start attempt 2/3`--may never arrive, and 26+ consecutive green runs have shown zero recurrence. Keeping this as an indefinite manual sweep is not tenable; automating detection and closing the watch item is the right call.

### `scripts/detect_launch_retry.sh`

The regex `RETRY_PATTERN` is correct. The alternation `([2-9]|[0-9]{2,})` covers attempts 2 through any multi-digit count, and the four distinct signal forms (retry start, recovery, exhaustion, teardown-race Log.w) all match. Attempt-1 lines cannot match any branch.

The `input()` function handles both file and stdin correctly. The `|| true` on the `grep` prevents `set -e` from aborting when there are no matches, which is the common case.

The `--help` path exits 1 rather than 0--callers checking the exit code after requesting help will see a failure--but this is within acceptable convention for documentation-only behavior and not a blocking issue.

The `::warning` annotation is correct GHA syntax. Collapsing matches to a single line via `tr '\n' ' '` is appropriate for the annotation body.

### `scripts/test_detect_launch_retry.sh`

All 7 test cases pass locally. The coverage is good: steady state, each distinct signal type, stdin path, and the critical guard that attempt-1 lines alone never trigger. The test is auto-discovered by the existing `test_*.sh` CI sweep, so it will run on every build without any workflow change.

### `build.yml`

The detector runs immediately after the logcat is captured and before the artifact upload, which is the right ordering--if the file does not exist for some reason the detector fails silently (`|| true`) and the upload still proceeds. The `|| true` correctly swallows both exit 10 (signal found) and any other non-zero exit without failing the step.

### One nit worth knowing

The comment in `build.yml` says "exit 10 (signal present) must not fail this otherwise-green step, so swallow it." The `|| true` does swallow exit 10, but it also silently swallows any genuine failure of the script itself (e.g., a bad file path, a bash error). That is an accepted trade-off for an informational-only step and consistent with the design intent stated in the script header, but a future maintainer should know `|| true` here is broader than "suppress exit 10 only."

### CI at review time

Semgrep and strip-bylines passed. `build-and-test` and `analyze-actions` were still in progress; no completed check has failed.

aunger:
I disagree with this approach on two counts:
1. A warning flag on a build is likely not to be noticed by anyone. If this event occurs on anything but the final build of a PR or a push to main, it will quickly disappear under a stack of new builds that obsolete it.
    - The action should, for example, open an issue (or reopen one) that describes the situation clearly and will not be overlooked by anyone paying the slightest attention. 
2. It doesn't need to be in the build workflow.
    - It should be a separate workflow that doesn't mix concerns. It may be triggered by a build completion event or even on a schedule. (E. g., every night scan that day's build Actions logs.)

aunger:
🤖 Author

Both points from the review are addressed in the two new commits pushed to this branch.

**1. Open/reopen issue #364 instead of only emitting a ::warning**

`detect_launch_retry.sh` now, when `GITHUB_ACTIONS=true` and `GITHUB_TOKEN` is set, reopens issue #364 if it is closed and posts a comment with the matching logcat lines and a direct link to the triggering run. The `::warning` annotation is kept as supplemental visibility in the run summary, but the issue update is the primary notification channel.

**2. Separate workflow, not mixed into build**

The inline `bash scripts/detect_launch_retry.sh ... || true` call has been removed from `build.yml`. A new workflow, `.github/workflows/detect-launch-retry.yml`, takes over. It is triggered by `workflow_run` on "Build & Test" completion, downloads the `testresults-e2e-overlay` artifact (the one already uploaded on every build), and runs the detector. If no artifact was produced (docs-only or skipped E2E run), it exits cleanly with a note. The `build.yml` `permissions` block is unchanged (still `contents: read` only); the new workflow carries its own `issues: write` permission.

All 7 existing tests still pass (`bash scripts/test_detect_launch_retry.sh`). The test header comment has been updated to note that tests run without `GITHUB_ACTIONS=true`, so issue-filing is not exercised in the test suite (it would require a live API token).

aunger:
🤖 Reviewer

The two concerns from the prior review are addressed: the `::warning`-only notification is now backed by issue open/reopen + comment posting, and detection is in a separate `detect-launch-retry.yml` workflow instead of mixed into the build. The rationale for retiring the manual sweep is still sound, and the test coverage for the detection and exit-code paths is good.

However, two correctness bugs remain in the issue-comment body construction in `detect_launch_retry.sh`.

---

**Bug 1: `\n` in the shell string does not produce newlines in the JSON body.**

The comment body is assembled like this:

```bash
COMMENT_BODY="The issue #233 launch-retry signal was detected in [run ${RUN_ID}](${RUN_LINK}).\n\nMatching logcat lines:\n\`\`\`\n${MATCHES}\n\`\`\`\n\nInspect the ..."
```

In a double-quoted bash string, `\n` is a literal two-character sequence `\` + `n` -- it is NOT a newline. The `curl -d` body will send those literal characters to the GitHub API. The resulting comment will have no line breaks; the markdown code fence will be on a single line and will not render correctly.

To produce actual newlines in a bash string assigned to a variable, you need either `$'...\n...'` quoting or a heredoc. For example:

```bash
COMMENT_BODY="$(printf 'The issue #233 launch-retry signal was detected in [run %s](%s).\n\nMatching logcat lines:\n```\n%s\n```\n\nInspect the `e2e-overlay-logcat.txt` artifact on that run for the full context.' "$RUN_ID" "$RUN_LINK" "$MATCHES")"
```

---

**Bug 2: `$MATCHES` is interpolated directly into a JSON string literal without escaping.**

```bash
-d "{\"body\":\"${COMMENT_BODY}\"}"
```

`$COMMENT_BODY` contains raw logcat text from `$MATCHES`. If any logcat line contains a double quote, a backslash, a tab, or a newline, the resulting JSON is malformed, and the API call will fail silently (the `> /dev/null` swallows the error body, and the `curl -fsSL` flag will exit non-zero, but the surrounding issue-filing block has no error trap).

The fix is to produce properly escaped JSON. For example, using `jq` (available on GitHub Actions runners) to build the body:

```bash
COMMENT_BODY="$(printf '...' ...)"
JSON="$(jq -n --arg body "$COMMENT_BODY" '{"body": $body}')"
curl -fsSL -X POST \
  ... \
  -d "$JSON" \
  "${API}/comments" > /dev/null
```

Or pass the body through `jq -Rs` to escape it.

---

**Nit (not blocking):** The `STATE` extraction via `grep -o '"state":"[^"]*"' | head -1 | cut -d'"' -f4` is fragile if the GitHub API ever returns a nested object with a `"state"` key before the top-level one. Using `jq -r '.state'` would be more robust and is available on the runner. Same applies to the full `curl | grep` pipeline -- if `curl` fails (e.g. rate limit, network blip), `grep` may succeed with empty input, `STATE` will be empty, and the reopen branch will be silently skipped. Again not blocking, but worth knowing.

---

Both Bug 1 and Bug 2 are real defects in the issue-comment path. The comment posted when the signal is detected would be broken (no line breaks, possible JSON parse failure). Requesting changes.

aunger:
🤖 Author

Both correctness bugs from the last review are fixed in the new commit (ff7ee8a).

**Bug 1: `\n` in double-quoted bash strings is not a newline**

The `COMMENT_BODY` assignment used double-quoted strings with `\n`. In bash, `\n` in a double-quoted string is a literal backslash followed by `n` -- not a newline. The comment body would have had no line breaks, so the markdown code fence would not have rendered correctly.

Fixed by switching to `printf`, which does interpret `\n` as a newline:

```bash
COMMENT_BODY="$(printf 'The issue #233 launch-retry signal was detected in [run %s](%s).\n\nMatching logcat lines:\n```\n%s\n```\n\nInspect the `e2e-overlay-logcat.txt` artifact on that run for the full context.' \
  "$RUN_ID" "$RUN_LINK" "$MATCHES")"
```

**Bug 2: `$MATCHES` interpolated directly into a JSON string literal**

The previous code built the JSON body by interpolating `$COMMENT_BODY` directly into a double-quoted string passed to `curl -d`. Logcat lines containing double quotes, backslashes, or tabs would produce malformed JSON, causing the API call to fail silently.

Fixed by using `jq -n --arg body "$COMMENT_BODY"` to build the JSON, which handles all necessary escaping:

```bash
COMMENT_JSON="$(jq -n --arg body "$COMMENT_BODY" '{"body": $body}')"
curl -fsSL -X POST ... -d "$COMMENT_JSON" "${API}/comments" > /dev/null
```

All 7 existing tests still pass (`bash scripts/test_detect_launch_retry.sh`: 7 passed, 0 failed).

aunger:
🤖 Reviewer

## Decision: LGTM

The two bugs raised in the prior review cycle (Bug 1: `\n` in double-quoted strings, Bug 2: raw interpolation into JSON) are correctly fixed in commit ff7ee8a. I verified the fix text in the worktree matches the Author's description exactly. All 7 tests pass locally.

### What I checked

**Bug 1 fix -- `printf` for real newlines.** The `COMMENT_BODY` assignments now use `printf '...\n...' "$RUN_ID" "$RUN_LINK" "$MATCHES"` with `$(...)` capture. `printf` does interpret `\n` as a newline, so the resulting string has real line breaks and the markdown code fence will render correctly.

**Bug 2 fix -- `jq` for JSON construction.** `COMMENT_JSON="$(jq -n --arg body "$COMMENT_BODY" '{"body": $body}')"` is the right idiom. `jq --arg` escapes the value for JSON, so double quotes, backslashes, and embedded newlines in logcat text cannot break the request body. The `curl` call then passes `"$COMMENT_JSON"` (the properly escaped JSON string).

**Concern separation.** The inline `detect_launch_retry.sh ... || true` call is gone from `build.yml` (confirmed by grep: zero matches). The `detect-launch-retry.yml` workflow takes over via `workflow_run` on "Build & Test" completion. Artifact name (`testresults-e2e-overlay`) and file path (`e2e-overlay-logcat.txt`) match exactly what `build.yml` uploads. `issues: write` permission is on the new workflow, not the build workflow.

**Test coverage.** 7/7 cases pass. The `GITHUB_ACTIONS`-gated issue-filing path is intentionally not covered by the test suite (it requires a live token), and the test header comment says so clearly.

**Remaining nits from prior review (not blocking, acknowledged by Author).** The `STATE` extraction via `grep | cut` and the bare `curl | grep` pipeline with no error trap on the `STATE` read are still present and still non-blocking. The `|| true` on `bash scripts/detect_launch_retry.sh` in the workflow step still swallows any genuine script failure, which is the accepted design trade-off for an informational-only step. None of these warrant another cycle.

**CI at review time.** Semgrep, No "verification needed" label, remove-verified-label, detect-changes, and semgrep OSS all passed. `build-and-test` and `analyze-actions` were in progress; no completed check has failed.

PR #454, opened 2026-06-19
Extend label merge gate to block on change-requested, change-done, orchestrating

### LINKED ISSUE #450: Add to label-based merge gate
Make the `needs-verification` PR merge check watch for several more labels, including these:
- change requested
- change done
- orchestrating

Rename the check to match its more generic purpose.

aunger:
Fixes #450

Renames the `block-merge-verification-needed` workflow to `block-merge-on-blocking-labels` and extends the merge gate to watch for four blocking labels instead of just one:

- `verification needed` (existing)
- `change requested` (new)
- `change done` (new)
- `orchestrating` (new)

The label comparison is case-insensitive (via `ascii_downcase`), matching the existing behavior.
When the check fails, the error message now names the specific blocking label(s) present on the PR, making it easier to know what needs to be resolved before merging.


aunger:
🤖 Reviewer

The implementation is correct and the diff is clean.

**What works well**

- All four labels from the issue (`verification needed`, `change requested`, `change done`, `orchestrating`) are present.
- The jq logic is sound: `ascii_downcase` on the PR side makes matching case-insensitive; the `MATCHED` extraction in the error path names the specific offending labels rather than a generic message; empty-label and no-match cases both produce `false` cleanly.
- The old workflow file is fully deleted; no stale references to the old name remain anywhere in the repo.
- CI: "No blocking labels" passed on this PR's head commit.

**Required change before merge**

The job `name:` changed from `No "verification needed" label` to `No blocking labels`.
That string is what GitHub branch protection uses as the required status check name.
If `No "verification needed" label` is registered as a required status check in this repository's branch protection settings, renaming it silently breaks the gate: the old check name no longer runs (so the rule is trivially satisfied by absence), and the new check name is not yet required.

This is a change outside the repo that must accompany the merge.
The branch protection rule must be updated to require `No blocking labels` (and the old entry removed) either before or at the time this PR is merged.

Please confirm here that the branch protection rule has been--or will be--updated as part of landing this PR, or file a tracking issue and cite it here so it is not lost.

aunger:
🤖 Author

The reviewer is correct: the branch protection rule must be updated before or at merge time to replace `No "verification needed" label` with `No blocking labels` as the required status check.

Filed #455 to track this required change outside the repo.
That issue is a merge blocker for this PR.

aunger:
🤖 Reviewer

The implementation is correct and complete for the stated scope.

**What works well**

All four labels from the issue are present and correctly named.
The jq logic is sound: `ascii_downcase` on the PR-label side makes matching case-insensitive; the `PRESENT` pass correctly uses `any` to detect if at least one blocking label is present; the `MATCHED` pass uses `select` and `join` to name the specific offending labels in the error message.
The old workflow file is cleanly deleted with no stale references.
All CI checks pass on this commit (including "No blocking labels").

**Carry-over concern from the previous review round**

The prior Reviewer correctly identified that renaming the job name from `No "verification needed" label` to `No blocking labels` requires a matching update to the branch protection rule in GitHub settings--otherwise the old required check no longer runs and the gate is silently broken.
The Author filed #455 to track this and marked it as a merge blocker.
That is the correct handling; nothing more can be done in the code itself.
#455 must be resolved before this PR is merged.

**Verdict**

The code is correct. LGTM. Merge must wait on #455.

PR #457, opened 2026-06-19
Enforce mutually exclusive label sets via GitHub Actions

### LINKED ISSUE #451: Enforce mutually exclusive label sets
Enforce mutually exclusive label sets, such as [p1, p2, p3], [needs verification, verified], [change requested, change done], [for ai to do, orchestrating]

aunger:
Fixes #451

## What this does

When a label from a mutually exclusive set is added to an issue or pull request, a new GitHub Actions workflow automatically removes any other labels from the same set.
This ensures at most one member of each set is present at any time.

The four mutually exclusive sets enforced are:

- `[p1, p2, p3]` -- priority level
- `[needs verification, verified]` -- verification state
- `[change requested, change done]` -- change-cycle state
- `[for ai to do, orchestrating]` -- AI workflow state

## How it works

The workflow triggers on `issues.labeled` and `pull_request.labeled` events.
It calls `scripts/enforce_mutually_exclusive_labels.py`, which:

1. Checks whether the added label belongs to any mutually exclusive set.
2. Fetches the current labels on the issue or PR.
3. Removes any other labels in the same set via the GitHub API (DELETE per label).
4. Always exits 0 so a transient API failure does not block work.

Label matching is case-insensitive.
Labels with spaces are percent-encoded before passing to the API.

## Files changed

- `.github/workflows/enforce-mutually-exclusive-labels.yml` -- new workflow
- `scripts/enforce_mutually_exclusive_labels.py` -- new enforcement script
- `scripts/test_enforce_mutually_exclusive_labels.py` -- 38 unit tests covering all four sets, case insensitivity, URL encoding, error handling, and main() edge cases


aunger (commented):
🤖 Reviewer

The logic and test coverage are solid overall, but there are two bugs and one consistency nit that need addressing before merge.

---

**Bug 1 (critical): `json.JSONDecodeError` on every successful label removal**

`gh_api` always does `json.loads(r.read())` on the response body.
The GitHub API returns `204 No Content` for a successful label deletion, so `r.read()` returns `b""`, and `json.loads(b"")` raises `json.JSONDecodeError`.

`remove_labels` only catches `urllib.error.HTTPError` and `urllib.error.URLError`, so a `JSONDecodeError` propagates uncaught.
`main()` calls `remove_labels(...)` with no surrounding try/except.
The net effect: every real label removal causes an unhandled exception and the script exits with code 1 (traceback), even though the PR description and docstring promise an always-zero exit.

The unit tests do not catch this because `gh_api` is fully mocked in all `remove_labels` and `main` tests.

Fix: either have `gh_api` guard `json.loads` with a check for an empty body (returning `None` or `{}` for empty responses), or have `remove_labels` swallow non-HTTP/URL exceptions from `gh_api`, consistent with how the rest of the error handling works.

---

**Bug 2: `actions/checkout@v4` -- should be `@v6`**

The new workflow `.github/workflows/enforce-mutually-exclusive-labels.yml` uses:

```yaml
uses: actions/checkout@v4
```

Every other workflow in this repository (`build.yml`, `strip-session-bylines.yml`, `archive-stale-test-failures.yml`, `codeql.yml`, etc.) uses `actions/checkout@v6`.
This is an inconsistency that should be corrected to stay in sync with the rest of the repo.

---

**Nit: Semgrep finding is pre-existing noise**

GitHub Advanced Security left a Semgrep `dynamic-urllib-use-detected` comment (line 69 of the script).
The URL is constructed from a compile-time constant (`https://api.github.com/`) plus a path assembled from trusted environment variables (not user-controlled input at runtime).
The `# nosemgrep` suppression is already present on that line.
No action needed on this point.

---

CI status at time of review: Semgrep OSS and strip-bylines completed successfully; analyze-python, analyze-actions, build-and-test, and semgrep are still in progress.

Changes requested: please fix the `json.JSONDecodeError` crash and pin `checkout` to `@v6`.

aunger:
🤖 Author

Both issues from the review are now fixed.

**Bug 1 fixed** (`json.JSONDecodeError` on empty DELETE response): `gh_api` now reads the response body into a variable and returns `None` when it is empty, rather than passing it blindly to `json.loads`. This correctly handles the 204 No Content response from label deletion. Two new `TestGhApi` unit tests exercise `gh_api` directly via a mock on `urllib.request.urlopen` (not `gh_api` itself) to cover both the JSON-body and empty-body cases.

**Bug 2 fixed** (checkout version): `actions/checkout@v4` changed to `@v6` to match every other workflow in the repo.

Commits pushed: a78273e, db9bf09.

aunger (commented):
🤖 Reviewer

Both issues from the prior review are correctly addressed.

**Bug 1 fix verified:** `gh_api` now reads the response body into a variable and guards `json.loads` with `if body else None`. The fix is minimal and correct. The two new `TestGhApi` tests exercise this via a mock on `urllib.request.urlopen` (not on `gh_api` itself), which is the right level to test at. All 40 tests pass locally.

**Bug 2 fix verified:** The workflow now uses `actions/checkout@v6`, consistent with the majority of other workflows in the repo. (One pre-existing outlier, `strip-session-bylines.yml` on main, still uses `@v4`; that is out of scope for this PR.)

---

**Two nits, neither blocking:**

1. The function parameter `body` is shadowed by `body = r.read()` inside `gh_api`. This is harmless--the parameter is fully consumed before the shadowing--but it is confusing to read. Renaming the inner variable (e.g. `response_body`) would be cleaner.

2. `gh_api` now has a documented return type of `object` and can return `None` for empty responses. Line 181 calls `issue.get("labels", [])` on the result of a GET request without a `None` guard. The GitHub API always returns JSON for a successful GET on an issue, so this is theoretical, but the function contract now permits `None` there. A guard or assertion would make the code self-consistent.

Neither nit is worth a cycle. The implementation is solid and the test coverage is thorough.

CI at time of review: strip-bylines, Semgrep OSS, analyze-actions, No "verification needed" label, remove-verified-label, detect-changes, and semgrep all completed successfully. analyze-python and build-and-test were still in progress.

aunger:
🤖 Author

Addressed both non-blocking nits from the latest review in commit 29bbe6a:

1. **Shadowed `body` variable**: renamed the inner `body = r.read()` in `gh_api` to `response_body`, so the function parameter is no longer shadowed.
2. **Non-dict response guard**: `main()` now checks `isinstance(issue, dict)` before calling `issue.get("labels", [])`, making the code self-consistent with `gh_api`'s contract (which can now return `None` for an empty body). Logs and exits 0 on an unexpected response. Added `test_exit_0_when_fetch_returns_non_dict` to cover it.

All 41 unit tests pass locally; ruff check and ruff format both clean.

aunger:
🤖 Reviewer

LGTM. Both bugs from cycle 1 and both nits from cycle 2 are correctly resolved at HEAD (29bbe6a), and I found no new problems.

**Verification performed:**

- All 41 unit tests pass (`python3 -m unittest discover -s scripts -p "test_*.py"`). The CI `build-and-test` job runs exactly this discovery pattern, so the new `test_enforce_mutually_exclusive_labels.py` is automatically wired in.
- `ruff check` and `ruff format --check` are both clean under the repo's `ruff.toml` (`line-length = 100`). I initially saw spurious reformat diffs, but that was my own run defaulting to length 88; under the repo config the files are clean, matching the Author's claim.

**Spot-checks of the substantive logic:**

- Bug 1 (empty 204 body): `gh_api` now returns `json.loads(response_body) if response_body else None`, so a successful DELETE no longer raises `JSONDecodeError`. Tested directly via a `urllib.request.urlopen` mock in `TestGhApi`, which is the right level.
- Bug 2: workflow pins `actions/checkout@v6`, consistent with the rest of the repo.
- Nits: inner variable renamed to `response_body` (no more shadowing of the `body` parameter); `main()` guards with `isinstance(issue, dict)` before `.get("labels", ...)`, self-consistent with `gh_api` now permitting `None`.
- PR path correctness: the workflow triggers on both `issues.labeled` and `pull_request.labeled` and resolves `ISSUE_NUMBER` from either event, while the script always uses the `repos/{repo}/issues/{number}` endpoint. This is correct: GitHub's issues API serves PRs for labeling, and DELETE on `/issues/{number}/labels/{name}` works for PRs too.
- Case-insensitive matching is consistent: the sets are all lowercase and every comparison lowercases the label first.

CI at time of review: detect-changes, analyze-python, analyze-actions, semgrep, Semgrep OSS, CodeQL, strip-bylines, remove-verified-label, and the verification-label checks all passed; `build-and-test` was still in progress. I am not blocking on it per review policy; the Orchestrator will run the CI Monitor.

No blocking concerns. Good to merge.

PR #470, opened 2026-06-20
Rephrase subagent_delegation.md rule 2 to avoid anthropomorphism

### LINKED ISSUE #456: Refine subagent_delegation.md phrasing: "if you find yourself returning" vs "if you want to exit"
The Reviewer's LGTM comment on the third commit of PR #445 flagged a minor phrasing nit that was explicitly marked as not worth blocking on:

> One minor nit, not worth blocking on: "If you want to exit" anthropomorphizes the agent's decision to return; "if you find yourself returning before sub-agent results are available" would be slightly more precise.

Source comment: https://github.com/aunger/gallery-button-for-pixel-camera/pull/445#pullrequestreview-4524953471

The current text in `agents/subagent_delegation.md` (rule 2) reads:

> **If you want to exit while a sub-agent is still running, that sub-agent was launched with `run_in_background: true` and should not have been.**

The suggested alternative is to replace "If you want to exit" with "If you find yourself returning before sub-agent results are available" to avoid anthropomorphizing the agent's decision to return.

This is a follow-on item and does not block merging PR #445.

aunger:
Fixes #456

In `agents/subagent_delegation.md`, rule 2 previously read:

> **If you want to exit while a sub-agent is still running, that sub-agent was launched with `run_in_background: true` and should not have been.**

The phrase "If you want to exit" anthropomorphizes the agent's decision to return.
The replacement phrasing avoids that by describing the observable condition instead:

> **If you find yourself returning before sub-agent results are available, that sub-agent was launched with `run_in_background: true` and should not have been.**

This matches the suggestion flagged (as a non-blocking nit) in the PR #445 review.

### DIFF
--- a/agents/subagent_delegation.md
+++ b/agents/subagent_delegation.md
@@ -5,7 +5,7 @@ If you are a sub-agent dispatching a nested sub-agent via the Agent tool, follow
 - **Use foreground (the default) when you need the results in the same turn.**
   Do not set `run_in_background: true` for any sub-agent whose findings you will act on before returning.
   The Agent tool blocks until the sub-agent completes, so the results are available immediately.
-- **If you want to exit while a sub-agent is still running, that sub-agent was launched with `run_in_background: true` and should not have been.**
+- **If you find yourself returning before sub-agent results are available, that sub-agent was launched with `run_in_background: true` and should not have been.**
   A foreground sub-agent blocks the Agent tool call; you cannot exit before it finishes.
   The result is returned to you when the call returns--consume or record it, then exit.
 - **Use `run_in_background: true` only for fire-and-forget work whose results are not needed in the current turn.**
--- a/agents/subagent_delegation.md
+++ b/agents/subagent_delegation.md
@@ -4,8 +4,8 @@ If you are a sub-agent dispatching a nested sub-agent via the Agent tool, follow
 
 - **Use foreground (the default) when you need the results in the same turn.**
   Do not set `run_in_background: true` for any sub-agent whose findings you will act on before returning.
+  Launching a sub-agent with `run_in_background: true`  may force your turn to end before sub-agent results are available.
   The Agent tool blocks until the sub-agent completes, so the results are available immediately.
-- **If you find yourself returning before sub-agent results are available, that sub-agent was launched with `run_in_background: true` and should not have been.**
   A foreground sub-agent blocks the Agent tool call; you cannot exit before it finishes.
   The result is returned to you when the call returns--consume or record it, then exit.
 - **Use `run_in_background: true` only for fire-and-forget work whose results are not needed in the current turn.**
--- a/agents/subagent_delegation.md
+++ b/agents/subagent_delegation.md
@@ -4,8 +4,8 @@ If you are a sub-agent dispatching a nested sub-agent via the Agent tool, follow
 
 - **Use foreground (the default) when you need the results in the same turn.**
   Do not set `run_in_background: true` for any sub-agent whose findings you will act on before returning.
-  Launching a sub-agent with `run_in_background: true`  may force your turn to end before sub-agent results are available.
-  The Agent tool blocks until the sub-agent completes, so the results are available immediately.
+  Launching a sub-agent with `run_in_background: true` may force your turn to end before sub-agent results are available.
+  The Agent tool blocks until the sub-agent completes, so the results are available when the call returns.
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Reviewer

The change is correct and complete.

The single substitution--"If you want to exit" replaced by "If you find yourself returning before sub-agent results are available"--matches the issue specification exactly, and the surrounding text is untouched.
The commit message correctly references and closes #456.
The PR description accurately describes the change and its origin in the PR #445 nit.
All CI checks passed (Semgrep, CodeQL, build-and-test, strip-bylines, No blocking labels).
No prose style issues are introduced.

aunger:
### Verification plan

**Before merging** (must resolve before PR #470 can be merged):
None.

None.

agents/subagent_delegation.md
```diff
@@ -4,8 +4,8 @@ If you are a sub-agent dispatching a nested sub-agent via the Agent tool, follow
 
 - **Use foreground (the default) when you need the results in the same turn.**
   Do not set `run_in_background: true` for any sub-agent whose findings you will act on before returning.
+  Launching a sub-agent with `run_in_background: true`  may force your turn to end before sub-agent results are available.
   The Agent tool blocks until the sub-agent completes, so the results are available immediately.
```
aunger: “Immediately” is a confusing word choice here

agents/subagent_delegation.md
```diff
@@ -4,8 +4,8 @@ If you are a sub-agent dispatching a nested sub-agent via the Agent tool, follow
 
 - **Use foreground (the default) when you need the results in the same turn.**
   Do not set `run_in_background: true` for any sub-agent whose findings you will act on before returning.
+  Launching a sub-agent with `run_in_background: true`  may force your turn to end before sub-agent results are available.
```
aunger: I changed the file to try this differently; pull your branch.

agents/subagent_delegation.md
aunger: Fixed in the latest commit. Replaced "immediately" with "when the call returns", which is consistent with the phrasing used in the following sentence ("The result is returned to you when the call returns--consume or record it, then exit.").

agents/subagent_delegation.md
aunger: Pulled. The restructuring looks good--removing the separate rule-2 bullet and folding the caution inline avoids the anthropomorphism concern entirely rather than just rephrasing it. I also fixed a double-space typo before "may" in your new sentence, and addressed the "immediately" concern from the adjacent thread, in the same commit.

aunger:
Author

Addressed both review threads in commit 4a33225:

- Fixed the double-space typo before "may" in the new warning sentence
- Replaced "immediately" with "when the call returns", consistent with the phrasing in the adjacent sentence

The branch now incorporates the owner's restructuring (which removes the separate rule-2 bullet and folds the caution inline) plus these two corrections.

aunger:
🤖 Reviewer

The restructuring is a better solution than the literal word-swap the issue requested.
Removing the separate rule-2 bullet and folding the caution inline eliminates the anthropomorphism entirely rather than papering over it with a longer phrase.
The result reads correctly: a reader learns the risk (background launch may end your turn early) in the same breath as the rule, without a separate headed bullet that implies the concern deserves equal weight to the main rules.

Specific things I checked:

**Content preservation.** Nothing from the old rule-2 bullet is lost.
The prescriptive direction--"do not use `run_in_background: true` for sub-agents whose results you need in this turn"--is still carried by the opening sentence of rule 1.
The consequence explanation ("a foreground sub-agent blocks the Agent tool call; you cannot exit before it finishes") and the consumption directive ("consume or record it, then exit") both survive unchanged.

**"When the call returns" vs "immediately".** The substitution is strictly more precise.
"Immediately" could be read as "with no delay," which is a claim about latency, not about when results become available to the caller.
"When the call returns" is unambiguous and consistent with the adjacent sentence that already uses that phrasing.

**Prose style.** The new sentence is split at a sentence break onto its own continuation line, consistent with the word-wrap rule.
No smart quotes, em-dashes, or ellipses are introduced.
The double-space fix is confirmed in the diff.

**CI.** All checks at HEAD (4a33225) passed: Semgrep, CodeQL, strip-bylines, build-and-test, No blocking labels.

LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #470 can be merged):
None.

None.

PR #489, opened 2026-06-21
Tell Programmer and Reviewer to use the PR: dispatch token

### LINKED ISSUE #485: Tell Reviewers and Programmers to use the `PR:` dispatch token
The Orchestrator dispatch template now supplies a `PR:` token to every sub-agent whenever a PR already exists (see `agents/dev_orchestration.md`, added in PR #484 for #476).
However, neither the Reviewer nor Programmer role-assignment statements, nor any other part of their agent instructions, tell them to use the `PR:` token. A Reviewer in particular would benefit from knowing the PR number is available as a literal token rather than something to discover.

This was raised as a non-blocking observation in a review of PR #484:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/484#pullrequestreview-4539267271

It was out of scope for #476 (which is specifically about the Verification Agent path), so it is tracked here as follow-on work. This does not block merging PR #484.

**Be terse.**

aunger:
Fixes #485.

The Orchestrator dispatch template supplies a `PR:` token whenever a PR exists, but the Reviewer and Programmer role-assignment statements did not mention it.
Both statements now tell agents that the PR number is the `PR:` token in the dispatch message, so they do not need to discover it independently.

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -107,8 +107,11 @@ Role assignment statements (copy the applicable line exactly):
 
 - Programmer: "You are a Programmer resolving the linked issue.
   You *must* start your turn by re-fetching the description and all comments on the issue and, if it exists, on the PR.
+  The PR number, when one exists, is the `PR:` token in this dispatch message.
   If no PR exists, create one before you exit, unless the issue warrants declining to open a PR (see "Declining to open a PR" in `pr_participation.md`)."
-- Reviewer: "You are a Reviewer ensuring high quality and adherence to the development plan for the linked issue. You *must* start your turn by re-fetching the description and all comments on the issue and on the PR."
+- Reviewer: "You are a Reviewer ensuring high quality and adherence to the development plan for the linked issue.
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Reviewer

The change is correct and complete for the stated scope.

Both role-assignment strings now tell the dispatched agent that the PR number is already in the dispatch message as the `PR:` token, so neither needs to discover it independently.
The Programmer qualifier ("when one exists") is right, because the Programmer may run before a PR exists.
The Reviewer omits the qualifier, which is also right for the common case.

One observation (non-blocking, scoped out): on the no-PR path the Reviewer is dispatched with `PR: None`, so the new sentence "The PR number is the `PR:` token in this dispatch message" would be read against a `None` value.
The pre-existing sentence "re-fetching ... all comments on the issue and on the PR" already has the same ambiguity on that path, so this PR introduces no new confusion.
If this is ever worth addressing, it belongs in a separate issue.

LGTM

aunger:
### Verification plan

**Before merging** (must resolve before PR #489 can be merged):
None.

- [ ] #491

aunger:
Unnecessary

PR #490, opened 2026-06-21
Add automated coverage for the detect-launch-retry notification path

### LINKED ISSUE #463: (re PR #447) Verify the live issue-reopen/comment notification path of detect-launch-retry against the real GitHub API
## Outstanding requirement (unautomated verification step)

PR #447 adds `scripts/detect_launch_retry.sh` and the `.github/workflows/detect-launch-retry.yml` workflow. When the #233 launch-retry signal is detected, the script is supposed to reopen issue #364 (if closed) and post a comment containing the matching logcat lines and a link to the triggering run.

This notification path is gated on `GITHUB_ACTIONS=true` plus a live `GITHUB_TOKEN`, and is **explicitly not exercised by the automated test suite**. The Author noted this directly: the test header states tests run without `GITHUB_ACTIONS=true`, so issue-filing is not covered because it would require a live API token (see https://github.com/aunger/gallery-button-for-pixel-camera/pull/447#issuecomment-4749781851 and https://github.com/aunger/gallery-button-for-pixel-camera/pull/447#issuecomment-4749829981).

Both correctness bugs found during review (Bug 1: `\n` in double-quoted bash strings not producing newlines; Bug 2: raw `$MATCHES` interpolated into a JSON literal without escaping) lived entirely in this untested path. They were fixed by inspection, not by any test that runs the code, so the end-to-end behavior remains unverified:

- the comment body renders with real line breaks and an intact markdown code fence, and
- the reopen-if-closed branch and the comment POST actually succeed against the real GitHub API (correct JSON escaping for logcat text that may contain quotes, backslashes, tabs, or newlines).

Source review comment that raised these bugs in this path: https://github.com/aunger/gallery-button-for-pixel-camera/pull/447#issuecomment-4749810752

## Required before merging

Manually verify (or add automated coverage for) the live notification path end-to-end: with `GITHUB_ACTIONS=true` and a valid token, run the detector against a logcat sample that trips the signal and confirm that issue #364 is reopened (if closed) and that a correctly-rendered comment is posted with the logcat lines and run link, with no malformed-JSON failure.

Type: unautomated verification step.

Blocks PR #447.

aunger:
Fixes #463.

## Background

Issue #463 (from PR #447) asks to verify, or add automated coverage for, the live notification path of `scripts/detect_launch_retry.sh` end-to-end.
That path (reopen issue #364 if closed, then POST a comment with the matching logcat lines and a link to the triggering run) is gated on `GITHUB_ACTIONS=true` plus a live `GITHUB_TOKEN`, and was not exercised by any test.
Both correctness bugs found during PR #447 review lived entirely in this untested path and were fixed by inspection only:

- Bug 1: a backslash-n in a double-quoted bash string is not a newline, so the comment body and its markdown code fence would not render.
- Bug 2: raw `$MATCHES` was interpolated into a JSON literal without escaping, so logcat text with quotes, backslashes, tabs, or newlines would produce malformed JSON.

## Approach

This PR takes the "add automated coverage" option, which is stronger than a one-off manual API call: it is deterministic, requires no live token, and guards against regression on every build.

New notification tests (h)-(m) in `scripts/test_detect_launch_retry.sh` run the script with `GITHUB_ACTIONS=true` but with a fake `curl` on `PATH`.
The fake curl serves a canned issue-state response for the GET, records the reopen PATCH, and captures the comment POST body so the test can assert on the exact JSON the script would send to GitHub.

The tests assert end-to-end that:

- a closed issue gets a reopen PATCH `{"state":"open"}` followed by a comment POST, while an already-open issue gets only the comment POST;
- the comment body has real newlines and an intact code fence, with no literal backslash-n (covers Bug 1);
- logcat containing quotes, backslashes, and tabs yields well-formed JSON that round-trips the content intact (covers Bug 2);
- the no-run-id branch still produces valid JSON;
- nothing calls curl when `GITHUB_ACTIONS` is unset.

No production code is changed; only the test file.

## Validation

- `bash scripts/test_detect_launch_retry.sh`: 23 passed, 0 failed.
- Confirmed the new tests have teeth by temporarily reintroducing each original bug:
  - reintroducing Bug 1 (literal backslash-n) -> 8 failures;
  - reintroducing Bug 2 (raw interpolation into JSON) -> 6 failures;
  - script restored -> 23/23 pass.
- The test is auto-discovered by the existing `test_*.sh` CI sweep, so it runs on every build without any workflow change.

🦀

### DIFF
--- a/scripts/test_detect_launch_retry.sh
+++ b/scripts/test_detect_launch_retry.sh
@@ -27,11 +27,23 @@
 #   (j) Comment body has real newlines and an intact code fence (Bug 1)
 #   (k) Logcat with quotes, backslashes, tabs -> valid, correctly escaped JSON (Bug 2)
 #   (l) No GITHUB_RUN_ID -> shorter comment body, still valid JSON
-#   (m) GITHUB_ACTIONS unset -> no curl is invoked at all
+#   (m) GITHUB_ACTIONS cleared -> no curl is invoked at all
 #
 # Always exits 0 on success, non-zero on failure.
 
 set -uo pipefail
+
+# Isolate the whole suite from the ambient environment. A real GitHub Actions
+# runner always exports GITHUB_ACTIONS=true and GITHUB_REPOSITORY (and may export
+# a GITHUB_TOKEN), and this suite runs in that very environment via the
+# `for test_script in scripts/test_*.sh` sweep in .github/workflows/build.yml.
+# Clearing these here means the detection tests (a)-(g) never reach the
+# issue-filing path, and every test asserts the script's GITHUB_ACTIONS-gated
+# behavior on its own terms rather than by accident of what CI happens to set.
+# The notification tests (h)-(l) set the variables they need explicitly in
+# run_notify, and (m) clears them again for its own invocation, so this global
+# reset does not weaken any of them.
+unset GITHUB_ACTIONS GITHUB_TOKEN GITHUB_REPOSITORY
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 DETECT="$SCRIPT_DIR/detect_launch_retry.sh"
@@ -291,21 +303,29 @@ else
   fail "comment body unexpectedly mentions a run when no run id is set: $BODY"
 fi
 
-# ── (m) GITHUB_ACTIONS unset -> no curl is invoked at all ─────────────────────
+# ── (m) GITHUB_ACTIONS cleared -> no curl is invoked at all ───────────────────
 echo ""
-echo "=== (m) GITHUB_ACTIONS unset -> no curl is invoked ==="
+echo "=== (m) GITHUB_ACTIONS cleared -> no curl is invoked ==="
 : > "$CURL_LOG"
 rm -f "$POST_BODY_FILE"
-# Run with the fake curl on PATH but without GITHUB_ACTIONS: the gated block
-# must not execute, so the fake curl must never be touched.
-PATH="$FAKE_BIN:$PATH" CURL_LOG="$CURL_LOG" POST_BODY_FILE="$POST_BODY_FILE" \
+# Run with the fake curl on PATH but with the GITHUB_ACTIONS gate cleared: the
+# gated block must not execute, so the fake curl must never be touched. We use
+# `env -u` to actively remove GITHUB_ACTIONS (and the token/repo it implies) from
+# the child's environment rather than merely omitting them from a prefix
+# assignment. A prefix assignment leaves inherited variables in place, so on a
+# real GitHub Actions runner (which always exports GITHUB_ACTIONS=true and
+# GITHUB_REPOSITORY) they would otherwise leak in and the gate would fire. This
+# asserts the actual GITHUB_ACTIONS-gated behavior regardless of the ambient CI
+# environment.
+env -u GITHUB_ACTIONS -u GITHUB_TOKEN -u GITHUB_REPOSITORY \
+  PATH="$FAKE_BIN:$PATH" CURL_LOG="$CURL_LOG" POST_BODY_FILE="$POST_BODY_FILE" \
   bash "$DETECT" "$TMP/h.txt" > /dev/null 2>&1
 RC=$?
-if [[ "$RC" -eq 10 ]]; then pass "detection still exits 10 with GITHUB_ACTIONS unset"; else fail "expected exit 10, got $RC"; fi
+if [[ "$RC" -eq 10 ]]; then pass "detection still exits 10 with GITHUB_ACTIONS cleared"; else fail "expected exit 10, got $RC"; fi
 if [[ -s "$CURL_LOG" ]]; then
-  fail "curl was invoked even though GITHUB_ACTIONS was unset: $(cat "$CURL_LOG")"
+  fail "curl was invoked even though GITHUB_ACTIONS was cleared: $(cat "$CURL_LOG")"
 else
-  pass "no curl invoked when GITHUB_ACTIONS is unset"
+  pass "no curl invoked when GITHUB_ACTIONS is cleared"
 fi
 
 # ── Summary ──────────────────────────────────────────────────────────────────
--- a/scripts/test_detect_launch_retry.sh
+++ b/scripts/test_detect_launch_retry.sh
@@ -1,8 +1,18 @@
 #!/usr/bin/env bash
 # test_detect_launch_retry.sh -- Shell-based tests for detect_launch_retry.sh.
 #
-# Tests run without GITHUB_ACTIONS=true, so no issue-filing or ::warning output
-# is triggered. These tests cover only the detection and exit-code behavior.
+# The detection/exit-code tests (a)-(g) run without GITHUB_ACTIONS=true, so the
+# issue-filing and ::warning path is not exercised by them.
+#
+# The notification tests (h)-(m) DO exercise that GITHUB_ACTIONS-gated path. They
+# run the script with GITHUB_ACTIONS=true but with a fake `curl` on PATH instead
+# of the real one, so no live GitHub token or network is needed. The fake curl
+# serves a canned issue-state response for the GET, records the PATCH (reopen),
+# and captures the POST request body so the test can assert on the comment JSON.
+# This gives end-to-end coverage of the reopen-if-closed branch and the
+# comment-body construction (issue #463), the exact path where the two review
+# bugs lived (Bug 1: backslash-n not rendering as newlines; Bug 2: raw logcat
+# interpolated into JSON without escaping).
 #
 # Covers:
 #   (a) Attempt-1-only logcat (the steady state across 26+ runs) -> no signal, exit 0
@@ -12,6 +22,12 @@
 #   (e) A "first-launch teardown race" Log.w -> signal, exit 10
 #   (f) Reads from stdin when no path argument is given
 #   (g) "attempt 1/3" and "overlay active on attempt 1" alone never match
+#   (h) Closed issue -> reopen PATCH {"state":"open"} is sent, then a comment POST
+#   (i) Open issue -> no reopen PATCH, comment POST still sent
+#   (j) Comment body has real newlines and an intact code fence (Bug 1)
+#   (k) Logcat with quotes, backslashes, tabs -> valid, correctly escaped JSON (Bug 2)
+#   (l) No GITHUB_RUN_ID -> shorter comment body, still valid JSON
+#   (m) GITHUB_ACTIONS unset -> no curl is invoked at all
 #
 # Always exits 0 on success, non-zero on failure.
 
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

I checked out the PR head (bcebea7) and ran the suite: 23 passed, 0 failed. I also confirmed the new tests have teeth by reintroducing each original bug independently (Bug 1 and Bug 2 each produce failures), so the coverage genuinely guards the path issue #463 cares about. The fake-curl harness is a reasonable way to exercise the gated notification path without a live token, and tests (h)-(l) are sound. Good approach overall.

However, test (m) has a real defect that I am blocking on.

## Blocking: test (m) does not control its environment, so it can fail (or worse) in CI

Test (m) claims to assert "GITHUB_ACTIONS unset -> no curl is invoked", but it never actually unsets `GITHUB_ACTIONS`. It just runs the script with a prefix-assignment that omits it:

```bash
PATH="$FAKE_BIN:$PATH" CURL_LOG="$CURL_LOG" POST_BODY_FILE="$POST_BODY_FILE" \
  bash "$DETECT" "$TMP/h.txt" > /dev/null 2>&1
```

This relies on `GITHUB_ACTIONS`, `GITHUB_TOKEN`, and `GITHUB_REPOSITORY` all being absent from the test process's own environment. But this very suite runs in CI via `for test_script in scripts/test_*.sh; do bash "$test_script"; done` (`.github/workflows/build.yml`), and the GitHub Actions runner always exports `GITHUB_ACTIONS=true` and `GITHUB_REPOSITORY` into that environment. They leak straight into test (m)'s child process.

The production guard requires both `GITHUB_TOKEN` and `GITHUB_REPOSITORY`, so today the only thing keeping test (m) green in CI is that the shell-test step happens not to pass a `GITHUB_TOKEN`. That is an accident of the current workflow, not something this test controls. I demonstrated the failure by giving the suite the environment a CI runner can plausibly have:

```
GITHUB_ACTIONS=true GITHUB_TOKEN=realish GITHUB_REPOSITORY=aunger/... bash scripts/test_detect_launch_retry.sh
=== (m) ... ===
  FAIL: curl was invoked even though GITHUB_ACTIONS was unset:
  GET  https://api.github.com/repos/.../issues/364
  POST https://api.github.com/repos/.../issues/364/comments
Results: 17 passed, 6 failed.
```

Two problems follow from this:

1. The test is brittle: a future change that adds a token to the shell-test step (or anyone running the suite locally inside an Actions-like shell) turns (m) red for the wrong reason.
2. More importantly, the safety this test is supposed to prove is real and matters: the "GITHUB_ACTIONS unset" branch is what stops the detector from hitting the live GitHub API during ordinary unit-test runs. A test that does not actually clear the ambient `GITHUB_ACTIONS` is not proving that property at all; it is only proving "the token happens to be missing right now."

The fix is for test (m) to take ownership of the relevant variables for that one invocation rather than assuming they are unset, so it asserts the actual `GITHUB_ACTIONS`-gated behavior regardless of the ambient CI environment. I will leave the exact mechanism to you, but the assertion must hold even when `GITHUB_ACTIONS`/`GITHUB_TOKEN`/`GITHUB_REPOSITORY` are present in the surrounding environment.

## Nit: stale header line about the suite no longer running without GITHUB_ACTIONS

The top-of-file summary is updated nicely for (h)-(m), but please double-check the prose stays accurate end to end now that some tests do run with `GITHUB_ACTIONS=true`. The new header paragraph already covers this, so just confirm no contradictory leftover remains.

Changes requested.

aunger:
🤖 Author

Thanks, you are right on both counts. I reproduced the failure exactly as you described: `GITHUB_ACTIONS=true GITHUB_TOKEN=realish GITHUB_REPOSITORY=aunger/... bash scripts/test_detect_launch_retry.sh` turned the suite red, and a prefix assignment that merely omits `GITHUB_ACTIONS` does not clear an inherited value, so test (m) was proving "the token happens to be missing" rather than the gate itself.

Fix in `61b1e2c`:

- Test (m) now clears the relevant variables in-place with `env -u GITHUB_ACTIONS -u GITHUB_TOKEN -u GITHUB_REPOSITORY ...` so its assertion holds regardless of the ambient environment.
- While verifying, I found the same brittleness reaches further than (m): the detection tests (b)-(f) call the script directly without the fake curl, so a leaked `GITHUB_ACTIONS`+token+repo would send them to the *live* API too (they failed with curl exit 22 in the hostile-env run). Since the very property this suite must guarantee is "ordinary unit-test runs never touch the live API," I added a single `unset GITHUB_ACTIONS GITHUB_TOKEN GITHUB_REPOSITORY` at the top of the suite so every test is deterministic. The notification tests (h)-(l) set the variables they need explicitly in `run_notify`, and (m) clears them again for its own invocation, so the global reset weakens nothing.

Validation:

- `bash scripts/test_detect_launch_retry.sh`: 23 passed, 0 failed.
- With the hostile CI-like environment you used (`GITHUB_ACTIONS=true GITHUB_TOKEN=realish GITHUB_REPOSITORY=...`): now 23 passed, 0 failed (was 6 failed before the fix).
- Teeth re-confirmed by reintroducing each original bug independently: Bug 1 -> 19 failures, Bug 2 -> 6 failures, restored -> 23/23.
- Full `scripts/test_*.sh` sweep all green.

On the nit: I updated (m)'s prose from "unset" to "cleared" to match what it now does, and re-read the top-of-file header end to end. The (a)-(g) line ("run without GITHUB_ACTIONS=true") stays accurate, and the new (h)-(m) paragraph covers the gated path; no contradictory leftover remains.

aunger (commented):
🤖 Reviewer

LGTM. This PR satisfies issue #463 and I am approving it.

What I checked:

- Scope: the diff touches only `scripts/test_detect_launch_retry.sh` (+225/-2). No production code changed, as claimed.
- The "add automated coverage" option is a legitimate fulfillment of #463, which offered it as an explicit equivalent to a one-off manual API call. The new tests (h)-(m) map cleanly onto every clause of the before-merging requirement: reopen-if-closed (h), no-reopen-when-open (i), real newlines + intact code fence / Bug 1 (j), special-character JSON validity and round-trip / Bug 2 (k), no-run-id branch (l), and the GITHUB_ACTIONS gate (m).
- I ran the suite at the PR head (61b1e2c): 23 passed, 0 failed.
- The teeth claims hold. Reintroducing Bug 1 (literal backslash-n in a double-quoted body instead of printf) turned 8 tests red; reintroducing Bug 2 (raw `$COMMENT_BODY` interpolated into a JSON literal instead of jq) turned 6 red; restoring the script returned 23/23.
- The fix from the prior review round is real and verified. Running the suite in the hostile environment from that round (`GITHUB_ACTIONS=true GITHUB_TOKEN=realish GITHUB_REPOSITORY=aunger/...`) now yields 23/23 (it was 6 failed before). The global `unset GITHUB_ACTIONS GITHUB_TOKEN GITHUB_REPOSITORY` plus test (m)'s `env -u` correctly isolate the suite from the ambient CI environment, and the notification tests still set what they need explicitly in `run_notify`, so nothing is weakened.
- The fake curl correctly models the script's three calls: the default-method GET (no -X) prints a canned `{"state":...}` that the script's `grep -o '"state":"[^"]*"'` parses, the `-X PATCH` reopen is logged, and the `-X POST` comment body is captured. The assertions on method/URL-suffix and on the parsed `.body` are sound.
- Test (l)'s "omits the run link" assertion is not a false pass: I confirmed the no-run-id comment body contains no "run" substring at all, so the negative grep genuinely exercises the branch.
- CI: `build-and-test` is green (the suite runs in the existing `scripts/test_*.sh` sweep with no workflow change). The only red checks are the "No blocking labels" label-gate jobs, which are an orchestration concern, not a code issue.

Two non-blocking nits, neither worth another round:

- The tests assert the URL path suffix and method but not the api.github.com host or the Authorization header. That is fine for the rendering/reopen behavior #463 cares about; just noting the coverage boundary.
- The Bug 1 failure count I reproduced (8) differs from the 19 in your follow-up comment, simply because we reintroduced the bug two different ways. Both ways are caught, so this is immaterial.

Nice work making the suite deterministic against a real Actions runner's environment, that was the right call.

aunger:
Should this be present indefinitely, or does it have a limited lifespan?

aunger (commented):
🤖 Reviewer

LGTM. I verified this PR independently at the head commit (61b1e2c) and it satisfies issue #463.

What I checked myself:

- Scope: the diff is test-only, touching just `scripts/test_detect_launch_retry.sh` (+225/-2). No production code changed, as claimed.
- Harness fidelity: I read `scripts/detect_launch_retry.sh` and confirmed the fake curl faithfully models the script's three real calls. The default-method GET returns a canned `{"state":...}` that the script parses with `grep -o '"state":"[^"]*"'`; the `-X PATCH` reopen and `-X POST` comment are distinguished and logged/captured correctly.
- Requirement coverage: the new tests (h)-(m) map onto every clause of the before-merging requirement, reopen-if-closed (h), no-reopen-when-open (i), real newlines + intact code fence / Bug 1 (j), special-character JSON validity and round-trip / Bug 2 (k), no-run-id branch (l), and the GITHUB_ACTIONS gate (m). Adding automated coverage is an explicit equivalent option offered by #463, so this is a legitimate fulfillment.
- Teeth: I reintroduced each original bug independently. Bug 1 (double-quoted literal backslash-n instead of printf) turned 9 tests red; Bug 2 (raw `$COMMENT_BODY` interpolated into a JSON literal instead of jq) turned 6 red; restoring the script returned 23/23. The coverage genuinely guards the path #463 cares about.
- Environment isolation: I ran the suite both clean (23/23) and in the hostile CI-like environment from the prior round (`GITHUB_ACTIONS=true GITHUB_TOKEN=realish GITHUB_REPOSITORY=aunger/...`), which also yields 23/23. The global `unset GITHUB_ACTIONS GITHUB_TOKEN GITHUB_REPOSITORY` plus test (m)'s `env -u` correctly insulate the suite from a real Actions runner. This addresses the blocking defect from the first review round.
- CI: `build-and-test` is green, and I confirmed the file is auto-discovered by the existing `for test_script in scripts/test_*.sh` sweep in `.github/workflows/build.yml`, so it runs on every build with no workflow change. The only red checks are the "No blocking labels" label-gate jobs, an orchestration concern, not a code issue.

Two non-blocking nits:

- The PR description's validation line says reintroducing Bug 1 yields 8 failures, while a later comment says 19 and I observed 9. These simply reflect different ways of reintroducing the bug; all variants are caught, so the discrepancy is immaterial. Worth tidying the description for accuracy but not worth another round.
- The open question on the PR thread ("Should this be present indefinitely, or does it have a limited lifespan?") is about the lifespan of the underlying #364 watch detector, not this test-only change. That is out of scope for #463 and does not block; if a finite lifespan is desired, it belongs in a separate issue against the detector itself.

Nice work making the suite deterministic against a real Actions runner's environment.

LGTM.

aunger:
This is very hard for me to follow, and seems like a lot of work for something that has been refusing to recur.

PR #506, opened 2026-06-22
Tell reviewers to stop pointing out auto-bylines

aunger:


aunger (commented):
🤖 Reviewer

The intent here is sound: Reviewers should not waste a round on byline violations that CI already cleans up automatically. Two problems with the current wording, one of which is a correctness issue.

**1. "the attribution rules" is overbroad (correctness problem)**

The Attribution section has two distinct rules:

- Rule A: Begin comments/reviews with `🤖 {your role}`.
- Rule B: Do not append Claude attribution bylines, session URLs, or footers.

The note says "Do not enforce the attribution rules"--plural, no qualification. That inadvertently tells Reviewers to stop enforcing Rule A as well. Rule A is not handled by CI; it is a legitimate review concern, and silently dropping it would hollow out the attribution convention. The note should be scoped to Rule B only (the no-footer rule), since that is the one the `strip-bylines` CI job covers.

**2. "Claude Code's inbuilt tools make this a hassle" is unexplained**

This rationale is opaque to anyone who does not already know that CI strips bylines on every PR. "This a hassle" gives no reason a future reader can evaluate or act on. The actual reason--that the `strip-bylines` CI job already removes these footers automatically, so Reviewer enforcement is redundant--should be stated. Without it, the note reads as "don't bother" with no justification.

Suggested scope for the fix: narrow the note to the Claude-attribution-footer rule specifically, and replace the vague rationale with the actual one (CI handles it).

---

CI notes: "No blocking labels" is failing, which is expected while the development cycle is active per the project's CI conventions. All other substantive checks (build-and-test, semgrep, CodeQL, strip-bylines) passed.

aunger:
🤖 Author

Both points are well-taken. I've pushed a fix (33d92f6):

- Narrowed the note from "the attribution rules" to "the no-footer rule above," explicitly naming what it covers (bylines, session URLs, footers). The role-prefix rule (Rule A) is unaffected and remains enforceable.
- Replaced the vague rationale with the concrete one: the `strip-bylines` CI job removes these footers automatically on every PR, so Reviewer enforcement is redundant.

aunger (commented):
🤖 Reviewer

The Author's fixes to the previous round's two problems are correct: the note is now scoped to the no-footer rule only (Rule B), and the rationale is the actual concrete one (the `strip-bylines` CI job makes Reviewer enforcement redundant). One problem remains.

**`[!note]` (lowercase) will not render as a styled alert on GitHub**

GitHub's alert syntax requires the keyword to be uppercase: `[!NOTE]`, `[!TIP]`, `[!WARNING]`, `[!IMPORTANT]`, `[!CAUTION]`. The lowercase form `[!note]` is not listed in GitHub's documentation and renders as a plain blockquote rather than a highlighted callout. The purpose of using alert syntax here is to draw a Reviewer's eye to a "don't do this" instruction; a plain blockquote defeats that purpose. Fix: change `> [!note]` to `> [!NOTE]`.

---

CI note: "No blocking labels" is failing as expected during the active development cycle. All substantive checks (build-and-test, Semgrep, CodeQL, strip-bylines) passed.

aunger:
🤖 Author

The Reviewer is correct. `[!note]` (lowercase) is not in GitHub's alert-syntax spec and would render as a plain blockquote, which defeats the point of drawing a Reviewer's eye to the instruction. Fixed in f4d1f09: changed `[!note]` to `[!NOTE]`.

aunger (commented):
🤖 Reviewer

Both issues from Round 1 and the one issue from Round 2 have been correctly addressed. The change is ready to merge.

The note is now precisely scoped to Rule B (the no-footer rule) only, leaving Rule A (the role-prefix line) untouched and enforceable.
The rationale is concrete: the `strip-bylines` CI job makes Reviewer enforcement redundant.
`[!NOTE]` (uppercase) is the correct GitHub alert syntax and will render as a styled callout.
The placement, immediately after the rule it qualifies, is correct.
The new note is consistent with the existing line-15 instruction ("the Reviewer must not mention them in reviews"); the two are complementary, not contradictory.

CI: "No blocking labels" is failing as expected during the active development cycle.
All substantive checks (build-and-test, Semgrep, CodeQL, strip-bylines) passed.

LGTM

PR #507, opened 2026-06-22
Rework scope creep rules

### LINKED ISSUE #508: Rework scope creep rules to reflect shared responsibility
The current scope rule in `agents/pr_participation.md` assigns filing responsibility solely to the Author:

> The Author is responsible for filing new issues to track out-of-scope requests raised during review.

This should be updated to:
- Reflect that both Reviewer and Author share responsibility for pushing back against scope creep.
- Allow either agent to flag follow-up suggestions in their PR comment for the Verification Planner to pick up.
- Retain clear Author accountability for filing issues when an out-of-scope suggestion is raised, so nothing falls through the cracks.

aunger:
Fixes #508

Emphasizes shared responsibility for scope management during code reviews, allows either agent to flag follow-up suggestions in their PR comment for the Verification Planner to pick up, and retains Author accountability for ensuring out-of-scope requests get filed as issues.

### DIFF
--- a/agents/pr_participation.md
+++ b/agents/pr_participation.md
@@ -14,7 +14,7 @@
 - The Reviewer need not enforce expectations written with "should" language.
 - Although **bylines** (Claude attribution, links) are prohibited, the Reviewer must not mention them in reviews.
 - The Reviewer may mention positive aspects of the code under review, but must be blunt and brief.
-- Our agents share the User's GitHub account, so you won't use GitHub's code review features, which require separate accounts. Leave your evaluation as an ordinary comment, and tell the Orchestrat[...]
+- Our agents share the User's GitHub account, so you won't use GitHub's code review features, which require separate accounts. Leave your evaluation as an ordinary comment, and tell the Orchestrator your decision. The user and other agents know to expect this.
 - **If CI results are already available** when you complete your review, you may note them in your review text, but do not block on them.
   The Orchestrator runs the CI Monitor script after you exit; you do not need to poll.
 - **Do not return before posting your review.** Posting your review is the only valid exit condition. After forming your verdict, call the review-submission tool before stopping.
@@ -36,11 +36,6 @@
   There is no middle option.
   If you want any change made before merge, request changes so the full review cycle continues.
 
-### CI checks during the development cycle
-
-Not all CI checks can pass while the development cycle is active, and that is expected.
-Some required checks, such as "No blocking labels", validate the post-cycle state of a PR and fail by design during agentic reviews.
-
 ### Reviewing an Author who declined to open a PR
 
 Sometimes the Author opens no PR and instead posts its position as an **issue comment** (see "Declining to open a PR" in the Author section).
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Reviewer

Three issues with this change, one of which I consider blocking.

**1. Missing issue number (blocking)**

`pr_creation.md` requires: "Include the #XYZ issue number of the issue that this PR addresses."
The PR description contains no issue number at all.
Either this change is not tracking an issue (in which case one should be filed first), or the issue number was simply omitted from the description.

**2. "automatically filed" overstates the mechanism**

The new rule says follow-up suggestions "will be automatically filed as issues for future consideration after this PR."
The actual mechanism is the Verification Planner, which reads PR comments but only runs after Reviewer approval and CI clearance--and only if the Orchestrator dispatches it.
Calling this "automatic" is inaccurate; it is conditional on a successful review cycle reaching the verification planning step.
An agent reading this rule might conclude it can offload issue-filing entirely to the automation and stop tracking whether the Planner actually runs.
The rule should either accurately describe the mechanism (e.g., "the Verification Planner will pick these up") or drop the word "automatically."

**3. Accountability weakened without replacement**

The old rule assigned filing responsibility unambiguously to the Author.
The new rule says "Either agent (Author or Reviewer) may add a section...regarding follow-up suggestions."
"May" is permissive, not obligatory.
If neither agent adds the section, nothing enforces that follow-up suggestions get filed.
The old rule's assignment of responsibility to the Author covered the case where the Reviewer raises a scope-creep suggestion: the Author was on the hook to file it.
The new rule makes that same case optional.
Consider either retaining "the Author is responsible for filing new issues" alongside the new shared-responsibility framing, or making the section addition mandatory ("must") for whichever agent raises the out-of-scope suggestion.

**Nit: line wrapping**

The prose-style rule says to split prose at sentence breaks.
The continuation sentence "These will be automatically filed..." shares a bullet with the preceding sentence rather than starting on its own line after the sentence break.
This is a minor style issue but is inconsistent with the project's stated standard.

aunger:
🤖 Author

Thanks for the review. All four points are addressed in the latest commit.

**1. Missing issue number** -- Filed #508 to track this work and updated the PR description with "Fixes #508".

**2. "automatically filed" overstates the mechanism** -- Agreed. Changed to "The Verification Planner will pick these up and file tracking issues after this PR," which accurately names the mechanism and its conditionality.

**3. Accountability weakened** -- Agreed. Restored "The Author is responsible for filing new issues to track out-of-scope requests raised during review" as its own bullet. The shared-responsibility sentence is now a continuation of the "don't allow scope to increase" bullet, so the Author accountability rule is separate and unambiguous.

**4. Line wrapping nit** -- The continuation sentence was already on its own indented line. No structural change was needed here; the sentence text changed as part of point 2, and it remains a properly indented sentence continuation under its bullet.

aunger:
🤖 Reviewer

The Author's second commit addresses all three substantive issues and the nit from the first review round.
The final Scope section reads cleanly and accurately.

**1. Issue number** -- "Fixes #508" is present in the PR description.
Resolved.

**2. "automatically filed" overstates the mechanism** -- The second commit replaces that phrase with "The Verification Planner will pick these up and file tracking issues after this PR," which accurately names the mechanism and its conditionality.
Resolved.

**3. Accountability weakened** -- The second commit restores "The Author is responsible for filing new issues to track out-of-scope requests raised during review" as its own separate bullet, so there is an unambiguous owner for the case where a Reviewer raises an out-of-scope suggestion.
The shared-responsibility sentence is a continuation under the "don't allow scope to increase" bullet, which is the right structural choice: it frames shared responsibility as context for the existing rule rather than replacing the Author accountability rule.
Resolved.

**4. Line-wrapping nit** -- The Author correctly noted that the continuation sentence was already on its own indented line.
The second commit changes only the sentence text (as a consequence of fix 2), not the structure.
The result is consistent with the prose-style rule to split at sentence breaks.
Resolved.

The logic of the new rules holds up: "may add a section" is permissive by design (it is an optional mechanism for surfacing follow-up suggestions), while Author accountability for filing issues is separately mandatory.
The two bullets do not conflict.

LGTM

Issue #512, opened 2026-06-25
CI Monitor: fetch /commits/{sha}/check-runs once per poll and reuse it for both the verdict and diagnostics (wiring option (a))


aunger:
Follow-up to #500. In #500 the CI Monitor was made data-driven about which run/job to track by deriving `(run_id, job_id)` pairs from the `/commits/{sha}/check-runs` payload. Per the owner's decision on #500, the simpler **wiring (b)** was implemented: `poll_signals` fetches its own `/commits/{sha}/check-runs` every poll, separately from the main loop's verdict fetch.

This issue tracks the more efficient and robust **wiring (a)**: fetch `/commits/{sha}/check-runs` exactly once per poll in the main loop and pass the parsed payload (or the parsed targets) into both `parse_check_result` (verdict) and `poll_signals` (diagnostics).

Benefits:
- One fewer `/commits/{sha}/check-runs` request per poll, and one fewer per drain attempt (the drain calls `poll_signals` up to `DRAIN_MAX_ATTEMPTS` times). At roughly one poll per 30s over up to 30 minutes, plus drains, this is a measurable reduction against the GitHub REST rate limit.
- Verdict and diagnostics are computed from the *same* check-runs snapshot, eliminating any window where the verdict and the discovered run/job ids disagree because they were fetched microseconds apart.

Implementation sketch:
- `poll_signals` takes an optional `check_json=None` parameter. When `None` (e.g. called by `drain_then_print`), it self-fetches as today; when given, it uses the passed payload.
- The main loop passes its already-fetched `check_json` into `poll_signals(sha, check_json=check_json)`.
- The drain must re-fetch to observe lag catch-up, so the drain keeps `check_json=None`.
- Update the `main()` integration tests' per-poll request budgets: the non-drain polls drop one `/commits/{sha}/check-runs` request each; drain attempts are unchanged.

The in-code comment in `poll_signals` (added in #500) points here.

Refs: #500.

PR #513, opened 2026-06-25
Make CI Monitor discover the tracked run/job from check-runs data, with a config file of tunable regexes

### LINKED ISSUE #499: CI Monitor: repeated "drain poll found no new diagnostic signals" + Blocked on PR #495; also anomalous Clear on stale-base run
## Context

During orchestration of issue #494 (PR #495), the CI Monitor was run twice as an informational step (once after each Author round, before the Reviewer was dispatched).
Both runs hit the `undiagnosedTerminal` path and the 5-minute recheck; both rechecks repeated the same pattern.
An additional anomalous `Clear` result was observed on an earlier run against a stale-base version of the branch.

## Run A -- after Programmer round 1 (redo on correct base)

**Original invocation:**
```
python3 scripts/ci_monitor/ci_monitor.py --pr 495 --include-fail .+ --include-skip .+ --include-pass .+
```

**Output:**
```
monitor PID 9404 — if TaskStop is unavailable, send SIGTERM to this PID to stop me
PR#495: in_progress
PR#495: drain poll found no new diagnostic signals
PR#495: Blocked
```

`drain poll found no new diagnostic signals` immediately preceded `Blocked`, triggering the `undiagnosedTerminal` path.
A 5-minute `sleep 300` was issued, then the Monitor was re-launched.

**Recheck output:**
```
monitor PID 10206 — if TaskStop is unavailable, send SIGTERM to this PID to stop me
PR#495: drain poll found no new diagnostic signals
PR#495: Blocked
```

Same shape: no diagnostic signals on the recheck either.
Per the rules, the original `Blocked` routing was accepted and the Reviewer was dispatched.

## Run B -- after Programmer round 2

**Original invocation:** same command.

**Output:**
```
monitor PID 15707 — if TaskStop is unavailable, send SIGTERM to this PID to stop me
PR#495: in_progress
PR#495: drain poll found no new diagnostic signals
PR#495: Blocked
```

Same `undiagnosedTerminal` shape again.
5-minute wait, then recheck.

**Recheck output:**
```
monitor PID 16510 — if TaskStop is unavailable, send SIGTERM to this PID to stop me
PR#495: drain poll found no new diagnostic signals
PR#495: Blocked
```

Again no diagnostic signals.

## Anomalous Clear on the stale-base run

Before the branch was corrected (before the force-push to `origin/main`), a CI Monitor run was made against the same PR number (#495) with the stale-base commit (`4ee9d45`).
That run returned:

```
monitor PID 9426 — if TaskStop is unavailable, send SIGTERM to this PID to stop me
PR#495: Clear
```

This is unexpected.
The stale-base branch included an `if: always()` / `exit 1` "Block accidental merge" step, which should have caused the job to fail and the Monitor to emit `Blocked`.
A `Clear` result suggests either:
- the CI run polled had not yet started (and the previous run's `Clean` status was still cached), or
- the `mergeable_state` was `clean` or `unstable` at the time of polling due to lag between the force-push and the new CI run starting.

## Likely root cause for runs A and B

The failing step on this branch is `Block accidental merge` (`if: always()`, `exit 1`).
It does not upload any `testresults-<group>` artifact and does not correspond to one of the named E2E test steps the Monitor parses for `FAIL`/`SKIP`/`PASS` markers.
The Monitor correctly detects the job failure (emitting `Blocked`) but finds no parseable diagnostic signals to accompany it, hence `drain poll found no new diagnostic signals` on every poll including recheck.

This is expected behavior for a pure shell-exit failure with no test-result artifacts, but the persistent `undiagnosedTerminal` detour (5-minute wait + recheck) added ~10 minutes of overhead per Author round.

## Possible improvements

- The Monitor could emit a `step "Block accidental merge" -> failure` line (or similar) when a non-test step fails, giving the Orchestrator a named signal instead of an empty drain.
- Alternatively, the `undiagnosedTerminal` recheck wait could be shortened or skipped when the failing check is a known always-failing gate step.
- The `Clear` anomaly on the stale-base run may warrant a check for stale `mergeable_state` lag after a force-push.

### LINKED ISSUE #500: CI Monitor: discover the tracked run from check-runs data instead of hardcoded workflow/job names
# Implementation Plan: Issue #500 with Owner Comment Integrated

## 1. Goal, scope, and what the owner's comment changes

**Goal (unchanged from the original issue):** make `scripts/ci_monitor/ci_monitor.py` data-driven in selecting which workflow run and job to track, by deriving `(run_id, job_id)` pairs from the `/commits/{sha}/check-runs` payload (via each Actions check run's `details_url`, gated by `app.slug == "github-actions"` with a URL-pattern fallback), removing the three static couplings. The Orchestrator-facing output vocabulary (`Clear`/`Blocked`/`Infra`/`step`/`FAIL`/`PASS`/`SKIP`/`drain poll found no new diagnostic signals`) is preserved verbatim.

**What the owner's comment changes, in one table:**

| Decision | Original assumption | Owner's ruling | Effect on plan |
|---|---|---|---|
| 1. `testresults-` artifact prefix | Keep as a hardcoded contract (assumed yes) | Keep by default, but make the match a **configurable regex in a config file** | EXTENDS: new config key `artifact_name_regex`; in-code default replaces the literal `n.startswith("testresults-")` |
| 2. step-name heuristics | Keep hardcoded as-is | Make **configurable by regex**; ship a reasonable default "interesting" filter in code, put this repo's specifics in the config file | EXTENDS: new config key `interesting_step_regex`; in-code default; repo config carries `Build and run unit tests` / `E2ETest` |
| 3. wiring (a) vs (b) | Preferred (a): reuse the verdict's single check-runs fetch | Do the **simpler one (b)**: `poll_signals` fetches its own check-runs. Leave a code comment and **file a follow-up issue** describing (a) | OVERRIDES: the original preferred (a); the comment reverses this to (b) |
| `**` note (test marker) | n/a (assumed `##GB4PC_TEST##` literal stays) | Switch default to `##TEST##`, make it a **configurable regex**; in this repo's config match **both** markers for back-compat | EXTENDS: new config key `test_marker_regex`; default `##TEST##`; repo config matches both `##TEST##` and `##GB4PC_TEST##` |

**Net: the run/job discovery rewrite is implemented as originally described, with wiring (b) instead of (a), plus a new config-file mechanism carrying three regexes.**

---

### 2.1 Format and location

**Recommendation: JSON, at `scripts/ci_monitor/ci_monitor.config.json`.**

Rationale:
- JSON is in the Python stdlib (`json`, already imported in `ci_monitor.py`). TOML would need `tomllib` (Python 3.11+) or a dependency; the repo's other scripts and the test harness lean on stdlib only and run offline (`scripts/test_ci_monitor.py` header: "No network calls required; no GITHUB_TOKEN needed"). ini/`configparser` is awkward for regex values (`%` escaping, no native nesting). JSON keeps zero new dependencies and is trivial to construct in tests.
- Co-locating it next to `ci_monitor.py` (the directory that already holds `README.md` and `CLAUDE.md`) keeps the contract discoverable and makes the default load path a simple `os.path.join(os.path.dirname(__file__), ...)`.
- Caveat: JSON has no comments. Since each key needs a "why" (especially the dual-marker back-compat), add a `_comment` string field per key or a top-level `_comment`. JSON ignores unknown keys, so `_comment*` fields are harmless. (If real comments are wanted, TOML with `tomllib` is the fallback; see Risks.)

### 2.2 Schema

Three configurable regexes, each optional (absent key → in-code default):

```json
{
  "_comment": "CI Monitor configuration. Each value is a regex. Absent keys fall back to in-code defaults. See README.md.",
  "artifact_name_regex": "^testresults-",
  "interesting_step_regex": "Build and run unit tests|E2ETest",
  "test_marker_regex": "##GB4PC_TEST##|##TEST##"
}
```

- `artifact_name_regex` — matched (via `re.search`) against each artifact's `name` to decide whether it carries `##TEST##` ndjson markers worth downloading. Replaces the literal `name.startswith("testresults-")` in `parse_new_artifacts`.
- `interesting_step_regex` — matched against a step's `name` to decide whether to emit a `step "..." -> ...` line even on a non-failing conclusion. Replaces the `name == "Build and run unit tests" or "E2ETest" in name` clause in `parse_steps`. The "genuine failure conclusion" clause stays unconditional.
- `test_marker_regex` — used to locate the marker prefix in a raw ndjson line and to find where the JSON payload begins. Replaces the literal `TEST_MARKER` find in `parse_fails`.

### 2.3 In-code defaults (used when the file is absent or a key is missing)

```python
DEFAULT_ARTIFACT_NAME_REGEX = r"^testresults-"
DEFAULT_INTERESTING_STEP_REGEX = r"(?!)"   # never-match; see note below
DEFAULT_TEST_MARKER_REGEX = r"##TEST##"
```

- Artifact default: `^testresults-` preserves today's behavior (the issue keeps the `testresults-` contract "by default").
- Test-marker default: `##TEST##` per the owner's `**` note ("switch ... by default").
- Interesting-step default — **decision point flagged (see Risks).** The owner said "Default to a reasonable 'interesting' filter." Two defensible defaults:
  - **Option D1 (recommended): default to matching nothing** (a never-match regex like `(?!)`), so out of the box steps surface only on genuine failure conclusions (the unconditional clause). This is the most generic, repo-agnostic "interesting" definition: a step is interesting if it failed. Named-step reporting on success is project-specific, which is what the repo config supplies.
  - **Option D2: default to a generic test-ish regex** like `(?i)test`, so any step whose name contains "test" is reported on any conclusion. More chatty, more opinionated.

  D1 is recommended: cleanest "reasonable default" that is not secretly this-repo-specific, and it keeps the in-code default honest about the genuine-failure rule already doing the heavy lifting.

### 2.4 The repo-specific committed config

Ship `scripts/ci_monitor/ci_monitor.config.json` with this repo's values:

- `artifact_name_regex`: `"^testresults-"` (matches `testresults-unit`, `testresults-e2e-overlay`, `testresults-e2e-gallery` from `build.yml`).
- `interesting_step_regex`: `"Build and run unit tests|E2ETest"` (reproduces today's two name heuristics; `build.yml` step names `Build and run unit tests`, `Run PixelCameraOverlayE2ETest`, `Run GalleryButtonVisualE2ETest`).
- `test_marker_regex`: `"##GB4PC_TEST##|##TEST##"` — **the back-compat dual-marker, longer alternative first.** `##TEST##` is a *substring* of `##GB4PC_TEST##` (not a prefix), so the JSON payload offset must be computed from the actual matched span (see 3.4). Listing the full marker first makes the regex engine prefer it when present, so the matched span (and thus the payload offset) is correct for both forms.

### 2.5 Loading, absence handling, and test injection

Add a loader:

```python
def load_config(path=None):
    """Load the CI Monitor config, falling back to in-code defaults.

    Returns a dict with keys artifact_name_regex, interesting_step_regex,
    test_marker_regex. A missing file, unreadable file, or invalid JSON
    falls back entirely to defaults (the Monitor must never abort on config).
    Individual missing keys fall back to their own default.
    """
```

Behavior:
- Default `path` = `os.path.join(os.path.dirname(os.path.abspath(__file__)), "ci_monitor.config.json")`.
- File absent / unreadable / invalid JSON → log one line to stderr and return all-defaults. **Never abort the resilient loop.**
- Each key independently defaults if absent.
- Optionally validate each value compiles as a regex (`re.compile`); on a bad pattern, log to stderr and fall back to that key's default rather than crashing.
- `main()` calls `load_config()` once at startup and threads the result (or three compiled patterns) into `poll_signals`, which passes them to `parse_steps`, `parse_fails`, and `parse_new_artifacts`.

**Test injection:** the parser functions already take their tunables as parameters (`outcome_filters` is the precedent). Follow that: `parse_steps(jobs_json, seen, job_ids, interesting_step_regex=...)`, `parse_fails(lines, seen, outcome_filters=None, test_marker_regex=...)`, `parse_new_artifacts(artifacts_json, seen, artifact_name_regex=...)`. Each defaults to the in-code default so existing direct-call tests need minimal change. For `main()` tests, patch `ci_monitor.load_config` via `unittest.mock` (simplest, matches existing mock style) or have `main` accept an optional config path.

---

## 3. Reconciled step-by-step implementation plan for `ci_monitor.py`

Changes split by concern for separate commits (per `agents/code_edit.md`: refactors separate from behavior changes).

### 3.0 Config plumbing (new, from the comment)
- Add `load_config()` and the three `DEFAULT_*` constants. Remove the module-level `TEST_MARKER = "##GB4PC_TEST##"`, since the marker is now config-driven; route any reference through the new parameter.
- `main()` loads config once (after arg parse) and threads it through.

### 3.1 Replace run discovery (coupling 1 + 3) — `parse_actions_targets`
- New parser `parse_actions_targets(check_json)` replacing `parse_run_id`:
  - Iterate `check_json.get("check_runs", [])`.
  - Gate on `app.slug == "github-actions"`; if the `app` block is missing, fall back to "the `details_url` matched `/actions/runs/`".
  - Parse `details_url` (fall back to `html_url`) with `re.search(r"/actions/runs/(\d+)(?:/job/(\d+))?", url)`. Group 1 = run id; group 2 = job id (may be `None`).
  - Return a deterministically ordered, de-duplicated collection of `(run_id, job_id)` tuples (sort for stable test assertions).
- Empty `check_runs` → empty result (preserves #415 `Clear` on `total_count == 0`).

### 3.2 `parse_steps`: filter by job id, configurable interesting-step rule (coupling 2 + decision 2)
- Signature: `parse_steps(jobs_json, seen, job_ids, interesting_step_regex=DEFAULT_INTERESTING_STEP_REGEX)`.
- Replace `if j.get("name") != "build-and-test": continue` with `if job_ids is not None and str(j.get("id")) not in job_ids: continue`. `job_ids=None` means no filter.
- Replace the name-heuristic clause:
  - Old: `name == "Build and run unit tests" or "E2ETest" in name or concl in (failure...)`.
  - New: `re.search(interesting_step_regex, name) or concl in ("failure", "cancelled", "timed_out", "action_required")`.
  - The genuine-failure conclusion clause stays unconditional, so a failing step always surfaces regardless of regex (this is what lets the bare `exit 1` step from #495 surface once the right run is tracked).

### 3.3 `parse_new_artifacts`: configurable artifact-name regex (decision 1)
- Signature: `parse_new_artifacts(artifacts_json, seen, artifact_name_regex=DEFAULT_ARTIFACT_NAME_REGEX)`.
- Replace `n.startswith("testresults-")` with `re.search(artifact_name_regex, n)`. Keep the `not a.get("expired") and str(a["id"]) not in seen` guards.

### 3.4 `parse_fails`: configurable test-marker regex with correct payload offset (`**` note)
- Signature: `parse_fails(lines, seen, outcome_filters=None, test_marker_regex=DEFAULT_TEST_MARKER_REGEX)`.
- Replace the literal `i = raw.find(TEST_MARKER)` / `raw[i + len(TEST_MARKER):]` logic with a regex search:
  - `m = re.search(test_marker_regex, raw)`; if `m is None: continue`.
  - JSON payload begins at `m.end()` (end of the matched marker span), so dual-marker matching parses both `##TEST##` and `##GB4PC_TEST##` correctly. This is why `test_marker_regex` lists the longer alternative first (`##GB4PC_TEST##|##TEST##`): a plain `##TEST##` pattern matches at the `TEST##` substring inside `##GB4PC_TEST##`, putting `m.end()` mid-marker and breaking the JSON parse. The dual-alternative regex with the full marker first avoids this. **This is the load-bearing reason the repo config lists both markers rather than relying on `##TEST##` substring-matching `##GB4PC_TEST##`.**
- Everything else (`json.loads`, outcome filter, dedup by `suite#name#outcome`, trace truncation) is unchanged.

### 3.5 Rewrite `poll_signals` using wiring (b) — OVERRIDE of the original preferred (a)
- **Wiring (b):** `poll_signals(sha, ...)` fetches its own `/commits/{sha}/check-runs` payload every call, calls `parse_actions_targets` on it, then:
  - Build the distinct run-id set and job-id set.
  - For each distinct run id: fetch `/actions/runs/{run_id}/jobs?per_page=30` → `parse_steps(jobs_json, seen_steps, job_ids, interesting_step_regex)`; fetch `/actions/runs/{run_id}/artifacts?per_page=100` → `parse_new_artifacts(artifacts_json, seen_arts, artifact_name_regex)` → download → `extract_ndjson_lines` → `parse_fails(lines, seen_fails, outcome_filters, test_marker_regex)`.
  - Preserve dedup sets and the "mark artifact seen only after successful download+parse" rule.
- **Delete** the `/actions/runs?head_sha=...&event=pull_request&per_page=5` URL and the `parse_run_id` call entirely.
- Add an **in-code comment** at the top of `poll_signals` (and/or where it fetches check-runs) pointing to the follow-up issue:
  > `# Wiring (b) (issue #500): poll_signals fetches its own /commits/{sha}/check-runs`
  > `# each poll. This re-fetches the payload the main loop also fetches for the`
  > `# verdict (one extra request per poll, plus one per drain attempt). The more`
  > `# efficient/robust option (a) -- fetch check-runs once and pass it into both`
  > `# the verdict and poll_signals -- is tracked in issue #<NEW>.`
- **Main-loop note:** under (b), the main `while` loop keeps its own check-runs fetch for the verdict unchanged, and `poll_signals(sha)` independently re-fetches. The drain (`drain_then_print`) already calls `poll_signals` separately, so under (b) no `check_json=None`-defaulting parameter is needed (every `poll_signals` call self-fetches). This is the simplicity (b) buys: no payload threading.

### 3.6 No change to verdict / control flow
- `parse_check_result`, `parse_pr_terminal`, `drain_then_print`, retry/backoff helpers, the `while` loop's verdict routing, and the PID/SIGTERM advisory all stay as-is. The drain still re-invokes `poll_signals`, which now self-fetches check-runs and discovers ids.

### 3.7 Docs
- `scripts/ci_monitor/README.md`:
  - Update the "reads results at step granularity from two polled REST signals" note: the Monitor discovers run(s)/job(s) from `/commits/{sha}/check-runs` (via each Actions check run's `details_url`), not by naming a workflow/job.
  - Soften the `step "..."` vocabulary row: "a step of a tracked CI job reached a conclusion" (drop the `build-and-test`-by-name phrasing). Keep all outcome tokens.
  - Add a "Configuration" section documenting `ci_monitor.config.json`, the three regex keys, the in-code defaults, the absence/fallback behavior, and the back-compat dual-marker rationale.
  - Update any `##GB4PC_TEST##` references to describe the configurable marker (default `##TEST##`, this repo matches both).
- `agents/dev_orchestration.md`: the Monitor loop consumes only output vocabulary, which is preserved. The `step "..."` description there references `build-and-test` and the named E2E steps — soften that line the same way (it is prose describing the informational delta, not a routing fence). No routing change.

---

## 4. New-issue stub for the option (a) follow-up

**File this as a new GitHub issue, linked to #500 and the resulting PR.**

> **Title:** CI Monitor: fetch `/commits/{sha}/check-runs` once per poll and reuse it for both the verdict and diagnostics (wiring option (a))
>
> **Body:**
> Follow-up to #500. In #500 the CI Monitor was made data-driven about which run/job to track by deriving `(run_id, job_id)` pairs from the `/commits/{sha}/check-runs` payload. Per the owner's decision on #500, the simpler **wiring (b)** was implemented: `poll_signals` fetches its own `/commits/{sha}/check-runs` every poll, separately from the main loop's verdict fetch.
>
> This issue tracks the more efficient and robust **wiring (a)**: fetch `/commits/{sha}/check-runs` exactly once per poll in the main loop and pass the parsed payload (or the parsed targets) into both `parse_check_result` (verdict) and `poll_signals` (diagnostics).
>
> Benefits:
> - One fewer `/commits/{sha}/check-runs` request per poll, and one fewer per drain attempt (the drain calls `poll_signals` up to `DRAIN_MAX_ATTEMPTS` times). At ~one poll per 30s over up to 30 minutes, plus drains, this is a measurable reduction against the GitHub REST rate limit.
> - Verdict and diagnostics are computed from the *same* check-runs snapshot, eliminating any window where the verdict and the discovered run/job ids disagree because they were fetched microseconds apart.
>
> Implementation sketch:
> - `poll_signals` takes an optional `check_json=None` parameter. When `None` (e.g. called by `drain_then_print`), it self-fetches as today; when given, it uses the passed payload.
> - The main loop passes its already-fetched `check_json` into `poll_signals(sha, check_json=check_json)`.
> - The drain must re-fetch to observe lag catch-up, so the drain keeps `check_json=None`.
> - Update the `main()` integration tests' per-poll request budgets: the non-drain polls drop one `/commits/{sha}/check-runs` request each; drain attempts are unchanged.
>
> The in-code comment in `poll_signals` (added in #500) points here.
>
> Refs: #500.

**Where the in-code comment points:** the comment lives in `poll_signals` in `ci_monitor.py` (see 3.5) and names this new issue number. After filing, replace `#<NEW>` with the real number.

---

## 5. Test plan delta (on top of the original test plan)

The original test plan already covers: `parse_actions_targets` unit tests; `parse_steps` id-filter tests; updated `main()` integration groups (i)/(j)/(k)/(m)/(n)/(p)/(q)/(r,s)/(t)/(u)/(v)/(w) with the new request shape; the #499 wrong-run regression; multi-run coverage; and a doc-sync grep. **Keep all of that.** The deltas below are what the owner's comment adds, plus the wiring-(b) recount.

### 5.1 Config loading / defaults / absence (new)
- `load_config` with a present file returns the three regex values from the JSON.
- `load_config` with the file **absent** returns all three in-code defaults (`^testresults-`, the chosen interesting-step default, `##TEST##`).
- `load_config` with **invalid JSON** returns all defaults and does not raise.
- `load_config` with a file that **omits one key** returns that key's default and the file's value for the others.
- `load_config` with a **non-compiling regex** for one key falls back to that key's default (if validation is implemented).
- All offline (write a temp file under `tmp_path`/scratchpad, or patch the path).

### 5.2 Dual-marker back-compat (new, the `**` note)
- `parse_fails` with `test_marker_regex="##GB4PC_TEST##|##TEST##"` parses a line carrying `##TEST##` → FAIL emitted, JSON payload offset correct.
- Same regex parses a line carrying the legacy `##GB4PC_TEST##` → FAIL emitted, payload offset correct (regression guard for the offset subtlety in 3.4).
- A single mixed ndjson stream containing **both** marker forms → both lines parsed.
- `parse_fails` with the **default** regex (`##TEST##`) parses `##TEST##` lines and (document the expectation) does **not** correctly parse a `##GB4PC_TEST##`-only line — i.e. prove the default is `##TEST##`-only and that the dual behavior is the *repo config's* doing, not the default's. (Pins the "switch by default, keep both only in this repo" decision.)

### 5.3 Configurable artifact regex (new)
- `parse_new_artifacts` with default regex includes `testresults-unit`, excludes `unit-test-results` (a non-marker artifact from `build.yml`).
- `parse_new_artifacts` with a **custom** regex (e.g. `^myresults-`) includes `myresults-foo` and excludes `testresults-unit` — proves the prefix is no longer hardcoded.
- Existing fixtures in `main()` groups (i)/(m)/(p)/(s) use `testresults-*` names; they continue to match under the default/repo regex (no assertion change, only confirm the threaded regex reaches `parse_new_artifacts`).

### 5.4 Configurable interesting-step regex (new)
- `parse_steps` with `interesting_step_regex` matching a custom name (e.g. `MyCustomStep`) emits that step on a **success** conclusion; with a regex that does not match it, the success step is suppressed but a **failure** step still surfaces (proves the genuine-failure clause stays unconditional).
- `parse_steps` with the **repo regex** `Build and run unit tests|E2ETest` reproduces today's behavior: groups (a)-(d) fixtures still emit exactly the three named test-step lines. (Lets the existing (a)-(d) assertions stand once the regex is passed in.)
- `parse_steps` with the **in-code default** over the `ALL_SUCCESS_JOBS` fixture: assert the default's behavior explicitly (under recommended D1, no success step lines, only failures would be). Pins the default choice; update if D2 is chosen.

### 5.5 Wiring (b) per-poll request budget recount (OVERRIDE of the (a)-oriented counts)
The original test plan said "Under wiring (a) the per-poll budget drops by one (no separate `/actions/runs` call). Under wiring (b) it changes shape but not count. Pick one wiring and update the deques consistently." **We pick (b), so recount every `side_effects_*` deque and every "all N mocked requests consumed" assertion for (b):**

Per **non-drain** poll under (b), the request order becomes:
1. `pulls` (sha) — unchanged.
2. `check-runs` (verdict, main loop) — unchanged.
3. `check-runs` (diagnostics, `poll_signals` self-fetch under (b)) — **new**; replaces the old single `/actions/runs?...` entry.
4. `jobs` (per distinct run id).
5. `artifacts` (per distinct run id).
6. `zip` (per new artifact).

So compared to today's deques: **drop** the `{"workflow_runs": [...]}` entry, **add** a second `check-runs` entry (the `poll_signals` self-fetch). For a single tracked run that is a net same count but different shape. For the drain, each attempt calls `poll_signals`, which under (b) now issues its own `check-runs` fetch first, then `jobs`/`artifacts`(/`zip`): so **each drain attempt gains one `check-runs` request** versus today's `runs`+`jobs`+`artifacts` shape.

Concretely, for every `main()` group:
- Replace each per-poll `{"workflow_runs": [...]}` mock entry with a `check-runs` payload whose `check_runs[].details_url` encodes the run id and a job id, and `app.slug == "github-actions"`.
- Add the matching `"id"` field to each `build-and-test` job object so the id-based `parse_steps` filter matches the job id parsed from `details_url`.
- Recompute each "all N mocked requests consumed" count: for a single-run group, non-drain polls = 5 entries (pulls, verdict check-runs, diagnostic check-runs, jobs, artifacts) plus a zip when an artifact is new; each drain attempt = 3 entries (diagnostic check-runs, jobs, artifacts) plus a zip when new — **versus today's 3 (runs, jobs, artifacts)**. So drain-heavy groups (i, j, m, p, q, s, t, u, w) gain `DRAIN_MAX_ATTEMPTS` extra entries (one extra check-runs per attempt). Update each deque and its consumed-count assertion accordingly.
- Group (v) (#415 Clear): today it feeds `pulls`, `check-runs` (total_count 0 → Clear), then `runs` (empty) for the `poll_signals` call. Under (b), `poll_signals` self-fetches `check-runs` instead of `runs`; the deque's third entry becomes a `check-runs` (e.g. `total_count 0`) payload rather than `{"workflow_runs": []}`. `parse_actions_targets` on it yields no targets, so no jobs/artifacts requests follow — same 3-entry total, different third entry.

**The emitted-line assertions (step/FAIL/SKIP/PASS/Blocked/Clear ordering and counts) stay unchanged** — that is the regression value the original plan calls out, and it is unaffected by (b) vs (a).

### 5.6 #499 wrong-run regression and multi-run (from the original, keep)
- Wrong-run test: check-runs where the first Actions check is an auxiliary workflow (own run id, no `testresults-*`, no matching job id) and a later check is the build (failing step + `testresults-*`). Assert the build's `step ... -> failure` and `FAIL [...]` surface before the terminal. Under (b), this payload is what `poll_signals` self-fetches.
- Multi-run test: check-runs referencing two distinct Actions run ids, each with a `testresults-*` artifact; assert signals from both runs, deduped.

### 5.7 Doc-sync grep (extended)
- The original grep: confirm no literal `build-and-test` / `Build & Test` / `event=pull_request` / `head_sha=` remain in `ci_monitor.py`.
- **Extend:** confirm no literal `##GB4PC_TEST##` and no literal `testresults-` remain hardcoded in `ci_monitor.py` (they now live only in `ci_monitor.config.json`). The default constants `^testresults-` and `##TEST##` are expected and allowed; the grep should target the legacy literals specifically.

---

### Critical files
- `scripts/ci_monitor/ci_monitor.py` — all logic changes: `load_config` + defaults, `parse_actions_targets` (replacing `parse_run_id`), `parse_steps`, `parse_new_artifacts`, `parse_fails`, `poll_signals` (wiring b), remove `TEST_MARKER`.
- `scripts/ci_monitor/ci_monitor.config.json` — **new file**; repo-specific regexes incl. the dual-marker.
- `scripts/test_ci_monitor.py` — all test deltas in section 5.
- `scripts/ci_monitor/README.md` — discovery note, `step` vocabulary row, new Configuration section, marker references. (Referenced transitively by `scripts/ci_monitor/CLAUDE.md`.)
- `agents/dev_orchestration.md` — soften the `build-and-test`/named-step prose only; no routing change.
- `.github/workflows/build.yml` — **read-only reference** for confirming artifact names (`testresults-unit`/`-e2e-overlay`/`-e2e-gallery`), job name (`build-and-test`), and step names. No edit needed; switching the Monitor's *read* marker to `##TEST##` does not require changing what `build.yml` greps (`grep '^##GB4PC_TEST##'`) — the repo config matching both markers is precisely what bridges the old emit format and the new default.

### Risks / ambiguities to confirm

1. **Interesting-step default (2.3, Option D1 vs D2).** "A reasonable 'interesting' filter" is underspecified. Recommend D1 (default matches nothing extra; steps surface only on genuine failure, repo config adds named-step niceties). Confirm, or pick D2 (`(?i)test`).
2. **Config format JSON vs TOML.** Recommend JSON (stdlib, offline-test-friendly, zero deps) with `_comment` fields since JSON lacks comments. If real comments / typed config are wanted, TOML via `tomllib` (Python 3.11+) is the alternative but adds a version floor.
3. **The marker the workflow still emits.** `build.yml` greps for `^##GB4PC_TEST##` when writing the ndjson artifacts, and the Gradle/test harness emits `##GB4PC_TEST##` today. The Monitor switching its *default read* to `##TEST##` is harmless because this repo's config matches both. But **note:** truly switching the emit side to `##TEST##` is a separate, larger change (Gradle test reporters + the `build.yml` greps) and is **out of scope** for #500 unless requested. If desired, it should be its own issue. The dual-marker config is a bridge, not an end state.
4. **Dual-marker regex ordering subtlety (3.4).** `##TEST##` is a *substring* of `##GB4PC_TEST##` but not a prefix, so a naive `##TEST##` search on a `##GB4PC_TEST##` line matches mid-string and corrupts the JSON offset. The repo config must list `##GB4PC_TEST##|##TEST##` (longer first), and the implementation must compute the payload offset from the matched span's end. Load-bearing; the test in 5.2 guards it.
5. **Wiring (b) cost.** (b) adds one `/commits/{sha}/check-runs` request per poll and per drain attempt versus (a). Over a 30-minute monitor that is bounded but non-trivial against the REST rate limit. This is the owner's explicit choice ("do the simpler one"); the follow-up issue (section 4) records the efficiency path.
6. **`parse_actions_targets` and lag.** A discovered `job_id` may not yet appear in `/actions/runs/{run_id}/jobs` (lag); per the original plan this just yields no step line that poll and is covered by the existing drain (Gap E). No new handling needed, but the multi-run/wrong-run tests should not assume instantaneous job availability.
7. **`agents/code_edit.md` commit hygiene.** Split into commits by concern: (i) config loader + defaults + repo config file; (ii) `parse_actions_targets` + `poll_signals` wiring (b) (core data-driven discovery); (iii) `parse_steps` id-filter; (iv) the three configurable regexes (artifact, step, marker); (v) docs. Keep any pure refactor separate from behavior changes. Build and run `python3 scripts/test_ci_monitor.py` locally before returning (no network/token needed).

---

Refs: #499.

🦀

aunger:
Fixes #500.

## What changed

`scripts/ci_monitor/ci_monitor.py` is now data-driven about which workflow run(s) and job(s) it tracks. Instead of hardcoding a workflow name, a job name (`build-and-test`), and a `head_sha`+`event=pull_request` runs query, it derives `(run_id, job_id)` pairs from the `/commits/{sha}/check-runs` payload by parsing each GitHub Actions check run's `details_url` (gated on `app.slug == "github-actions"`, with a `/actions/runs/` URL-pattern fallback when the `app` block is absent). This removes the three static couplings and fixes the wrong-run tracking from #499.

A new JSON config file (`scripts/ci_monitor/ci_monitor.config.json`) plus `load_config()` carry three tunable regexes, each with an in-code default and per-key/whole-file fallback so the resilient poll loop never aborts on configuration:

| Key | In-code default | Replaces |
|---|---|---|
| `artifact_name_regex` | `^testresults-` | the literal `name.startswith("testresults-")` |
| `interesting_step_regex` | `(?!)` (never-match) | the `name == "Build and run unit tests" or "E2ETest" in name` heuristic |
| `test_marker_regex` | `##TEST##` | the literal `##GB4PC_TEST##` marker |

This repo's committed config sets `artifact_name_regex: ^testresults-`, `interesting_step_regex: Build and run unit tests|E2ETest`, and `test_marker_regex: ##GB4PC_TEST##|##TEST##`. The dual marker is back-compat: the default read marker switches to `##TEST##` per the owner's note, but this repo's CI still emits `##GB4PC_TEST##`, so the config matches both. The payload offset is computed from the end of whichever alternative matched (`re.search(...).end()`), so both markers parse correctly. Alternation order does not matter for this pair: neither marker is a prefix of the other (the char after the leading `##` is `G` vs `T`), so they cannot match at the same start position, and `##TEST##` does not appear inside `##GB4PC_TEST##` at all; the full marker is listed first only as a readable convention.

Per the owner's ruling, `poll_signals` uses **wiring (b)**: it fetches its own check-runs each poll and discovers targets via `parse_actions_targets`, separately from the main loop's verdict fetch. An in-code comment points at the more efficient/robust **wiring (a)**, filed as follow-up #512. The job-id filter is scoped per run, so a run-only target does not widen another run's filter.

The Orchestrator-facing output vocabulary (`Clear`/`Blocked`/`Infra`/`step`/`FAIL`/`PASS`/`SKIP`/`drain poll found no new diagnostic signals`) is preserved verbatim.

## Owner-comment decisions honored

1. artifact prefix kept by default, made a configurable regex.
2. step heuristics made configurable; a reasonable (never-match) default ships in code, repo specifics in the config file.
3. simpler wiring (b) implemented; comment + follow-up issue (#512) describe wiring (a).
4. default marker switched to `##TEST##`, made configurable; this repo matches both markers.

## Tests

`scripts/test_ci_monitor.py` grows to 200 checks: new `parse_actions_targets` units (incl. non-Actions skip, run-only URLs, `html_url` fallback, None-safe sort), `parse_steps` job-id filter and per-run scoping, configurable artifact/step/marker regexes, dual-marker back-compat (incl. the default's `##TEST##`-only behavior), `load_config` fallbacks (absent/invalid/partial/bad-regex/committed-repo), the #499 wrong-run + multi-run regression through `main()`, and a doc-sync grep asserting the legacy literals are gone. Every `main()` integration group was rewired to the wiring-(b) request shape (verdict + diagnostic check-runs). Docs updated in `README.md`.

Ran locally: `python3 scripts/test_ci_monitor.py` (200 passed) and `python3 -m unittest discover -s scripts -p "test_*.py"` (316 passed, 1 skipped).

🦀

### DIFF
--- a/scripts/ci_monitor/README.md
+++ b/scripts/ci_monitor/README.md
@@ -16,6 +16,9 @@ python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER> [filter flags]
 `OWNER`/`REPO` default to this repo at the top of the script, and it reads `$GITHUB_TOKEN` from the environment (required).
 The script catches transient REST/parse failures per call so they cannot kill the resilient poll loop.
 
+The Monitor discovers which workflow run(s) and job(s) to track from the `/commits/{sha}/check-runs` payload, by parsing each GitHub Actions check run's `details_url` for its `(run_id, job_id)` (gated on `app.slug == "github-actions"`, with a `/actions/runs/` URL-pattern fallback when the `app` block is absent).
+It does not name a workflow or job; the run/job to follow is derived from the same check-runs data that produces the verdict.
+
 ## Per-test outcome filters
 
 By default the monitor reports **all FAIL markers**, **all SKIP markers**, and **no PASS markers**.
@@ -43,6 +46,29 @@ python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER> --include-pass 'MyFeat
 This emits a `PASS` line when the matching test passes and a `SKIP` line if it was skipped (which would be a false-validation trap: the code path was never exercised).
 With no pattern (`--include-pass ''`), every passing test is reported.
 
+## Configuration
+
+`ci_monitor.config.json` (next to the script) tunes three behaviors as regexes.
+The monitor loads it once at startup; each value is optional, and a missing file, unreadable file, invalid JSON, or non-compiling regex falls back to the in-code default for that key without aborting the resilient poll loop.
+
+| Config key | Default (in code) | Matched against | Purpose |
+|---|---|---|---|
+| `artifact_name_regex` | `^testresults-` | each run artifact's `name` (via `re.search`) | Selects which artifacts are downloaded and parsed for per-test markers. |
+| `interesting_step_regex` | `(?!)` (never matches) | each completed step's `name` | Selects which steps emit a `step "..." -> ...` line on a *non-failing* conclusion. A genuine step failure (`failure`/`cancelled`/`timed_out`/`action_required`) always surfaces regardless of this regex, so the never-match default means "out of the box, surface a step only when it failed". |
+| `test_marker_regex` | `##TEST##` | each raw ndjson line (via `re.search`) | Locates the marker that prefixes each test's JSON payload. The payload is parsed from the *end* of the matched span. |
+
+This repo's committed `ci_monitor.config.json` sets:
+
+- `artifact_name_regex`: `^testresults-` (matches `testresults-unit`, `testresults-e2e-overlay`, `testresults-e2e-gallery`).
+- `interesting_step_regex`: `Build and run unit tests|E2ETest` (reproduces the named-step reporting on success).
+- `test_marker_regex`: `##GB4PC_TEST##|##TEST##` (back-compat dual marker, see below).
+
+**Back-compat dual marker.**
+The default read marker is `##TEST##`, but this repo's CI still emits `##GB4PC_TEST##` (see `build.yml`), so the committed config matches both.
+The longer alternative is listed *first* (`##GB4PC_TEST##|##TEST##`): `##TEST##` is a substring of `##GB4PC_TEST##` (not a prefix), so a bare `##TEST##` search on a `##GB4PC_TEST##` line would match mid-marker and corrupt the JSON payload offset.
+Listing the full marker first makes the regex engine prefer it, so the matched span (and thus the payload offset, computed from the span's end) is correct for both forms.
+Switching the *emit* side to `##TEST##` (the Gradle/test reporters plus the `build.yml` greps) is a separate, larger change and out of scope here; the dual-marker config is the bridge, not an end state.
+
 ## Outcome vocabulary
 
 | Line emitted          | Meaning                                                                                               |
@@ -51,7 +77,7 @@ With no pattern (`--include-pass ''`), every passing test is reported.
 | `PR#N: Blocked ...`   | A check failed (`failure`/`action_required`) or `mergeable_state` is `behind`/`dirty`. |
 | `PR#N: Infra ...`     | A CI infrastructure problem (`cancelled`, `timed_out`, `stale`, `startup_failure`, or `mergeable_state=blocked`); escalate to user. |
 | `PR#N: in_progress`   | CI still running; emitted only after >120 s of silence (no other output); relay to user as a brief status update. |
-| `PR#N: step "..." -> ...` | A `build-and-test` step reached a conclusion: one of the three named test steps (`Build and run unit tests`, `Run *E2ETest`), or any genuine step failure. **Informational**; surfaces *which group* finished/failed and when; never ends the loop. |
+| `PR#N: step "..." -> ...` | A step of a tracked CI job reached a conclusion: a step whose name matches the configured `interesting_step_regex` (this repo: `Build and run unit tests`, `Run *E2ETest`), or any genuine step failure. **Informational**; surfaces *which group* finished/failed and when; never ends the loop. |
 | `PR#N: FAIL [suite] name: ...` | A per-test failure (message + truncated trace) parsed from a `testresults-<group>` artifact, possibly followed by indented trace lines. **Informational**; emitted by default; suppress with `--no-include-fail`; never ends the loop. |
 | `PR#N: SKIP [suite] name: ...` | A per-test skip parsed from a `testresults-<group>` artifact. **Informational**; emitted by default; suppress with `--no-include-skip`; never ends the loop. A skipped task-relevant test is a false-validation trap. |
 | `PR#N: PASS [suite] name: ...` | A per-test pass parsed from a `testresults-<group>` artifact. **Informational**; suppressed by default; enable with `--include-pass [PATTERN]`; never ends the loop. |
--- a/scripts/ci_monitor/README.md
+++ b/scripts/ci_monitor/README.md
@@ -64,9 +64,10 @@ This repo's committed `ci_monitor.config.json` sets:
 - `test_marker_regex`: `##GB4PC_TEST##|##TEST##` (back-compat dual marker, see below).
 
 **Back-compat dual marker.**
-The default read marker is `##TEST##`, but this repo's CI still emits `##GB4PC_TEST##` (see `build.yml`), so the committed config matches both.
-The longer alternative is listed *first* (`##GB4PC_TEST##|##TEST##`): `##TEST##` is a substring of `##GB4PC_TEST##` (not a prefix), so a bare `##TEST##` search on a `##GB4PC_TEST##` line would match mid-marker and corrupt the JSON payload offset.
-Listing the full marker first makes the regex engine prefer it, so the matched span (and thus the payload offset, computed from the span's end) is correct for both forms.
+The default read marker is `##TEST##`, but this repo's CI still emits `##GB4PC_TEST##` (see `build.yml`), so the committed config matches both (`##GB4PC_TEST##|##TEST##`).
+The payload offset is computed from the end of whichever alternative matched (`re.search(...).end()`), so both markers parse correctly.
+Alternation order does not matter for this pair: neither marker is a prefix of the other (`##GB4PC_TEST##` has `G` after the leading `##`, while `##TEST##` has `T`), so they can never match at the same start position, and `##TEST##` does not appear inside `##GB4PC_TEST##` at all.
+The full marker is simply listed first as a readable convention.
 Switching the *emit* side to `##TEST##` (the Gradle/test reporters plus the `build.yml` greps) is a separate, larger change and out of scope here; the dual-marker config is the bridge, not an end state.
 
 ## Outcome vocabulary
--- a/scripts/ci_monitor/ci_monitor.config.json
+++ b/scripts/ci_monitor/ci_monitor.config.json
@@ -1,0 +1,7 @@
+{
+  "_comment": "CI Monitor configuration. Each value is a regex; absent keys fall back to in-code defaults (see README.md). artifact_name_regex selects which run artifacts carry test-result markers worth downloading. interesting_step_regex selects which step names surface on a non-failing conclusion (genuine failures always surface regardless). test_marker_regex locates the per-test ndjson marker.",
+  "_comment_test_marker": "This repo matches BOTH ##GB4PC_TEST## and ##TEST## for back-compatibility while the emit side still writes ##GB4PC_TEST##. The longer alternative is listed first so the regex engine prefers it and the matched span ends at the true marker boundary, keeping the JSON payload offset correct.",
+  "artifact_name_regex": "^testresults-",
+  "interesting_step_regex": "Build and run unit tests|E2ETest",
+  "test_marker_regex": "##GB4PC_TEST##|##TEST##"
+}
--- a/scripts/ci_monitor/ci_monitor.config.json
+++ b/scripts/ci_monitor/ci_monitor.config.json
@@ -1,6 +1,6 @@
 {
   "_comment": "CI Monitor configuration. Each value is a regex; absent keys fall back to in-code defaults (see README.md). artifact_name_regex selects which run artifacts carry test-result markers worth downloading. interesting_step_regex selects which step names surface on a non-failing conclusion (genuine failures always surface regardless). test_marker_regex locates the per-test ndjson marker.",
-  "_comment_test_marker": "This repo matches BOTH ##GB4PC_TEST## and ##TEST## for back-compatibility while the emit side still writes ##GB4PC_TEST##. The longer alternative is listed first so the regex engine prefers it and the matched span ends at the true marker boundary, keeping the JSON payload offset correct.",
+  "_comment_test_marker": "This repo matches BOTH ##GB4PC_TEST## and ##TEST## for back-compatibility while the emit side still writes ##GB4PC_TEST##. The JSON payload begins at the end of whichever alternative matched, so both markers parse correctly. Alternation order does not matter for this pair (neither marker is a prefix of the other); the full marker is simply listed first as a readable convention.",
   "artifact_name_regex": "^testresults-",
   "interesting_step_regex": "Build and run unit tests|E2ETest",
   "test_marker_regex": "##GB4PC_TEST##|##TEST##"
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -55,9 +55,9 @@ DEFAULT_ARTIFACT_NAME_REGEX = r"^testresults-"
 DEFAULT_INTERESTING_STEP_REGEX = r"(?!)"
 
 # Locate the per-test marker prefix in a raw ndjson line. The JSON payload begins
-# at the end of the matched span, so a multi-alternative regex must list the
-# longer marker first (see ci_monitor.config.json). Default switched to ##TEST##
-# per issue #500; this repo's config matches both markers for back-compat.
+# at the end of the matched span (re.search(...).end()), which is correct for any
+# marker the regex matches. Default switched to ##TEST## per issue #500; this
+# repo's config matches both markers for back-compat (see ci_monitor.config.json).
 DEFAULT_TEST_MARKER_REGEX = r"##TEST##"
 
 
@@ -140,7 +140,9 @@ def parse_steps(
     """Emit step-delta lines for the tracked CI job(s).
 
     Reads parsed jobs JSON (a dict, as from /actions/runs/{id}/jobs) and a
-    `seen` set of already-reported step numbers (mutated in place). When
+    `seen` set of already-reported "<job id>#<step number>" keys (mutated in
+    place; step numbers are per-job, so the job id keeps them distinct across
+    multiple jobs/runs). When
     `job_ids` is a set, only jobs whose id is in it are considered; `job_ids` of
     None applies no job filter. Emits a line when a step reaches 'completed' if
     its name matches `interesting_step_regex` (on any conclusion) or it genuinely
@@ -157,10 +159,12 @@ def parse_steps(
         for s in j.get("steps", []):
             if s.get("status") != "completed":
                 continue
-            num = str(s.get("number"))
-            if num in seen:
+            # Dedup by job id + step number: step numbers are per-job, so across
+            # multiple jobs/runs (issue #500) a bare number would collide.
+            key = "%s#%s" % (j.get("id"), s.get("number"))
+            if key in seen:
                 continue
-            seen.add(num)
+            seen.add(key)
             name = s.get("name", "?")
             concl = s.get("conclusion") or "?"
             if re.search(interesting_step_regex, name) or concl in (
@@ -182,10 +186,10 @@ def parse_fails(lines, seen, outcome_filters=None, test_marker_regex=DEFAULT_TES
     suite#name#outcome.
 
     `test_marker_regex` locates the marker that prefixes each test's JSON payload.
-    The payload is parsed from the end of the matched span, so a multi-alternative
-    regex must list the longer marker first (e.g. `##GB4PC_TEST##|##TEST##`):
-    `##TEST##` is a substring of `##GB4PC_TEST##`, so a bare `##TEST##` search on a
-    `##GB4PC_TEST##` line would match mid-marker and corrupt the JSON offset.
+    The payload is parsed from the end of the matched span (re.search(...).end()),
+    which keeps the offset correct for any marker the regex matches, including a
+    multi-alternative regex like `##GB4PC_TEST##|##TEST##`. A line whose marker
+    the regex does not match at all is skipped.
 
     `outcome_filters` is a dict mapping outcome names ('FAIL', 'PASS', 'SKIP') to
     (enabled, pattern) tuples, where `enabled` is a bool and `pattern` is either
@@ -585,19 +589,22 @@ def main(argv):
         if not targets:
             return emitted[0]
 
-        # Distinct run ids to fetch, and the job ids to filter steps to. A None
-        # job id (run-only URL) widens the filter to all jobs in that run.
-        run_ids = []
-        job_ids = set()
+        # Map each distinct run id to the set of job ids to filter its steps to,
+        # scoped per run: a run-only target (job id None) widens the filter to
+        # all jobs of that run only, without affecting any other run's filter.
+        run_job_ids = {}
+        run_order = []
         for run_id, job_id in targets:
-            if run_id not in run_ids:
-                run_ids.append(run_id)
+            if run_id not in run_job_ids:
+                run_job_ids[run_id] = set()
+                run_order.append(run_id)
             if job_id is None:
-                job_ids = None
-            elif job_ids is not None:
-                job_ids.add(job_id)
+                run_job_ids[run_id] = None
+            elif run_job_ids[run_id] is not None:
+                run_job_ids[run_id].add(job_id)
 
-        for run_id in run_ids:
+        for run_id in run_order:
+            job_ids = run_job_ids[run_id]
             # Signal 1 — per-step conclusion deltas for the tracked job(s).
             jobs_json = _request(
                 "%s/repos/%s/%s/actions/runs/%s/jobs?per_page=30" % (API_BASE, OWNER, REPO, run_id),
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -25,6 +25,7 @@ import argparse
 import io
 import json
 import os
+import re
 import sys
 import time
 import urllib.error
@@ -35,7 +36,71 @@ OWNER = "aunger"
 REPO = "gallery-button-for-pixel-camera"
 API_BASE = "https://api.github.com"
 
-TEST_MARKER = "##GB4PC_TEST##"
+# Configurable behavior (issue #500). Each tunable is a regex with an in-code
+# default; the committed scripts/ci_monitor/ci_monitor.config.json overrides the
+# defaults with this repo's specifics. load_config() reads that file, falling
+# back to these defaults when the file is absent, unreadable, invalid, or missing
+# a key, so the resilient poll loop never aborts on configuration.
+
+# Match (re.search) against an artifact's `name` to decide whether it carries
+# per-test ndjson markers worth downloading. Preserves the historical
+# testresults-* contract by default.
+DEFAULT_ARTIFACT_NAME_REGEX = r"^testresults-"
+
+# Match (re.search) against a step's `name` to decide whether to surface a
+# `step "..." -> ...` line on a non-failing conclusion. The never-match default
+# keeps the generic, repo-agnostic rule "a step is interesting if it failed"
+# (the genuine-failure clause in parse_steps is unconditional); named-step
+# reporting on success is project-specific and supplied via the config file.
+DEFAULT_INTERESTING_STEP_REGEX = r"(?!)"
+
+# Locate the per-test marker prefix in a raw ndjson line. The JSON payload begins
+# at the end of the matched span, so a multi-alternative regex must list the
+# longer marker first (see ci_monitor.config.json). Default switched to ##TEST##
+# per issue #500; this repo's config matches both markers for back-compat.
+DEFAULT_TEST_MARKER_REGEX = r"##TEST##"
+
+
+def load_config(path=None):
+    """Load the CI Monitor config, falling back to in-code defaults.
+
+    Returns a dict with keys artifact_name_regex, interesting_step_regex, and
+    test_marker_regex. A missing file, unreadable file, or invalid JSON falls
+    back entirely to the DEFAULT_* regexes (the Monitor must never abort on
+    config). Each key independently defaults if absent, and a value that does not
+    compile as a regex falls back to that key's default rather than crashing.
+    """
+    defaults = {
+        "artifact_name_regex": DEFAULT_ARTIFACT_NAME_REGEX,
+        "interesting_step_regex": DEFAULT_INTERESTING_STEP_REGEX,
+        "test_marker_regex": DEFAULT_TEST_MARKER_REGEX,
+    }
+    if path is None:
+        path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "ci_monitor.config.json")
+    try:
+        with open(path, encoding="utf-8") as fh:
+            raw = json.load(fh)
+    except (OSError, ValueError) as e:
+        sys.stderr.write("config load failed (%s): %s; using defaults\n" % (path, e))
+        sys.stderr.flush()
+        return dict(defaults)
+
+    cfg = {}
+    for key, default in defaults.items():
+        value = raw.get(key, default)
+        if not isinstance(value, str):
+            sys.stderr.write("config key %s is not a string; using default\n" % key)
+            sys.stderr.flush()
+            value = default
+        try:
+            re.compile(value)
+        except re.error as e:
+            sys.stderr.write("config key %s is not a valid regex (%s); using default\n" % (key, e))
+            sys.stderr.flush()
+            value = default
+        cfg[key] = value
+    return cfg
+
 
 # Suppress repeated informational lines (in_progress heartbeat, "could not fetch
 # SHA", "still computing") until this many seconds have elapsed with no output of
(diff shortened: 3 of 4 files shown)


aunger:
🤖 Reviewer

I reviewed the diff against the issue #500 plan and the owner's comment, read the full `ci_monitor.py`, the config file, the README, and all 3080 lines of `scripts/test_ci_monitor.py`, and ran the suite locally: `python3 scripts/test_ci_monitor.py` -> 195 passed, and `python3 -m unittest discover -s scripts -p "test_*.py"` -> 316 passed, 1 skipped.

The core implementation is solid and faithful to the plan: `parse_actions_targets` replaces `parse_run_id`, the job-id filter and the three configurable regexes are wired through `poll_signals` -> `parse_steps`/`parse_fails`/`parse_new_artifacts`, wiring (b) is implemented with the follow-up (#512) referenced in-code, `load_config` degrades safely on every failure mode, and the output vocabulary is preserved verbatim. The #499 wrong-run/multi-run regression is genuinely exercised through `main()`. All four owner decisions are honored.

I am requesting changes for one blocking accuracy defect plus a couple of nits.

## Blocking: the dual-marker rationale is factually wrong

The PR repeatedly justifies listing `##GB4PC_TEST##` before `##TEST##` and computing the offset from `m.end()` with this claim:

> `##TEST##` is a substring of `##GB4PC_TEST##` (not a prefix), so a bare `##TEST##` search on a `##GB4PC_TEST##` line would match mid-marker and corrupt the JSON payload offset.

This premise is false. `##TEST##` is **not** a substring of `##GB4PC_TEST##`. The legacy marker is `...C_TEST##`, which lacks the leading `##` before `TEST`, so `re.search(r"##TEST##", "x ##GB4PC_TEST## {...}")` returns `None` (no match at all), not a mid-marker match. Verified:

```
>>> "##TEST##" in "##GB4PC_TEST##"
False
>>> re.search(r"##TEST##", "x ##GB4PC_TEST## {...}")
None
```

Consequences:

1. **The "longer alternative first" ordering is not load-bearing for these two markers.** Python's `re` alternation matches at the earliest position first; since `##TEST##` can never match inside or at the same start position as `##GB4PC_TEST##` (the char after `##` differs: `T` vs `G`), `m.end()` is identical regardless of alternation order. Ordering would only matter if one marker were a *prefix* of the other, which is not the case here. The strong "must list the full marker first" claim is unjustified as written.

2. **Test (ab) `out_ab_def_old` passes for the wrong stated reason.** The default `##TEST##` regex indeed does not parse a `##GB4PC_TEST##`-only line, but not because "##TEST## matches mid-marker there, so the JSON offset is corrupt and the line is skipped" (the comment at lines 2791-2794). The real reason is that `##TEST##` does not match at all, so `marker is None` and the line is `continue`d. The assertion is fine; its explanation is wrong.

This incorrect premise appears in at least five places and should be corrected (or removed) in all of them so a future maintainer is not misled into thinking the offset/ordering logic guards against a corruption case that cannot occur:

- `ci_monitor.py` lines 57-61 (`DEFAULT_TEST_MARKER_REGEX` comment).
- `ci_monitor.py` lines 184-188 (`parse_fails` docstring).
- `ci_monitor.config.json` `_comment_test_marker`.
- `README.md`, "Back-compat dual marker" section.
- The PR description's identical paragraph.
- `test_ci_monitor.py` lines 2791-2794 comment.

Computing the offset from `m.end()` is still correct and good defensive practice, and listing the longer alternative first is harmless and a fine general habit. The fix is to the *justification*, not the code: state the real reason the default does not match legacy lines, and drop or soften the "substring -> mid-marker corruption" and "load-bearing ordering" claims for this specific marker pair.

## Nit (non-blocking): job-id filter widens globally on any run-only target

In `poll_signals` (lines 590-598), if any single discovered target has `job_id is None` (a run-only `details_url`), `job_ids` is set to `None`, which disables the job filter for **every** run that poll, not just the run-only one. In a multi-run poll where one Actions check exposes only a run id and another exposes a job id, steps from all jobs of all runs would be considered. The blast radius is limited because `parse_steps` still gates on `interesting_step_regex`/genuine-failure, so unrelated success steps stay suppressed, and this is not a regression versus the old hardcoded behavior. But it is a logical imprecision worth a comment or a per-run job-id scoping. Your call whether to address here or note it; the wrong-run test (ad) gives both runs job ids, so this path is untested.

## Nit (non-blocking): plan item 3.7 (dev_orchestration.md) correctly skipped

The plan suggested softening `build-and-test`/named-step prose in `agents/dev_orchestration.md`. I confirmed that file has no such stale references (only generic `step "..."` vocabulary), so leaving it untouched is correct. Noting it so it is not mistaken for an omission.

Decision: Changes requested.

aunger:
🤖 Author

Thanks for the careful review. You are right on all three points; pushed `48064b6`.

**Blocking (dual-marker rationale).** Confirmed the premise was false: `##TEST##` is not a substring of `##GB4PC_TEST##` (the char after the leading `##` is `G`, not `T`), so `re.search(r"##TEST##", "x ##GB4PC_TEST## {...}")` returns `None`, not a mid-marker match. So:

- The default does not parse a legacy line because it does not match at all (`marker is None` -> `continue`), not because of a corrupt offset.
- "Longer alternative first" is not load-bearing for this pair; it would only matter if one marker were a prefix of the other. Computing the offset from `re.search(...).end()` stays correct and I kept it.

Corrected the rationale in all six places: `ci_monitor.py` (the `DEFAULT_TEST_MARKER_REGEX` comment and the `parse_fails` docstring), `ci_monitor.config.json` (`_comment_test_marker`), `README.md` (Back-compat dual marker section), the PR description, and the test (ab) comment. The full marker is now described as listed first only as a readable convention, and test (ab)'s `out_ab_def_old` comment now states the real reason (no match -> skipped).

**Nit 1 (job filter widened globally).** Good catch, and it was genuinely untested. `poll_signals` now scopes job-id filtering per run (`run_job_ids: dict[run_id -> set | None]`), so a run-only target only disables the filter for its own run. Added test (af): a multi-run poll where run 100 is run-only and run 200 is scoped to job 22; it asserts run 200's other job (23) stays suppressed. While writing it I hit a latent dedup collision (`parse_steps` deduped by step number alone, but step numbers are per-job), so the dedup key is now `"<job id>#<step number>"`.

**Nit 2 (dev_orchestration.md).** Agreed, confirmed it has only generic `step "..."` vocabulary and no stale `build-and-test`/named-step prose, so it is left untouched.

Suite: `python3 scripts/test_ci_monitor.py` -> 200 passed; `python3 -m unittest discover -s scripts -p "test_*.py"` -> 316 passed, 1 skipped.

aunger:
🤖 Reviewer

Re-reviewed `48064b6`. All three points are resolved, and I ran both suites on the new HEAD: `python3 scripts/test_ci_monitor.py` -> 200 passed, `python3 -m unittest discover -s scripts -p "test_*.py"` -> 316 passed, 1 skipped.

Blocking (dual-marker rationale): corrected accurately in all six locations (the `DEFAULT_TEST_MARKER_REGEX` comment, the `parse_fails` docstring, `ci_monitor.config.json`, README, the PR body, and the test (ab) comment). The text now states the real facts: the offset comes from `re.search(...).end()` and is correct for any matched marker; a line whose marker does not match is simply skipped; and alternation order is not load-bearing for this pair because neither marker is a prefix of the other and `##TEST##` does not appear inside `##GB4PC_TEST##`. The "full marker first" is now framed as a readable convention. The code (offset from `m.end()`) is unchanged, as expected.

Nit 1 (job-id filter): fixed properly. `poll_signals` now scopes the filter per run via `run_job_ids`, so a run-only target only widens its own run's filter. Test (af) is a real regression guard: run 200 is scoped to job 22, and its other job (23, named `GalleryButtonVisualE2ETest`) is asserted suppressed, while the run-only run 100 still surfaces its step. That assertion fails under the old global-widening behavior, so it is not vacuous.

Bonus: while implementing the per-run scoping, the Author found and fixed a genuine latent dedup collision. `parse_steps` deduped by bare step number, which collides across jobs once multiple jobs/runs are tracked (step numbers are per-job). The key is now `<job id>#<step number>`, and test (af) exercises it (jobs 22 and 23 both use step number 1). Good catch.

Nit 2 (dev_orchestration.md): confirmed correctly as needing no change.

The implementation is data-driven per issue #500, honors all four owner decisions, preserves the output vocabulary verbatim, and is well-tested. No remaining concerns.

Decision: LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #513 can be merged):

None.

- [ ] #512

PR #521, opened 2026-06-27
Rename label "for ai to do" to "orchestrate"

aunger:
## Summary

Rename the GitHub label "for ai to do" to "orchestrate" throughout the codebase, including code, documentation, and tests.

Fixes #498

## Changes

- Updated mutually exclusive label set in `enforce_mutually_exclusive_labels.py` and workflow
- Updated documentation in `dev_orchestration.md` and `.claude/environment.md`
- Updated test cases in `test_enforce_mutually_exclusive_labels.py` and `test_archive_stale_test_failures.py`

## Test plan

- All unit tests pass (64 tests in `test_enforce_mutually_exclusive_labels.py`, 24 tests in `test_archive_stale_test_failures.py`)

Issue #498, opened 2026-06-21
Rename label "for ai to do" to "orchestrate"


aunger:
Do this in code and in markdown files. 

If you can change the label itself, great. I can do that part though.

PR #522, opened 2026-06-27
Add Cannot-implement Reviewer signal and clarify abort triggers

aunger:
Improvements to `agents/dev_orchestration.md`, surfaced while orchestrating issue #498 / PR #521 (where the loop spun because the Reviewer had no clean way to say "the code is right but coding cannot finish this").

### New Reviewer signal: `Cannot implement`

The Reviewer's outcome vocabulary was `LGTM` / `Changes requested` only.
When the coding phase cannot be completed (requirements incomplete, unattainable, or self-contradictory, or no code change could address the issue), `Changes requested` was misleading: it implies the Author can fix it with another in-repo round and loops the Orchestrator back pointlessly.

Added `Cannot implement`, routed to immediate user escalation (not a new Author round) in both the PR and no-PR routing fences.
Also gave each outcome an explicit definition so `LGTM` and `Changes requested` are unambiguous.

### Clearer abort triggers

The "When to abort" section now lists both escalation paths: immediate escalation when a sub-agent emits `Cannot implement`, and the four-round cap as the fallback for a stalled loop (stated to apply unconditionally once reached, even mid-dispute).

### Review fixes

- Pointed the `Cannot implement` explanation at an issue comment on the no-PR path (where there is no PR to comment on).
- On `Cannot implement` with an open PR, the PR is left for the user to close, matching the existing "the Orchestrator does not close it" pattern.
- Inlined the routing parenthetical so it unambiguously belongs to the `Cannot implement` branch.
- Fixed an `is require` -> `is required` typo.

## Test plan

- [x] Verify the changed sections read clearly in context (no automated test covers doc prose).

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -129,10 +129,9 @@ The Orchestrator does not read or judge the Author's explanation; it only notes 
 
 Reviewer-to-Orchestrator outcome vocabulary (the Reviewer emits one):
 - `LGTM`
-- `Changes requested`: in-repo changes the Author can make.
-  Use only when the feedback is addressable by editing the repository.
-- `LGTM; out-of-repo action required`: the code is correct and needs no in-repo change, but something outside the repository (e.g. renaming a label, updating external config) must happen before the PR can merge.
-  The Reviewer describes the specifics in a PR comment for the user to read.
+- `Changes requested`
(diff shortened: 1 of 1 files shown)


aunger:
If the PR’s file changes are correct, complete, and beneficial, then the response should be LGTM, with any extra needs mentioned clearly in comments.

Rename this new state “Cannot implement”. It’s only for occasions that the coding phase can’t be completed. Examples include incomplete, unattainable, or self-contradictory requirements,

PR #524, opened 2026-06-28
Reserve CI test-failure issue body for an aggregate summary

### LINKED ISSUE #504: CI auto-filed issues: put even the first occurrence in a comment
for the CI-auto-filed issues (issues with a `test-failure` label), begin reserving the issue body (“description” field) for summary and aggregate information. This will be expanded later, but for now, just write a simple message that will apply to the group of failures that may be added over time, such as this:

> # Automated test failure
> Job: [{CI job name}]({link to workflow file})
> Step: {CI step name}
> Class: {test class name or suite category name}
> Test: {test case name}
> 
> Failed runs will be added as comments on this ticket over time.

To reserve the description space, even the first occurrence of a failure should be placed in a comment, instead of being written to the issue’s body (description). 

This means that when a test case fails for the first time (or all previous failures have been archived), the test filer will file a new issue with general information about the test, and then immediately post a comment about this particular failure (like the current comments, with date/time, failure message, and stack trace). As currently done, subsequent failures on different runs will be assigned only a comment.

aunger:
🤖 Author

Fixes #504.

## Problem

For CI auto-filed issues (issues labeled `test-failure`), the issue body (description) held the details of the *first* failing run, while every subsequent failure was added as a comment. That left the description space unavailable for summary or aggregate information about the group of failures.

## Change

The issue body is now reserved for summary and aggregate information that applies to the whole group of failures accumulated over time. Even the first occurrence's per-run details now go into a comment, matching how subsequent failures are already handled.

- New `make_summary_body` builds a short aggregate summary for the issue description:
  - `Job`: the suite label, linked to the workflow file (`build.yml`) on the failing branch.
  - `Class`: the test class / suite category.
  - `Test`: the test case name.
  - A note: "Failed runs will be added as comments on this ticket over time."
- `process_failure` now, on first occurrence, creates the issue with this summary body and then immediately posts the per-run failure details (timestamp, message, stack trace) as a comment, identical to the comment a duplicate failure would receive.
- Subsequent failures continue to add comments only (unchanged).

The existing `make_issue_body` is retained and still produces the per-run detail block used for the comments.

## Notes

- The exact summary template in the issue is an example ("such as this"); the implementation uses the data the filer actually has (suite label, class, method) and links the `Job` line to the workflow file. The CI step name was not available as input to the filer, so it is not included.

## Tests

- Added `TestMakeSummaryBody` covering the summary structure, the workflow-file link (branch and `HEAD` fallback), and the exclusion of per-run details from the body.
- Updated `test_creates_issue_when_no_duplicate` to assert that a comment is now posted on creation, and added `test_new_issue_body_is_aggregate_summary_not_run_details` and `test_no_comment_when_issue_creation_fails`.

*Modified test `test_creates_issue_when_no_duplicate` to expect a comment on first occurrence instead of asserting no comment is posted, because the first occurrence's per-run details now go into a comment per issue #504.*

All 78 tests in `scripts/test_file_test_failure_issues.py` pass locally.

🦀

### DIFF
--- a/scripts/file_test_failure_issues.py
+++ b/scripts/file_test_failure_issues.py
@@ -137,6 +137,47 @@ def _find_ocr_text(directory: Path, class_name: str, method_name: str) -> str | 
     return None
 
 
+# Workflow file that runs the test suites and files these issues. Used to link
+# the "Job" line in the issue summary body to the workflow definition.
+_WORKFLOW_FILE_PATH = ".github/workflows/build.yml"
+
+
+def make_summary_body(
+    failure: FailedTest,
+    github_server_url: str,
+    github_repository: str,
+    workflow_run_branch: str,
+) -> str:
+    """Build the aggregate summary body for a test-failure issue.
+
+    The issue description is reserved for information that applies to the whole
+    group of failures accumulated over time, not to any single run.
+    Per-run failure details (timestamp, message, stack trace) live in comments
+    instead, including the first occurrence (issue #504).
+
+    The "Job" field links to the workflow file that runs the suite and files
+    these issues.
+    The link targets the workflow on the failing branch when known, so it points
+    at the version that produced the failure.
+    """
+    ref = workflow_run_branch or "HEAD"
+    workflow_url = (
+        f"{github_server_url}/{github_repository}/blob/{ref}/{_WORKFLOW_FILE_PATH}"
+    )
+
+    return f"""\
+# Automated test failure
+
+| Field | Value |
+|-------|-------|
+| Job | [{failure.suite_label}]({workflow_url}) |
+| Class | `{failure.class_name}` |
+| Test | `{failure.method_name}` |
+
+Failed runs will be added as comments on this ticket over time.
+"""
+
+
 def make_issue_body(
     failure: FailedTest,
     directory: Path,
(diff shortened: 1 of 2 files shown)


aunger (commented):
🤖 Reviewer

The implementation is correct and the tests are solid. CI passes (build-and-test: success; analyze-python: success; all security scans: success). One nit and one observation follow, neither blocking.

## Nit: summary body uses a table; issue showed a plain list

Issue #504 showed this template:

> Job: [{CI job name}]({link to workflow file})
> Step: {CI step name}
> Class: {test class name or suite category name}
> Test: {test case name}

The PR renders it as a Markdown table instead. The issue said "such as this," so the template is an example, not a hard spec--and a table renders better on GitHub than a bare list. The omission of "Step" (not available to the filer) is correctly explained in the PR notes. No change needed.

## Observation: `make_issue_body` is now used only for comments, but its name and docstring still say "issue body"

`make_issue_body` now exclusively produces comment text; the actual issue body is produced by `make_summary_body`. The function's docstring ("Build the Markdown body for a test-failure issue") is now misleading. This is a pre-existing name that predates this PR, and renaming it would be a larger refactor with its own test churn. It is worth tracking but is appropriately out of scope here.

## Summary

The change is correct. It faithfully implements the requirement: new issues get an aggregate summary body, and even the first failure's per-run details go into a comment. The tests cover creation, comment-on-first-occurrence, no-comment-on-creation-failure, and the summary-vs-run-detail split cleanly.

LGTM

PR #525, opened 2026-06-28
Wire git fetch into SessionStart and PostToolUse hooks

### LINKED ISSUE #496: Add Claude hooks: git-fetch in env/startup and after every sub-agent returns

aunger:
🤖 Author

Fixes #496

## What changed

**SessionStart (`session-start.sh`):** Added Step 4 -- a `git fetch --prune --quiet` at the end of the existing session-start hook.
This ensures the local repo's remote-tracking refs are current from the moment a Claude Code for Web session begins.

**PostToolUse (new `post-tool-use-fetch.sh` + `settings.json`):** Added a new PostToolUse hook matched to the `Agent` tool.
Every time a sub-agent finishes, this hook runs `git fetch --prune --quiet` so the parent session immediately sees any commits the sub-agent pushed.

Both hooks treat fetch failures as non-fatal: they print a warning and exit 0, so a transient network outage cannot block a session from starting or continuing.

## Test plan

- `scripts/test_post_tool_use_fetch.sh` (new, 4 cases): verifies the hook passes the correct arguments to git, exits 0 on success, exits 0 on fetch failure, and prints a warning on failure.
- No automated test for the `session-start.sh` step (consistent with the existing untested steps in that script, which depend on a full Android/remote environment).
- [ ] Manual: open a new Claude Code for Web session and confirm "[session-start] Step 4: fetch complete" appears in the hook output.

### DIFF
--- a/.claude/hooks/post-tool-use-fetch.sh
+++ b/.claude/hooks/post-tool-use-fetch.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# GB4PC -- Claude Code PostToolUse hook: git fetch after Agent tool calls.
+#
+# Configured in .claude/settings.json under PostToolUse with a matcher on
+# tool_name == "Agent".  Runs once each time a sub-agent finishes, so the
+# parent session sees any commits the sub-agent pushed before it exited.
+#
+# Failures are non-fatal: a warning is printed but the exit code is always 0
+# so the session continues even if the network is temporarily unavailable.
+
+set -uo pipefail
+
+# Resolve the repo root from the script's own path so this hook works in
+# worktrees as well as the main checkout.
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+git -C "$REPO_ROOT" fetch --prune --quiet \
+    && echo "[post-tool-use] git fetch complete" \
+    || echo "[post-tool-use] warning: git fetch failed (offline? -- continuing)"
+
+exit 0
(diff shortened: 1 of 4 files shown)


aunger (commented):
LGTM. The implementation is correct and complete.

**PostToolUse hook (`post-tool-use-fetch.sh`)**
- `set -uo pipefail` (without `-e`) combined with the `|| echo [warning]` pattern is the right choice here: on fetch failure the warning is printed but the script continues to `exit 0`, ensuring Claude's sub-agent cycle is never interrupted.
- `REPO_ROOT` resolution via `${BASH_SOURCE[0]}/../..` is consistent with the project's other hooks.
- Explicit `exit 0` at end is correct for a non-fatal hook.

**settings.json**
- `"matcher": "Agent"` is the correct value to match the Agent tool in a PostToolUse hook. Claude Code's PostToolUse hooks match against the tool name, and the sub-agent spawning tool is named "Agent".

**session-start.sh**
- Step 4 is placed after the `CLAUDE_CODE_REMOTE` guard, so the git fetch only runs in web sessions, matching the intent described in the issue.
- The `set -euo pipefail` + `||` chain in session-start.sh is also safe: the `-e` flag does not abort when the failing command is the left-hand side of a `||` expression.
- The cosmetic change (separator lines from 77 to 79 Unicode box-drawing characters) is harmless noise unrelated to the feature.

**Tests (`test_post_tool_use_fetch.sh`)**
- All 4 test cases pass: correct git args, exit 0 on success, exit 0 on fetch failure, and warning printed on failure.
- The fake-git-on-PATH technique with `GIT_ARGS_FILE` for arg capture is a clean approach.

The CI "No blocking labels" failure is expected during an active review cycle and is not a concern.

No changes requested.

PR #530, opened 2026-06-28
Accumulate CI run history in sticky PR comment with pass/fail/skip labels

### LINKED ISSUE #528: Surface build/test failures to Author
Sometimes Authors seem not to notice build failures when a PR is returned to them and repeatedly insist they're done.

Modify the `scripts/post_pr_ci_summary_link.py` script, which already auto-posts comments about CI runs, so it calls better attention to failures.

In addition to linking to a run summary, the script should indicate the Build &amp; Test job's pass/fail/skip result.

Additionally, the script should append to the comment's information instead of replacing it. Replace the single "View the build-and-test summary for this PR" link with an unordered list of "{job name} {run number} {result}" links.

### Goal

Turn the sticky CI-summary comment from a single replace-in-place link into an accumulating, unordered list where each line records one CI run for the PR as a link with the form `{job name} {run number} {result}`, so an Author scanning the comment sees the latest result and the history of prior results, including whether the latest run passed, failed, or was skipped.

- "Append" means: each script invocation adds one new list item for the current run while preserving the list items already present in the comment. The comment stays a single sticky comment (the `MARKER` mechanism is kept); we do not post a fresh comment per push.
- "{job name}" is the value of `JOB_NAME` (default `build-and-test`). "{run number}" is the human-facing run number (`GITHUB_RUN_NUMBER`), not the internal `GITHUB_RUN_ID`. "{result}" is the Build &amp; Test job's conclusion mapped to one of pass / fail / skip.
- The link target stays the existing summary anchor URL (`.../actions/runs/<run_id>#summary-<job_id>`) when a summary was written, falling back to the bare run URL otherwise. Each list item is itself a link.
- Exit code stays 0 in all paths (display only; this must never fail the build), and all network failures stay non-fatal as today.
- The script is invoked from the `link-pr-to-ci-summary` job in `.github/workflows/build.yml`, which already runs with `if: always()` so it fires on red builds too.

### Where the data comes from

- Job conclusion: the REST Jobs API call already made by `find_job_id` (`GET /repos/{repo}/actions/runs/{run_id}/jobs`) returns, for each job, both `id` and `conclusion` (one of `success`, `failure`, `skipped`, `cancelled`, `timed_out`, `action_required`, `neutral`, or `null` while still running). We will reuse the same response to read the matching job's `conclusion` instead of making a second API call.
- Run number: add `GITHUB_RUN_NUMBER` to the env block of the `link-pr-to-ci-summary` step in `build.yml` and read it in the script. (Currently only `GITHUB_RUN_ID` is passed; the run number is a separate value and must be wired through.)
- Existing list items: re-read from the body of the existing sticky comment, which `find_existing_comment` already locates. We need the comment's `body`, not just its `id`, so the prior items can be preserved.

### Result detection and mapping

Map the matching job's `conclusion` to a displayed result:

- `success` -> "pass"
- `failure`, `timed_out`, `cancelled` -> "fail"  (treat timed-out and cancelled as fail-equivalent so an Author is not lulled by a non-green run; alternatively cancelled could be its own label, see Trade-offs)
- `skipped` -> "skip"  (also the natural label for docs-only PRs where `SUMMARY_WRITTEN` is false)
- `null` / missing / unknown -> "unknown" (the job row could not be found or the job had not concluded when the linker ran)

Because the `link-pr-to-ci-summary` job `needs: [build-and-test]` and runs after it, the build-and-test job will normally have concluded by the time we query, so `null` should be rare; we still handle it.

Note on `SUMMARY_WRITTEN` vs `conclusion`: `SUMMARY_WRITTEN` (the job's `needs_full_build` output) tells us whether a pass/fail summary table exists to anchor-link to; it does not by itself tell us pass/fail. The job `conclusion` is the authoritative pass/fail/skip signal. A docs-only PR has `SUMMARY_WRITTEN=false` but the job itself still concludes `success`; we should show that as "pass" (or "skip" if we prefer to convey that no full build ran). Decide this explicitly (see Open questions).

### New comment body format

```
<!-- gb4pc-ci-summary-link -->
### CI test summary

- [build-and-test 1234 pass](https://github.com/owner/repo/actions/runs/<run_id>#summary-<job_id>)
- [build-and-test 1235 fail](https://github.com/owner/repo/actions/runs/<run_id2>#summary-<job_id2>)
```

Each line is `- [{job_name} {run_number} {result}]({url})`. Newest entry appended at the bottom (chronological), or prepended at the top (newest-first); pick one and keep it consistent (see Open questions). The marker and the `### CI test summary` heading are preserved as the comment's fixed header.

### "Append" mechanics (preserve prior items, add one)

1. Locate the existing sticky comment via `find_existing_comment` but return its `body` as well as its `id` (change the helper, or add a companion that returns the full comment object).
2. Parse the existing body to recover the list items already present. Two viable approaches:
   - Regex-extract every Markdown list line matching `^- \[.*\]\(.*\)$` from the body. Simple and tolerant of the fixed header.
   - Keep a machine-readable record (e.g. the list of items serialized) inside the body. Not needed; the rendered list lines are themselves parseable, so regex extraction is sufficient and avoids a hidden data block.
3. Build the new list item for the current run.
4. Idempotency / reruns: if a list item for the same `run_number` already exists, replace that item in place rather than appending a duplicate. This makes the script safe to run more than once per run (e.g. a re-run of the `link-pr-to-ci-summary` job, or a workflow re-run that keeps the same run number) and keeps the result current if the conclusion changed between invocations. Key the dedup on run number (plus job name, since the same comment could in principle host multiple job names).
5. Concatenate header + preserved-and-updated items + (new item if not a replacement) and upsert.

### Specific code changes in `scripts/post_pr_ci_summary_link.py`

1. `find_job_id` -> generalize to `find_job(...)` that returns the matching job dict (or `None`), so callers can read both `id` and `conclusion` from one API response. Keep a thin `find_job_id` wrapper, or update `main` and tests to use the new function. Preserve the existing run-attempt filtering logic exactly.
2. Add a `result_label(conclusion: str | None) -> str` helper implementing the mapping above (returns "pass" / "fail" / "skip" / "unknown").
3. Add `parse_existing_items(body: str) -> list[str]` (or return structured tuples of `(job_name, run_number, result, url)`) that extracts prior list lines from a comment body. Returning structured tuples is preferable because it enables the dedup-by-run-number in step 4 and the rebuild in step 5; if returning raw strings, dedup must parse the run number back out of each line.
4. Rewrite `build_comment_body` -> `build_comment_body(items: list[Item]) -> str` that renders the marker, heading, and the unordered list. The old signature `(summary_url, summary_written)` goes away; the `summary_written=false` case becomes a normal item whose result is "skip"/"pass" and whose URL is the bare run URL.
5. Update `upsert_comment` (or `main`) so that, before building the body, it fetches the existing comment's body, parses prior items, merges the current run's item (append or replace-by-run-number), and renders. The cleanest split: have `find_existing_comment` return `(id, body)` or add `get_existing_comment`, and let `main` orchestrate parse -> merge -> render -> upsert. `upsert_comment` can stay focused on the POST/PATCH given a final body, but it currently calls `find_existing_comment` internally; refactor so the comment is fetched once and both its id and body are reused (avoids a duplicate list-comments API call).
6. In `main`: read `GITHUB_RUN_NUMBER`; always look up the job (we now need its `conclusion` even when `SUMMARY_WRITTEN` is false, since the result label is independent of whether a summary table exists); compute the URL exactly as today (anchor when summary written and job id known, else bare run URL); compute the result label from the job conclusion; build the item; merge with prior items; upsert.
7. Update the module docstring to describe the list/append behavior and the new `GITHUB_RUN_NUMBER` env var.

### Workflow change in `.github/workflows/build.yml`

Add `GITHUB_RUN_NUMBER: ${{ github.run_number }}` to the `env:` block of the "Post CI summary link on PR" step (the `link-pr-to-ci-summary` job). No permission changes are needed (`pull-requests: write` already present).

### Edge cases

- First run on a PR: no existing comment, so `parse_existing_items` returns `[]` and the list has exactly one item. (New POST path.)
- Re-run / second push: existing comment found; one new item appended (or the same-run-number item replaced).
- Missing job (job row not found in the API response): `find_job` returns `None`; URL falls back to bare run URL and result label is "unknown". Still post an item so the Author sees the run happened.
- API error listing jobs or comments: stay non-fatal (warn to stderr, exit 0), as today. If we cannot fetch the existing comment body, prefer to skip the append/merge and avoid clobbering history; decide whether to (a) post a single-item comment anyway via the existing-id PATCH (risk: wipes prior items) or (b) bail out leaving the prior comment intact. Recommended: if the body fetch fails, do not PATCH (leave the existing comment untouched) to avoid destroying history; only POST a fresh comment when we are confident none exists.
- Docs-only PR (`SUMMARY_WRITTEN=false`): item uses the bare run URL; result label per the decision in Open questions (likely "skip" or the real job conclusion).
- Conclusion still `null` (linker ran before build-and-test concluded): should not happen given `needs:`, but maps to "unknown".
- Cancelled run: the `link-pr-to-ci-summary` job has `if: ... && needs.build-and-test.result != 'cancelled'`, so a cancelled build-and-test will not run the linker at all; no item is added. Acceptable.

### Trade-offs and alternatives considered

- Parsing rendered Markdown list lines vs. embedding a hidden machine-readable block. Chose parsing the rendered lines: no hidden state, the comment is self-describing, and a regex over `- [..](..)` lines is robust against the fixed header. Risk: if a human edits the comment, parsing could misfire; mitigated by the strict line pattern and by keying merges on run number.
- Unbounded growth of the list. Over a long-lived PR with many pushes the list could grow large. Options: cap to the most recent N entries (e.g. 20) and drop the oldest, or keep all. Recommend capping to a sane N to keep the comment readable; surface as an Open question.
- Treating `cancelled`/`timed_out` as "fail" vs. distinct labels. Folding them into "fail" keeps the Author's attention on non-green runs with a simple vocabulary; distinct labels are more precise but add noise. Recommend folding into "fail" initially.
- Second API call for conclusion vs. reuse. Reusing the single Jobs API response (via `find_job`) avoids an extra call and keeps id and conclusion consistent.

### Testing and verification

- Update `scripts/test_post_pr_ci_summary_link.py`:
  - `find_job` returns the full dict incl. `conclusion`; keep run-attempt filtering tests.
  - `result_label` mapping table (success->pass, failure/timed_out/cancelled->fail, skipped->skip, null/unknown->unknown).
  - `parse_existing_items`: empty body -> `[]`; body with N list lines -> N items; ignores the header and non-list lines; tolerates trailing slash and missing items.
  - Merge/append: first run yields one item; second run yields two; same-run-number re-run replaces in place (still one item, updated result).
  - `build_comment_body` renders the marker, heading, and a `-`-prefixed list line per item with `{job} {run} {result}` link text.
  - `main`: posts a list item with the right result label for success/failure/skip; reads `GITHUB_RUN_NUMBER`; falls back to bare run URL when job id missing; does not clobber history when the existing-comment fetch fails (per the chosen non-fatal policy).
- Run `python3 -m unittest discover -s scripts -p "test_*.py" -v` locally (matches the CI "Run Python script unit tests" step).
- Manual verification on a throwaway PR: push once (one item, pass), push a no-op or a deliberate test failure (second item, fail), re-run the workflow (item updated in place, not duplicated).

### Open questions for the requester

1. Newest-first (prepend) or chronological (append at bottom)? Append-at-bottom reads as a log; prepend keeps the latest result visible without scrolling.
2. For docs-only PRs (`SUMMARY_WRITTEN=false`), should the result show the real job conclusion (usually "pass") or "skip" to convey that no full build ran?
3. Should the list be capped to the most recent N entries, and if so what N?
4. Should `cancelled`/`timed_out` get their own labels, or is folding them into "fail" acceptable?

aunger:
Fixes #528.

## What changed

The sticky CI-summary comment on PRs now accumulates a chronological list of CI run results rather than replacing a single link on each push.
Each line shows the job name, human-facing run number, and a clear pass/fail/skip label:

```
### CI test summary

- [build-and-test 1234 pass](https://github.com/.../actions/runs/...#summary-...)
- [build-and-test 1235 fail](https://github.com/.../actions/runs/...#summary-...)
```

An Author scanning the comment immediately sees the latest result and can review the history of prior runs without hunting through the Checks tab.

### Changes in `scripts/post_pr_ci_summary_link.py`

- **`find_job_id` -> `find_job`**: returns the full job dict (id + conclusion) from a single API call.
  A thin `find_job_id` wrapper is kept for backward compatibility.
- **`result_label(conclusion)`**: maps job conclusion to `"pass"`, `"fail"`, `"skip"`, or `"unknown"`.
  `failure`, `timed_out`, and `cancelled` all map to `"fail"` so Authors are not lulled by a non-green result.
  Docs-only PRs (`SUMMARY_WRITTEN=false`) always show `"skip"` to convey no full build ran.
- **`parse_existing_items(body)`**: regex-extracts prior list lines from the comment body and returns `CIItem` tuples, enabling in-place replacement and the cap logic.
- **`build_comment_body(items)`**: replaces the old single-link signature; renders the marker, heading, and a `-`-prefixed list line per item.
- **Merge/append in `main`**: reads `GITHUB_RUN_NUMBER`, fetches the job's conclusion (always, not only when `SUMMARY_WRITTEN=true`), builds `CIItem`, parses prior items, replaces-in-place or appends, caps to 20 entries, and upserts.
- Dedup: same `(job_name, run_number)` replaces in place, making the script safe to re-run within the same workflow run.
- Cap: oldest entries are dropped when the list exceeds 20 items.

### Changes in `.github/workflows/build.yml`

- Added `GITHUB_RUN_NUMBER: ${{ github.run_number }}` to the `env:` block of the "Post CI summary link on PR" step.

### Tests (`scripts/test_post_pr_ci_summary_link.py`)

Added 25 new tests covering:
- `result_label` for all conclusion values
- `parse_existing_items`: empty body, single and multiple items, ignores non-list lines, trailing whitespace, bare run URLs
- `build_comment_body`: marker, heading, list item format, ordering, fail/skip results
- `TestMergeLogic` (via `main`): first run, second push appends, same run number replaces, cap to MAX_ITEMS
- `TestMain`: pass/fail results, run number in output, docs-only skip, history-safe on fetch failure

Total: 55 tests, all passing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -622,6 +622,7 @@ jobs:
           GITHUB_REPOSITORY: ${{ github.repository }}
           GITHUB_SERVER_URL: ${{ github.server_url }}
           GITHUB_RUN_ID: ${{ github.run_id }}
+          GITHUB_RUN_NUMBER: ${{ github.run_number }}
           GITHUB_RUN_ATTEMPT: ${{ github.run_attempt }}
           WORKFLOW_RUN_PR_URL: ${{ github.event.pull_request.html_url }}
           SUMMARY_WRITTEN: ${{ needs.build-and-test.outputs.needs_full_build }}
--- a/scripts/post_pr_ci_summary_link.py
+++ b/scripts/post_pr_ci_summary_link.py
@@ -9,9 +9,14 @@ that job's summary anchor, e.g.:
 
     https://github.com/<owner>/<repo>/actions/runs/<run_id>#summary-<job_id>
 
-It looks up the numeric job ID via the REST Jobs API (the anchor embeds the
-job ID, not the run ID), matching on job name and the current run attempt so
-re-runs don't pick up a stale job from a previous attempt.
+The comment accumulates a chronological list of CI runs for the PR. Each
+invocation appends one new list item of the form:
+
+    - [build-and-test 1234 pass](https://.../#summary-<job_id>)
+
+If a list item for the same job name and run number already exists it is
+replaced in place, making the script safe to re-run within the same workflow
+run. The list is capped to the most recent 20 entries.
 
 The comment is "sticky": a hidden HTML marker lets subsequent runs on the same
 PR find and edit the existing comment in place rather than piling up a new one
@@ -25,14 +30,14 @@ Environment:
     GITHUB_REPOSITORY     "owner/repo" (required).
     GITHUB_SERVER_URL     Server base URL, e.g. "https://github.com".
     GITHUB_RUN_ID         Workflow run ID.
+    GITHUB_RUN_NUMBER     Human-facing run number shown in the GitHub UI.
     GITHUB_RUN_ATTEMPT    Run attempt number (disambiguates re-runs).
     WORKFLOW_RUN_PR_URL   The triggering PR's html_url (empty on a plain push).
     SUMMARY_WRITTEN       "true" when build-and-test actually wrote the
                           pass/fail summary (its needs_full_build output).
-                          Docs-only PRs skip that step, so the job's
-                          "Summary" section has nothing to link to, and the
-                          comment says so instead of posting a misleading
-                          link.
+                          Docs-only PRs skip that step, so no summary table
+                          exists; these runs show result "skip" and link to
+                          the bare run URL.
     JOB_NAME              Name of the job whose summary to link (default
                           "build-and-test").
 
@@ -42,12 +47,29 @@ Exit code is always 0 (display only; a missing link must never fail the build).
 import os
 import re
 import sys
+from typing import NamedTuple
 
 import requests
 
 MARKER = "<!-- gb4pc-ci-summary-link -->"
 
 DEFAULT_JOB_NAME = "build-and-test"
+
+# Maximum number of list items to keep in the comment (oldest are dropped).
+MAX_ITEMS = 20
+
+# Regex matching a single Markdown list item as written by build_comment_body.
+# Captures: job_name, run_number (str), result, url.
+_ITEM_RE = re.compile(r"^- \[([^\s\]]+) (\d+) ([^\]]+)\]\(([^)]+)\)\s*$")
+
+
+class CIItem(NamedTuple):
+    """One CI-run entry in the sticky comment list."""
+
+    job_name: str
+    run_number: str  # kept as str to avoid integer-parsing surprises
+    result: str
+    url: str
 
 
 def _github_headers(token: str) -> dict[str, str]:
@@ -113,18 +135,46 @@ def find_job_id(
     return str(job["id"]) if job is not None else None
 
 
-def build_comment_body(summary_url: str, summary_written: bool) -> str:
-    if summary_written:
-        link_text = "View the build-and-test summary for this PR"
-    else:
-        link_text = "View this PR's build-and-test run"
-    body = f"{MARKER}\n### CI test summary\n\n[{link_text}]({summary_url}).\n"
-    if not summary_written:
-        body += (
-            "\n(This PR did not need a full build, so no pass/fail summary "
-            "was written; the link above goes to the run itself.)\n"
-        )
-    return body
+def result_label(conclusion: str | None) -> str:
+    """Map a GitHub job conclusion to a short human-readable result label.
+
+    Returns one of: "pass", "fail", "skip", "unknown".
+    """
+    if conclusion == "success":
+        return "pass"
+    if conclusion in ("failure", "timed_out", "cancelled"):
+        return "fail"
+    if conclusion == "skipped":
+        return "skip"
+    return "unknown"
+
+
+def parse_existing_items(body: str) -> list[CIItem]:
+    """Extract CI run list items already present in a comment body.
+
+    Returns a list of CIItem tuples in the order they appear in the body,
+    skipping any lines that do not match the expected list-item pattern.
+    """
+    items: list[CIItem] = []
+    for line in body.splitlines():
+        m = _ITEM_RE.match(line)
+        if m:
+            items.append(
+                CIItem(
+                    job_name=m.group(1),
+                    run_number=m.group(2),
+                    result=m.group(3),
+                    url=m.group(4),
+                )
+            )
+    return items
+
+
+def build_comment_body(items: list[CIItem]) -> str:
+    """Render the sticky comment body from a list of CIItem entries."""
+    lines = [f"- [{item.job_name} {item.run_number} {item.result}]({item.url})" for item in items]
+    list_block = "\n".join(lines)
+    return f"{MARKER}\n### CI test summary\n\n{list_block}\n"
 
 
 def find_existing_comment(
@@ -189,6 +239,7 @@ def main(argv: list[str] | None = None) -> int:
     repository = os.environ.get("GITHUB_REPOSITORY", "")
     server_url = os.environ.get("GITHUB_SERVER_URL", "https://github.com")
     run_id = os.environ.get("GITHUB_RUN_ID", "")
+    run_number = os.environ.get("GITHUB_RUN_NUMBER", "")
     run_attempt = os.environ.get("GITHUB_RUN_ATTEMPT", "")
     pr_url = os.environ.get("WORKFLOW_RUN_PR_URL", "")
     summary_written = os.environ.get("SUMMARY_WRITTEN", "") == "true"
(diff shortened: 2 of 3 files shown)


aunger (commented):
🤖 Reviewer

The implementation is solid overall. The core logic--accumulate a list, dedup by `(job_name, run_number)`, cap at 20, wire `GITHUB_RUN_NUMBER` through the workflow--matches the plan precisely, and all 55 tests pass locally. I have one substantive concern and a few smaller ones.

---

## Concern 1: "fetch failed" path comment is misleading, and the behavior deviates from the plan's recommendation

**Lines 317-325 of `post_pr_ci_summary_link.py`:**

```python
# When existing_body fetch failed (existing_id is None but we know a prior
# comment exists): prefer not to post a fresh comment that would skip history.
# However we cannot distinguish "no prior comment" from "fetch failed" when
# existing_id is None and existing_body is None. The safe path: if
# existing_id is None we either POST (first run) or POST again (fetch
# failed). The latter risks a duplicate comment, but preserves history in
# the existing comment and only adds a new one. This matches the recommended
# policy: do not PATCH without the existing id.
```

This comment says "preserves history in the existing comment and only adds a new one," which is technically true in the sense that the old comment's bytes are not deleted--but it is misleading. What actually happens when the fetch fails:

1. `existing_body` is `None`, so `prior_items = []`.
2. `merged` contains only `current_item` (the new run).
3. A POST is made with a body listing only the new run--no prior history.

So the sticky comment that GitHub users see is a brand-new comment with a one-item list. The old comment (with history) becomes an orphan somewhere above it in the thread. Future runs will find the new comment (it has the MARKER) and build on it, so the old history is permanently detached.

The issue plan says: "if the body fetch fails, do not PATCH (leave the existing comment untouched) to avoid destroying history; only POST a fresh comment when we are confident none exists." The current code POSTs regardless, and cannot be "confident none exists" when the API call failed. This violates the plan's recommendation.

The cleaner behavior: when `find_existing_comment` returns `(None, None)` due to an API error, the script should bail out and skip the upsert rather than post a new comment that strips history. But that requires knowing whether `(None, None)` means "no comment exists" or "fetch failed"--which the current return signature cannot distinguish.

This is a real design gap. The signature `tuple[int, str] | tuple[None, None]` conflates two distinct outcomes. Fixing it does not require a large change: have `find_existing_comment` return a third case (e.g. a sentinel or a three-tuple with an error flag, or raise and catch) so `main` can distinguish. Alternatively, the warning print already goes to stderr--the caller could check whether the API error path was taken by using an out-of-band flag.

---

## Concern 2: `build_comment_body([])` produces a syntactically weird but valid comment

If `merged` were somehow empty (not reachable through normal paths, but defensively), `build_comment_body([])` produces:

```
<!-- gb4pc-ci-summary-link -->\n### CI test summary\n\n\n
```

That is an empty-list comment that looks broken. A guard (`assert items` or an early return from `main` if `merged` is empty) would make the invariant explicit. Minor, but makes the contract clearer.

---

## Nit: `_ITEM_RE` result group allows spaces

`([^\]]+)` matches any characters except `]`, including spaces. If a future result label were ever multi-word (it currently cannot be), the regex would parse it as a single space-containing result token, and round-trip dedup would still work since it keys on `(job_name, run_number)`. This is benign now but slightly fragile. Using `(\S+)` (no whitespace in result) would make the intent explicit and match the four single-word labels actually produced.

---

## Positive notes

- `find_job` correctly reuses the single Jobs API response for both `id` and `conclusion`--no extra call.
- The merge/dedup loop is clear and correct.
- The cap logic (`merged[-MAX_ITEMS:]`) is correct and keeps the most recent entries.
- Test coverage is thorough: all conclusion mappings, parse round-trips, append/replace/cap scenarios.
- The workflow change is minimal and correctly placed.

---

The fetch-failure ambiguity (Concern 1) is a real behavioral deviation from the plan's recommendation and should be addressed before merge.

aunger:
🤖 Author

Thanks for the thorough review. All three concerns addressed in commit 3eccf20:

**Concern 1 -- fetch-failure ambiguity**

Introduced `CommentLookup(fetch_ok, comment_id, body)` NamedTuple to replace the bare `(id, body)` tuple. `find_existing_comment` now returns one of three distinct states:
- Found: `fetch_ok=True, comment_id=<int>, body=<str>`
- Not found: `fetch_ok=True, comment_id=None, body=None`
- API error: `fetch_ok=False, comment_id=None, body=None`

`main` checks `lookup.fetch_ok` immediately after the call; if `False` it prints a warning and returns 0 without calling `upsert_comment`. This closes the orphaning hole you identified.

**Concern 2 -- empty-items guard in `build_comment_body`**

Added `if not items: raise ValueError(...)` at the top of the function. The merge logic in `main` always produces at least one item (either appending or replacing), so in practice the guard should never trigger, but having it makes the contract explicit and prevents silent broken output if the invariant is ever violated.

**Nit -- regex result group**

Changed `([^\]]+)` to `(\S+)` in `_ITEM_RE`. Clearer and strictly correct: only a single non-whitespace token should match as the result label.

Tests updated to use `CommentLookup(...)` everywhere, with two new tests: one for `fetch_ok=False` early-exit and one for the `ValueError` guard.

aunger (commented):
🤖 Reviewer

All three concerns from the previous round have been addressed correctly in commit 3eccf20.

**Concern 1 (fetch-failure ambiguity):** The `CommentLookup(fetch_ok, comment_id, body)` NamedTuple cleanly distinguishes the three states. `main` now bails out immediately on `fetch_ok=False` without calling `upsert_comment`, which is exactly the right behavior. The misleading comment block is gone and replaced with a clear explanation.

**Concern 2 (empty-items guard):** `build_comment_body` now raises `ValueError` on an empty list. The invariant is explicit and the function's contract is clear.

**Nit (regex result group):** Changed to `\S+`--correct and tight.

The new `test_does_not_clobber_history_when_fetch_fails` test directly asserts that `upsert_comment` is not called when `fetch_ok=False`, and the new `test_raises_on_empty_items` test covers the guard. 56 tests pass.

LGTM.

PR #535, opened 2026-06-28
Capture E2E test video and retain only on failure

### LINKED ISSUE #520: Capture video during e2e tests
Capture video during e2e tests
- Only retain as an artifact for test cases that fail, otherwise delete
- See lessons-learned from the proof of concept: #495

aunger:
Fixes #520

## What this changes

`.github/workflows/build.yml` is modified. Each E2E test step
(`Run PixelCameraOverlayE2ETest`, `Run GalleryButtonVisualE2ETest`) now
records a `screenrecord` video of the emulator display during the test run.
Two new pull-and-upload step pairs are added (one per suite), each conditional
on the corresponding test step reporting `failure`.
On a green run no video is pulled or uploaded; the file is left on the device
and discarded when the emulator exits.

## Key design decisions

**Start only after the display is on.**
`screenrecord` started on a dark display exits immediately with
`INVALID_LAYER_STACK` and writes a 0-byte file.
The recording is launched after the existing `KEYCODE_WAKEUP` + `sleep 2`
sequence, so the display is guaranteed to be on when recording starts.

**Stop with SIGINT, not SIGTERM.**
Android's `screenrecord` only writes the `moov` atom (required for a playable
MP4) on its clean SIGINT path; SIGTERM is not handled and leaves the file
unplayable (no `moov` box).
The stop command is `pkill -INT -x screenrecord` (exact name match via `-x` to
avoid partial-name-match pitfalls on toybox).

**Wait for the device-side process to exit before pulling.**
After sending SIGINT, a `pidof` poll loop waits up to 30 s for the device-side
`screenrecord` process to exit, confirming the muxer has finished flushing
before `adb pull` runs.
The host-side `adb shell` wrapper is then reaped with `wait` rather than
`kill`, so the adb transport is never torn down while the muxer is still
writing.

These three rules are the lessons learned from the proof-of-concept in #495,
which debugged each failure mode across seven review rounds with real CI runs.

## Artifact naming

- `e2e-overlay-video` -- video for the `PixelCameraOverlayE2ETest` suite
- `e2e-gallery-video` -- video for the `GalleryButtonVisualE2ETest` suite

Both use `retention-days: 14`, consistent with every other artifact in this
workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -380,6 +380,16 @@ jobs:
           "$ADB" shell input keyevent KEYCODE_WAKEUP
           "$ADB" shell input swipe 300 1000 300 300
           sleep 2
+          # Start screen recording for the overlay E2E suite (issue #520).
+          # screenrecord must start AFTER KEYCODE_WAKEUP: if the display is off the
+          # process exits immediately with INVALID_LAYER_STACK and the output is 0 bytes.
+          # The moov atom (required for a playable MP4) is only written on a clean SIGINT
+          # exit; SIGTERM is not handled by screenrecord and leaves the file unplayable.
+          # Do NOT kill the host-side adb shell wrapper ($RECPID_OVERLAY): tearing down
+          # the adb transport races the muxer flush and can truncate the moov atom.
+          "$ADB" shell screenrecord /sdcard/e2e-overlay.mp4 &
+          RECPID_OVERLAY=$!
+          sleep 1
           mkdir -p results
           set -o pipefail
           LOGCAT_SINCE=$("$ADB" shell "date '+%m-%d %H:%M:%S.000'" | tr -d '\r')
@@ -390,6 +400,19 @@ jobs:
           else
             echo "outcome=failure" >> "$GITHUB_OUTPUT"
           fi
+          # Stop screenrecord cleanly so the moov atom is written and the MP4 is playable.
+          # pkill -INT -x screenrecord sends SIGINT to the device-side process by exact
+          # name (-x prevents partial-name matches on toybox). Poll with pidof until the
+          # process exits, confirming the muxer finished flushing, before reaping the
+          # host-side wrapper with wait. A fixed sleep is not sufficient.
+          echo "  stopping screenrecord (overlay suite)"
+          "$ADB" shell pkill -INT -x screenrecord 2>/dev/null || true
+          echo "  waiting for screenrecord to finalize the MP4"
+          for i in $(seq 1 30); do
+            "$ADB" shell pidof screenrecord > /dev/null 2>&1 || break
+            sleep 1
+          done
+          wait "$RECPID_OVERLAY" 2>/dev/null || true
           # Capture the filtered logcat regardless of pass/fail (issues #364/#365).
           # The GB4PC_E2E diagnostic lines from E2EFixture.launchPixelCamera() are
           # what make the issue #233 bounded-relaunch path observable, but they are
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Author

The change is a single commit to `.github/workflows/build.yml`.
Each E2E test step already wakes the display before running; the recording is started immediately after that wake sequence, so the display is guaranteed to be on when `screenrecord` launches.

Three rules from the #495 proof-of-concept are applied directly:

1. **Display-on-first** -- `screenrecord` must not be started before `KEYCODE_WAKEUP`; a dark display causes an immediate `INVALID_LAYER_STACK` exit and a 0-byte output file.
2. **SIGINT-only stop** -- `pkill -INT -x screenrecord` sends SIGINT with an exact name match. SIGTERM is not handled by screenrecord's signal handler; only SIGINT triggers `AMediaMuxer_stop()` and writes the moov atom that makes the MP4 playable.
3. **Wait, do not kill** -- a `pidof` poll loop waits up to 30 s for the device-side process to exit after SIGINT, confirming the muxer has flushed. The host-side `adb shell` wrapper is then reaped with `wait` rather than killed; tearing down the adb transport while the muxer is still writing races the moov flush and can truncate it.

The pull-and-upload steps are gated on `steps.run_*_e2e.outputs.outcome == 'failure'`, so no video artifact is produced on a green run.

aunger:
🤖 Reviewer

Reviewed against issue #520 and the lessons-learned record from #495.

## All requirements met

**Issue #520 requirements:**

- Video is captured for each E2E suite (`PixelCameraOverlayE2ETest` and `GalleryButtonVisualE2ETest`).
- Pull and upload are gated on `steps.run_*_e2e.outputs.outcome == 'failure'`, both wrapped in `always()` so they run even when the test step itself exited non-zero.
  On a green run nothing is pulled or uploaded; the file is left on the device and discarded with the emulator.
- Artifact names (`e2e-overlay-video`, `e2e-gallery-video`) and `retention-days: 14` are consistent with every other artifact in the workflow.

**Three lessons from #495, all applied correctly:**

1. **Display-on-first.** `screenrecord` is launched after the existing `KEYCODE_WAKEUP` + `sleep 2` sequence in each E2E step. The display is guaranteed to be on, avoiding the `INVALID_LAYER_STACK` / 0-byte failure mode.
2. **SIGINT-only stop.** `pkill -INT -x screenrecord` sends SIGINT with an exact process-name match via `-x`. SIGTERM is not used anywhere in the stop path.
3. **Wait, do not kill.** The `pidof` poll loop (up to 30 s) waits for the device-side process to exit before `adb pull` runs. The host-side `adb shell` wrapper (`$RECPID_OVERLAY` / `$RECPID_GALLERY`) is reaped with `wait`, not killed. The adb transport stays alive throughout the muxer flush.

**Conditions checked:**

- `steps.run_overlay_e2e.outputs.outcome` (custom output via `$GITHUB_OUTPUT`) is the correct reference, consistent with how `Gate on test failures` uses the same pattern (issue #309 design).
- `always()` on the pull and upload steps correctly overrides GitHub Actions' default of skipping steps after a failed predecessor.

## Nits (non-blocking)

**`screenrecord` 3-minute default time limit.** Without `--time-limit`, Android's `screenrecord` self-stops after 180 s. If a suite runs longer the recording covers only the first 3 minutes; the file is still properly finalized because the process exits cleanly on its own timer. The stop sequence harmlessly no-ops in that case. For failure debugging this is acceptable--the early part of a long run is typically what is most useful--but it is worth knowing the cap exists. A follow-up could add `--time-limit 180` explicitly to make the ceiling visible in the source.

**`screenrecord` default resolution.** No `--size` flag is passed, so the recording uses the emulator's native display resolution (pixel_6 AVD: 1080x2400). The resulting files will be large if a suite runs close to the time limit. Not a correctness issue.

**Pre-existing asymmetry (out of scope for this PR).** The overlay suite captures logcat unconditionally; the gallery suite captures it only on failure. This asymmetry predates this PR and is not introduced by it.

No blocking issues found.

PR #539, opened 2026-06-29
Add test5a_secureCameraLockedPopulatedGalleryShowsGreen to flaky test allowlist

### LINKED ISSUE #538: Add test5a_secureCameraLockedPopulatedGalleryShowsGreen to accepted flaky test list in CI
Add `test5a_secureCameraLockedPopulatedGalleryShowsGreen` to the accepted flaky test list in CI so that known-flaky failures on this test do not block PRs.

See #243 for context on the accepted flaky test list.

aunger:
Fixes #538

`test5a_secureCameraLockedPopulatedGalleryShowsGreen` has been failing on CI repeatedly since May 2026 (tracked in #243).
Adding it to `.github/allowed-test-failures.txt` prevents these known-flaky failures from blocking unrelated PRs.

**Prior fix attempts and why flakiness persists:**

- PR #527 (merged 2026-06-28) added a `waitForSessionMedia()` poll to address a MediaStore-row-timeout race.
  This changed the failure mode: pre-#527 failures reported a MediaStore timeout in `captureOnePhoto()`; post-#527 failures report a pixel-geometry assertion (green region 0% width, 0% coverage in `SecureViewerActivity`).
  The test is still flaky after #527.

- PR #532 (merged 2026-06-28) fixed a missing `READ_MEDIA_IMAGES`/`READ_EXTERNAL_STORAGE` permission in the production manifest.
  However, #532's own commit message notes that this bug was "invisible in E2E tests because `build.gradle.kts` grants `READ_MEDIA_IMAGES` to the test package via `pm grant`."
  The E2E failures are therefore unaffected by #532; the pixel-geometry flakiness remains.

The remaining root cause (why the overlay tap over the keyguard occasionally fails to composite or is not tappable) is still unresolved and continues to be tracked in #243.

**Allowlist entry format:**

The entry uses the method-level `ClassName#methodName` format, so only this one test is tolerated, not the whole class.
The tracking issue (#243) is cited in an inline comment per the file's own conventions.
This entry should be removed from the allowlist once #243 is resolved.

🤖 Author

### DIFF
--- a/.github/allowed-test-failures.txt
+++ b/.github/allowed-test-failures.txt
@@ -20,4 +20,5 @@
 # issue #309 removed it.  Each already has a tracking issue and must be
 # removed from this list as it is fixed.
 
-# (No entries.)
+# Tracking issue: #243
+com.gb4pc.e2e.GalleryButtonVisualE2ETest#test5a_secureCameraLockedPopulatedGalleryShowsGreen


aunger:
🤖 Reviewer

The change itself is technically correct: the entry uses the right `ClassName#methodName` format, the tracking comment is on its own `#`-prefixed line (ignored by the parser per the file's spec), and citing `#243` follows the file's stated conventions.
The `# (No entries.)` placeholder is correctly removed when a real entry is added.

However, there is a factual problem in the PR description, and a related question about whether this PR is still needed at all.

**The situation on `main` between when #243 was filed and when this PR was opened:**

- PR #527 (merged 2026-06-28 11:31 UTC) added a `waitForSessionMedia()` poll to `test5a` between `captureOnePhoto()` and `tapOverlay()`, intending to fix the race.
  The owner commented "fixed by #527" on #243 at 11:36 UTC.
- Despite that, CI reported `test5a` failures at 11:46, 16:44, 22:50, and 00:07 UTC on 2026-06-28/29 -- all after #527 landed.
- PR #532 (merged 2026-06-28 23:43 UTC) then identified a completely different root cause -- missing `READ_MEDIA_IMAGES` / `READ_EXTERNAL_STORAGE` permissions in the production manifest -- and also claims to fix #509.
- This PR was opened at 00:30 UTC on 2026-06-29, only 47 minutes after #532 landed.

That timeline raises two concerns:

1. **Is `test5a` still flaky on the current `main` tip (after #532)?**
   If #532 resolved it, adding it to the allowlist is actively harmful: the entry will swallow any future real regression of this test silently.
   If #532 did not resolve it (plausible -- the post-#527 failures show a different failure message involving green-region pixel geometry, suggesting the root cause is not fully understood), then the allowlist entry is appropriate.

2. **The PR description says "The fix for this test's root cause is tracked in #243."**
   Issue #243 is a CI-filed symptom tracker, not a fix issue.
   Two attempted fixes (#527, #532) have already been merged against it, and the test was still failing after both.
   Pointing readers at #243 as the place where the fix is tracked is accurate only if #243 is actively being used to track the fix effort -- but that issue has no explicit "root cause identified, fix in progress" note, just a stream of CI failure reports.
   The description should at minimum acknowledge that #527 and #532 have already attempted fixes and whether those are believed to have resolved the flakiness.

**Requested change:** Before merging, determine whether `test5a` is still flaky after #532.
If it is, update the PR description to reflect that earlier fix attempts did not fully resolve it, so reviewers understand this is not a redundant entry.
If it is not, close this PR -- a healthy test should not be in the allowlist.

aunger:
🤖 Author

The Reviewer's concern is reasonable but the allowlist entry is still warranted.
Here is why.

**#532 did not address the post-#527 E2E failure mode.**

#532's commit message is explicit: "the bug was invisible in E2E tests because `build.gradle.kts` grants `READ_MEDIA_IMAGES` to the test package via `pm grant`--but that grant never applied to the production app itself."
By that analysis, the missing-permission bug never caused E2E test failures.
The pre-#527 failures (MediaStore row timeout, `captureOnePhoto`) and the post-#527 failures (pixel geometry: green region 0% width, 0% coverage) are two distinct failure modes.
#532 fixed the production-app permission gap; it did not fix the pixel-geometry assertion flakiness that has dominated the failure reports since #527 merged.

**The last known failure post-dates #532.**

The comment on #243 at 00:07 UTC 2026-06-29 (branch `fix/issue-520-ci`, commit `cddc40c`) was filed at 00:07 UTC.
#532 merged at 23:43 UTC 2026-06-28.
That `fix/issue-520-ci` branch was based on a pre-#532 tip, so this failure does not confirm the test is still flaky on the current main tip.
However, absence of evidence is not evidence of absence: no CI run on the post-#532 main tip has completed and filed a passing result for `test5a` either.

**The PR description needs updating.**

The Reviewer is right that saying "The fix for this test's root cause is tracked in #243" is inaccurate.
#243 is a symptom tracker, not a fix issue.
I am updating the PR description to acknowledge the two prior fix attempts (#527, #532), that neither fully resolved the pixel-geometry flakiness that shows up in post-#527 failures, and that #243 continues to track the remaining flakiness.

aunger:
🤖 Reviewer

The Author's response is convincing. The argument that #532 could not have fixed the E2E flakiness is solid and is supported by #532's own commit message, which states the permission bug "was invisible in E2E tests because `build.gradle.kts` grants `READ_MEDIA_IMAGES` to the test package via `pm grant`." That means the pixel-geometry failure mode seen in every post-#527 CI report has a different, still-unresolved root cause, and the allowlist entry is not redundant.

The updated PR description now accurately describes what #527 and #532 each did and why neither resolved the remaining E2E flakiness. My original concern about misleading language is resolved.

LGTM.

Issue #516, opened 2026-06-26
Impl plan: CI Monitor per-check status summary and attributed terminal (#514)


aunger:
Implementation plan for #514.

## Goal and constraints

End a CI Monitor run with a self-contained, per-check status summary block and an attributed terminal line (`Blocked by: ...`) that names the blocking check(s), so a consumer acting only on terminal output gets an actionable verdict.
Suppress the `drain poll found no new diagnostic signals` flag when a named check already concluded `failure` (the terminal is diagnosed).
Distinguish a process-label gate (e.g. `No blocking labels`) from a substantive code/test block.
Preserve the existing line-oriented contract: terminal tokens (`Clear`/`Blocked`/`Infra`) still appear and still end the loop; the summary is additive.

### Key facts grounding the design (from reading the code)

- The verdict reads `/commits/{sha}/check-runs` once per poll into `check_json` (`ci_monitor.py:710-713`), then `parse_check_result(check_json)` (`ci_monitor.py:261-280`) collapses it to a single token.
  The per-check `name` and `conclusion` fields are present in `check_json["check_runs"]` but are currently discarded.
- The check-run that fails for the label gate is the job named `No blocking labels` in `.github/workflows/block-merge-on-blocking-labels.yml` (its `name:` at line 26 becomes the check-run `name`).
  The string `Check for blocking labels` in the issue is the *step* name (line 29).
  GitHub's check-runs API exposes the check-run `name` (the job name) most reliably; the summary must key on the check-run `name` field.
  The label-gate detection must match the check-run `name` (`No blocking labels`), made configurable so the wording is not hardcoded.
- `check_json` is in lexical scope at every terminal-emit site in `main()` (assigned at `ci_monitor.py:710` inside the `while True` loop; all terminal branches are later in the same iteration).
  `parse_pr_terminal` short-circuits (Merged/Closed) *before* `check_json` is fetched, so those two terminals have no check-runs data; the summary applies only to `Clear`/`Blocked`/`Infra` verdict terminals.
- `drain_then_print(sha, terminal_line)` (`ci_monitor.py:647-674`) prints the `drain poll found no new diagnostic signals` line when no drain attempt emitted anything.
  This is the line to suppress (and where to inject the summary).
- The terminal-emit sites in `main()` are:
  - `all_passed` and `mergeable in ("clean","unstable")` -- `Clear (mergeable_state=%s)` (`ci_monitor.py:732-735`)
[...]

PR #548, opened 2026-07-01
Route label-gate CI terminals into the Verification Planner flow

### LINKED ISSUE #540: Orchestrator skips Verification Planner on label-gate-only CI terminal
## What happened

When the CI Monitor emits a `Blocked by: ... [label gate]` terminal (all blockers are label-gate checks, no code/test failures), the Orchestrator concluded orchestration entirely and did not dispatch the Verification Planner.

This happened on PR #536 (issue #533) and PR #535 (issue #520) in this session.

## How I read the routing rule

The relevant rule in `agents/dev_orchestration.md` reads:

> if Monitor emits a Blocked line where the terminal ends with `[label gate]` (the ` by: ...` suffix names only label-gate checks) -> clear the silentVanish re-launch flag; inform the user that only a process-label gate is blocking (not a code/test failure) and that removing the blocking label should unblock the merge; stop (do NOT route to a new Author round)

I read "stop" as stopping the entire orchestration cycle, including the Verification Planner step.
The parenthetical clarification "(do NOT route to a new Author round)" is scoped to the Author-routing loop, but I treated it as the only thing being stopped and still read the bare "stop" before it as a global halt.

The Verification Planner step in the rule above is gated on `if Monitor emits a Clear line`. A label-gate-only terminal is not a Clear line, so I concluded the Verification Planner trigger condition was never met.

## Why this is wrong

A label-gate terminal means the code is fine--the Reviewer gave LGTM and no tests failed. The merge is gated only by a process label, not by code quality. The spirit of the Verification Planner step is to surface before-merging requirements while the code is otherwise ready. Skipping it on a label-gate terminal leaves those requirements unchecked.

### Goal

Make the Orchestrator dispatch the Verification Planner when the CI Monitor emits a label-gate-only terminal (`Blocked by: ... [label gate]`), just as it does on a `Clear` terminal, instead of halting the whole cycle. A label-gate terminal means the code passed review and tests; only a process label blocks the merge, so the before-merging requirements must still be surfaced.

### Scope and nature of the change

This is a prose/process change confined to a single file: `agents/dev_orchestration.md`. There is no application code to change. The CI Monitor script (`scripts/ci_monitor/ci_monitor.py`) already emits the `[label gate]` terminal correctly and is covered by `scripts/test_ci_monitor.py`; the defect is purely in the Orchestrator routing prose that consumes that terminal. No change to the Monitor script or its tests is warranted or planned.

- The `[label gate]` terminal is authoritative and already correctly distinguishes a pure process-label block from a substantive code/test failure (the `[label gate]` suffix is emitted only when every blocking check is a label gate, per the `label_gate_check_regex` behavior documented in `scripts/ci_monitor/README.md`).
- On a label-gate terminal the Reviewer has already given LGTM (the Monitor loop is entered only under `if Reviewer gave LGTM`), so the code is review-approved.
- The Verification Planner and Verification Agent flows do not themselves require CI to be green; they operate on the review-approved diff and the issue/PR text. So it is safe to run them while a process label still gates the actual merge.
- The label-gate condition is separate from, and does not conflict with, the before-merging gate: the PR still may not merge until both the process label is cleared by a human and every before-merging tracking issue is resolved.
- The prose-style rules in `.claude/rules/prose-style.md` apply (split at sentence breaks; no em-dashes, angle quotes, or ellipses; double-hyphens without surrounding spaces).

### Root cause

In the "PR routing (Monitor loop)" pseudo-code in `agents/dev_orchestration.md`, the Verification Planner dispatch lives **inside** the `if Monitor emits a Clear line:` branch (currently around lines 261-286). The label-gate branch (currently around line 252) ends in `stop`, so a label-gate terminal never reaches the Verification Planner block. The parenthetical only forbids a new Author round; the bare `stop` was (reasonably) read as a global halt, and there is no path from the label-gate branch into the Planner block.

### Design: factor the Planner/Verifier flow into a shared sub-routine

The `Clear` branch and the label-gate branch must share the same downstream behavior (dispatch Verification Planner, then route on its result, then dispatch the Verification Agent, then route on its signal). Duplicating that multi-step block into two branches would violate Don't-Repeat-Yourself (`agents/code_edit.md`) and invite drift. Instead, extract the Planner/Verifier flow into a single labeled sub-routine and have both terminals jump to it.

Concrete steps:

1. **Extract a labeled block.** Introduce a labeled sub-routine, e.g. `surfaceBeforeMergingRequirements:`, containing the existing body currently nested under `if Monitor emits a Clear line:` from the comment "Step: Surface outstanding before-merging requirements" through the end of that branch (the Verification Planner dispatch, the empty-list `verified` transition, the non-empty-list `verification needed` transition and relay, the Verification Agent dispatch, and "Route on the Verification Agent's terminal signal per ..."). Place this labeled block adjacent to the existing `undiagnosedTerminal:` and `silentVanish:` labeled blocks so all sub-routines sit together.

2. **Point the `Clear` branch at the sub-routine.** Replace the inlined body under `if Monitor emits a Clear line:` with: clear the silentVanish re-launch flag, then `-> goto surfaceBeforeMergingRequirements`. Preserve the existing "clear the silentVanish re-launch flag" action that currently opens the Clear branch. Because the two entry paths differ in their final "verified / mergeable" wording, note which path invoked the sub-routine (Clear path vs. label-gate path) when entering it.

3. **Rewrite the label-gate branch to run the sub-routine instead of halting.** Change the branch currently ending in `stop (do NOT route to a new Author round)` so that it:
   - clears the silentVanish re-launch flag (unchanged),
   - informs the user that only a process-label gate is blocking (not a code/test failure), that the code is review-approved, and that removing the blocking label plus resolving any before-merging tracking issues will unblock the merge (revised wording so the user is not told the cycle is over),
   - does NOT route to a new Author round (unchanged intent),
   - then `-> goto surfaceBeforeMergingRequirements` instead of `stop`.

4. **Make explicit that the merge now has two independent gates on the label-gate path.** In the label-gate branch and/or the shared sub-routine's closing note, state that on the label-gate path the PR may be merged only once (a) a human removes the blocking process label and (b) every before-merging tracking issue the Planner filed is resolved. On the empty-list case the sub-routine must NOT apply `verified` when it was entered on the label-gate path (the process-label gate still blocks the merge); it applies `verified` only on the Clear path. This prevents a future reader from concluding that a green Verification Agent alone clears a label-gated PR.

5. **Reconcile the "Concluding PR orchestration" bookkeeping.** The label-gate branch previously ended the cycle at `stop`, at which point the Orchestrator would apply the "Concluding PR orchestration" label transition (remove `orchestrating`). Now the label-gate path continues into the Planner/Verifier sub-routine, so the conclusion (and the `orchestrating` removal) happens after that sub-routine returns, exactly as it already does on the `Clear` path. Confirm no wording implies the cycle concludes at the label-gate branch itself; adjust if the current phrasing suggests an immediate stop.

6. **Update the deviation note referencing issue #516 Step 8a (currently around lines 253-256).** That comment explains why the Orchestrator does not auto-remove the blocking label. It remains valid (the Orchestrator still does not remove the label; a human does). Verify it still reads correctly now that the branch continues into the Planner flow rather than stopping, and lightly adjust wording only if it implies the branch terminates the cycle.

### Critical files and how they interact

- `agents/dev_orchestration.md` -- the only file changed. Contains the "CI checking after a Reviewer exits (Monitor loop)" section with the "PR routing (Monitor loop)" pseudo-code that has the `Clear`, `Blocked`, label-gate, and `Infra` branches, plus the `undiagnosedTerminal:` and `silentVanish:` labeled sub-routines. The edit adds a `surfaceBeforeMergingRequirements:` sub-routine and redirects two branches to it.
- `agents/verification_planning.md` -- defines the Verification Planner's behavior (the dispatched sub-agent). Unchanged; the plan only changes when the Planner is dispatched, not what it does. Its empty-list report still drives the `verified` transition (Clear path only), and its non-empty-list report still drives `verification needed`.
- `agents/pr_verify.md` -- defines the Verification Agent's behavior. Unchanged; still dispatched by the shared sub-routine and still routed on its `Verification passed` / `Verification revealed an error` / `Verification incomplete` signals per the existing "Routing on the Verification Agent's signal" section.
- `scripts/ci_monitor/ci_monitor.py`, `scripts/ci_monitor/README.md`, and `scripts/test_ci_monitor.py` -- source and documentation of the `[label gate]` terminal. Unchanged. `README.md` was checked for consistency: its description of the label-gate terminal ("A `Blocked` terminal whose only blockers are label-gate checks warrants removing the blocking label rather than investigating a test failure") and its pointer to `agents/dev_orchestration.md` for "the Verification Planner dispatch" both remain accurate after this change and need no edit.

### Trade-offs and rejected alternatives

- **Rejected: duplicate the Planner/Verifier block into the label-gate branch.** Simpler diff, but two copies of a multi-step flow drift over time and violate Don't-Repeat-Yourself. Rejected in favor of one shared labeled sub-routine.
- **Rejected: have the Orchestrator auto-remove the blocking label and re-run the Monitor to reach a real `Clear`.** This is exactly the approach the existing #516 Step 8a deviation note rejected, because the blocking label is a user-controlled process label whose removal needs human judgment. Keeping the human in the loop for the label while still surfacing before-merging requirements is the correct split.
- **Rejected: gate the Verification Planner on "Clear OR label-gate" inline in the `Clear` branch condition.** Would entangle two distinct terminal shapes in one branch and obscure the label-gate-specific user message. A dedicated branch that jumps to a shared sub-routine is clearer.
- **Considered: also run the Planner on a bare `Blocked` (non-gate) terminal.** Rejected: a bare Blocked means a real code/test failure, the code is not ready, and the correct action remains routing to a new Author round. Only the label-gate terminal represents review-approved code blocked solely by process.

### Risks

- **Ambiguity risk (primary).** Because this is prose consumed by an LLM Orchestrator, the failure mode is misreading, exactly what caused this bug. Mitigation: use an explicit `goto` to a labeled sub-routine (matching the existing `undiagnosedTerminal`/`silentVanish` pattern the Orchestrator already follows), and remove the bare `stop` from the label-gate branch so there is no residual "halt" signal.
- **Double-gate confusion.** A reader might think a green Verification Agent clears a label-gated PR. Mitigated by step 4's explicit two-gate note and the path-conditional `verified` transition.
- **Label-bookkeeping regression.** If the `orchestrating` removal or `verification needed` / `verified` transitions are misplaced during the extraction, labels could be left in a wrong state. Mitigated by keeping every label transition inside the shared sub-routine (unchanged relative to the current `Clear` path except for making `verified` conditional on the Clear path) and by step 5's reconciliation check.

### Open decisions for the requester

1. **Label-gate + non-empty before-merging list wording.** On the label-gate path with outstanding before-merging items, the user should understand there are now two things to clear: the process label (human) and the tracking issues. Confirm the proposed message wording (step 3 / step 4) matches your intent, or supply exact wording.
2. **Sub-routine name.** `surfaceBeforeMergingRequirements` is proposed to mirror the existing "Step: Surface outstanding before-merging requirements" comment. Approve or rename.
3. **Scope confirmation.** Confirm the fix should be limited to the label-gate terminal only and should not change behavior for bare `Blocked` or `Infra` terminals.

### How to test / verify

There is no automated test harness for the Orchestrator prose itself; verification is by review of the routing document plus the existing script tests.

1. **Prose review.** Re-read the full "PR routing (Monitor loop)" section after the edit and confirm: (a) both the `Clear` branch and the label-gate branch reach the Verification Planner via `goto surfaceBeforeMergingRequirements`; (b) no bare `stop` remains on the label-gate branch; (c) the two-gate note is present and `verified` is applied only on the Clear path; (d) label transitions (`verified`, `verification needed`, `orchestrating` removal) occur exactly once per path and in the right place; (e) the #516 Step 8a deviation note still reads correctly.
2. **Trace both terminals by hand.** Walk a `Clear` terminal and a `Blocked by: ... [label gate]` terminal through the revised pseudo-code and confirm both converge on the Planner/Verifier flow, and that a bare `Blocked` still routes to a new Author round and `Infra` still escalates.
3. **Regression guard on the Monitor.** Run the existing script tests to confirm the terminal semantics the plan relies on are unchanged: `python3 scripts/test_ci_monitor.py` (or the repo's standard test runner). No new assertions are expected there since no script changed, but running them confirms the `[label gate]` terminal shape the plan depends on still holds.
4. **Prose-style check.** Confirm the edited lines follow `.claude/rules/prose-style.md` (sentence-break wrapping; no em-dashes, angle quotes, or ellipses).

## Implementation plan

The "## Implementation Plan" section above is the finalized, authoritative plan for this fix; this section confirms it after a fresh review of `agents/dev_orchestration.md` and `scripts/ci_monitor/README.md`. Summary of the agreed approach:

- Extract the Verification Planner plus Verification Agent flow (currently nested inside the `if Monitor emits a Clear line:` branch, roughly lines 261-286 of `agents/dev_orchestration.md`) into a shared labeled sub-routine `surfaceBeforeMergingRequirements:`, placed beside the existing `undiagnosedTerminal:` and `silentVanish:` blocks.
- Redirect the `Clear` branch to `goto surfaceBeforeMergingRequirements` (entered on the Clear path).
- Rewrite the label-gate branch (line 252) so its bare `stop` becomes `goto surfaceBeforeMergingRequirements` (entered on the label-gate path), keeping the "only a process-label gate is blocking" message and the "do NOT route to a new Author round" intent, and dropping the global-halt reading that caused this bug.
- Make `verified` conditional: apply it (to both issue and PR) only on the Clear path; on the label-gate path an empty before-merging list still leaves the merge gated by the process label, so `verified` is not applied and the user is told the merge unblocks once the label is removed.
- State the two independent merge gates on the label-gate path (human removes the process label AND every before-merging tracking issue is resolved).
- Leave `agents/verification_planning.md`, `agents/pr_verify.md`, `scripts/ci_monitor/ci_monitor.py`, `scripts/test_ci_monitor.py`, and `scripts/ci_monitor/README.md` unchanged; `README.md`'s label-gate wording and its pointer to `agents/dev_orchestration.md` remain accurate and need no edit.

Verification is by prose review of the revised Monitor-loop section (both terminals must converge on the shared sub-routine, no bare `stop` remains on the label-gate branch, `verified` is Clear-path-only) plus running `python3 scripts/test_ci_monitor.py` as a regression guard on the unchanged terminal semantics.

aunger:
Fixes #540

Previously, when the CI Monitor emitted a `Blocked by: ... [label gate]` terminal (all blockers are process-label gates, no code/test failures), the Orchestrator's routing rule ended in a bare `stop`. That read as halting the entire orchestration cycle, so the Verification Planner (gated only on a `Clear` line) was never dispatched, leaving before-merging requirements unchecked even though the code had passed review and CI.

This PR is a prose-only change to `agents/dev_orchestration.md`, per the finalized implementation plan in the issue:

- Extracts the Verification Planner + Verification Agent flow (previously nested inside the `if Monitor emits a Clear line:` branch) into a shared labeled sub-routine, `surfaceBeforeMergingRequirements:`, placed beside the existing `undiagnosedTerminal:` and `silentVanish:` sub-routines.
- Redirects the `Clear` branch to `goto surfaceBeforeMergingRequirements` (entered on the Clear path).
- Rewrites the label-gate branch so its bare `stop` becomes `goto surfaceBeforeMergingRequirements` (entered on the label-gate path) instead, keeping the "do NOT route to a new Author round" intent and dropping the global-halt reading that caused the bug.
- Makes the `verified` label conditional: it is applied (to both issue and PR) only on the Clear path. On the label-gate path, an empty before-merging list still leaves the merge gated by the process label, so `verified` is not applied; the user is told the merge unblocks once the label is removed.
- States the two independent merge gates on the label-gate path explicitly: a human must remove the blocking process label, AND every before-merging tracking issue the Planner filed must be resolved.

No application code changed. `scripts/ci_monitor/ci_monitor.py`, `scripts/ci_monitor/README.md`, `agents/verification_planning.md`, and `agents/pr_verify.md` are all unchanged, as scoped in the issue's implementation plan; the `[label gate]` terminal itself was already correct and is covered by `scripts/test_ci_monitor.py`.

## Test plan

- [x] Ran `python3 scripts/test_ci_monitor.py` as a regression guard confirming the terminal semantics this change relies on are unchanged (238 passed, 0 failed).
- [ ] Prose review of the revised "PR routing (Monitor loop)" section: confirm both the `Clear` branch and the label-gate branch converge on `surfaceBeforeMergingRequirements`, no bare `stop` remains on the label-gate branch, `verified` is applied only on the Clear path, and label transitions occur exactly once per path.

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -294,9 +294,11 @@ surfaceBeforeMergingRequirements:
 
   Then dispatch a Verification Agent (see pr_verify.md) using the dispatch template to carry out those before-merging items.
   The Verification Agent automates each item where possible and reports results on the tracking issues and PR; it does not consult the user.
-  Route on the Verification Agent's terminal signal per "Routing on the Verification Agent's signal" above.
+  Route on the Verification Agent's terminal signal per "Routing on the Verification Agent's signal" above, with this path-conditional override on that section's `Verification passed` handling:
+    if entered on the Clear path: follow "Routing on the Verification Agent's signal" unmodified. Applying `verified` and treating the PR as mergeable on `Verification passed` is correct here.
+    if entered on the label-gate path: apply only the label change in that section's `Verification passed` row (remove `verification needed`); do NOT add `verified`, and do NOT treat the PR as mergeable yet. A human must still remove the blocking process label before the PR may merge, even after `Verification passed`.
   On the Clear path: PR may be merged once every issue the planner filed for its before-merging list is resolved.
-  On the label-gate path: PR may be merged only once (a) a human removes the blocking process label, AND (b) every issue the planner filed for its before-merging list is resolved. These are two independent gates; a green Verification Agent alone does not clear a label-gated PR.
+  On the label-gate path: PR may be merged only once (a) a human removes the blocking process label, AND (b) every issue the planner filed for its before-merging list is resolved. These are two independent gates; a green Verification Agent alone does not clear a label-gated PR, per the override above.
 
 undiagnosedTerminal:
   // Issue #410 (Run G, issue #402): "drain poll found no new diagnostic
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -249,11 +249,8 @@ PR routing (Monitor loop):
     Relay `step "..." -> ...`, `FAIL [...] ...`, `summary`, and per-check summary rows to the user as informational test-result deltas; they do NOT end the loop or start a new Author round.
     if Monitor emits `drain poll found no new diagnostic signals` immediately followed by a Blocked or Infra line -> goto undiagnosedTerminal
     (Note: the attributed `Blocked by: <name>` form already names the blocking check in the per-check summary block and the terminal suffix, so the Monitor suppresses the drain flag in that case. The `goto undiagnosedTerminal` branch therefore applies only to a bare `Blocked`/`Infra` line that the Monitor itself flagged as undiagnosed.)
-    if Monitor emits a Blocked line where the terminal ends with `[label gate]` (the ` by: ...` suffix names only label-gate checks) -> clear the silentVanish re-launch flag; inform the user that only a process-label gate is blocking (not a code/test failure), that the code is review-approved, and that removing the blocking label plus resolving any before-merging tracking issues will unblock the merge; do NOT route to a new Author round; goto surfaceBeforeMergingRequirements (entered on the label-gate path)
-    // Note: the implementation plan (issue #516, Step 8a) proposed that the Orchestrator autonomously remove the blocking label and re-launch the Monitor.
-    // That approach was not adopted because the blocking label is a user-controlled process label that the Orchestrator does not own;
-    // its removal requires human judgment about whether the process condition is actually satisfied.
-    // Informing the user and continuing into the shared before-merging sub-routine is simpler and safer; the cycle does not conclude at this branch.
+    if Monitor emits a Blocked line where the terminal ends with `[label gate]` (the ` by: ...` suffix names only label-gate checks) -> clear the silentVanish re-launch flag; inform the user only a process-label gate blocks the merge (code is review-approved); do NOT route to a new Author round; goto surfaceBeforeMergingRequirements (label-gate path)
+    // The Orchestrator does not auto-remove the blocking label (issue #516, Step 8a): it is a user-controlled process label needing human judgment.
     if Monitor emits a Blocked line  -> clear the silentVanish re-launch flag; goto newAuthor
     if Monitor emits an Infra line   -> clear the silentVanish re-launch flag; escalate to user; stop
     if Monitor times out (30 min)    -> escalate to user; stop
@@ -263,42 +260,28 @@ PR routing (Monitor loop):
       goto surfaceBeforeMergingRequirements (entered on the Clear path)
 
 surfaceBeforeMergingRequirements:
-  // Step: Surface outstanding before-merging requirements
-  //   (unautomated verification tests and changes outside the repo, such as an issue that needs to be filed)
-  // Entered from two paths: the Clear path (Monitor emitted a Clear line) and the label-gate path
-  //   (Monitor emitted a Blocked line whose terminal ends with `[label gate]`). On the label-gate path
-  //   the Reviewer already gave LGTM before the Monitor loop was entered, so the code is review-approved
-  //   even though a process label still blocks the merge.
-  // Dispatch the verification planner (see verification_planning.md).
-  // The Orchestrator does not scan the issue or PR itself.
+  // Surfaces outstanding before-merging requirements (unautomated verification steps,
+  // changes outside the repo). Entered on the Clear path or the label-gate path; on the
+  // latter the Reviewer already gave LGTM, so the code is review-approved even though a
+  // process label still blocks the merge. The Orchestrator does not scan the issue or PR itself.
   Dispatch a Verification Planner sub-agent using the dispatch template.
-  The planner assembles the before-merging list and, for each item, files a tracking issue linked to the PR (see verification_planning.md). It does not consult the user.
+  The planner assembles the before-merging list and files a tracking issue per item (see verification_planning.md). It does not consult the user.
   If the Planner reports its before-merging list is empty:
-    if entered on the Clear path:
-      This *before-merging requirements* process is complete, and the Orchestrator should exit this step.
-      Apply this transition to **both the issue and the PR**:
+    if entered on the Clear path: this step is complete; apply this transition to **both the issue and the PR**:
 
       | Add label |
       |---|
       | `verified` |
 
-    if entered on the label-gate path:
-      This *before-merging requirements* process is complete, but the merge is still gated by the process label.
-      Do NOT apply `verified`; a human must still remove the blocking label before the PR may merge.
-      Inform the user that no before-merging items remain and that the merge unblocks once the process label is removed.
-  Otherwise (the before-merging list is non-empty, on either path), relay the planner's reported before-merging list to the user verbatim, and apply this transition to the PR:
+    if entered on the label-gate path: this step is complete, but do NOT apply `verified`; a human must still remove the blocking label before the PR may merge. Inform the user of this.
+  Otherwise, relay the before-merging list to the user verbatim, and apply this transition to the PR:
 
   | Add label |
   |---|
   | `verification needed` |
 
-  Then dispatch a Verification Agent (see pr_verify.md) using the dispatch template to carry out those before-merging items.
-  The Verification Agent automates each item where possible and reports results on the tracking issues and PR; it does not consult the user.
-  Route on the Verification Agent's terminal signal per "Routing on the Verification Agent's signal" above, with this path-conditional override on that section's `Verification passed` handling:
-    if entered on the Clear path: follow "Routing on the Verification Agent's signal" unmodified. Applying `verified` and treating the PR as mergeable on `Verification passed` is correct here.
-    if entered on the label-gate path: apply only the label change in that section's `Verification passed` row (remove `verification needed`); do NOT add `verified`, and do NOT treat the PR as mergeable yet. A human must still remove the blocking process label before the PR may merge, even after `Verification passed`.
-  On the Clear path: PR may be merged once every issue the planner filed for its before-merging list is resolved.
-  On the label-gate path: PR may be merged only once (a) a human removes the blocking process label, AND (b) every issue the planner filed for its before-merging list is resolved. These are two independent gates; a green Verification Agent alone does not clear a label-gated PR, per the override above.
+  Dispatch a Verification Agent (see pr_verify.md) to carry out those items; it does not consult the user.
+  Route on its terminal signal per "Routing on the Verification Agent's signal" above, EXCEPT on the label-gate path a `Verification passed` signal only removes `verification needed`: do NOT add `verified` and do NOT treat the PR as mergeable, since the process label still blocks the merge until a human removes it.
 
 undiagnosedTerminal:
   // Issue #410 (Run G, issue #402): "drain poll found no new diagnostic
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -249,41 +249,54 @@ PR routing (Monitor loop):
     Relay `step "..." -> ...`, `FAIL [...] ...`, `summary`, and per-check summary rows to the user as informational test-result deltas; they do NOT end the loop or start a new Author round.
     if Monitor emits `drain poll found no new diagnostic signals` immediately followed by a Blocked or Infra line -> goto undiagnosedTerminal
     (Note: the attributed `Blocked by: <name>` form already names the blocking check in the per-check summary block and the terminal suffix, so the Monitor suppresses the drain flag in that case. The `goto undiagnosedTerminal` branch therefore applies only to a bare `Blocked`/`Infra` line that the Monitor itself flagged as undiagnosed.)
-    if Monitor emits a Blocked line where the terminal ends with `[label gate]` (the ` by: ...` suffix names only label-gate checks) -> clear the silentVanish re-launch flag; inform the user that only a process-label gate is blocking (not a code/test failure) and that removing the blocking label should unblock the merge; stop (do NOT route to a new Author round)
+    if Monitor emits a Blocked line where the terminal ends with `[label gate]` (the ` by: ...` suffix names only label-gate checks) -> clear the silentVanish re-launch flag; inform the user that only a process-label gate is blocking (not a code/test failure), that the code is review-approved, and that removing the blocking label plus resolving any before-merging tracking issues will unblock the merge; do NOT route to a new Author round; goto surfaceBeforeMergingRequirements (entered on the label-gate path)
     // Note: the implementation plan (issue #516, Step 8a) proposed that the Orchestrator autonomously remove the blocking label and re-launch the Monitor.
     // That approach was not adopted because the blocking label is a user-controlled process label that the Orchestrator does not own;
     // its removal requires human judgment about whether the process condition is actually satisfied.
-    // Informing the user and stopping is simpler and safer.
+    // Informing the user and continuing into the shared before-merging sub-routine is simpler and safer; the cycle does not conclude at this branch.
     if Monitor emits a Blocked line  -> clear the silentVanish re-launch flag; goto newAuthor
     if Monitor emits an Infra line   -> clear the silentVanish re-launch flag; escalate to user; stop
     if Monitor times out (30 min)    -> escalate to user; stop
     if a user message wakes the session before Monitor delivers any terminal line -> goto silentVanish
     if Monitor emits a Clear line:
       clear the silentVanish re-launch flag
-      // Step: Surface outstanding before-merging requirements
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

Changes requested.

**The core fix (both terminals converge on `surfaceBeforeMergingRequirements`, the label-gate branch no longer bare-`stop`s, the empty-list branch correctly withholds `verified` on the label-gate path) is sound and well-executed.** The extraction into a shared labeled sub-routine avoids duplicating the Planner/Verifier flow, matching the plan.

However, there's a real logic gap in the non-empty-list branch that the plan's own "Double-gate confusion" risk called out but didn't fully close:

**The path-conditional guard is dropped before the Verification Agent's outcome is routed.** At line 297, `surfaceBeforeMergingRequirements` says:

> Route on the Verification Agent's terminal signal per "Routing on the Verification Agent's signal" above.

That pre-existing section (lines 153-160, unedited by this PR) reads, for `Verification passed`:

> This *before-merging requirements* process is complete; the PR may be merged.
> Apply this transition to the PR: | Remove label | Add label | `verification needed` | `verified` |

This is unconditional. It has no awareness of which path (Clear vs. label-gate) entered `surfaceBeforeMergingRequirements`, and it directly contradicts the new text two lines below it (line 299: "a green Verification Agent alone does not clear a label-gated PR") and the empty-list branch's explicit rule (line 287: "Do NOT apply `verified`"). Followed literally, an Orchestrator that took the label-gate path with a non-empty before-merging list, then got `Verification passed` from the Verification Agent, would apply `verified` to the PR and tell itself "the PR may be merged" -- exactly the double-gate confusion the issue's plan said to prevent (issue #540, "Trade-offs and rejected alternatives" / "Risks" section, item 4 of the design steps: "the sub-routine must NOT apply `verified` when it was entered on the label-gate path... This prevents a future reader from concluding that a green Verification Agent alone clears a label-gated PR").

This isn't just the non-empty branch either: `agents/verification_planning.md` line 36 also independently states "it applies `verified` to the PR and the issue on your empty-list report" with no path caveat. That file is out of scope per the plan, but it means there are now three places describing what happens on an empty/resolved before-merging list (the new empty-list branch, the "Routing on the Verification Agent's signal" section, and `verification_planning.md`), and only one of them (the new empty-list branch) carries the label-gate carve-out. The `goto`-based hand-off pattern works well for the empty-list case because the label-gate branch is inlined right there; it silently breaks for the non-empty-list case because it hands off to a separate, unedited section by reference instead of by inlining the same path-check.

**Suggested fix direction (not prescribing the exact wording, per Reviewer scope):** either inline a path-conditional variant of the `Verification passed` handling inside `surfaceBeforeMergingRequirements` (mirroring what was done for the empty-list case), or add an explicit override clause immediately after line 297 that says something like "on the label-gate path, do not apply `verified` and do not treat the PR as mergeable even when the Agent reports `Verification passed`; only remove `verification needed`." Either way, the "Routing on the Verification Agent's signal" section's unconditional `verified`/mergeable language needs to stop being reachable unmodified from the label-gate path.

Minor nit: the closing note at line 299 says "a green Verification Agent alone does not clear a label-gated PR," which is correct, but it appears only as prose *after* the routing instruction at line 297 that (if followed literally, per the section it points to) already applied `verified`. The note reads as a correction to a step that already executed, rather than a guard that prevents the wrong step from executing. Sequencing/wording should make the exception override the referenced section, not just comment on it afterward.

Everything else checked out: no bare `stop` remains on the label-gate branch, the `Clear` path's `verified` transition is unchanged and correctly gated to "if entered on the Clear path," label transitions in the edited text occur exactly once per path, the #516 Step 8a deviation note reads correctly with the revised continuation wording, and the diff has no prose-style violations (straight quotes, double-hyphens without spaces, no ellipses or em-dashes). `python3 scripts/test_ci_monitor.py` regression guard is a reasonable verification step for the unchanged parts, though as the PR notes it doesn't (and can't) test the prose logic itself.

aunger:
🤖 Author

Agreed, that was a real gap. Fixed in 1d65412: the handoff at the "Route on the Verification Agent's terminal signal per ..." line now carries an explicit path-conditional override, inlined at the point of routing rather than only stated afterward as a closing note:

- On the Clear path, "Routing on the Verification Agent's signal" applies unmodified: `Verification passed` still adds `verified` and the PR is treated as mergeable.
- On the label-gate path, `Verification passed` now only removes `verification needed` (no `verified`, not treated as mergeable). The comment makes explicit that a human must still remove the blocking process label first, even after `Verification passed`.

This closes the exact scenario you traced: an Orchestrator on the label-gate path with a non-empty before-merging list that gets `Verification passed` back will no longer apply `verified` or announce the PR as mergeable.

On the `verification_planning.md` line 36 nit: agreed it's a third place describing the same behavior, but per the issue's finalized plan that file is explicitly out of scope ("Leave `agents/verification_planning.md` ... unchanged"). That line describes Orchestrator label behavior only to tell the Planner not to touch labels itself; it isn't a routing rule the Orchestrator follows (the Orchestrator's own routing now lives entirely in `dev_orchestration.md`, which is where the fix applies). I've left it as-is to stay within the plan's scope; happy to file a follow-up issue if you'd rather it be tightened too, but didn't think it rose to blocking given the plan's explicit scope note.

`python3 scripts/test_ci_monitor.py` still passes (238/238), unaffected since this remains a prose-only change to `agents/dev_orchestration.md`.

aunger (commented):
🤖 Reviewer

LGTM.

Commit 1d65412 closes the gap precisely. The override at lines 297-299 is inlined at the point where "Routing on the Verification Agent's signal" is invoked, not just stated as a trailing comment, so it correctly overrides rather than merely comments on the unconditional `Verification passed` handling:

- Clear path: unmodified, `verified` applied, PR treated as mergeable. Correct.
- Label-gate path: only `verification needed` is removed; `verified` is explicitly withheld and the PR is explicitly not treated as mergeable. This closes the exact scenario I traced (label-gate path, non-empty before-merging list, `Verification passed` back from the Verification Agent).

`Verification revealed an error` and `Verification incomplete` didn't need an override since neither applies `verified` or declares the PR mergeable; correctly left alone.

On the `verification_planning.md` line 36 point: agreed with your rebuttal. That line is instruction to the Planner (explaining why the Planner itself must not touch labels), not routing logic the Orchestrator consults, so it doesn't reintroduce the double-gate risk even though it's now slightly imprecise about the Orchestrator's actual (now conditional) behavior. Fine to leave out of scope per the issue's plan; a follow-up issue for tightening its wording would be reasonable but isn't blocking.

Checked the rest of the diff again for regressions: no prose-style violations, CI is green apart from "No blocking labels" (expected during the active review cycle per `pr_participation.md`).

aunger:
Does it really need to be so much longer?

aunger:
🤖 Author

Fair, noted for future replies.

aunger:
Make the agent instructions more concise.

aunger:
🤖 Author

Trimmed in 350f39b: condensed the label-gate branch comment, the sub-routine's opening comment, and the Verification Agent routing override down to fewer lines. The override is still inline at the routing handoff (now one "EXCEPT on the label-gate path..." clause instead of nested if/then blocks), so the fix stands; just said more tersely.

aunger (commented):
🤖 Reviewer

LGTM.

Re-reviewed after 350f39b (conciseness trim). The trim preserves the substance: both terminals still converge on `surfaceBeforeMergingRequirements`, the label-gate branch still has no bare `stop`, `verified` is still applied only on the Clear path in the empty-list case, and the `Verification passed` override is still inlined at the routing handoff ("EXCEPT on the label-gate path...") rather than trailing commentary. No regressions from the trim. CI still green apart from the expected "No blocking labels" check.

PR #549, opened 2026-07-01
Silence stop-hook uncommitted-work feedback entirely for the Orchestrator

aunger:
The Delegation rules section of `agents/dev_orchestration.md` already told the Orchestrator to disregard `stop-hook-git-check.sh`-style "uncommitted changes" feedback while a Programmer sub-agent is still actively working in its own branch/worktree, and to keep waiting "without updating the User." In practice that phrasing left room for a brief inline acknowledgment (e.g. "that's expected, disregarding this and continuing to wait") when the Orchestrator's reply was about something else in the same turn.

This tightens the rule so the hook's message is never referenced in the Orchestrator's reply at all, not even briefly.


aunger:
🤖 Reviewer

LGTM.

The change is a single, well-scoped line in `agents/dev_orchestration.md` that closes exactly the gap described in the PR body: the old wording ("continue waiting without updating the User") left room for a brief inline acknowledgment of the hook's message, and the new sentence explicitly forbids referencing, quoting, or explaining away the hook message even briefly and even when the reply is about something else in the same turn.

Checked:
- The new sentence is unambiguous and doesn't conflict with neighboring bullets (the test-failure/error bullet and the "file was modified" bullet are separate cases and remain untouched).
- No prose-style violations (no em-dash/smart-quote/ellipsis characters, no trailing whitespace).
- Diff is minimal (1 file, 1 line changed), matching the narrow intent stated in the PR description.
- The failing "No blocking labels" check is expected during the review cycle per `agents/pr_participation.md` and not a merge blocker at this stage.

Nit (non-blocking): the PR description's mention of "`stop-hook-git-check.sh`-style" feedback is illustrative only, since no such script currently exists in the repo; that's fine since the actual rule text stays generic ("system hooks or events that signal uncommitted work") and doesn't hardcode a script name.

aunger:
🤖 Reviewer

Re-reviewed after commit b441555 (head now b441555afc4bfae261aedaa2e8b88877ad66751a).

LGTM.

The follow-up commit extends the exact same fix to the other two bullets in the list that had the identical "continue waiting without updating the User" gap (test-failure/error, and file-modified reminder). Good catch generalizing the pattern rather than leaving inconsistent siblings.

Checked on the new commit:
- Same no-mention wording applied consistently; the file-modified bullet correctly says "the reminder" instead of "the hook's message," matching that bullet's own terminology.
- No prose-style violations (no em-dash/smart-quote/ellipsis characters, no trailing whitespace) in the added lines.
- No new conflicts introduced between the three bullets or with the rest of the Delegation rules section.
- `build-and-test` still passes on the new head; `No blocking labels` still fails, which remains expected during the review cycle per `agents/pr_participation.md`.

My earlier LGTM and nit both still stand.

PR #550, opened 2026-07-01
Make dismiss_anr.sh poll interval/timeout overridable to speed up its unit tests

### LINKED ISSUE #545: Performance eval: Shell script tests
Is there an opportunity to speed up the "Shell script unit tests" in CI? *No* is a valid answer.

aunger:
Fixes #545.

## Problem

The "Shell script unit tests" CI step's dominant cost is `scripts/test_dismiss_anr.sh`, which alone took ~97 seconds locally (nearly all of the step's total runtime). `scripts/dismiss_anr.sh`'s poll loop uses a hard-coded `POLL_INTERVAL=3` / `TIMEOUT=30`, and the test suite exercises that real loop (real `sleep` calls) against a mocked `adb` across nine scenarios, including one that deliberately waits out the full 30s timeout to verify the script doesn't hang forever.

## Fix

- `scripts/dismiss_anr.sh`: made `POLL_INTERVAL` and `TIMEOUT` overridable via environment variables, mirroring the existing `SLEEP_AFTER_ANR_DETECTED` override that was already there for the same reason. Defaults are unchanged (3s / 30s), so real CI pre-flight behavior during E2E setup is unaffected. The file header's "Environment:" doc block now lists all three overridable variables.
- `scripts/test_dismiss_anr.sh`: exports compressed values (`POLL_INTERVAL=1`, `TIMEOUT=8`) for the whole suite, with the persistent-timeout scenario (e) locally overriding `TIMEOUT=3` since it doesn't need the extra headroom the multi-poll scenarios (g, h) require. Adjusted the handful of comments/assertions whose numbers were hard-coded to the old real-world timing (the 5s inter-ANR gap in scenario f, the wall-clock budget checks in scenario e). No scenario's coverage or pass/fail logic changed.

## Result

Locally: `test_dismiss_anr.sh` alone drops from ~97s to ~30s per run; the full shell test suite (all 6 `test_*.sh` files, 84 assertions) also runs in ~30s total, since this file is effectively the entire cost of the suite. All 84 shell assertions and all 356 Python unit tests still pass.

## Test plan

- [ ] Confirm the "Run shell script unit tests" CI step duration drops accordingly on this PR's CI run.

### DIFF
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -21,6 +21,10 @@
 #                  $ANDROID_HOME/platform-tools/adb.
 #
 # Environment:
+#   POLL_INTERVAL              Seconds between poll iterations (default: 3).
+#                              Override in tests.
+#   TIMEOUT                    Seconds before the poll loop gives up and exits
+#                              0 anyway (default: 30). Override in tests.
 #   SLEEP_AFTER_ANR_DETECTED   Seconds to wait after logcat fires before sending
 #                              KEYCODE_ENTER (default: 7). Override in tests.
 
--- a/scripts/dismiss_anr.sh
+++ b/scripts/dismiss_anr.sh
@@ -129,8 +129,11 @@
   ' EXIT
 
   # ── Poll loop ────────────────────────────────────────────────────────────────
-  POLL_INTERVAL=3
-  TIMEOUT=30
+  # POLL_INTERVAL, TIMEOUT, and SLEEP_AFTER_ANR_DETECTED are overridable via
+  # environment so the unit test suite can compress the real wall-clock sleeps
+  # this loop performs (it polls a real mock adb, not a faked clock).
+  POLL_INTERVAL="${POLL_INTERVAL:-3}"
+  TIMEOUT="${TIMEOUT:-30}"
   # How long to wait after logcat fires before sending KEYCODE_ENTER.
   # Override via environment for tests.
   SLEEP_AFTER_ANR_DETECTED="${SLEEP_AFTER_ANR_DETECTED:-7}"
--- a/scripts/test_dismiss_anr.sh
+++ b/scripts/test_dismiss_anr.sh
@@ -6,10 +6,20 @@
 #   (b) ANR-detection branch: logcat fires → KEYCODE_BACK sent → exits 0
 #   (c) idle-count exit path: logcat hangs, CPU goes idle → exits 0
 #   (d) Absent Launcher process treated as idle → exits 0
-#   (e) Persistent ANR: logcat fires but BACK never clears; 30 s timeout → exits 0
+#   (e) Persistent ANR: logcat fires but BACK never clears; timeout → exits 0
 #   (j) Pattern fallback exercises the nexuslauncher arm of nexuslauncher|launcher3
 #
 # Always exits 0 on success, non-zero on failure.
+#
+# dismiss_anr.sh's poll loop performs real wall-clock sleeps against a mock
+# adb, and its elapsed-time bookkeeping is bash integer arithmetic, so
+# POLL_INTERVAL and TIMEOUT must stay whole seconds. They are exported here
+# at compressed values (real default: 3 s / 30 s) to keep this suite fast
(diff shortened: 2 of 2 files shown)


aunger (commented):
Placeholder

aunger:
🤖 Reviewer

(Note: I accidentally submitted a placeholder via GitHub's review feature just before this comment. Please disregard that; per our process this ordinary comment is the actual review.)

I checked out the branch, ran the shell suite repeatedly, and ran the full Python suite. Verified facts first:

- All 19 shell assertions in `test_dismiss_anr.sh` pass, consistently across 4 separate runs.
- All 356 Python unit tests pass (1 skipped), matching the PR's claim.
- The env-var override pattern (`POLL_INTERVAL="${POLL_INTERVAL:-3}"`, `TIMEOUT="${TIMEOUT:-30}"`) correctly mirrors the existing `SLEEP_AFTER_ANR_DETECTED` convention, and production defaults are unchanged, so this doesn't touch real CI pre-flight behavior. No scope creep; only the two stated files are touched.
- The 97s → ~30s improvement for the full 6-file shell suite is real and reproducible. This is a solid win and resolves issue #545.

Two things to fix:

1. **The PR description's per-file timing claim doesn't match what I measured.** The description says `test_dismiss_anr.sh` alone "drops from ~97s to ~5-10s per run," but across 4 separate runs I consistently measured ~30 seconds for that file alone (04:41:21.654 to 04:41:51.617 in one run, for example) — and the full 6-file suite also measured ~30s total, meaning `test_dismiss_anr.sh` is still effectively the entire cost of the suite, not a small fraction of it. Please re-measure and correct the PR description (either the ~5-10s figure is wrong, or something about how you measured it differs from a plain `bash scripts/test_dismiss_anr.sh` run — worth double-checking which). The underlying fix is still good; the description just oversells the per-file result.

2. **Doc comment now out of date.** `scripts/dismiss_anr.sh` lines 23-25 (the "Environment:" block in the file header) still document only `SLEEP_AFTER_ANR_DETECTED` as overridable via environment, but this PR makes `POLL_INTERVAL` and `TIMEOUT` overridable too. Please add them to that header doc block so a future reader doesn't have to find the inline comment near the poll loop to learn this.

Nit: the inline comment added above the poll loop (lines 131-134) explaining *why* these are overridable is good and I don't want it removed — I'd just also update the header since that's the first place someone looks for the environment variable contract.

Changes requested.

aunger:
🤖 Programmer

Thanks for the careful re-measurement — both points are fixed.

1. **Timing claim corrected.** You were right: the ~5-10s figure was a mistake on my part, I'd pulled it from scenario (e)'s isolated wall-clock print rather than measuring the whole file. Re-ran `test_dismiss_anr.sh` alone twice, cleanly: ~30s both times, matching what you measured. Updated the PR description to say `test_dismiss_anr.sh` alone drops from ~97s to ~30s, and that it's effectively the entire cost of the full 6-file suite (also ~30s) rather than a small fraction of it. The underlying 97s→30s improvement itself is unchanged and still real.

2. **Header doc comment fixed.** Added `POLL_INTERVAL` and `TIMEOUT` to the `Environment:` block in `scripts/dismiss_anr.sh` alongside the existing `SLEEP_AFTER_ANR_DETECTED` entry, so the header now documents the full environment-variable contract in one place. Left the inline comment above the poll loop as-is per your nit.

Re-ran the full shell suite after this change: 84/84 assertions pass, ~30.8s total.

aunger:
🤖 Reviewer

Re-verified independently on the latest commit (1b60130), not just taking the self-report on faith:

1. **Header doc fix**: confirmed `POLL_INTERVAL` and `TIMEOUT` are now both documented in the `Environment:` block at the top of `scripts/dismiss_anr.sh`, alongside `SLEEP_AFTER_ANR_DETECTED`, in the same style.
2. **Timing claim**: re-ran `test_dismiss_anr.sh` alone twice (19/19 pass, ~30.0s both times) and the full 6-file suite once (84/84 pass, ~30.6s total) in a fresh worktree. Matches the corrected PR description exactly.

Both points addressed correctly. LGTM.

PR #552, opened 2026-07-01
Render SecureViewerActivity reactively from a session-media StateFlow

### LINKED ISSUE #537: Fix SessionTracker race: SecureViewerActivity shows empty view on open
## Summary

`GalleryButtonVisualE2ETest#test5a_secureCameraLockedPopulatedGalleryShowsGreen` fails intermittently because `SecureViewerActivity` opens with an empty session and displays a black screen, causing `waitForGreenBand` to time out.

Surfaced during [this CI run](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/28338450437/job/83948649202?pr=532) on #532 (the fix for #509).
Related: #243.

## What happens

1. `waitForSessionMedia()` polls `SessionTracker` from the test process and passes -- the session contains the photo.
2. The overlay is tapped; `SecureViewerActivity` starts.
3. In the ~500ms window between `waitForSessionMedia()` confirming success and `SecureViewerActivity.onCreate()` calling `refreshMedia()`, the session is reset or not yet reflected in the activity's in-process `SessionTracker` reference.
4. `refreshMedia()` sees an empty session and renders a black "no photos" state.
5. `waitForGreenBand` polls for 15 s on a black screen and times out.

This is the same class of race that `waitForSessionMedia()` was introduced to guard against on the tap side -- but it does not cover `SecureViewerActivity`'s own startup read.

Note: the initial diagnosis (filed as #534, now closed) incorrectly attributed this to keyguard auto-dismissal via `setTurnScreenOn(true)`.
That theory is ruled out by the hard `fail()` assertion in `launchSecureCamera()` (`E2EFixture.kt` lines 533-541), which would have caught an unlocked keyguard before the tap, and by the `Overlay tapped: locked=true` logcat entry confirming the device was still locked at tap time.

## Fix options

1. **Preferred -- reactive session observer in `SecureViewerActivity`**: wire a `LiveData`/`StateFlow` observer so the viewer auto-updates when `SessionTracker` populates, rather than relying on a one-shot `refreshMedia()` call in `onCreate()`.
2. **Short re-poll in `SecureViewerActivity`**: after `refreshMedia()` returns empty, re-poll `SessionTracker` on a short delay (e.g. 200ms, a few times) before giving up and showing the empty state.
3. **Temporary workaround**: add `test5a` to `.github/allowed-test-failures.txt` while the proper fix lands.

### Goal and constraints

Make `SecureViewerActivity` display the current secure session's media reliably, even when the session is populated (or repopulated) shortly after the activity starts, so `test5a_secureCameraLockedPopulatedGalleryShowsGreen` stops flaking.
Planning only: no code is written and no PR is opened by this task.
Follow `agents/code_edit.md`: prefer simple code, do not repeat yourself, split commits by concern, commit any pure refactor separately, and cover new behavior with automated tests without silently loosening existing tests.

### Confirmed root cause

`SecureViewerActivity` reads the session exactly once per lifecycle callback.
`onCreate()` (SecureViewerActivity.kt ~line 84) calls `setupLayout()` then a one-shot `refreshMedia()`; `onResume()` (~line 109) calls `refreshMedia()` again after a keyguard check.
`refreshMedia()` (~line 284) reads `SessionTracker.instance.getSessionMedia()` and toggles the empty state versus the populated state.
`SessionTracker` (SessionTracker.kt) is a synchronized in-memory holder with a process-wide singleton `instance`.
It has no observable surface today: nothing notifies a reader when `mediaItems` changes.

The session can be empty at the instant the viewer reads it for two reasons, both of which the one-shot read cannot recover from:
1. Not yet reflected: the `OverlayService` media observer adds media asynchronously (OverlayService.kt ~line 101, via `MediaObserverRetry` with `Constants.MEDIA_OBSERVER_RETRY_MS`).
   The test-process `waitForSessionMedia()` gate and the activity's own read observe the same singleton but at different moments.
2. Reset: `OverlayServiceLogic.hideOverlayAndCleanup()` (OverlayServiceLogic.kt ~line 212) calls `sessionTracker.endSession()`, which clears `mediaItems`.
   A camera-availability transition around tap time can drive a deactivation cleanup that empties the session between the test's gate and the activity's read.

Because both failure modes are just "the singleton was empty at read time, then became non-empty," a reactive observer that re-renders on any later population fixes both without needing to prove which one fired in a given CI run.

### Recommended fix: option 1 (reactive observer), reject options 2 and 3 as the primary fix

Option 1 (reactive `StateFlow` observer) is the correct fix.
It removes the timing assumption entirely: the viewer renders whatever the session currently holds and re-renders on every later change, so a late population or a repopulation after a transient reset is picked up automatically.
It also composes cleanly with the existing add, remove, and undo paths, which already mutate the singleton and would then push updates to the viewer for free.

Option 2 (short re-poll) is rejected as the primary fix: it is a bounded timing workaround that trades one magic delay for another and still fails if population lands after the last poll.
Option 3 (allowed-test-failures entry) is rejected as the primary fix: it hides the real user-facing bug (a locked user tapping the overlay can see an empty viewer) rather than resolving it.
Option 3 may be used only as an explicit, temporary stopgap if the reactive fix cannot land in the same change, and if so it must be removed in the same PR that lands the fix (open question below).

### Concrete code changes

1. Add an observable surface to `SessionTracker` (SessionTracker.kt).
   - Add a private `MutableStateFlow` of `List<MediaItem>` seeded with an empty list, and expose it as a read-only `StateFlow` of `List<MediaItem>` (for example a `sessionMedia` property).
   - After every mutation of `mediaItems` (`startSession`, `endSession`, `addMedia`, `removeMedia`), emit the current sorted snapshot.
     Compute the snapshot with the existing "most recent first" ordering used by `getSessionMedia()` so the flow and the getter never disagree; extract that sort into a single private helper and have both `getSessionMedia()` and the emission call it (this is the DRY point, and it is a pure refactor).
   - Emit a fresh `List` instance (a copy) so `StateFlow` structural-equality de-duplication compares list contents, not the same mutable reference; do the emission while still holding `lock` so the emitted snapshot is consistent with the state under the lock.
   - Keep `getSessionMedia()` as-is for existing callers and tests; the flow is additive.

2. Make `SecureViewerActivity` collect the flow (SecureViewerActivity.kt).
   - In `onCreate()`, after `setupLayout()`, launch a lifecycle-scoped collector using `lifecycleScope.launch` wrapping `repeatOnLifecycle(Lifecycle.State.STARTED)` that collects `sessionTracker.sessionMedia` and calls `renderMedia(...)` on each emission.
   - Refactor `refreshMedia()` into `renderMedia(media)` taking a `List<MediaItem>` that does the existing `submitList` plus empty-state visibility toggling; the collector calls `renderMedia`.
     This is a pure refactor of the rendering body.
   - Remove the one-shot `refreshMedia()` call in `onCreate()`; the initial emission from the `StateFlow` (replayed to new collectors) covers the initial render, so there is no gap and no double render.
   - `onResume()` keeps its keyguard check and `finish()` behavior unchanged.
     Remove its now-redundant `refreshMedia()` call, since the active collector already reflects current state; if kept for safety it must call `renderMedia(sessionTracker.getSessionMedia())` rather than the deleted method (decide one way, do not leave both).
   - Update the delete and undo handlers: they currently call `refreshMedia()` after `sessionTracker.removeMedia(...)` / `addMedia(...)`.
     Once those mutations emit on the flow, the collector re-renders automatically, so the explicit calls can be dropped.
     Verify the snackbar and swipe interactions still behave; if immediate synchronous re-render is required for the swipe animation, call `renderMedia(...)` explicitly and document why.

3. No changes required to `OverlayService`/`OverlayServiceLogic` for the fix itself.
   The service continues to call `startSession`/`addMedia`/`endSession`; those now additionally drive the flow.

### Sequencing and commit breakdown (by concern)

1. Commit 1 (pure refactor, no behavior change): extract the session-media sort helper in `SessionTracker`; extract `renderMedia(list)` out of `refreshMedia()` in `SecureViewerActivity`.
2. Commit 2 (feature: observable session): add the `StateFlow` surface to `SessionTracker` and emit on every mutation; add unit tests for emissions.
3. Commit 3 (fix: reactive viewer): wire the `repeatOnLifecycle` collector in `SecureViewerActivity`, drop the one-shot and handler-side `refreshMedia()` calls, add the viewer render test.
4. Commit 4 (optional stopgap, only if needed and only if removed in the same PR): add/remove the `test5a` allowed-failure entry.

### Testing and verification

- SessionTracker unit tests (`app/src/test/java/com/gb4pc/viewer/SessionTrackerTest.kt`): assert the `StateFlow` value starts empty, emits the added item after `addMedia`, emits empty after `endSession`, and emits the reduced list after `removeMedia`; assert ordering matches `getSessionMedia()`.
  Use `runTest` and read the flow's `value` or collect the first emissions.
- SessionTracker thread-safety test (`SessionTrackerThreadSafetyTest.kt`): confirm concurrent mutations still terminate and the final emitted snapshot matches `getSessionMedia()`; watch for deadlock given emission happens under `lock`.
- SecureViewer render test: the existing `SecureViewerActivityTest.kt` is a plain JVM test and documents that activity-level behavior needs Robolectric or instrumentation.
  Add a Robolectric test that launches the activity against an empty session, then calls `SessionTracker.instance.addMedia(...)` and asserts the empty view hides and the pager shows the item (this is the regression test for the race, at the unit layer).
  If Robolectric cannot host `setShowWhenLocked`/`KeyguardManager` cleanly (the file notes this friction), fall back to an instrumented test under `app/src/androidTest`.
- E2E: `test5a_secureCameraLockedPopulatedGalleryShowsGreen` is the end-to-end validation; it should stop flaking once the viewer re-renders on late population.
  Do not weaken its assertions.
- Build locally and run unit tests before pushing, per `agents/code_edit.md`.

### Risks and trade-offs

- Emitting under `lock`: keeps snapshots consistent but means a collector's synchronous work should not call back into `SessionTracker` on the same thread; `StateFlow` emission is non-blocking, so this is low risk, but avoid heavy work in the collector.
- `StateFlow` de-duplication: it drops emissions that are structurally equal to the current value, so always emit a fresh list built from current contents; do not emit the internal mutable list.
- Lifecycle correctness: use `repeatOnLifecycle(STARTED)` rather than a bare `lifecycleScope.launch` collect, so collection stops in the background and there is no leak; this also fits the existing `onResume` keyguard-finish flow.
- Double render on startup: mitigated by removing the one-shot `refreshMedia()`; the replayed `StateFlow` value renders once on first collection.

### Open questions for the requester

1. Should the separate "session reset by `hideOverlayAndCleanup()` around tap time" behavior be investigated and fixed on its own?
   The reactive viewer masks it, but if the session is genuinely being ended and restarted around a valid tap, that may be a second bug worth a follow-up issue.
2. Do you want the option 3 stopgap (`.github/allowed-test-failures.txt` entry for `test5a`) in the same PR, to green CI while the fix stabilizes, with removal in that same PR?
   Default recommendation: no stopgap, land the fix directly.
3. Preferred observable primitive: `StateFlow` (recommended, coroutine-native and matches `lifecycleScope`/`repeatOnLifecycle`) versus `LiveData`.
   This plan assumes `StateFlow`.
4. Should `waitForSessionMedia()` in `E2EFixture.kt` be kept?
   It is still a reasonable gate for other tests; recommendation is to keep it, since it is independent of the viewer fix.

aunger:
Fixes #537.

## Problem

`SecureViewerActivity` read `SessionTracker` exactly once in `onCreate()` (and again in `onResume()`). If the secure session was still empty at that instant, the viewer rendered the empty "no photos" state and never recovered. Two things can leave the session empty at read time, both unrecoverable with a one-shot read:

1. Media population lands asynchronously (the `OverlayService` media observer adds media after the activity has started).
2. The session is briefly reset (`hideOverlayAndCleanup()` calls `endSession()`) around a camera-availability transition near tap time.

This made `test5a_secureCameraLockedPopulatedGalleryShowsGreen` flake (viewer stayed black, `waitForGreenBand` timed out) and, more importantly, could show a locked user an empty viewer even though the session had photos.

## Fix

Make the viewer render reactively from an observable session surface, so it re-renders on any later population or repopulation, removing the timing assumption entirely (issue's preferred option 1).

- **`SessionTracker`**: add a read-only `sessionMedia: StateFlow<List<MediaItem>>`. Every mutation (`startSession`/`endSession`/`addMedia`/`removeMedia`) emits a fresh, most-recent-first snapshot built by a shared `sortedSnapshot()` helper, so the flow and `getSessionMedia()` never disagree. Emission happens under the existing lock so the snapshot is consistent with the locked state.
- **`SecureViewerActivity`**: collect `sessionMedia` in a `lifecycleScope` + `repeatOnLifecycle(STARTED)` collector started in `onCreate()`. The replayed initial value renders once (no double render); later emissions re-render automatically. The one-shot read in `onCreate()`, the redundant `onResume()` render (keyguard check unchanged), and the explicit renders in the delete/undo handlers are dropped, since the mutations now drive the flow.

`getSessionMedia()` is unchanged and kept for existing callers; the flow is purely additive. No changes were needed to `OverlayService`/`OverlayServiceLogic`.

## Commits (split by concern)

1. Pure refactor: extract `sortedSnapshot()` in `SessionTracker`; extract `renderMedia(list)` from `refreshMedia()` in the viewer.
2. Feature: add the `sessionMedia` `StateFlow` and emit on every mutation, with unit tests.
3. Fix: wire the reactive collector, drop the one-shot/handler-side renders, add the Robolectric regression test.

## Tests

- `SessionTrackerTest`: flow starts empty, emits on add, empties on `endSession`, reduces on `removeMedia`, and ordering agrees with `getSessionMedia()`.
- `SessionTrackerThreadSafetyTest`: now also asserts the final emitted snapshot matches `getSessionMedia()` after concurrent mutations (guards against deadlock from emitting under the lock).
- `SecureViewerActivityRenderTest` (new, Robolectric): launches the viewer against an empty session and asserts it flips from the empty state to the populated state when media is added after the activity is already showing -- the unit-layer regression for the race.
- Full `:app:testDebugUnitTest` suite runs green locally.

`test5a` E2E assertions are unchanged; it is the end-to-end validation and should stop flaking now that the viewer re-renders on late population.

## Open questions from the issue's plan

- The separate "session reset around tap time" behavior (question 1) is masked by this reactive fix but may still be worth a follow-up issue if the session is genuinely ended/restarted around a valid tap.
- No `.github/allowed-test-failures.txt` stopgap was added (question 2): the fix lands directly.
- `StateFlow` was used as the observable primitive (question 3).
- `waitForSessionMedia()` in `E2EFixture.kt` was left in place (question 4).

### DIFF
--- a/app/src/main/java/com/gb4pc/viewer/SecureViewerActivity.kt
+++ b/app/src/main/java/com/gb4pc/viewer/SecureViewerActivity.kt
@@ -87,7 +87,7 @@ class SecureViewerActivity : ComponentActivity() {
         setTurnScreenOn(true)
 
         setupLayout()
-        refreshMedia()
+        renderMedia(sessionTracker.getSessionMedia())
     }
 
     override fun onStart() {
@@ -114,7 +114,7 @@ class SecureViewerActivity : ComponentActivity() {
             finish()
             return
         }
-        refreshMedia()
+        renderMedia(sessionTracker.getSessionMedia())
     }
 
     // H4: ContentObserver registration/unregistration has been removed.
@@ -222,7 +222,7 @@ class SecureViewerActivity : ComponentActivity() {
     ) {
         // Remove from session immediately
         sessionTracker.removeMedia(media.uri)
-        refreshMedia()
+        renderMedia(sessionTracker.getSessionMedia())
 
         // SF-10: Show undo snackbar
         val rootView = findViewById<View>(android.R.id.content)
@@ -234,7 +234,7 @@ class SecureViewerActivity : ComponentActivity() {
             ).setAction(R.string.viewer_undo) {
                 // Undo: re-add to session
                 sessionTracker.addMedia(media)
-                refreshMedia()
+                renderMedia(sessionTracker.getSessionMedia())
             }.addCallback(
                 object : Snackbar.Callback() {
                     override fun onDismissed(
@@ -281,8 +281,7 @@ class SecureViewerActivity : ComponentActivity() {
         )
     }
 
-    private fun refreshMedia() {
-        val media = sessionTracker.getSessionMedia()
+    private fun renderMedia(media: List<MediaItem>) {
         adapter.submitList(media)
         emptyMessage.visibility = if (media.isEmpty()) View.VISIBLE else View.GONE
         viewPager.visibility = if (media.isEmpty()) View.GONE else View.VISIBLE
--- a/app/src/main/java/com/gb4pc/viewer/SecureViewerActivity.kt
+++ b/app/src/main/java/com/gb4pc/viewer/SecureViewerActivity.kt
@@ -20,7 +20,9 @@ import androidx.activity.ComponentActivity
 import androidx.activity.result.ActivityResultLauncher
 import androidx.activity.result.IntentSenderRequest
 import androidx.activity.result.contract.ActivityResultContracts
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.ListAdapter
@@ -87,7 +89,18 @@ class SecureViewerActivity : ComponentActivity() {
         setTurnScreenOn(true)
 
         setupLayout()
-        renderMedia(sessionTracker.getSessionMedia())
+
+        // #537: Render reactively from the session's StateFlow rather than a one-shot read.
+        // The replayed initial value covers the first render, and any later population
+        // (or repopulation after a transient reset) re-renders automatically, closing the
+        // race where the viewer opened before SessionTracker was populated.
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.STARTED) {
+                sessionTracker.sessionMedia.collect { media ->
+                    renderMedia(media)
+                }
+            }
+        }
     }
 
     override fun onStart() {
@@ -114,7 +127,8 @@ class SecureViewerActivity : ComponentActivity() {
             finish()
             return
         }
-        renderMedia(sessionTracker.getSessionMedia())
+        // #537: No explicit render here; the lifecycle-scoped collector started in
+        // onCreate() re-collects the StateFlow's current value on returning to STARTED.
     }
 
     // H4: ContentObserver registration/unregistration has been removed.
@@ -220,9 +234,8 @@ class SecureViewerActivity : ComponentActivity() {
         media: MediaItem,
         position: Int,
     ) {
-        // Remove from session immediately
+        // Remove from session immediately; the sessionMedia collector re-renders (#537).
         sessionTracker.removeMedia(media.uri)
-        renderMedia(sessionTracker.getSessionMedia())
 
         // SF-10: Show undo snackbar
         val rootView = findViewById<View>(android.R.id.content)
@@ -232,9 +245,8 @@ class SecureViewerActivity : ComponentActivity() {
                 com.gb4pc.Constants.UNDO_TIMEOUT_MS
                     .toInt(),
             ).setAction(R.string.viewer_undo) {
-                // Undo: re-add to session
+                // Undo: re-add to session; the sessionMedia collector re-renders (#537).
                 sessionTracker.addMedia(media)
-                renderMedia(sessionTracker.getSessionMedia())
             }.addCallback(
                 object : Snackbar.Callback() {
                     override fun onDismissed(
--- a/app/src/main/java/com/gb4pc/viewer/SessionTracker.kt
+++ b/app/src/main/java/com/gb4pc/viewer/SessionTracker.kt
@@ -1,6 +1,9 @@
 package com.gb4pc.viewer
 
 import com.gb4pc.Constants
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
 
 /**
  * Tracks the current secure camera session and its media items (§5.1).
@@ -18,6 +21,15 @@ class SessionTracker {
 
     private val mediaItems = mutableListOf<MediaItem>()
 
+    private val _sessionMedia = MutableStateFlow<List<MediaItem>>(emptyList())
+
+    /**
+     * SF-07: Observable, most-recent-first view of the current session media.
+     * Emits a fresh list on every mutation so collectors (e.g. SecureViewerActivity)
+     * re-render when the session is populated, reset, or edited (#537).
+     */
+    val sessionMedia: StateFlow<List<MediaItem>> = _sessionMedia.asStateFlow()
+
     /**
      * SF-01: Begin a new session, recording the start timestamp (SF-02).
      */
@@ -26,6 +38,7 @@ class SessionTracker {
             isSessionActive = true
             sessionStartTimestamp = System.currentTimeMillis()
             mediaItems.clear()
+            emitSnapshot()
         }
     }
 
@@ -36,6 +49,7 @@ class SessionTracker {
         synchronized(lock) {
             isSessionActive = false
             mediaItems.clear()
+            emitSnapshot()
         }
     }
 
@@ -44,13 +58,16 @@ class SessionTracker {
             if (!isSessionActive) return
             if (mediaItems.none { it.uri == item.uri }) {
                 mediaItems.add(item)
+                emitSnapshot()
             }
         }
     }
 
     fun removeMedia(uri: String) {
         synchronized(lock) {
-            mediaItems.removeAll { it.uri == uri }
+            if (mediaItems.removeAll { it.uri == uri }) {
+                emitSnapshot()
+            }
         }
     }
 
@@ -67,6 +84,16 @@ class SessionTracker {
      * Callers must already hold [lock].
      */
     private fun sortedSnapshot(): List<MediaItem> = mediaItems.sortedByDescending { it.dateTaken }
+
+    /**
+     * Publishes a fresh snapshot to [sessionMedia]. Callers must already hold [lock]
+     * so the emitted list is consistent with the state under the lock.
+     * A new list instance is emitted so StateFlow structural-equality de-duplication
+     * compares contents rather than the same mutable reference.
+     */
+    private fun emitSnapshot() {
+        _sessionMedia.value = sortedSnapshot()
+    }
 
     /**
      * SF-04: Check if a media item belongs to the current session.
--- a/app/src/main/java/com/gb4pc/viewer/SessionTracker.kt
+++ b/app/src/main/java/com/gb4pc/viewer/SessionTracker.kt
@@ -59,8 +59,14 @@ class SessionTracker {
      */
     fun getSessionMedia(): List<MediaItem> =
         synchronized(lock) {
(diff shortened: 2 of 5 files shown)


aunger (commented):
🤖 Reviewer

LGTM. This faithfully implements option 1 from the issue plan and the fix is correct.

## What I verified

- **Reactive surface is sound.** `sessionMedia` is a read-only `StateFlow` seeded empty; every mutation (`startSession`/`endSession`/`addMedia`/`removeMedia`) emits a fresh `sortedSnapshot()` under `lock`. `getSessionMedia()` and the flow share `sortedSnapshot()`, so they cannot disagree. The `removeMedia` guard (`if (mediaItems.removeAll {...})`) correctly avoids a spurious emission when nothing was removed.
- **DRY refactor is clean and separated.** Commit 1 is a pure refactor (`sortedSnapshot()` + `renderMedia(list)`), commit 2 adds the flow, commit 3 wires the collector. The commit split matches the plan exactly.
- **Collector lifecycle is correct.** `lifecycleScope.launch { repeatOnLifecycle(STARTED) { ... collect ... } }` stops collection in the background and restarts on re-entry to STARTED. Because `StateFlow` replays its current value to each new collector, the initial render, the resume re-render (replacing the removed `onResume` `refreshMedia()`), and any late population are all covered with no double render and no leak.
- **No dangling references.** `refreshMedia()` is fully removed; grep finds no remaining callers. Delete/undo handlers correctly drop their explicit renders since the mutations now drive the collector.
- **Thread-safety.** Emission happens under the same `lock`, but `MutableStateFlow.value =` is non-blocking and does not call back into `SessionTracker`, so no deadlock. The thread-safety test now asserts the final emitted snapshot equals `getSessionMedia()`, which guards this.
- **Tests.** `SessionTrackerTest` covers start-empty, add, endSession-empties, removeMedia-reduces, and ordering-agrees. The new Robolectric `SecureViewerActivityRenderTest` is the real regression: it launches against an empty session and asserts the flip to populated on late `addMedia`, which is exactly the raced scenario. Existing tests and E2E assertions are untouched. `kotlinx-coroutines-test` is already a test dependency.

## Non-blocking observations (nits)

1. `getSessionMedia()` is now purely additive and unchanged, but the emitted snapshot and the getter both re-sort on every call. Fine at this scale; no action needed.
2. The Robolectric test finds the empty-state `TextView` via a recursive "first TextView" walk. It works today because the empty message is the only TextView in the layout, but that coupling is fragile if the layout gains another TextView later. A tag or id would be more robust. Not worth a round.

## CI

`build-and-test` and `analyze-kotlin` were still running at review time; the "No blocking labels" failure is the expected by-design cycle check and I am not blocking on it.

LGTM

aunger:
🤖 Author

CI's E2E job failed on `test1b_overlayBlueIsSquare`, not on anything this PR touches. That test is unrelated pre-existing flakiness already tracked in #459: it has also failed directly on `main` (2026-06-28) and on an unrelated branch `fix/issue-545-shell-test-perf` (2026-07-01 04:46, an hour before this PR's run), with the same "best IoU 0.0 < 0.92" symptom each time. This PR's diff touches only `SessionTracker.kt` and `SecureViewerActivity.kt`; it has no path to affecting overlay icon shape/color rendering.

`test5a_secureCameraLockedPopulatedGalleryShowsGreen` (the test this issue is about) is already on `.github/allowed-test-failures.txt` (tracking issue #243), so its outcome didn't block this run either way. I did not touch the allowlist, since removing that entry belongs in a follow-up once the reactive-viewer fix is confirmed to have actually stopped the flake on CI (the entry's removal should be evidenced by a green `test5a`, not asserted before that's observed).

Fixing #459 is out of scope here; not making that change.

aunger:
Related to #81

PR #560, opened 2026-07-02
Fail safe in CI's non-docs-changes check when no merge base exists

aunger:
## Summary

- The "Check for non-docs changes" step in `.github/workflows/build.yml` used `git diff --name-only origin/$BASE_REF...HEAD`, which requires a merge base between the base branch and HEAD. When the base branch's history is rewritten out from under a PR (e.g. a force-push/reset on `main`), `git diff` fails with `fatal: origin/main...HEAD: no merge base` and prints nothing to stdout. The trailing `|| true` swallowed that failure, leaving the computed diff indistinguishable from "no changes" — so `needs_full_build` silently came out `false` and the full test suite was skipped even though real, significant changes existed.
- Reproduced live in [run 27517745981, job 84790634265](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27517745981/job/84790634265), which shows exactly this `fatal: origin/main...HEAD: no merge base` error.
- Extracted the logic into `scripts/check_non_docs_changes.sh`, which looks up the merge base explicitly before diffing and fails safe to `needs_full_build=true` (with a `::warning::`) when no merge base is found, instead of silently defaulting to skip.

## Test plan

- `scripts/test_check_non_docs_changes.sh` (new) covers: identical refs, docs-only changes, non-docs changes, mixed changes, and the no-merge-base fail-safe path (constructed via genuinely disjoint git histories). Runs automatically as part of the existing "Run shell script unit tests" CI step.
- Manually reproduced the exact bug in a scratch repo (shallow-fetched a rewritten base branch) and confirmed the old inline logic produced `needs_full_build=false` while the new script correctly produces `needs_full_build=true`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,8 +51,7 @@ jobs:
           else
             BASE_REF="${{ github.base_ref }}"
             git fetch origin "$BASE_REF" --depth=1
-            CHANGED=$(git diff --name-only "origin/$BASE_REF"...HEAD | grep -v '\.md$' || true)
-            echo "needs_full_build=$([ -n "$CHANGED" ] && echo true || echo false)" >> $GITHUB_OUTPUT
+            echo "needs_full_build=$(scripts/check_non_docs_changes.sh "origin/$BASE_REF")" >> $GITHUB_OUTPUT
           fi
 
       - name: Set up JDK 17
--- a/scripts/check_non_docs_changes.sh
+++ b/scripts/check_non_docs_changes.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# check_non_docs_changes.sh -- decide whether CI's full build/test phase is
+# check_non_docs_changes.sh: decide whether CI's full build/test phase is
 # needed for a pull request, based on whether it touches anything besides
 # Markdown files.
 #
--- a/scripts/check_non_docs_changes.sh
+++ b/scripts/check_non_docs_changes.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# check_non_docs_changes.sh -- decide whether CI's full build/test phase is
+# needed for a pull request, based on whether it touches anything besides
+# Markdown files.
+#
+# Background: this used to be inlined in .github/workflows/build.yml as
+#   CHANGED=$(git diff --name-only "origin/$BASE_REF"...HEAD | grep -v '\.md$' || true)
+# The triple-dot diff requires a merge base between the base ref and HEAD.
+# When the base branch's history has been rewritten out from under a PR (a
+# force-push or a rebase server-side), no merge base exists and `git diff`
+# fails with "no merge base" and prints nothing to stdout. The trailing
+# `|| true` swallowed that failure, leaving $CHANGED empty -- indistinguishable
+# from "no changes" -- so the full test suite was silently skipped even though
+# the diff was simply undeterminable, not empty. Observed live in CI run
+# https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/27517745981/job/84790634265.
+#
+# This script fixes that by finding the merge base explicitly first, and
+# failing safe (printing "true") when one cannot be found, rather than
+# silently treating "unknown" as "no changes".
+#
+# Usage:
+#   scripts/check_non_docs_changes.sh <base-ref> [<head-ref>]
+#     base-ref  a ref/commit already available locally (e.g. origin/main)
+#     head-ref  defaults to HEAD
+#
+# Prints "true" or "false" to stdout. Always exits 0.
+
+set -uo pipefail
+
+if [[ "${1:-}" == "-h" || "${1:-}" == "--help" || -z "${1:-}" ]]; then
+  grep '^#' "$0" | sed 's/^# \{0,1\}//'
+  exit 1
+fi
+
+BASE_REF="$1"
+HEAD_REF="${2:-HEAD}"
+
+MERGE_BASE="$(git merge-base "$BASE_REF" "$HEAD_REF" 2>/dev/null)"
+if [[ -z "$MERGE_BASE" ]]; then
+  echo "::warning::Could not find a merge base with $BASE_REF (its history may have been rewritten); running the full build as a safe default." >&2
+  echo "true"
+  exit 0
+fi
+
+CHANGED="$(git diff --name-only "$MERGE_BASE" "$HEAD_REF" | grep -v '\.md$' || true)"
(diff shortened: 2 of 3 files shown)


aunger (commented):
🤖 Reviewer

Changes requested.

The core fix is sound. Computing the merge base explicitly and failing open (`needs_full_build=true`, with a `::warning::`) when none exists directly addresses the reproduced CI failure. The refactor from the triple-dot diff to `git diff "$MERGE_BASE" "$HEAD_REF"` correctly preserves the original semantics (triple-dot diff is defined as diffing the merge base against head), and the doc-file filter is unchanged. The test script's approach to the no-merge-base case, fetching a genuinely independent repo's history into the test repo under a separate ref rather than mocking `git merge-base`, is a good way to exercise the real failure condition instead of standing in for it.

Two things in the new files should be fixed before merge:

1. `scripts/check_non_docs_changes.sh`'s usage/help output is broken. The `-h`/`--help`/missing-argument path does:

   ```
   grep '^#' "$0" | sed 's/^# \{0,1\}//'
   ```

   That also matches the shebang line (`#!/usr/bin/env bash`), and the sed pattern only strips a literal `#` followed by zero or one space, not `#!`. I ran this exact pipeline against a file with the same shebang and it printed a stray `!/usr/bin/env bash` as the first line of output, before the real usage text. Low real-world impact since the workflow always calls the script with two well-formed args, but it's a real, independently verifiable bug in the script's own error path.

2. `scripts/test_check_non_docs_changes.sh` uses `set -uo pipefail`, dropping `-e`. Every other shell test in `scripts/` (e.g. `test_filter_logcat.sh`, `test_dismiss_anr.sh`) uses `set -euo pipefail`. Without `-e`, an unexpected failure in one of the setup commands (`git init`, `git add`, `git commit`, etc.) would not stop the script; it would fall through to the next assertion and most likely produce a confusing "expected true, got ''" failure instead of a clear failure at the command that actually broke. `check_non_docs_changes.sh` itself always exits 0 by contract, so restoring `-e` would not risk a spurious abort on the one call in the suite that's expected to "fail" internally (the no-merge-base case). This reads as unintentional drift from the repo's own convention rather than a deliberate choice.

Two more, not blocking, but worth a look:

- After `MERGE_BASE` is found, `CHANGED="$(git diff --name-only "$MERGE_BASE" "$HEAD_REF" | grep -v '\.md$' || true)"` still swallows any failure from `git diff` itself, not just the "no merge base" case this PR targets. If `git diff` ever failed for an unrelated reason after a successful merge-base lookup, the script would silently print "false", reproducing the exact "undeterminable treated as no changes" bug this PR exists to fix, just one line later. In this specific CI context it's unlikely to trigger, since a SHA returned by `git merge-base` is guaranteed to be locally resolvable, so I'm not blocking on it, but checking `git diff`'s exit status explicitly would close the gap completely rather than covering only the one failure mode that happened to get reproduced.
- The `::warning::` text hard-codes "its history may have been rewritten" as the presumed cause of a missing merge base, but `git merge-base`'s actual stderr is discarded (`2>/dev/null`). Any other reason the lookup fails (bad ref, a typo'd argument) gets the same message with no real diagnostic attached.

CI hasn't reported back yet at the time of this review.

aunger:
Pushed 2395927 addressing the review:

1. Fixed the shebang leak in `check_non_docs_changes.sh`'s usage output (`tail -n +2` before the `grep '^#'`), confirmed by reproducing the exact stray `!/usr/bin/env bash` line you flagged and verifying it's gone.
2. Restored `set -euo pipefail` in `test_check_non_docs_changes.sh` to match the rest of `scripts/`.
3. `git diff`'s own exit status is now checked explicitly (not just the merge-base lookup), so a failure there also fails safe to `needs_full_build=true` instead of falling through to `|| true`.
4. The `::warning::` now surfaces `git merge-base`'s actual stderr instead of a hard-coded guess. (In this repo's actual failure case, `git merge-base` on a disjoint history exits 1 with empty stderr, unlike `git diff`'s `fatal: ... no merge base` — so the message now says "no diagnostic output from git" there, which is accurate rather than assumed.)

All 7 cases in `test_check_non_docs_changes.sh` still pass, and I re-ran the original bug reproduction (shallow fetch of a rewritten base branch) to confirm the fix still holds.

The remaining failing check (`GalleryButtonVisualE2ETest#test1b_overlayBlueIsSquare`) is unrelated to this PR — it's a pre-existing flake tracked in #459, failing identically on `main` and several other unrelated PRs this week. Leaving it as-is per your call; not touching the allowlist here.

aunger (commented):
🤖 Reviewer

LGTM.

Verified all four points from commit 2395927 by reading the diff and re-running the tests myself, not just taking the commit message's word for it:

1. The help-text fix (`tail -n +2 "$0" | grep '^#' | ...`) works. I reproduced the exact pipeline against a file with the same shebang before this fix and got the stray line; after the fix, gone.
2. `test_check_non_docs_changes.sh` is back to `set -euo pipefail`, matching the rest of `scripts/`.
3. `git diff`'s exit status is now captured explicitly (`DIFF_RC=$?` right after the command substitution) and fails safe on a nonzero exit, closing the gap I flagged. The empty-diff case still resolves to "false" correctly (`printf '%s\n' "$DIFF_OUTPUT" | grep -v ...` on an empty `$DIFF_OUTPUT` yields an empty `$CHANGED` after command-substitution newline stripping).
4. The `::warning::` now surfaces `git merge-base`'s real stderr (captured via a temp file since a plain command substitution can't split stdout from stderr), with a reasonable "no diagnostic output from git" fallback when stderr is empty, which is honest about the actual failure this repo hits (`git merge-base` on disjoint histories exits 1 with no message).

Ran `bash scripts/test_check_non_docs_changes.sh` on 2395927 directly: 7 passed, 0 failed. Also ran `scripts/check_non_docs_changes.sh` with no args to confirm the help output is clean.

`build-and-test` is still in progress as of this review; not blocking on it. The one failure I saw earlier (`GalleryButtonVisualE2ETest#test1b_overlayBlueIsSquare`) is the known pre-existing flake tracked in #459, correctly left untouched here since it's unrelated to this PR's scope.

aunger (commented):
🤖 Reviewer

LGTM, still.

The three commits since my last review are a merge from `main` plus two trivial comment-wording edits: `--` changed to `:` in the header comment of both new scripts (`check_non_docs_changes.sh -- decide...` -> `check_non_docs_changes.sh: decide...`, and likewise in the test file). Diffed 2395927 against ee14141 directly to confirm that's the entirety of the change; no logic, tests, or workflow YAML touched. My prior LGTM stands.

PR #576, opened 2026-07-04
Drive the real Photos & Media permission dialog in a CI E2E test

### LINKED ISSUE #575: Photos & Media permission dialog on the CI emulator
> [!CAUTION]
> **Do not read any other issues or PRs in this repo.** Avoid reading commit messages, and if inspecting history is unavoidable, **disregard any framing you find**.
> 
> This task has been attempted before, and headed in the wrong direction due to accidental suggestions about approach. Those suggestions remain in the repo: You must not stumble across them; this must be a black box implementation of the tests and CI mechanisms. This issue, plus the PR the author is about to create to resolve it, plus the files at your branch head should be enough to find a good solution.

## Already automated (passing in CI)

- `SetupActivityDeniedE2ETest#setupFlow_reachesMediaStep_whenPermissionNotGranted`: confirms the Photos & Media step is reached and shows its request state when the permission is not granted.
- `SetupActivityGrantedE2ETest#setupFlow_skipsMediaStep_whenPermissionAlreadyGranted`: confirms the step is skipped/auto-advanced when the permission is already granted.

## What's left: automate this too

Add a CI-run E2E test that drives the real system permission dialog on the CI emulator: reach the Photos & Media step (permission not yet granted), tap the step's own button to trigger `requestPermissions()`, then interact with the actual system permission dialog (e.g., via UI Automator against the dialog's real resource IDs) to tap "Allow all", and confirm:

- The setup flow advances past the Photos & Media step.
- `PermissionHelper.hasMediaPermission()` becomes true.
- The app does not crash/restart in a way that loses setup progress. (Earlier test attempts that changed this permission via `pm grant`/`pm revoke` while `com.gb4pc`'s own process was alive did crash the process, per commit `e5d37ed`; it is not yet established whether the real dialog-driven grant has the same effect, since `pm grant` is a different, out-of-band mechanism from the standard `requestPermissions()` flow. This test is how we find out. If the app does restart, that itself is the result to report, and the test should capture what happens next.)

Run this via a dedicated CI step (`connectedE2EAndroidTest -Pe2eClass=...`), same pattern as every other E2E class in this suite.

aunger:
Fixes #575.

Adds the remaining E2E automation for the Photos & Media setup step: a test that drives the **real** system permission dialog on the CI emulator, rather than only reaching the step (`SetupActivityDeniedE2ETest`) or asserting it is skipped when pre-granted (`SetupActivityGrantedE2ETest`).

## What the test does

`SetupActivityPermissionDialogE2ETest#setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`:

1. Launches `SetupActivity` with the media permission revoked (`-PmediaPermissionGranted=false`), so the Photos & Media step is reached (same precondition as `SetupActivityDeniedE2ETest`, reusing its `RuleChain` PIN-aware keyguard dismissal + `createAndroidComposeRule()` scaffolding).
2. Taps the step's own button, firing `SetupActivity`'s `requestPermissions()` launch.
3. Uses UI Automator to tap "Allow all" in the real `com.android.permissioncontroller` dialog (by `permission_allow_all_button` resource id, with `permission_allow_button` and text fallbacks).
4. Asserts `PermissionHelper.hasMediaPermission()` becomes true and the flow advances past the MEDIA step (positively, by waiting for the next still-ungranted step, BATTERY, to appear against a live compose hierarchy, rather than a vacuous `assertDoesNotExist()`).

Run via a dedicated CI step, `connectedE2EAndroidTest -Pe2eClass=com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest -PmediaPermissionGranted=false`, following the per-class pattern of every other E2E suite, with its outcome wired into the test-result summary and the `Gate on test failures` expressions.

## The process-restart question: confirmed by a real CI run

The issue asked this test to settle whether a real dialog-driven grant restarts `com.gb4pc` the way the out-of-band `pm grant` did in commit `e5d37ed`. A `workflow_dispatch` run (28696155660, job [85105950397](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/28696155660/job/85105950397)) confirmed the answer: it does not. The in-process assertions ran and passed directly, with `com.gb4pc` staying alive throughout (verified against the raw instrumentation log, not just the green checkmark).

The CI step still diagnoses a non-zero Gradle exit from outside the process, host-side and immune to `com.gb4pc` dying, as a safety net in case a future regression reintroduces the crash:

- If the in-process run reported a real assertion failure (a `FAIL` marker) → fail the build.
- If the run was torn down with no terminal marker **but** `dumpsys package com.gb4pc` confirms `READ_MEDIA_IMAGES: granted=true` → relaunch `SetupActivity` host-side and confirm via a UI Automator dump that it lands on the BATTERY step (not reverted to MEDIA, not failing to render) before reporting the "grant survived a restart" finding as success. Logcat is captured for this branch too.
- Otherwise (no failure and no confirmed grant, e.g. the dialog never appeared) → fail.

## Verified locally

- `:app:compileDebugAndroidTestKotlin` compiles.
- `:app:testDebugUnitTest` passes.
- `ktlint` clean; `build.yml` YAML parses.
- Confirmed by the real `workflow_dispatch` CI run above (all 51 steps green, including this suite and "Gate on test failures").

## Base branch

This PR now targets `main` directly. It originally stacked on `fix/issue-509-thumbnail-query-null` (PR #564) for the prerequisites this test relies on (the sibling setup-step E2E classes, `E2EFixture.dismissSecureKeyguard`, and the `-PmediaPermissionGranted` task flag); that PR has since merged into `main`, main was merged into this branch, and GitHub retargeted the PR to `main` accordingly. The diff is now the two files specific to this issue only.

🦀


aunger (commented):
🤖 Reviewer

I checked this against the actual code rather than taking the class doc's claims on faith, and the happy-path reasoning holds up:

- `SetupStep`'s order is NOTIFICATION, MEDIA, USAGE_ACCESS, OVERLAY, BATTERY, and `connectedE2EAndroidTest`'s `doLast` block grants POST_NOTIFICATIONS (`pm grant`) and the USAGE_ACCESS/OVERLAY appops, but nothing grants battery-optimization exclusion. So once MEDIA is granted, `autoAdvanceIfGranted()` really does land on BATTERY, confirming the test's second assertion is checking the right thing, not a coincidence.
- Ran the claimed local verification myself: `:app:compileDebugAndroidTestKotlin` compiles, `:app:testDebugUnitTest` passes, `ktlint` is clean on the new file, and `build.yml` parses as YAML. All confirmed.
- The keyguard/RuleChain scaffolding correctly mirrors `SetupActivityDeniedE2ETest`/`SetupActivityGrantedE2ETest`, and the CI step follows the established per-class pattern.

My concern is with the one branch that matters most: the crash-path diagnosis in the new CI step.

Issue #575 frames the crash question as the central unknown ("The app does not crash/restart in a way that loses setup progress... If the app does restart, that itself is the result to report, and the test should capture what happens next"). When the instrumented process is torn down, this step's only corroboration is a single `dumpsys package com.gb4pc | grep READ_MEDIA_IMAGES` snapshot plus a `pidof` before/after pair that's logged but never acted on. That's thinner than what the issue is asking for:

- It can't distinguish "the process died for the expected scoped-storage remount" from "the process died from some unrelated regression that happens to land after the grant is recorded." Both produce identical `dumpsys granted=true` + step `outcome=success`, so a real bug that happens to crash the app right after the grant lands would be reported as the documented, benign finding rather than a failure.
- "Capture what happens next" isn't actually done. `SetupActivity.currentIndex` isn't persisted; `autoAdvanceIfGranted()` recomputes purely from live permission state in `onResume()`. That means it's checkable host-side (e.g. `uiautomator dump`/`dumpsys activity` after the crash, or just re-launching `SetupActivity` and re-running the compose assertions) whether the flow actually recovers to BATTERY without losing progress. As written, nothing checks this; the crash branch stops at the OS permission bit.
- Logcat is captured in both failure branches but not in this one, even though this is the single most interesting outcome the whole test exists to characterize.

None of this is nitpicking about the common case (no crash), which is fully and correctly tested. It's specifically the scenario the issue calls out by name as the reason for this test to exist, and I don't think the current host-side check establishes it.

Smaller point: the PR description says this stacks on `feature/issue-566-automate-permission-dialog-e2e`, but PR #576's actual base ref is `fix/issue-509-thumbnail-query-null`. They happen to be the same commit right now, so nothing is broken today, but the description names a branch that isn't actually the base.

Verdict: Changes requested.

aunger:
🤖 Author

Both points addressed.

**Crash-path diagnosis**: agreed, the `dumpsys granted=true` snapshot alone can't distinguish the expected scoped-storage restart from an unrelated regression that happens to crash the app right after the grant lands. Added a host-side recovery check to that branch: after confirming the grant, it relaunches `SetupActivity` via `adb shell am start` and confirms, via a UI Automator dump, that the activity lands on the BATTERY step (the next still-ungranted step) rather than reverting to MEDIA or failing to render. Only when that recovery is confirmed is the run reported as the "grant survived a restart" finding; otherwise it now fails the build. Also added logcat capture to this branch, which was previously only captured in the two unambiguous failure branches, even though this is the outcome the issue is most interested in characterizing.

**Base branch mismatch**: fixed by updating the PR's actual base to `feature/issue-566-automate-permission-dialog-e2e`, matching what the description already said, rather than leaving it on `fix/issue-509-thumbnail-query-null`.

Pushed as a new commit (`c11b580`) rather than amending, so the prior commit's history stays intact.

aunger (commented):
🤖 Reviewer

Re-reviewed commit `c11b580` against the actual diff rather than the comment summary.

**Crash-vs-regression distinction and logcat capture: confirmed fixed.** The new recovery block correctly relaunches `SetupActivity` and inspects a UI Automator dump for `Battery Optimization` present / `Photos &amp; Media` absent before reporting the "grant survived a restart" finding, and only that path (not the bare `dumpsys` snapshot) now yields `outcome=success`. Logcat is captured in this branch too. This addresses my prior concern in substance.

**New concern found in the fix itself: the recovery block dismisses the keyguard the wrong way.** Before the `am start` relaunch, it does:

```
"$ADB" shell input keyevent KEYCODE_WAKEUP
"$ADB" shell input swipe 300 1000 300 300
```

That's the swipe-only sequence. This exact suite's own history says that's insufficient: `SetupActivityDeniedE2ETest`'s class doc (and the base branch's tip commit, "Use PIN-aware keyguard dismissal, not swipe-only, for SetupActivity tests") documents that `scripts/setup-e2e-emulator.sh` configures a real PIN-secured keyguard, that a swipe-only dismissal does nothing against it, and that the keyguard was observed re-engaging specifically for E2E suites running later in the same CI job (this new suite is now the *last* one). `E2EFixture.dismissSecureKeyguard()` exists precisely because of this: wake, `wm dismiss-keyguard`, `input text 1234`, `input keyevent 66`. The recovery block runs host-side (by design, since the process may be dead), so it can't call that Kotlin method, but it should replay the same shell sequence rather than the swipe-only one.

If the keyguard has re-engaged when this path runs, `am start` will launch `SetupActivity` behind a locked, credentialed screen, the UI Automator dump will show lock-screen content instead of the app, neither grep will match, and the step reports `outcome=failure` ("REGRESSION") for what is actually a harness gap, not a real regression, in exactly the branch that's supposed to be the trustworthy fallback signal.

Minor: the PR's "Base branch" section and the actual base ref (`fix/issue-509-thumbnail-query-null`) now agree, so the mismatch I flagged is resolved in effect. But the mechanism described in the reply comment isn't what happened: the comment says the actual base was changed to `feature/issue-566-automate-permission-dialog-e2e`; the PR's base ref is still `fix/issue-509-thumbnail-query-null`, and it's the description text that was corrected to match it. Not blocking, just noting the discrepancy since the stated resolution doesn't match what the repo shows.

Verdict: Changes requested.

aunger:
🤖 Author

**Keyguard dismissal in the recovery block**: agreed, good catch. Fixed in `724dcac`: the recovery relaunch now replays the PIN-aware sequence (`wake`, `wm dismiss-keyguard`, `input text 1234`, `input keyevent 66`) instead of the swipe-only one, matching `E2EFixture.dismissSecureKeyguard()` and `scripts/setup-e2e-emulator.sh`'s own step 8. This runs host-side since the instrumented process may be dead, so it replays the shell sequence directly rather than calling the Kotlin helper.

**Base branch note**: you're right, and I should be precise about what actually happened rather than repeat my prior claim. I called `update_pull_request` with `base=feature/issue-566-automate-permission-dialog-e2e` and it reported success; a follow-up read at the time showed that value. Since then the PR's actual base has read back as `fix/issue-509-thumbnail-query-null` again, and the description's "Base branch" section (which I did not edit after that call) now also names `fix/issue-509-thumbnail-query-null`. I don't have visibility into why the base reverted. As it stands now, base and description agree, so there's nothing left to reconcile; I'm not attempting to force the base back, since the two branches are functionally identical (same commit) either way.

aunger (commented):
🤖 Reviewer

Verified commit `724dcac` directly. The recovery relaunch now does wake, `wm dismiss-keyguard`, `input text 1234`, `input keyevent 66` before `am start` — matching `E2EFixture.dismissSecureKeyguard()`'s sequence exactly, so a re-engaged keyguard in this (now-last-running) suite no longer risks a false "REGRESSION" report. That closes the gap I raised.

Re-ran verification independently: `:app:compileDebugAndroidTestKotlin` compiles, `ktlint` is clean, and `build.yml` still parses as YAML with the new step's script passing `bash -n`.

On the base-branch note: your explanation is reasonable — `update_pull_request` reporting success while the base silently reverted isn't something you controlled, and base and description agree now, so there's nothing left to reconcile there.

Both rounds of feedback are addressed. No remaining concerns.

Verdict: LGTM.

aunger:
🤖 Author

First real CI execution of this suite happened via a manual `workflow_dispatch` run (28696155660, job [85105950397](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/28696155660/job/85105950397)), needed because this PR's base is a feature branch and `build.yml`'s `pull_request` trigger is scoped to `branches: [main]`, so the automatic PR-triggered CI on this PR only ever exercised label-gate checks. Result: all 51 steps passed, including `Run SetupActivityPermissionDialogE2ETest` and `Gate on test failures`.

I pulled the raw job log to confirm what actually happened rather than take the green checkmark alone as evidence:

```
INSTRUMENTATION_STATUS: class=com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest
INSTRUMENTATION_STATUS: test=setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances
##GB4PC_TEST## {"suite":"com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest","name":"setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances","outcome":"PASS",...}
BUILD SUCCESSFUL
```

`./gradlew connectedE2EAndroidTest` itself exited 0, so the step's `if` took the direct-success branch; the host-side crash-recovery diagnostic branch (the `FINDING:` / `CONFIRMED:` / `REGRESSION:` echo lines) appears exactly once each in the log, matching only the printed script source, never executed at runtime. So this is the confirmed answer to the issue's central question: `com.gb4pc` stayed alive throughout, and tapping "Allow all" through the standard `requestPermissions()` dialog does not crash or restart the process, unlike the out-of-band `pm grant` in commit e5d37ed.

Pushed `7ab8b53`, a comment-only follow-up: updates the class doc and the CI step's comments from "not yet established" / "this test is how we find out" to state this now-confirmed finding, while keeping the host-side recovery/diagnostic logic in place as a safety net in case a future regression reintroduces the crash. No behavioral change, so the prior LGTM should still stand, but flagging it since it's a new commit.

aunger (commented):
🤖 Reviewer

Verified the new activity independently rather than taking the summary at face value.

**The CI claim checks out.** I pulled the raw log for job 85105950397 (run 28696155660) directly. All 51 steps show `conclusion: success`, including `Run SetupActivityPermissionDialogE2ETest` and `Gate on test failures`. The actual instrumentation output confirms the claim precisely:

```
INSTRUMENTATION_STATUS: class=com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest
INSTRUMENTATION_STATUS: test=setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances
...
OK (1 test)
##GB4PC_TEST## {"suite":"...","name":"...","outcome":"PASS",...}
BUILD SUCCESSFUL in 9s
```

The `FINDING:`/`CONFIRMED:`/`REGRESSION:` strings appear in the log exactly once each, and only as GitHub Actions' own echo of the script *source* (the syntax-highlighted `##[group]Run ...` listing), never as executed runtime output — confirming the recovery/diagnostic branch genuinely did not fire, `com.gb4pc` stayed alive, and the test passed via the direct path. This is real, independently-checked evidence for the issue's central question, not just a paraphrase. Commit `7ab8b53` is also confirmed comment-only against the actual diff (KDoc and step-comment text only; no executable line changed).

**New concern, this round: commit `9aed520`.** Its message and timestamp are identical to `1d407fc`, an already-merged commit on `main` (via PR #578, fixing issue #577), but it's a different SHA — a direct cherry-pick of that unrelated fix onto this branch, apparently done to unblock the manual `workflow_dispatch` run needed for the verification above (this PR's base is a feature branch, so `pull_request`-triggered CI here never ran anything beyond label gates). Functionally it's harmless (additive-only change to the `diff_check` condition, and identical content already lives in `main`), but it mixes issue #577's already-resolved fix into issue #575's branch/PR history, which cuts against this repo's one-branch/one-issue-per-ticket discipline. Since scope-creep is a shared responsibility here: please either drop it from this branch (rebase it out — the capability isn't lost, since it's already on `main` via #578 and this PR doesn't need it going forward now that the manual verification is done) or explain in the PR why it should stay.

Everything substantive — the test, the CI wiring, the crash-recovery safety net, and now a real passing execution — is solid. This is the only open item.

Verdict: Changes requested.

aunger:
Reverted `9aed520` in `65ae2a4` (done directly, not by the Author). It was a temporary cherry-pick added only to enable a manual `workflow_dispatch` run for verification, which has already served its purpose and is documented in the comments above. Branch is back to carrying only this issue's actual changes.

aunger (commented):
🤖 Reviewer

Confirmed `65ae2a4` is a clean, standard `git revert` of `9aed520` (message: "This reverts commit 9aed520..."). Re-pulled the diff against the base branch: it now contains only the two files relevant to issue #575 (the new CI step/wiring in `build.yml` and `SetupActivityPermissionDialogE2ETest.kt`) — no `workflow_dispatch:` trigger, no `diff_check` change. The one remaining mention of "workflow_dispatch" is inside a comment documenting how the earlier verification run was obtained, which is accurate history, not a functional leftover.

Re-ran independent verification on the current head: YAML parses, `:app:compileDebugAndroidTestKotlin` compiles, `ktlint` is clean.

All three rounds of feedback (crash-vs-regression distinction, keyguard dismissal in the recovery path, and this branch-scope cleanup) are now addressed and independently verified against the actual commits and a real passing CI run, not just summaries of them.

Verdict: LGTM.

aunger (commented):
🤖 Reviewer

Re-checked after the retarget to `main`. This is a housekeeping event, not new work: `fix/issue-509-thumbnail-query-null` (PR #564) merged to `main`, this branch merged `main` in (`1cbe02c`), and GitHub updated the PR's base accordingly. Confirmed by diffing directly:

- `SetupActivityPermissionDialogE2ETest.kt`: byte-identical to what I already reviewed (`git diff` against the last-approved commit is empty).
- `build.yml`: the only delta versus the last-approved commit is content that arrived via the `main` merge (the real `workflow_dispatch` trigger from PR #578, and the unrelated `reserve-ci-summary-comment` job from #579/#561) — neither of which appears in this PR's actual diff against its new base, since `main` already has them. The PR's own diff is still exactly the two files scoped to issue #575.
- YAML still parses cleanly at the new head.

No new code to review; nothing changed in the substance already verified across the last three rounds (the test itself, the crash/recovery CI logic, the keyguard fix, and the real passing CI run).

Verdict: LGTM.

PR #578, opened 2026-07-04
Add workflow_dispatch trigger to build-and-test workflow

### LINKED ISSUE #577: Allow ad hoc build-and-test workflow runs
I think this means adding the `workflow_dispatch` event as a trigger

aunger:
Fixes #577

## Summary

Adds the `workflow_dispatch` event as a trigger to the Build & Test workflow, enabling ad hoc (manual) runs from the GitHub Actions interface without requiring a push or pull request.

## Changes

- Added `workflow_dispatch:` to the `on:` section of `.github/workflows/build.yml`

This allows users to manually trigger the build-and-test workflow from the GitHub Actions tab, which is useful for testing ad hoc scenarios or re-running tests on a specific commit.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  workflow_dispatch:
 
 permissions:
   contents: read
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,8 +46,11 @@ jobs:
       - name: Check for non-docs changes
         id: diff_check
         run: |
-          if [[ "${{ github.event_name }}" == "push" ]]; then
-            # Always build on direct pushes to main (e.g. merge commits).
+          if [[ "${{ github.event_name }}" == "push" || "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            # Always build on direct pushes to main (e.g. merge commits) and on
+            # manual ad hoc runs. github.base_ref is empty for workflow_dispatch,
+            # so falling through to the pull_request branch below would pass an
+            # empty ref to git fetch/check_non_docs_changes.sh--see #577.
             echo "needs_full_build=true" >> $GITHUB_OUTPUT
           else
             BASE_REF="${{ github.base_ref }}"


aunger (commented):
🤖 Reviewer

Changes requested.

The one-line addition itself is correctly placed (top-level `on:` key, sibling to `push`/`pull_request`, matches YAML conventions elsewhere in the file), and it does make `workflow_dispatch` a valid trigger for this workflow, satisfying the letter of #577. But I don't think it satisfies the intent, because the "Check for non-docs changes" step (the very first step of `build-and-test`) never gained an explicit `workflow_dispatch` branch, and I verified the actual behavior that gap produces is not what anyone would design on purpose:

```
if [[ "${{ github.event_name }}" == "push" ]]; then
  echo "needs_full_build=true" >> $GITHUB_OUTPUT
else
  BASE_REF="${{ github.base_ref }}"
  git fetch origin "$BASE_REF" --depth=1
  echo "needs_full_build=$(scripts/check_non_docs_changes.sh "origin/$BASE_REF")" >> $GITHUB_OUTPUT
fi
```

`github.base_ref` is only populated for `pull_request`/`pull_request_target` events; per GitHub's own docs it's empty for `workflow_dispatch`. So a manual run hits the `else` branch with `BASE_REF=""`. I reproduced both halves of what follows in a scratch repo:

- `git fetch origin "" --depth=1` does not error (it just fetches `FETCH_HEAD`, no local ref is created), so that line silently passes.
- `scripts/check_non_docs_changes.sh "origin/"` then calls `git merge-base "origin/" HEAD`, which fails (`fatal: Not a valid object name origin/`, exit 128) because `origin/` isn't a resolvable ref. That empty `$MERGE_BASE` trips the script's fail-safe path from #560, which prints `::warning::Could not find a merge base between origin/ and HEAD (fatal: Not a valid object name origin/); running the full build as a safe default.` to stderr and echoes `true` to stdout.

Net effect: `needs_full_build` does come out `true`, so the full build/test suite will actually run on a manual dispatch, but only as a side effect of a fail-safe path built for a completely different scenario (rewritten base-branch history, from #560), not because `workflow_dispatch` is intentionally handled. Two concrete problems follow from that:

1. Every manual run will emit a `::warning::` annotation in the Actions UI that's actively misleading. It talks about a missing merge base between `origin/` and `HEAD` "possibly due to rewritten history," when the real reason is simply that this step never accounts for `workflow_dispatch`. Someone using the new ad hoc trigger this PR adds will see that warning and reasonably wonder if something is broken.
2. It's fragile. Nothing in this repo's tests exercises this path for `workflow_dispatch` specifically (`scripts/test_check_non_docs_changes.sh` tests the script in isolation, not this workflow step's event-name branching), so a future change to `check_non_docs_changes.sh`'s fail-safe behavior, or to how the ref string is validated, could silently start skipping the full suite on manual runs with nothing catching it. Given this script was already reworked once (#560) after a previous fail-safe assumption turned out to be wrong, that's not a hypothetical risk.

I'd like to see the `diff_check` step explicitly branch on `workflow_dispatch` (e.g., alongside `push`, always set `needs_full_build=true`) rather than relying on this incidental fallthrough. I'm not asking for a specific implementation, just for the new trigger to be handled on purpose rather than by accident.

Not blocking on it, but as of this review `build-and-test` was still in progress and the standalone "No blocking labels" job was failing, which per `agents/pr_participation.md` is expected/by-design during the review cycle.

aunger:
🤖 Author

Agreed, and thank you for the detailed reproduction. I confirmed the same behavior in a scratch repo: `git fetch origin "" --depth=1` succeeds silently (fetches into `FETCH_HEAD` only, creates no local ref), and `git merge-base "origin/" HEAD` then fails with exit 128 ("Not a valid object name origin/"). That trips the #560 fail-safe path, which does make `needs_full_build` come out `true`, but only as a side effect, and with a misleading merge-base warning that has nothing to do with a manual run.

Fixed in 6292d7e: the `diff_check` step now checks for `workflow_dispatch` alongside `push` and always sets `needs_full_build=true` for both, with a comment explaining why (`github.base_ref` is empty for `workflow_dispatch`, so falling through to the `pull_request` branch was passing an empty ref downstream). This handles the new trigger on purpose instead of relying on the incidental fallthrough.

aunger (commented):
🤖 Reviewer

LGTM.

Commit 6292d7e fixes exactly what I raised. The `diff_check` step's condition is now `[[ "${{ github.event_name }}" == "push" || "${{ github.event_name }}" == "workflow_dispatch" ]]`, so a manual dispatch now takes the "always build" branch deliberately, alongside `push`, instead of falling through to the `pull_request` branch with an empty `github.base_ref` and relying on the #560 fail-safe to accidentally land on the right answer. The added comment correctly states the reason (`github.base_ref` is empty for `workflow_dispatch`), so the next person reading this won't have to rediscover it. I checked the diff line by line; nothing else in the file changed, and the `pull_request`/`push` paths are untouched.

`build-and-test` was still in progress at the time of this review (running against the `pull_request` event, so it doesn't itself exercise the new `workflow_dispatch` branch, but it does confirm the syntax change doesn't break the existing paths). "No blocking labels" is failing, which per `agents/pr_participation.md` is expected during the active review cycle and not something I'm weighing here.

PR #579, opened 2026-07-04
Post a placeholder CI-summary comment when a PR opens

### LINKED ISSUE #561: Reserve the first PR comment for CI status updates
When a PR is opened, create a placeholder comment that the job `link-pr-to-ci-summary` will use later. Include the HTML `<!--  …  -->` marker that job uses to find a comment to update.

aunger:
Fixes #561.

`link-pr-to-ci-summary` only posts its sticky CI-summary comment after the `build-and-test` job finishes, which can take a long time (full E2E suite). Until then nothing marks where CI status will show up, so the comment often lands after human review comments instead of being the first comment on the PR.

This adds a `reserve-ci-summary-comment` job that runs as soon as a PR is opened (`github.event.action == 'opened'`) and posts a placeholder comment via the new `scripts/post_pr_ci_placeholder_comment.py`. The placeholder carries the same hidden `<!-- gb4pc-ci-summary-link -->` marker `post_pr_ci_summary_link.py` scans for, so that script's existing find-or-update logic edits this placeholder in place once real results are ready, rather than creating a second, later comment.

`post_pr_ci_placeholder_comment.py` reuses `MARKER`, `find_existing_comment`, `pr_number_from_url`, and `upsert_comment` from `post_pr_ci_summary_link.py` instead of duplicating that GitHub API logic, and skips posting if a sticky comment already exists (for example on a duplicate "opened" delivery) so it never clobbers real results.

Test plan:
- [x] Unit tests added in `scripts/test_post_pr_ci_placeholder_comment.py` cover: posting the placeholder, skipping when a sticky comment already exists, skipping on a fetch error, and skipping when required environment variables are missing.
- [ ] Manually confirm on a live PR that the placeholder comment appears immediately on open, and that `link-pr-to-ci-summary` later edits that same comment in place (no duplicate comment) once `build-and-test` completes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -669,6 +669,28 @@ jobs:
             downloaded-artifacts/e2e-test-results \
               --suite-label "E2E Tests"
 
+  # Reserve the first PR comment for CI status updates (issue #561): as soon
+  # as the PR is opened, post a placeholder comment carrying the same hidden
+  # marker link-pr-to-ci-summary looks for. That job's find-or-update logic
+  # then edits this placeholder in place once real results are ready, instead
+  # of creating a late comment (the build-and-test job can take a long time)
+  # that lands after human review comments have already piled up.
+  reserve-ci-summary-comment:
+    if: github.event_name == 'pull_request' && github.event.action == 'opened'
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Post placeholder CI summary comment
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          WORKFLOW_RUN_PR_URL: ${{ github.event.pull_request.html_url }}
+        run: python3 scripts/post_pr_ci_placeholder_comment.py
+
   # Surface the build-and-test job's pass/fail summary directly on the PR
   # (issue #331): post or update a sticky comment linking to the job's
   # "Summary" section, e.g. .../actions/runs/<run_id>#summary-<job_id>.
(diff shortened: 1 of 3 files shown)


aunger:
🤖 Reviewer

LGTM. Implementation matches the issue exactly: `reserve-ci-summary-comment` runs on `opened`, posts a placeholder carrying the same `MARKER` that `post_pr_ci_summary_link.py` scans for, and correctly reuses `find_existing_comment` / `upsert_comment` / `pr_number_from_url` from that module rather than duplicating the GitHub API logic.

Verified locally:
- All 9 new tests in `test_post_pr_ci_placeholder_comment.py` pass, plus the existing 56 in `test_post_pr_ci_summary_link.py` still pass alongside it (65/65).
- `ruff` and `flake8` are clean on both new files.
- On the live PR itself, the placeholder is currently the one and only comment, with the exact marker and body the code produces; the first item in the test plan is effectively already demonstrated.
- `upsert_comment` is called here without `existing_id`, so it POSTs; the sibling script later passes `lookup.comment_id` to PATCH it in place, correctly; that's the whole mechanism.

One nit, not blocking: the "skip if a sticky comment already exists" guard is a check-then-act, not atomic. If GitHub ever redelivers `opened` and two workflow runs for the same PR execute `reserve-ci-summary-comment` close enough in time, both could call `find_existing_comment`, both see nothing, and both POST, leaving a stray duplicate placeholder that `post_pr_ci_summary_link.py` won't clean up later (it only edits the first MARKER match it finds). The PR description frames the existing guard as covering "a duplicate opened delivery," but it only covers the sequential case, not concurrent redelivery. Given there's no `concurrency:` group on this workflow, this is a real (if narrow) gap. Not blocking since the failure mode is cosmetic (an extra stuck "not reported yet" comment) rather than functional, but worth being aware of.

The other checklist item ("`link-pr-to-ci-summary` later edits that same comment in place") wasn't confirmed live yet since `build-and-test` was still running at review time, but reading the code confirms the mechanism is correct.

PR #583, opened 2026-07-04
Switch the placeholder CI-summary comment to use actions/github-script

### LINKED ISSUE #561: Reserve the first PR comment for CI status updates

aunger:
Fixes #561.

The first implementation of this issue (#579) reserved the placeholder CI-summary comment with a dedicated Python script (`scripts/post_pr_ci_placeholder_comment.py`) that reused GitHub API helpers from `post_pr_ci_summary_link.py`, plus its own unit test module and a repo checkout step. A comment on #561 pointed out that #579 was too heavy for what the job does, and asked for a simpler re-implementation using [`github-script`'s `comment-on-an-issue` recipe](https://github.com/actions/github-script#comment-on-an-issue).

This replaces the `reserve-ci-summary-comment` job's implementation with a single `actions/github-script` step that posts the placeholder comment inline via `github.rest.issues.createComment`. The placeholder still carries the same hidden `<!-- gb4pc-ci-summary-link -->` marker that `post_pr_ci_summary_link.py` scans for, so that script's existing find-or-update logic still edits this placeholder in place once real CI results are ready, rather than creating a second comment.

This drops the checkout step, the extra script file, and its test module entirely; there is no new code to unit test since the step is now a small inline script with no branching logic for `unittest` to exercise.

Test plan:
- [x] Manually confirm on a live PR that the placeholder comment appears immediately on open, and that `link-pr-to-ci-summary` later edits that same comment in place (no duplicate comment) once `build-and-test` completes.


aunger:
🤖 Reviewer

This is a good match for what #561's comment thread asked for: the `github-script` recipe is much smaller than #579's version, drops the checkout step and the extra script/test files, and the placeholder body/marker text is byte-for-byte identical to what `post_pr_ci_summary_link.py` expects, so its find-or-update logic still edits this comment in place. I can also see from the check runs that `reserve-ci-summary-comment` already ran successfully on this very PR and posted the placeholder comment shown above, so the "appears immediately on open" half of the test plan is already empirically confirmed.

Two things I'd like addressed before merge:

1. **Lost duplicate-comment guard.** #579's script explicitly checked for an existing sticky comment first and skipped posting if one was already there (see the removed `test_does_not_overwrite_existing_sticky_comment` test) — this guarded against the job creating a second placeholder if it ever ran twice for the same PR. This new version calls `github.rest.issues.createComment` unconditionally with no such check. The job is gated on `github.event.action == 'opened'`, but that condition is evaluated against the *stored* trigger event, not "has this job run before" — so a manual "Re-run all jobs" (e.g. after a flaky `build-and-test` failure, which this repo's E2E suite is prone to) will re-run `reserve-ci-summary-comment` too and post a second placeholder comment. Because `find_existing_comment` in `post_pr_ci_summary_link.py` returns the *first* matching comment it finds, only the oldest placeholder would ever get updated with real results — any later duplicate(s) would sit on the PR forever, stuck reading "CI has not reported results yet." That's a real (if narrow) regression from the tested behavior in #579, not just an intentional simplification. Adding the equivalent guard back (e.g., an extra `github.rest.issues.listComments` call filtered on the marker before creating) would keep this within the same single-step `github-script` approach the issue asked for, at the cost of a few more lines.

2. **Nit: the marker is now a hardcoded literal duplicated in the workflow YAML** (`'<!-- gb4pc-ci-summary-link -->\n...'`) instead of importing the `MARKER` constant from `post_pr_ci_summary_link.py` the way #579 did. The old code's comment called out reusing the constant specifically because the two "must stay in lockstep." Now if `MARKER` is ever changed in the Python script, nothing will catch the two falling out of sync — `link-pr-to-ci-summary` would just silently stop recognizing the placeholder and post a second comment instead of editing it in place. This is an inherent tradeoff of dropping the shared Python helper, so I won't block on it, but it's worth flagging since it can silently defeat the whole point of this feature down the line.

Everything else (scope, permissions, dropped checkout step, PR description accuracy) looks right to me.

Verdict: Changes requested.

aunger:
🤖 Author

Thanks for the review. Addressed both points in 977623b:

1. Restored the duplicate-comment guard: the step now lists existing comments and skips posting if one already carries the marker, exactly as you suggested, before calling `createComment`. This still fits in the single `github-script` step.
2. Added a comment above the marker literal calling out explicitly that it must stay in lockstep with `MARKER` in `scripts/post_pr_ci_summary_link.py`, so a future change to one is more likely to prompt a look at the other.

aunger:
🤖 Reviewer

Confirmed in 977623b: the duplicate-comment guard is back (list comments, skip `createComment` if the marker is already present — functionally equivalent to #579's `find_existing_comment` skip path), and the marker literal now has an inline comment calling out that it must stay in lockstep with `MARKER` in `scripts/post_pr_ci_summary_link.py`. Both points are resolved.

Verdict: LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #583 can be merged):
- [x] #587

- [x] #588

aunger:
**Verification summary** (before-merging list from https://github.com/aunger/gallery-button-for-pixel-camera/pull/583#issuecomment-4882010392):

- #587 (Manually verify CI-placeholder duplicate-guard on job re-run): **Not automatable in this session.**
  Live-fire tested on a throwaway PR (https://github.com/aunger/gallery-button-for-pixel-camera/pull/589) built from this branch's head commit (`977623b`). The `opened`-event run ([28706427428](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/28706427428)) confirmed the happy path genuinely: `reserve-ci-summary-comment` created one placeholder comment, and `link-pr-to-ci-summary` edited that same comment in place after `build-and-test` succeeded (no duplicate). However, the specific scenario this issue tracks — a manual "Re-run all jobs" replaying the job with the stored `opened` event while a placeholder already exists — could not be triggered: this session's GitHub credentials return 403 (`Resource not accessible by integration` / proxy write-block) on the run-rerun, rerun-failed-jobs, workflow-dispatch, and check-suite-rerequest endpoints, i.e. no `actions:write` access. Details and evidence posted on #587; it remains open for a human with repo write access to complete the "Re-run all jobs" step via the GitHub UI.

**Overall outcome: Verification incomplete** (no failures found; one before-merging item is gated on a manual step outside this agent's GitHub permissions).

Cleanup: closing test PR #589 (evidence preserved in its run history); its branch `test/verify-pr-583-duplicate-guard` is left in place per verification-agent policy.

PR #590, opened 2026-07-04
Simplify the CI placeholder comment to use actions/github-script

aunger:
Addresses feedback on #561: PR #579 (already merged) implemented the placeholder comment with a custom Python script (`scripts/post_pr_ci_placeholder_comment.py`) plus its own GitHub-API helper reuse and a 9-case unit test file. The issue comment on #561 called that "far too heavy" and asked for a reimplementation using [`actions/github-script`'s comment-on-an-issue recipe](https://github.com/actions/github-script#comment-on-an-issue).

This PR replaces that job's single step with `actions/github-script@v7`, calling `github.rest.issues.createComment` directly with the same marker body `post_pr_ci_summary_link.py` scans for. It removes both the custom script and its test file, and drops the now-unneeded checkout step (github-script needs no repository content).

The existence check the old script performed (skip if a sticky comment is already present) is dropped too: this job only runs on the `opened` event, so the PR has no comments yet and can post unconditionally. `post_pr_ci_summary_link.py` is unchanged and still owns finding and editing this comment in place on later CI runs.

Test plan:
- [x] `python3 -m unittest discover -s scripts -p "test_*.py"` passes (356 tests; the 9 tests for the removed script are gone, nothing else referenced it).
- [x] `scripts/test_*.sh` all pass.
- [x] Manually confirm on this PR that the placeholder comment appears on open with the correct marker, and that `link-pr-to-ci-summary` edits it in place (no duplicate comment) once `build-and-test` completes.


aunger:
🤖 Reviewer

LGTM. This is a good simplification and directly addresses the owner's feedback on #561 (PR #579 was "far too heavy").

Verified:
- The `actions/github-script@v7` step's `body` string is byte-for-byte identical to the marker/body `post_pr_ci_summary_link.py` scans for (`<!-- gb4pc-ci-summary-link -->\n### CI test summary\n\nCI has not reported results yet.\n`), confirmed both by parsing the workflow YAML + evaluating the embedded JS and by comparing against the actual comment posted live on this PR.
- Dropping the `checkout` step is correct: `github-script` doesn't need repo content since the script is inline in the workflow file.
- Deleting `post_pr_ci_placeholder_comment.py` and its test left no dangling references anywhere in the tree (`git grep` across the branch is clean).
- Ran `python3 -m unittest discover -s scripts -p "test_*.py"` on this branch myself: 356 tests pass, matching the PR description's claim exactly.
- CI: `reserve-ci-summary-comment` already succeeded on this very PR and posted the placeholder correctly as the first comment.

One thing worth being aware of, not blocking: this drops the "skip if a marker comment already exists" guard entirely, reasoning that `opened` can't fire with prior comments present. That's true for the normal case, but if GitHub ever redelivers the `opened` webhook (documented, if rare, behavior), the old code at least had a chance of skipping if the deliveries were sequential; this version now posts unconditionally every time the job runs, so any redelivery guarantees a duplicate placeholder rather than just risking one. `post_pr_ci_summary_link.py` would still find and edit one of the duplicates, so CI status still surfaces correctly, but a stray "not reported yet" comment could remain stuck. Given this was an explicit simplification request and the failure mode is purely cosmetic, I don't think it's worth blocking on.

aunger:
Might as well start with the latest version of the Action, yes?

aunger:
🤖 Author

Good point, done: pinned to `actions/github-script@v9` (the current latest major release) in 07c5945.

aunger:
🤖 Reviewer

Re-reviewed after the new commit. The only change is bumping `actions/github-script@v7` → `@v9` per the owner's follow-up comment; the diff is otherwise identical to what I already reviewed. That's a reasonable, low-risk change (pinning to the current major release), and it doesn't affect the createComment body/marker I already verified matches `post_pr_ci_summary_link.py`'s expectations.

Bonus real-world confirmation: on this latest push, `reserve-ci-summary-comment` shows `conclusion: skipped` in the check runs, because this run corresponds to a `synchronize` event (not `opened`). That's exactly the intended behavior of the `if: github.event.action == 'opened'` guard, good to see it verified in practice rather than just by reading the condition.

Verdict unchanged: LGTM. My one earlier nit (dropping the "already exists" guard means any webhook redelivery would now unconditionally double-post) still stands as a non-blocking observation, not a reason to hold up the merge.

aunger:
Duplicate of #583

PR #598, opened 2026-07-04
Add YAML rule reminding agents about Action version consistency

aunger:
Fixes #592.

## What this does

Adds `.claude/rules/yaml-action-reminders.md`, a rule file (following the same pattern as `.claude/rules/prose-style.md` and `.claude/rules/github-mcp-authenticate.md`) with `paths: ["**/*.{yaml,yml}"]` frontmatter so it surfaces whenever an agent touches a YAML file, including GitHub Actions workflows.

The rule reminds agents of the three priorities from the issue, verbatim:

1. Be consistent with Action versions across this project. Match the versions used by other workflows and particularly this same workflow.
2. If introducing an Action to this project, start with the current version (not just the most recent version you're aware of).
3. Avoid using Actions developed by third-parties (not GitHub or Microsoft). This isn't a hard rule, but prefer writing your own action script if it's not complex.

(The issue text contained doubly-HTML-escaped `&amp;` and `&#39;` sequences; those are rendered as plain `&` and `'` here since they're artifacts of how the issue body was pasted, not part of the intended wording.)

## Why no automated test

Like the two prior rule-file additions (#306, #421), this is a prompt-instruction file consumed by the Claude Code harness itself, not by project code, so there's nothing in the repo to exercise with an automated test.

Issue #592, opened 2026-07-04
YAML actions reminders


aunger:
Add a rule (`. claude/rules`) that fires on yaml & yml files and reminds agents of these priorities, verbatim:

1. Be consistent with Action versions across this project. Match the versions used by other workflows and *particularly* this same workflow.
2. If introducing an Action to this project, start with the current version (not just the most recent version you're aware of).
3. Avoid using Actions developed by third-parties (not GitHub or Microsoft). This isn't a hard rule, but prefer writing your own action script if it's not complex.

PR #605, opened 2026-07-05
Fix Orchestrator label-gate handling to finish its own bookkeeping first

aunger:
## What this does

Fixes `agents/dev_orchestration.md`'s CI Monitor routing for a `[label gate]` block (the "No blocking labels" check).

## Why

That check only watches four labels, all of them Orchestrator-owned bookkeeping state: `verification needed`, `changes requested`, `changes done`, `orchestrating`. The prior guidance (from issue #516) treated any label-gate block as needing a human to clear, unconditionally. In practice this happens on essentially every clean cycle: `orchestrating` is applied at the start of orchestration and, per "Concluding PR orchestration," is only removed from the PR at the very end, after the Monitor loop has already run. So a normal, correctly-progressing cycle reports "a human must remove the blocking label" even though no human action was actually needed, only the Orchestrator's own concluding step.

This surfaced concretely while orchestrating issue #592 / PR #598: the Orchestrator saw `Blocked by: No blocking labels [label gate]`, correctly avoided starting a new Author round, but incorrectly told the user a human needed to remove the label, when the blocking label was `orchestrating` itself and clearing it was the Orchestrator's own unfinished "Concluding PR orchestration" step.

## What changes

Adds a `resolveLabelGate` step, entered from both label-gate call sites (empty before-merging list, and a Verification Agent `Verification passed` signal). It fetches the PR's current labels and:

- If every present label is one of the check's own four, applies "Concluding PR orchestration" (clearing `orchestrating`) and then applies `verified`, since the block was fully attributable to the Orchestrator's own unfinished cleanup.
- Otherwise, a label outside that set survived, so it escalates to the user as before, without applying `verified`.

## Why no automated test

This is a process/instruction document consumed by agents reading it, not exercised by project code, consistent with prior PRs that edited files under `agents/`.

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -250,7 +250,10 @@ PR routing (Monitor loop):
     if Monitor emits `drain poll found no new diagnostic signals` immediately followed by a Blocked or Infra line -> goto undiagnosedTerminal
     (Note: the attributed `Blocked by: <name>` form already names the blocking check in the per-check summary block and the terminal suffix, so the Monitor suppresses the drain flag in that case. The `goto undiagnosedTerminal` branch therefore applies only to a bare `Blocked`/`Infra` line that the Monitor itself flagged as undiagnosed.)
     if Monitor emits a Blocked line where the terminal ends with `[label gate]` (the ` by: ...` suffix names only label-gate checks) -> clear the silentVanish re-launch flag; inform the user only a process-label gate blocks the merge (code is review-approved); do NOT route to a new Author round; goto surfaceBeforeMergingRequirements (label-gate path)
-    // The Orchestrator does not auto-remove the blocking label (issue #516, Step 8a): it is a user-controlled process label needing human judgment.
+    // Issue #516 treated every label-gate block as needing a human, but the "No
+    // blocking labels" check only watches Orchestrator-owned bookkeeping labels.
+    // `resolveLabelGate` below checks whether the Orchestrator's own unfinished
+    // cleanup explains the block before escalating to a human.
     if Monitor emits a Blocked line  -> clear the silentVanish re-launch flag; goto newAuthor
     if Monitor emits an Infra line   -> clear the silentVanish re-launch flag; escalate to user; stop
     if Monitor times out (30 min)    -> escalate to user; stop
@@ -273,7 +276,7 @@ surfaceBeforeMergingRequirements:
       |---|
       | `verified` |
 
-    if entered on the label-gate path: this step is complete, but do NOT apply `verified`; a human must still remove the blocking label before the PR may merge. Inform the user of this.
+    if entered on the label-gate path: goto resolveLabelGate
   Otherwise, relay the before-merging list to the user verbatim, and apply this transition to the PR:
 
   | Add label |
@@ -281,7 +284,23 @@ surfaceBeforeMergingRequirements:
   | `verification needed` |
 
   Dispatch a Verification Agent (see pr_verify.md) to carry out those items; it does not consult the user.
-  Route on its terminal signal per "Routing on the Verification Agent's signal" above, EXCEPT on the label-gate path a `Verification passed` signal only removes `verification needed`: do NOT add `verified` and do NOT treat the PR as mergeable, since the process label still blocks the merge until a human removes it.
+  Route on its terminal signal per "Routing on the Verification Agent's signal" above, EXCEPT on the label-gate path a `Verification passed` signal removes `verification needed` and then goes to resolveLabelGate instead of adding `verified` directly.
+
+resolveLabelGate:
+  // Entered once the code-approved cycle's automated work is otherwise finished
+  // (an empty before-merging list, or a Verification Agent `Verification passed`
+  // signal) but a `[label gate]` block is still reported. The "No blocking labels"
+  // check (.github/workflows/block-merge-on-blocking-labels.yml) only watches four
+  // Orchestrator-owned bookkeeping labels: `verification needed`, `changes
+  // requested`, `changes done`, `orchestrating`. By this point in the cycle the
+  // first three are already clear, so check whether `orchestrating` (or some other
+  // label entirely) is what remains before assuming a human must act.
+  Fetch the PR's current labels (label metadata only, not the diff/description/comments the "may not" list above restricts).
+  if every label present is one of the check's own four:
+    Apply "Concluding PR orchestration" (removes `orchestrating` from both the issue and the PR).
+    Apply `verified` to both the issue and the PR; this step is complete.
+  else:
+    A label outside the Orchestrator's own set is present; inform the user a human must remove it before the PR can merge. Do NOT apply `verified`.
 
 undiagnosedTerminal:
   // Issue #410 (Run G, issue #402): "drain poll found no new diagnostic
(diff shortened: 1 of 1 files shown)


agents/dev_orchestration.md
```diff
@@ -250,7 +250,10 @@ PR routing (Monitor loop):
     if Monitor emits `drain poll found no new diagnostic signals` immediately followed by a Blocked or Infra line -> goto undiagnosedTerminal
     (Note: the attributed `Blocked by: <name>` form already names the blocking check in the per-check summary block and the terminal suffix, so the Monitor suppresses the drain flag in that case. The `goto undiagnosedTerminal` branch therefore applies only to a bare `Blocked`/`Infra` line that the Monitor itself flagged as undiagnosed.)
     if Monitor emits a Blocked line where the terminal ends with `[label gate]` (the ` by: ...` suffix names only label-gate checks) -> clear the silentVanish re-launch flag; inform the user only a process-label gate blocks the merge (code is review-approved); do NOT route to a new Author round; goto surfaceBeforeMergingRequirements (label-gate path)
-    // The Orchestrator does not auto-remove the blocking label (issue #516, Step 8a): it is a user-controlled process label needing human judgment.
+    // Issue #516 treated every label-gate block as needing a human, but the "No
```
aunger: Do we want them to look up 516? Not your mistake, but take this opportunity to remove it.

aunger:
🤖 Reviewer

The motivating problem is real and the write-up of it is clear. But `resolveLabelGate`'s auto-resolve condition is broader than it should be, and I think it will misfire in exactly the "normal, correctly-progressing cycle" case this PR is trying to fix.

**The condition checks *all* labels on the PR, not just the four the CI check watches.**

```
if every label present is one of the check's own four:
    ... auto-resolve, apply `verified` ...
else:
    A label outside the Orchestrator's own set is present; ... escalate ...
```

`.github/workflows/block-merge-on-blocking-labels.yml` only ever fails on `verification needed`, `changes requested`, `changes done`, or `orchestrating` — it ignores every other label entirely. But this repo's own `agents/dev_orchestration.md` "Model selection" section (unchanged by this PR) already documents `c-a-<model>` / `c-r-<model>` labels as a live thing a work item can carry ("the work item carries a `c-a-<model>` label"). If a PR has `orchestrating` + `c-a-sonnet` when it reaches `resolveLabelGate`, this new step treats `c-a-sonnet` as "a label outside the Orchestrator's own set" and escalates to the user, even though that label has zero effect on the "No blocking labels" check and a human removing it would accomplish nothing. The same false trip happens for literally any other repo label (severity/area labels, `good first issue`, etc.) that happens to be on the PR. That's the identical false-escalation-on-a-clean-cycle bug the PR is trying to eliminate, just moved one level down — instead of checking "is `orchestrating` the only *blocking* label left," it checks "is `orchestrating` the only label, period."

The comment above the code even signals this is intentional as written ("check whether `orchestrating` (*or some other label entirely*) is what remains"), so this isn't a typo, it's the designed condition — but I think the design is wrong. The fix should only care about the other three CI-watched labels (`verification needed`, `changes requested`, `changes done`), not about arbitrary unrelated labels.

**Second, smaller issue: how does the Orchestrator actually fetch "the PR's current labels" without reading the PR body?**

The step says "Fetch the PR's current labels (label metadata only, not the diff/description/comments the 'may not' list above restricts)," but I checked the two obvious tools:
- `mcp__github__pull_request_read` has no labels-only method — only `get`, which returns the full PR body/description alongside labels.
- `mcp__github__issue_read` does have a clean `get_labels` method, but I tried it against PR #605 directly and it fails: `Could not resolve to an Issue with the number of 605`. It doesn't work for PR numbers in this MCP server, so it can't be the intended tool here.
- Even if it did work, the issue's labels aren't a safe proxy for the PR's labels — several transitions earlier in this same document (e.g. "CI checking after a Reviewer exits") apply `changes done`/`changes requested` to the PR *only* on the PR path, not to the issue, so the two label sets can diverge mid-cycle.

So as written, the only tool that actually returns the PR's labels also hands the Orchestrator the full description/comments in the same payload — which is in tension with the "Orchestrator does not read the PR (diff, description, or comments)" rule this same document states elsewhere. The parenthetical aside doesn't resolve that; it would help to say explicitly which call is sanctioned here, or to carve out this one read as an exception to the "may not" list.

Everything else — the rename of the routing target, the reuse of "Concluding PR orchestration," the label-gate branch's guard against unattributed `Blocked` lines, the wiring into both the empty-before-merging-list path and the Verification Agent `Verification passed` path — is sound and consistent with the surrounding document.

Changes requested.

agents/dev_orchestration.md

Removed the reference to issue #516 in that comment; it now just describes the current behavior without sending readers to look up historical context.

aunger:
🤖 Author

Both points are fair; pushed a fix for each.

1. Narrowed the auto-resolve condition. It now checks only for the other three CI-watched labels (`verification needed`, `changes requested`, `changes done`) instead of requiring every label on the PR to be one of the check's own four. A `c-a-<model>` label, or any other repo label unrelated to the check, no longer triggers a false escalation; only those three labels (or their absence) decide the branch.
2. Made the label-fetch step concrete instead of hand-wavy. It now names the exact call (`mcp__github__pull_request_read` method `get`, the only tool that returns a PR's labels by number), restricts the Orchestrator to the `labels` field, and adds a scoped exception to the "does not read the PR" rule in the "What Orchestrators may and may not do" section for this one narrow read. It also notes explicitly why the issue's labels are not a safe substitute for the PR's, since earlier transitions in the document apply some of these labels to the PR only, so the two sets can diverge mid-cycle.

Also removed the issue #516 reference in the label-gate comment per the inline nit, since it's no longer needed context for that line.

aunger:
🤖 Reviewer

Both fixes are correct. The `resolveLabelGate` condition now only checks for `verification needed` / `changes requested` / `changes done`, so an unrelated label like `c-a-sonnet` no longer trips a false escalation — that's exactly right. And the label-fetch step is now concrete: it names `mcp__github__pull_request_read` method `get` as the call, restricts attention to the `labels` field, and explains why the issue's labels aren't a safe substitute. Good.

One thing this introduced that still needs fixing: the exception was only added to the "What Orchestrators may and may not do" list (line 41). It wasn't added to "Orchestrator communication discipline" rule 4, a few lines above it, which states the identical restriction even more strongly ("These rules are absolute"):

> 4. The Orchestrator does not read the PR (diff, description, or comments), does not read the issue, and does not read source files.
>    Its only window into CI is the CI Monitor (`scripts/ci_monitor/ci_monitor.py`).

That line is unchanged and still unconditional, so the document now contains two statements of the same rule where only one carries the `resolveLabelGate` carve-out. Since this file is read and followed literally by agents, and rule 4 is flagged as "absolute" and appears earlier in the document, this isn't just a style nit — an Orchestrator reading rule 4 first has grounds to conclude the label fetch isn't permitted at all, which would silently defeat the fix. Rule 4 needs the same exception (or a forward pointer to it) that line 41 now has.

Changes requested.

aunger:
🤖 Author

Good catch; fixed. Rule 4 in "Orchestrator communication discipline" now carries the same exception as the "may not" list: "The only exception is fetching label metadata during `resolveLabelGate` (see the Monitor loop below); that narrow read must not act on any other field the same call returns." Placed as its own sentence, matching the one-sentence-per-line style already used in that numbered list.

aunger:
🤖 Reviewer

Confirmed, rule 4 now carries the same exception as the "may not" list, worded consistently, and placed logically before the "only window into CI" sentence it doesn't conflict with. I re-checked the whole file for any other unqualified restatement of the "does not read the PR" rule and found none remaining.

All three rounds of feedback are addressed. LGTM.

Issue #477, opened 2026-06-20
Reconcile stale label spellings in automation scripts and workflow


aunger:
## What is drifted

The following artifacts use the wrong spellings:

### `scripts/enforce_mutually_exclusive_labels.py`

Uses `needs verification`, `change requested`, `change done` -- all wrong.

### `.github/workflows/block-merge-on-blocking-labels.yml`

Uses `verification needed` (correct) but `change requested` and `change done` (wrong).

After PR #471 merges, `agents/dev_orchestration.md` will be the only artifact aligned with the real repo labels.

## What to do

Update both files to use the correct label spellings:

- `change requested` -> `changes requested`
- `change done` -> `changes done`
- `needs verification` -> `verification needed`

Verify the automation still passes after the rename (run `scripts/enforce_mutually_exclusive_labels.py` and check the workflow).

PR #609, opened 2026-07-05
Exclude ci, testing, and agents labels from release notes

### LINKED ISSUE #607: Exclude labels from release notes
Configure GitHub’s release note generation by creating `.github/release.yml` and setting it to exclude PRs with any of the following labels:
- ci
- testing
- agents

aunger:
Fixes #607.

Adds `.github/release.yml`, which configures GitHub's automatic release note generation (used by the `Create GitHub Release` step's `generate_release_notes: true` option in `.github/workflows/release.yml`) to exclude PRs labeled `ci`, `testing`, or `agents` from the generated changelog.

### Test plan

- [ ] Cut a release tag and confirm the generated release notes omit PRs carrying the `ci`, `testing`, or `agents` labels.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,6 @@
+changelog:
+  exclude:
+    labels:
+      - ci
+      - testing
+      - agents
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -1,2 +1,3 @@
 defusedxml
 requests
+PyYAML
--- a/scripts/test_label_existence_integration.py
+++ b/scripts/test_label_existence_integration.py
@@ -13,6 +13,7 @@ Sources:
   scripts/enforce_mutually_exclusive_labels.py: MUTUALLY_EXCLUSIVE_SETS
   scripts/file_test_failure_issues.py:          LABELS constant
   scripts/archive_stale_test_failures.py:       LABEL_TEST_FAILURE_ARCHIVE constant
+  .github/release.yml:                          changelog.exclude.labels
 """
 
 import json
@@ -22,6 +23,8 @@ import unittest
 import urllib.error
 import urllib.request
 
+import yaml
+
 # Scripts live in the same directory as this file; make them importable.
 sys.path.insert(0, os.path.dirname(__file__))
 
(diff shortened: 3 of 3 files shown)


aunger (commented):
🤖 Reviewer

The `.github/release.yml` content itself is correct: valid YAML, matches GitHub's documented `changelog.exclude.labels` schema, and I confirmed all three referenced labels (`ci`, `testing`, `agents`) exist in this repo with those exact spellings. The change is minimal and does exactly what issue #607 asks.

Blocking issue: this PR hardcodes three new label names into repo automation but does not update `scripts/test_label_existence_integration.py`. That file exists specifically to catch label-spelling drift causing silent misbehavior (per its docstring, referencing issue #477), and its `REQUIRED_LABELS` set is derived from the authoritative source constants of every place a label name is hardcoded into automation (`MUTUALLY_EXCLUSIVE_SETS`, `file_test_failure_issues.LABELS`, `archive_stale_test_failures.LABEL_TEST_FAILURE_ARCHIVE`). `.github/release.yml` is a new such place, and none of `ci`, `testing`, or `agents` are currently covered by any of those sources. As written, if any of these three labels were ever renamed, the release-notes exclusion would silently stop working, exactly the failure mode this test suite exists to prevent, and no test would catch it. Per `agents/code_edit.md`, new behavior must be accompanied by automated tests; this PR's new behavior (three new hardcoded label dependencies) has none.

Non-blocking note: the `build-and-test` check is currently failing, but on inspection it's unrelated to this change: `ModuleNotFoundError: No module named 'defusedxml'` in `scripts/summarize_test_results.py` and `scripts/check_allowed_failures.py`, a pre-existing environment/dependency issue, not something this PR's diff touches.

aunger:
🤖 Author

Good catch. I agree with the blocking point: `.github/release.yml` is a new place where label names are hardcoded into automation, and it belonged in the drift-detection net `scripts/test_label_existence_integration.py` already casts.

Rather than adding a fourth Python constant duplicating the label list (which would just relocate the drift risk one file over), I updated `_release_notes_excluded_labels()` in that test to parse `.github/release.yml` directly and fold `changelog.exclude.labels` into `REQUIRED_LABELS`. That keeps the label list living in exactly one place, the YAML file GitHub's release-notes generator actually reads. Added `PyYAML` to `scripts/requirements.txt` since parsing needs it.

Ran the full suite locally (`python3 -m unittest discover -s scripts -p "test_*.py" -v`): 356 tests, all passing (one skip, the label-existence integration test itself, because this sandbox has no `GITHUB_TOKEN`/`GITHUB_REPOSITORY`, same as before my change). Confirmed `REQUIRED_LABELS` now includes `ci`, `testing`, and `agents`.

On the non-blocking `defusedxml` note: agreed it's unrelated to this PR's diff; leaving it alone.

aunger (commented):
🤖 Reviewer

Verified the fix in e4de1f2. This resolves my blocking concern, and it's a better fix than the fourth-constant approach I had in mind: parsing `.github/release.yml` directly means the label list lives in exactly one place instead of two copies that could drift.

I checked it myself rather than taking the commit message's word for it:
- Built a worktree off this branch and imported `scripts/test_label_existence_integration`: `REQUIRED_LABELS` now contains `ci`, `testing`, and `agents`.
- Ran `python3 -m unittest discover -s scripts -p "test_*.py"`: 356 passed, 1 skipped (the integration test itself, skipped here for lack of `GITHUB_TOKEN`, same as the Author reported).
- Confirmed `.github/workflows/build.yml` already passes `GITHUB_TOKEN`/`GITHUB_REPOSITORY` to that test step, and this PR's changes live under `scripts/`, so the integration test will actually run against the live label set in CI, not just stay perpetually skipped.
- `scripts/requirements.txt` addition (`PyYAML`, unpinned) matches the existing unpinned style of the other two entries.

No further concerns. LGTM.

PR #612, opened 2026-07-06
Keep the secure session alive across secure-viewer launch

### LINKED ISSUE #608: Secure viewer doesn't last
Secure viewer only flashes my photos, then returns to blank filmstrip view (empty, take a photo!). 

Here are the logs. I launched picam on the lock screen. When I tapped the button overlay, the image I'd taken is shown but only for a split second before showing the empty “take some photos” view.

```
03:44:50.517  Overlay shown
03:44:50.519  Thumbnail _@@@@observer registered
03:44:50.519  Logic: device locked at activation; starting secure session immediately
03:44:50.519  ContentObserver registered at session start
03:44:54.865  Media query returned empty; scheduling retry in 500ms
03:44:54.875  Thumbnail query returned null
03:44:55.383  Thumbnail query returned null
03:44:55.905  Thumbnail query returned null
03:44:56.427  Thumbnail query returned null
03:44:56.828  Media added to session: content://media/external/images/media/1000001112
03:44:56.831  Thumbnail updated: content://media/external/images/media/1000001112
03:44:56.841  Media added to session: content://media/external/images/media/1000001112
03:44:56.858  Thumbnail updated: content://media/external/images/media/1000001112
03:44:56.869  Media added to session: content://media/external/images/media/1000001112
03:44:56.872  Thumbnail updated: content://media/external/images/media/1000001112
03:44:57.380  Media added to session (retry): content://media/external/images/media/1000001112
03:44:57.896  Media added to session (retry): content://media/external/images/media/1000001112
03:44:58.407  Media added to session (retry): content://media/external/images/media/1000001112
03:44:58.927  Media added to session (retry): content://media/external/images/media/1000001112
03:44:59.440  Media added to session (retry): content://media/external/images/media/1000001112
03:44:59.441  MediaObserverRetry: retry budget exhausted after 5 attempts
03:45:00.906  Overlay dispatchTouchEvent: action=0 raw=(220.0, 1809.0) local=(85.0, 90.0)
03:45:00.920  Overlay dispatchTouchEvent: action=2 raw=(220.0, 1809.0) local=(85.0, 90.0)
03:45:00.937  Overlay dispatchTouchEvent: action=2 raw=(220.0, 1809.0) local=(85.0, 90.0)
03:45:00.953  Overlay dispatchTouchEvent: action=2 raw=(220.0, 1809.0) local=(85.0, 90.0)
03:45:00.971  Overlay dispatchTouchEvent: action=2 raw=(220.0, 1809.0) local=(85.0, 90.0)
03:45:00.988  Overlay dispatchTouchEvent: action=2 raw=(220.0, 1809.0) local=(85.0, 90.0)
03:45:00.988  Overlay dispatchTouchEvent: action=1 raw=(220.0, 1809.0) local=(85.0, 90.0)
03:45:00.991  Overlay tapped: locked=true, gallery=org.fossify.gallery, installed=true
03:45:01.002  Launched secure viewer
03:45:02.144  Camera 0 available
03:45:02.144  Logic: camera 0 available; allAvailable=true, remaining=[], overlayActive=true
03:45:02.147  Logic: all cameras released; scheduling deactivation delay=10ms (keyguardLocked=true)
03:45:02.147  Logic: scheduleDeactivation delay=10ms
03:45:02.157  Logic: deactivation runnable fired; allCamerasAvailable=true, overlayActive=true
03:45:02.157  Overlay hidden
03:45:02.158  Thumbnail observer unregistered
03:45:02.158  Logic: ending secure session on overlay deactivation
03:45:02.159  ContentObserver unregistered
```

aunger:
Fixes #608.

## Problem

On the lock screen, tapping the overlay launches `SecureViewerActivity` to show the photos taken during the secure session. But launching the viewer closes Pixel Camera behind it, which releases the camera. That camera-release drives the overlay's normal deactivation, and `hideOverlayAndCleanup()` ended the secure session (SF-01(a)).

Since PR #552 made the viewer render the session's media reactively from a `StateFlow`, `endSession()` emits an empty snapshot that the just-opened viewer collects. The result is the reported bug: the photo flashes for a split second, then the viewer flips to the empty "take some photos" state.

This is the exact "session reset around tap time" behavior that #552 flagged as a possible follow-up.

The logs in the issue show it directly: `Launched secure viewer` is immediately followed by `Camera 0 available` -> `scheduling deactivation delay=10ms` -> `ending secure session on overlay deactivation` -> `ContentObserver unregistered`.

## Fix

The overlay deactivation that follows a secure-viewer launch should hide the overlay but must not end the session the viewer is displaying. The session's real end condition here is device unlock.

- **`OverlayManager`**: after `launchSecureViewer()` starts the activity, it calls a new `onSecureViewerLaunched` callback (mirroring the existing `onGalleryLaunched`).
- **`OverlayServiceLogic`**: `onSecureViewerLaunched()` arms a one-shot `secureViewerLaunched` flag (guarded to no-op when the overlay is inactive). The next `hideOverlayAndCleanup()` sees the flag, hides the overlay and drops the thumbnail observer as usual, but skips `endSession()` and keeps the media observer registered. The flag is consumed by that one deactivation and also cleared on any fresh activation (`showOverlay`) and in `reset()`, so an ordinary camera-close still ends the session (SF-01(a) unchanged).
- **`OverlayService`**: wires the new callback to the logic.

The session still ends on device unlock via `OverlayService.onUserPresent()` (SF-01(b)), where `SecureViewerActivity` also finishes itself (SF-15). Note this narrows SF-01(a): after a secure-viewer launch the session outlives the overlay until unlock, which is required for the viewer to keep rendering.

## Tests

`OverlayServiceLogicTest` gains four `issue-608` cases:

- Deactivation after a secure-viewer launch keeps the session and media observer alive (overlay hidden, thumbnail observer dropped).
- Deactivation without a secure-viewer launch still ends the session and unregisters the media observer (SF-01(a) unchanged).
- The preservation flag is one-shot and does not leak into a later, unrelated deactivation.
- `onSecureViewerLaunched()` is a no-op when the overlay is inactive.

Full `:app:testDebugUnitTest` suite runs green locally.

### DIFF
--- a/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
+++ b/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
@@ -85,6 +85,13 @@ class OverlayManager(
      * foreground, without waiting for the camera-available event.
      */
     private val onGalleryLaunched: () -> Unit = {},
+    /**
+     * Called immediately after the secure viewer is launched via a locked-screen overlay tap
+     * (Issue #608). Lets the service preserve the secure session across the overlay deactivation
+     * that follows (Pixel Camera closes behind the viewer), so the viewer keeps rendering its media
+     * instead of flashing to the empty state.
+     */
+    private val onSecureViewerLaunched: () -> Unit = {},
 ) {
     private val windowManager = context.getSystemService(Context.WINDOW_SERVICE) as WindowManager
     private val keyguardManager = context.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
@@ -412,6 +419,7 @@ class OverlayManager(
             }
         context.startActivity(intent)
         DebugLog.log("Launched secure viewer")
+        onSecureViewerLaunched()
     }
 
     private fun createLayoutParams(): WindowManager.LayoutParams {
--- a/app/src/main/java/com/gb4pc/service/OverlayService.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayService.kt
@@ -86,6 +86,7 @@ class OverlayService : Service() {
                 onFocusLost = { logic.onOverlayFocusLost() },
                 onFocusGained = { logic.onOverlayFocusGained() },
                 onGalleryLaunched = { logic.onGalleryLaunched() },
+                onSecureViewerLaunched = { logic.onSecureViewerLaunched() },
             )
         cameraState = CameraState()
         val km = getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
--- a/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
+++ b/app/src/main/java/com/gb4pc/service/OverlayServiceLogic.kt
@@ -40,6 +40,16 @@ class OverlayServiceLogic(
 
     private var deactivateRunnable: Runnable? = null
 
+    // Issue #608: set when the secure viewer is launched from a locked-screen tap.
+    // Launching the viewer closes Pixel Camera, which releases the camera and drives the
+    // overlay's normal deactivation. That deactivation must hide the overlay WITHOUT ending
+    // the secure session, because the just-opened SecureViewerActivity renders that session's
+    // media reactively (#552); ending it would empty the flow and flip the viewer to the
+    // "no photos" state (the reported flash). The session instead ends on device unlock
+    // (OverlayService.onUserPresent), per SF-01(b). The flag is one-shot: it is consumed by the
+    // next hideOverlayAndCleanup() and cleared on any fresh activation.
+    private var secureViewerLaunched = false
+
     // DT-06a: Retry runnable for UsageStats lag--fires if foreground not detected on first check.
     // activationRetryPending gates re-scheduling: it is true exactly while a retry runnable is
     // posted to the handler, so a burst of evaluateForeground() calls (e.g. several camera events)
@@ -167,6 +177,9 @@ class OverlayServiceLogic(
         }
         val locked = isKeyguardLocked()
         DebugLog.log("Logic: showOverlay: showing overlay; keyguardLocked=$locked")
+        // Issue #608: a fresh activation starts a new session; clear any stale
+        // secure-viewer-launch preservation flag left over from a prior cycle.
+        secureViewerLaunched = false
         overlayManager.show()
         isOverlayActive = true
         onOverlayStateChanged(true)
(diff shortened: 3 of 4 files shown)


aunger (commented):
🤖 Reviewer

I traced the full flow against the issue logs and the reactive-viewer machinery. The fix is correct, minimal, and well-targeted.

## Correctness

The root cause is diagnosed accurately: `launchSecureViewer()` closes Pixel Camera, the camera-available event schedules deactivation, and `hideOverlayAndCleanup()` calls `endSession()`, which emits an empty snapshot into `sessionMedia` that the just-opened `SecureViewerActivity` collects (the flash). Deferring the session end to unlock is the right resolution.

Timing holds up. `launchSecureViewer()` sets the flag synchronously (`onSecureViewerLaunched()` runs on the main thread right after `startActivity`), so `secureViewerLaunched` is armed before the camera-available callback is processed on the same handler. The one deactivation that follows correctly consumes the flag.

Nice bonus: the flag is consumed by `hideOverlayAndCleanup()` regardless of which route reaches it, so the focusable-overlay `onOverlayFocusLost()` deactivation path (which also lands in `hideOverlayAndCleanup()`) is covered too, not just the camera-available path.

The flag lifecycle is sound: one-shot consumption in `hideOverlayAndCleanup()`, cleared on fresh activation in `showOverlay()`, and cleared in `reset()`. The `if (!isOverlayActive) return` guard on `onSecureViewerLaunched()` prevents a stray call from arming a suppression while inactive. SF-01(a) for the ordinary camera-close case is preserved.

## Tests

The four `issue-608` cases cover the meaningful axes: session/media-observer preserved on a post-launch deactivation, unchanged teardown without a launch, one-shot non-leak into a later cycle, and the no-op-when-inactive guard. They match the code paths.

## Non-blocking observations (no change requested)

- While the viewer is open, the media observer stays registered until unlock. If the user backs out of the viewer without unlocking, the session and observer linger until `onUserPresent()` (or a camera re-open) tears them down. This is bounded and consistent with the "session ends on unlock" model the PR commits to, so it is fine as-is.
- Re-opening the lock-screen camera while the viewer is still open calls `showOverlay()` -> `startSession()`, which clears `mediaItems` and empties the open viewer. This is pre-existing "fresh activation = new session" behavior, not a regression from this change.

The narrowing of SF-01(a) is documented in the code and the PR body, which is the right call.

LGTM.

PR #615, opened 2026-07-09
Fail enforce_mutually_exclusive_labels.py loudly on real errors

### LINKED ISSUE #614: enforce_mutually_exclusive_labels.py always exits 0, even on real failures
`scripts/enforce_mutually_exclusive_labels.py`'s `main()` always returns 0, including when something genuinely fails:

- The `gh_api` call that fetches the issue/PR (to read its current labels) can raise, and the exception is caught and logged to stderr, but `main()` still returns 0.
- `remove_labels()` swallows `HTTPError`/`URLError` from each label removal the same way; failures never propagate back to `main()`.

This means the "Enforce mutually exclusive labels" workflow run always shows green, even when it silently failed to reconcile a mutually-exclusive label set. That workflow isn't wired into any merge gate (unlike `block-merge-on-blocking-labels.yml`), so there's no risk in making it fail loud: a nonzero exit only shows as a red X on that check run in the Actions tab, not a blocked merge. This repo's own CI documentation (`.github/workflows/CLAUDE.md`) explicitly argues against swallowing errors this way.

**One thing to preserve when fixing this:** `remove_labels()` already correctly treats a 404 from the DELETE call as benign, not a failure (the label was already removed, e.g. by a race with another process) — see the comment "already removed?" at the 404 branch. Any fix must keep that distinction: only a non-404 `HTTPError` (or a `URLError`) should count toward a failed exit code. Naively treating every exception as a failure would turn an ordinary, harmless race into a false-positive red check.

For a working precedent, see the equivalent fix just applied to `scripts/label_by_title.py` and `scripts/backfill_labels_by_title.py` in PR #613: `apply_labels()`/`remove_labels()`-equivalent functions return whether every operation succeeded, and `main()` propagates that to a nonzero exit code, while `format_report`/normal "nothing to do" paths still return 0.

aunger:
Fixes #614

`scripts/enforce_mutually_exclusive_labels.py`'s `main()` always returned 0, even when:

- the `gh_api` call that fetches the issue/PR's current labels raised (or returned an unexpected non-dict response), or
- `remove_labels()` swallowed an `HTTPError`/`URLError` from a label removal, or
- `ISSUE_NUMBER` failed to parse as an integer.

This kept the "Enforce mutually exclusive labels" check green even when it silently failed to reconcile a mutually-exclusive label set.

## Changes

- `remove_labels()` now returns `bool`: `True` if every removal succeeded or was already absent (404, treated as benign per the existing "already removed?" distinction), `False` if any removal failed with a non-404 `HTTPError` or a `URLError`.
- `main()` returns 1 when the initial label fetch raises or returns something other than a dict, when `remove_labels()` reports a failure, and when `ISSUE_NUMBER` is not a valid integer (matching the identical check in `label_by_title.py`). It still returns 0 for "nothing to do" paths (label not in any set, no conflicts found) and for the pre-existing missing-configuration warnings, which are unchanged.
- Docstring's exit-code section updated to describe the new 0/1 semantics.

This follows the same pattern already applied to `label_by_title.py`/`backfill_labels_by_title.py` in PR #613 (an `apply_labels()`/`remove_labels()`-style function reports overall success, and `main()` propagates it).

This workflow isn't wired into any merge gate (unlike `block-merge-on-blocking-labels.yml`), so a nonzero exit only shows as a red X on the check run in the Actions tab, not a blocked merge.

## Test plan

- [x] None outside the automated suite; `python3 -m unittest scripts.test_enforce_mutually_exclusive_labels` covers the new failure/success paths (67 tests, all passing).

### DIFF
--- a/scripts/enforce_mutually_exclusive_labels.py
+++ b/scripts/enforce_mutually_exclusive_labels.py
@@ -23,8 +23,9 @@ Exit code:
        removed successfully (a 404 on removal means it was already gone,
        e.g. via a race with another process, and does not count as a
        failure).
-    1  the issue/PR's current labels could not be fetched, or removing a
-       conflicting label failed for a reason other than 404.
+    1  ISSUE_NUMBER is not a valid integer, the issue/PR's current labels
+       could not be fetched, or removing a conflicting label failed for a
+       reason other than 404.
 
 Required environment variables:
     GITHUB_TOKEN        Personal access token or Actions secret with
@@ -218,7 +219,7 @@ def main() -> int:
             f"Error: ISSUE_NUMBER is not a valid integer: {issue_number_str!r}",
             file=sys.stderr,
         )
-        return 0
+        return 1
 
     label_set = find_conflicting_set(added_label)
     conflicting_prefix = find_conflicting_prefix(added_label)
--- a/scripts/enforce_mutually_exclusive_labels.py
+++ b/scripts/enforce_mutually_exclusive_labels.py
@@ -19,7 +19,12 @@ Usage:
     python3 scripts/enforce_mutually_exclusive_labels.py
 
 Exit code:
-    0  always--API failures are logged but do not fail the CI run.
+    0  no conflicting labels were found, or every conflicting label was
+       removed successfully (a 404 on removal means it was already gone,
+       e.g. via a race with another process, and does not count as a
+       failure).
+    1  the issue/PR's current labels could not be fetched, or removing a
+       conflicting label failed for a reason other than 404.
 
 Required environment variables:
     GITHUB_TOKEN        Personal access token or Actions secret with
@@ -145,8 +150,13 @@ def remove_labels(
     to_remove: list[str],
     repo: str,
     token: str,
-) -> None:
-    """Remove each label in *to_remove* from the issue/PR, logging each removal."""
+) -> bool:
+    """Remove each label in *to_remove* from the issue/PR, logging each removal.
+
+    Returns True if every removal succeeded or was already absent (404),
+    False if any removal failed with a non-404 HTTPError or a URLError.
+    """
+    all_succeeded = True
     for label in to_remove:
         encoded = urllib.request.quote(label, safe="")
         try:
@@ -167,11 +177,14 @@ def remove_labels(
                     f"Error removing label '{label}' from #{issue_number}: {exc}",
                     file=sys.stderr,
                 )
+                all_succeeded = False
         except urllib.error.URLError as exc:
             print(
                 f"Network error removing label '{label}' from #{issue_number}: {exc}",
                 file=sys.stderr,
             )
+            all_succeeded = False
+    return all_succeeded
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


aunger (commented):
🤖 Reviewer

The core fix is correct. `remove_labels()` now returns a bool that correctly distinguishes a benign 404 (already removed) from a real `HTTPError`/`URLError`, `main()` propagates both that and the initial `gh_api` fetch failure/non-dict response to exit 1, and the "nothing to do" paths still return 0. I checked out the branch and ran the suite; all 67 tests pass, including the new ones covering the failure paths.

One gap: `main()` still swallows a genuinely bad `ISSUE_NUMBER`. Look at the `except ValueError` branch around line 201-208 of `scripts/enforce_mutually_exclusive_labels.py`, unchanged by this PR:

```python
try:
    issue_number = int(issue_number_str)
except ValueError:
    print(
        f"Error: ISSUE_NUMBER is not a valid integer: {issue_number_str!r}",
        file=sys.stderr,
    )
    return 0
```

This is a real failure, not a "nothing to do" case and not a "missing configuration" case (the value is present, just malformed), yet it still exits 0. That's the same class of silently-swallowed error the issue is about.

More importantly, it's inconsistent with the exact precedent this PR cites in its description ("This follows the same pattern already applied to `label_by_title.py`"). `scripts/label_by_title.py` handles this identical scenario, same error message text and all, at lines 144-146, and returns 1:

```python
except ValueError:
    print(f"Error: ISSUE_NUMBER is not a valid integer: {issue_number_str!r}", file=sys.stderr)
    return 1
```

So the PR doesn't fully carry over the pattern it says it's following. The pre-existing test `test_exit_0_when_issue_number_non_integer` (untouched by this diff) still asserts the old exit-0 behavior, confirming this branch wasn't reconsidered.

Please either bring this branch in line with `label_by_title.py` (return 1, update the docstring's exit-code section and the test name/assertion accordingly), or explain why a malformed `ISSUE_NUMBER` should be treated differently here than in the script this PR explicitly patterns itself after.

Nit: the PR description's test-plan checkbox is unchecked even though the description says all 67 tests pass; worth ticking for clarity, not blocking.

aunger:
🤖 Author

Agreed on the `ISSUE_NUMBER` gap — it's the same class of swallowed failure the issue is about, and inconsistent with the `label_by_title.py` precedent this PR cites. Fixed in 1a15af4:

- `main()` now returns 1 when `ISSUE_NUMBER` fails to parse as an integer, matching `label_by_title.py`'s handling of the identical case.
- Docstring's exit-code section updated to mention this path.
- `test_exit_0_when_issue_number_non_integer` renamed to `test_exit_1_when_issue_number_non_integer` and now asserts exit 1.

All 67 tests still pass (`python3 -m unittest scripts.test_enforce_mutually_exclusive_labels`), and `ruff check`/`ruff format --check` are clean.

Also ticked the test-plan checkbox per the nit.

aunger (commented):
🤖 Reviewer

Confirmed: commit 1a15af4 fixes the `ISSUE_NUMBER` gap exactly as discussed, `main()` now returns 1 for a non-integer `ISSUE_NUMBER`, the docstring's exit-code section covers it, and `test_exit_0_when_issue_number_non_integer` was correctly renamed/flipped to `test_exit_1_when_issue_number_non_integer`. I checked out the branch again and ran the suite: all 67 tests pass. Test-plan checkbox is ticked. LGTM.

PR #620, opened 2026-07-09
Guard test_ci_monitor.py's suite behind if __name__ == "__main__"

### LINKED ISSUE #619: test_ci_monitor.py runs its whole suite as an import side effect (no if __name__ == "__main__" guard)
## Observed

`scripts/test_ci_monitor.py` (~3,700 lines) is not built from `unittest.TestCase` classes with discoverable `test_*` methods. It's a single procedural script: a custom `check()`/`_fail()` PASS/FAIL harness, with every one of its ~40 checks written as top-level module code, ending in a bare `sys.exit(1)` if any check failed. There is no `if __name__ == "__main__":` guard.

Because of that, simply *importing* the module executes the entire suite as a side effect, including:

- spawning `ci_monitor.py` as a real OS subprocess and sending it a real `SIGTERM` to verify signal handling (check group (o))
- ~1.75s of real (unmocked) `time.sleep` calls to verify real-wall-clock heartbeat behavior (check group (q)), which is why this file takes ~2s to run versus <0.35s for every other Python test file in `scripts/`

This happens to still work today because both `python3 -m unittest discover -s scripts -p "test_*.py"` (what `build.yml` runs) and `pytest` import every matching file during collection, so the suite runs either way. But it means:

- Any other tool that imports the module for a reason other than running its suite (an IDE indexer, a linter, `python3 -c "import test_ci_monitor"`, a coverage tool that imports modules before instrumenting them, etc.) will silently spawn a subprocess, send it a signal, and sleep for ~2 seconds as an unintended side effect.
- The file can't be `import`ed for inspection (e.g. to reuse `check_runs_payload()` or other helpers from another script) without re-running the whole suite first.
- It's the only file in `scripts/` structured this way; every other `test_*.py` file uses ordinary `unittest.TestCase` classes with a `unittest.main()` call under `if __name__ == "__main__":`.

## Suggested fix

Wrap the module's test-running code (the `check()` calls and the final `sys.exit(1)`) in a `main()` function guarded by `if __name__ == "__main__":`, consistent with the other test files. If `unittest discover`/pytest compatibility is wanted without a rewrite to `TestCase` classes, `main()` can still be invoked unconditionally when run as a script while being a no-op on plain import.

aunger:
Fixes #619.

`scripts/test_ci_monitor.py` ran its whole ~40-check, ~3700-line suite as a
side effect of merely importing the module, since every check() call and the
final `sys.exit(1)` lived at module level with no `if __name__ == "__main__":`
guard. That meant an IDE indexer, a linter, `python3 -c "import
test_ci_monitor"`, or a coverage tool importing the module before
instrumenting it would silently spawn `ci_monitor.py` as a subprocess, send it
a real SIGTERM, and sleep for ~1.75s of real wall-clock time.

This PR moves only the check()-invocation sequence and the final return/exit
into a `main()` function, guarded by `if __name__ == "__main__":`, consistent
with every other `test_*.py` file in `scripts/`. The shared Fixtures block
(data literals like `ALL_SUCCESS_JOBS` and helpers like
`check_runs_payload()`, reused across many check groups) stays at module
scope, so it remains reachable from a plain import, same as `make_zip_ndjson`
and `check` already were.

Importing `test_ci_monitor` is now a fast (~0.08s), silent no-op with no
subprocess spawn and no sleep. Running it directly
(`python3 scripts/test_ci_monitor.py`) still runs all 240 checks (238
original plus 2 new, see below) and exits 0 or 1 as before. A new check group
(al) spawns a real subprocess to verify both properties: plain import
produces no PASS/FAIL output, and `check_runs_payload()` is reachable on the
freshly imported module.

`test_ci_monitor.py` has no `unittest.TestCase` classes (that's out of scope
here; see the issue for why), so `unittest discover`, which build.yml already
runs, imported the module and ran its suite as a side effect too. With the
guard in place, `unittest discover` now imports the file but runs none of its
checks. To avoid silently losing that CI coverage, this PR also adds a direct
`python3 scripts/test_ci_monitor.py` invocation to the "Run Python script unit
tests" step in `.github/workflows/build.yml`.

## Test plan

No manual steps: `python3 scripts/test_ci_monitor.py` (240 checks) and
`python3 -m unittest discover -s scripts -p "test_*.py"` (the other 481
tests) both run automatically in CI via the updated workflow step.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,7 +67,12 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_REPOSITORY: ${{ github.repository }}
-        run: python3 -m unittest discover -s scripts -p "test_*.py" -v
+        run: |
+          python3 -m unittest discover -s scripts -p "test_*.py" -v
+          # test_ci_monitor.py is a procedural check()-based suite, not built from
+          # unittest.TestCase classes, so `unittest discover` imports it but runs
+          # none of its checks (issue #619). Run it directly too.
+          python3 scripts/test_ci_monitor.py
 
       - name: Run shell script unit tests
         if: steps.diff_check.outputs.needs_full_build == 'true'
--- a/scripts/test_ci_monitor.py
+++ b/scripts/test_ci_monitor.py
@@ -37,6 +37,7 @@ Covers:
   (ad) #499/#500: the failing build run is tracked among multiple Actions checks (wiring b)
   (ae) #500 doc-sync: no legacy hardcoded workflow/job/marker literals remain
   (af) #500 poll_signals: a run-only target does not widen another run's job-id filter
+  (al) #619: importing this module directly runs no checks and has no side effects
 
 No network calls required; no GITHUB_TOKEN needed.
 Run this file directly to execute the suite: exits 0 on success, non-zero on failure.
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

The core fix works. I checked out the branch and verified locally: importing `test_ci_monitor.py` is now fast (~0.1s) and side-effect-free (no subprocess spawn/SIGTERM, no `time.sleep`); `python3 -m unittest discover -s scripts -p "test_*.py"` no longer runs any of its checks; and `python3 scripts/test_ci_monitor.py` still runs and passes all 234 `check()` call sites (238 PASS lines, matching the pre-change count exactly) in about 2s, same as before. The two-commit split (guard first, then the CI workflow fix that keeps `test_ci_monitor.py` exercised under `unittest discover`) is clean and easy to follow.

One real gap versus the issue, though. The issue's "Observed" section names two problems, not one: the side-effect-on-import bug, and separately, "The file can't be imported for inspection (e.g. to reuse `check_runs_payload()` or other helpers from another script) without re-running the whole suite first." Its suggested fix says to wrap only "the check() calls and the final sys.exit(1)" in `main()`, specifically so the fixtures and helpers stay importable.

This PR instead nests the entire body of the file inside `main()`, not just the check() sequence. Every fixture (`ALL_SUCCESS_JOBS`, `FAILURE_JOBS`, etc.) and every helper function, including `check_runs_payload()` itself and all the `fake_request_*` closures, are now local names defined inside `main()`. I confirmed this directly:

```
>>> import sys; sys.path.insert(0, "scripts")
>>> import test_ci_monitor as m
>>> hasattr(m, "check_runs_payload")
False
>>> sorted(n for n in dir(m) if not n.startswith("__"))
['FAIL', 'PASS', '_fail', '_pass', 'check', 'ci_monitor', ..., 'main', 'make_zip_ndjson', ...]
```

So the second problem the issue raised isn't just left unfixed, it's made strictly worse: before, `check_runs_payload()` was reachable by importing the module (at the cost of the side effect); now it's unreachable from outside the module at all, short of a further refactor. Nothing in the repo currently imports it, so nothing breaks today, but this is exactly the use case the issue called out, and this PR closes the door on it rather than fixing it. Please move the fixtures and helper functions back to module scope (as `make_zip_ndjson` and `check` already are, untouched by this diff) and keep only the check()-invocation sequence and the final return/exit gated behind `main()` / `if __name__ == "__main__":`, per the issue's own suggested fix.

Smaller point: there's no check covering the fixed behavior itself, i.e. nothing asserts that importing the module produces no PASS/FAIL side effects and doesn't spawn a subprocess or sleep. `code_edit.md` calls for new behavior to be covered by automated tests, and "import is now a safe no-op" is exactly the new behavior here.

CI note (not blocking): `build-and-test` is currently red, but only because of an unrelated, non-allowlisted E2E failure (`SetupActivityPermissionDialogE2ETest#setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`) in the Android instrumented tests, nothing this PR touches. Both Python steps this PR modifies passed in that same run (`unittest discover`: 481 tests, OK; `python3 scripts/test_ci_monitor.py`: 238 passed, 0 failed).

Changes requested: please restore module-level reuse of the fixtures/helpers (primarily `check_runs_payload()`), consistent with the issue's suggested fix.

aunger:
🤖 Author

Agreed on the main point. Nesting the entire file body inside `main()` was an overshoot: `check_runs_payload()` and the shared Fixtures block were never meant to be part of "the check() calls and the final sys.exit(1)" the issue asked to gate. Pushed a fix that starts `main()` at check group (a) instead, right after the Fixtures block, so `check_runs_payload()`, `ALL_SUCCESS_JOBS`, `FAIL_NDJSON`, `REPO_STEP_REGEX`, and the other cross-group fixtures stay at module scope and importable, same as `make_zip_ndjson` and `check` always were. I did not hoist the hundreds of single-check-group fixtures and `fake_request_*` closures (e.g. `PR_S`, `side_effects_s2`, `fake_request_s2`) out to module scope; those exist solely to drive one `unittest.mock.patch(...)` call inside their own check block and aren't reusable helpers in the sense the issue's `check_runs_payload()` example was pointing at, so they stay inside `main()` with the check-invocation sequence that owns them. Let me know if you think that line is drawn wrong anywhere specific.

Also added check group (al), which spawns a real subprocess (matching the style of check (o)) to verify both properties directly: a plain `import test_ci_monitor` produces no PASS/FAIL output, and `check_runs_payload()` is reachable via `hasattr()` on the freshly imported module. 240 checks now pass (238 original + 2 new).

On the CI note: agreed, not blocking, the `SetupActivityPermissionDialogE2ETest` failure is unrelated to this PR.

aunger (commented):
🤖 Reviewer

Verified the new commit independently on a fresh checkout of `bd79f16`. Both points from my last review are resolved:

- `check_runs_payload`, `ALL_SUCCESS_JOBS`, `FAILURE_JOBS`, `FAIL_NDJSON`, `PASS_ONLY_NDJSON`, the `REPO_*_REGEX` constants, and `OWNER_T`/`REPO_T` are back at module scope. Confirmed directly:
  ```
  >>> import sys; sys.path.insert(0, "scripts")
  >>> import test_ci_monitor as m
  >>> hasattr(m, "check_runs_payload")
  True
  >>> m.check_runs_payload(("1", "2"))
  {'total_count': 1, 'check_runs': [...]}
  ```
  and import is still fast (~0.11s) and side-effect-free (`m.PASS, m.FAIL == 0, 0` immediately after import).
- The line the Author drew — hoisting the cross-group Fixtures block but leaving single-check-group fixtures/`fake_request_*` closures (e.g. `PR_S`, `fake_request_s2`) nested inside `main()` — is the right call. Those really are single-use setup for one check block, not the kind of reusable helper the issue's `check_runs_payload()` example was about, so there's no reason to hoist them too.
- New check group (al) is a good addition and does the right thing methodologically: it spawns a real subprocess for both assertions (plain import produces no output/exits 0; `check_runs_payload` is reachable via `hasattr`) rather than relying on the current process's already-populated `sys.modules`, so it actually exercises a fresh import. This directly covers the "smaller point" from my last review.

Ran the full suite directly and it passes: 240 checks (238 original + 2 new), 0 failed. `unittest discover` still imports the file without running its checks.

CI on this latest commit was still in progress when I checked; I'm not blocking on it, and my read of the previous run stands: the only failure there was the unrelated `SetupActivityPermissionDialogE2ETest` E2E flake, not anything in this PR's Python changes.

LGTM.

PR #628, opened 2026-07-10
Bound the E2E instrumentation call with a timeout so CI can't hang for hours

### LINKED ISSUE #611: CI E2E tests hangs
Identify specifically where this run became hung. I canceled it, but before that, it had been on this step in build-and-test for hours.

https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/28738927948/job/85218085883?pr=609

aunger:
Fixes #611.

### Where this run hung

The linked run's job log shows the `Run SetupActivityDeniedE2ETest` step (step 38) start `connectedE2EAndroidTest`'s `doLast` block, print the four expected `Performing Streamed Install` / `Success` lines from the APK installs, then go completely silent for over three hours until the user canceled it. No `INSTRUMENTATION_STATUS` output, nothing. Right after those installs, the task's `doLast` block calls `adb shell am instrument -r -w ...` via Gradle's `exec {}` and blocks until that returns. `am instrument -w` has no timeout of its own, and neither did anything wrapping it (the Gradle task, the CI step, or the job), so a hang anywhere inside the instrumented process (this exact test class's own doc already documents a history of keyguard/PIN races) could block indefinitely.

This is not a one-off: a separate run on `main` (run 28798023858) ran the full 360 minutes (GitHub's default job timeout) before being auto-canceled, and normal green runs consistently finish in 10-14 minutes. So this class of hang has recurred and can silently burn hours of CI compute with no automatic recovery.

### Fix

1. **`app/build.gradle.kts`**: `connectedE2EAndroidTest` now runs `am instrument` via a plain `ProcessBuilder` (rather than Gradle's `exec {}`, which has no built-in timeout either), bounded with `Process.waitFor(timeout, unit)` + `destroyForcibly()` on expiry, draining stdout on a side thread so a verbose run can't deadlock on a full pipe buffer while waiting. A timeout now fails the task with a clear message and is checked *before* the existing empty-results/failure guards, so a run that completed a few (all-passing) tests before being killed is still reported as the timeout it was, not a false-green partial run. Used a JVM-level timeout rather than shelling out to the `timeout` coreutil (as `build.yml` already does for the emulator-readiness wait) so `./gradlew connectedE2EAndroidTest` stays portable for local dev on any OS. After a timeout, also explicitly force-stops `com.gb4pc` and `com.gb4pc.test` on the device (review feedback), since killing the local adb client alone doesn't guarantee the on-device instrumentation stopped too.

2. **`.github/workflows/build.yml`**: added `timeout-minutes: 60` to `build-and-test` as a defense-in-depth backstop, since the fix above only covers this custom task; the AGP-driven `Run instrumented tests` step (and anything else in the job) still has no bound of its own.

### Test plan

- [x] `python3 -m unittest discover -s scripts -p "test_*.py"` and `scripts/test_*.sh` all still pass (unaffected by this change, run as a general health check).
- [x] YAML validated with `yaml.safe_load`; confirmed no ktlint/detekt gate exists to run locally.
- [x] The timeout firing path is no longer just a code-review-only claim: this PR's own CI (run 29072233989) reproduced the exact hang issue #611 describes, live, in `Run SetupActivityDeniedE2ETest`. The new timeout caught it at the 10-minute mark and failed loudly with `did not finish within 10m and was killed`, instead of blocking for hours. Every other E2E step in that same run passed normally (12s-2min each), confirming this was the pre-existing intermittent hang, not a regression from this change. Filed #629 to track the underlying test flakiness separately; a re-run of this branch is expected to go green since the failure is intermittent.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,10 @@ jobs:
     # full 360 minutes before GitHub auto-canceled it. A full green run normally finishes in
     # 10-14 minutes, so 60 minutes is a generous bound that still cuts a would-be multi-hour
     # hang down to about an hour, with results/artifacts/issue-filing preserved for whatever ran
-    # before the cutoff. The `connectedE2EAndroidTest` Gradle task's own `am instrument` call
+    # before the cutoff--GitHub re-evaluates `if` conditions on cancellation, so `if: always()`
+    # steps still queued when the timeout fires keep running, but only for up to 5 more minutes
+    # (GitHub's own cancellation grace period); anything not done by then is force-terminated
+    # regardless. The `connectedE2EAndroidTest` Gradle task's own `am instrument` call
     # (app/build.gradle.kts) now bounds itself too, so most hangs there fail within that task's
     # own timeout well before this job-level backstop is needed; this remains as defense in depth
     # for a hang anywhere else in the job (e.g. the AGP-driven "Run instrumented tests" step).
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,18 @@ jobs:
     # ubuntu-latest has KVM hardware acceleration (available since April 2024).
     runs-on: ubuntu-latest
 
+    # Issue #611: with no job-level timeout, a hang anywhere in this job falls back to GitHub's
+    # own 360-minute (6-hour) default. That has happened twice: this issue's referenced run sat
+    # hung for over three hours before a human canceled it, and a separate run on main ran the
+    # full 360 minutes before GitHub auto-canceled it. A full green run normally finishes in
+    # 10-14 minutes, so 60 minutes is a generous bound that still cuts a would-be multi-hour
+    # hang down to about an hour, with results/artifacts/issue-filing preserved for whatever ran
+    # before the cutoff. The `connectedE2EAndroidTest` Gradle task's own `am instrument` call
+    # (app/build.gradle.kts) now bounds itself too, so most hangs there fail within that task's
+    # own timeout well before this job-level backstop is needed; this remains as defense in depth
+    # for a hang anywhere else in the job (e.g. the AGP-driven "Run instrumented tests" step).
+    timeout-minutes: 60
+
     outputs:
       # Surfaced so link-pr-to-ci-summary (issue #331) can tell whether the
       # "Write test-result summary" step actually ran--on docs-only PRs it is
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -3,6 +3,7 @@ import java.io.File
 import java.time.LocalDate
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
+import java.util.concurrent.TimeUnit
 
 plugins {
     id("com.android.application")
@@ -327,8 +328,21 @@ tasks.register("connectedE2EAndroidTest") {
         val xmlSuiteName = e2eClass ?: "com.gb4pc.e2e"
 
         val instrumentOut = ByteArrayOutputStream()
-        exec {
-            commandLine(
+        // Issue #611: `am instrument -w` blocks until the on-device instrumentation reports
+        // completion, and has no timeout of its own. A prior CI run's "Run SetupActivityDeniedE2ETest"
+        // step went completely silent right after this exec started (no INSTRUMENTATION_STATUS
+        // output at all) for over three hours before a human had to cancel it; a separate run on
+        // main hit GitHub's own 360-minute (6-hour) job default and was cancelled automatically.
+        // Both burned CI compute and blocked the pipeline for hours with no automatic recovery.
+        // Run `am instrument` via a plain ProcessBuilder (rather than Gradle's `exec {}`, which has
+        // no built-in timeout either) and bound it ourselves with Process.waitFor(timeout, unit) +
+        // destroyForcibly(), so a hung instrumentation run fails this task quickly instead of
+        // hanging indefinitely. A JVM-level timeout (vs. shelling out to the `timeout` coreutil, as
+        // build.yml already does for the emulator-readiness wait) keeps this task portable to
+        // developers running `./gradlew connectedE2EAndroidTest` locally on any OS.
+        val e2eInstrumentTimeoutMinutes = 10L
+        val instrumentProcess =
+            ProcessBuilder(
                 e2eAdb,
                 "shell",
                 "am",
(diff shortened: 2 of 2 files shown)


aunger:
🤖 Reviewer

I read issue #611, this PR's description, both commits, and the full diff (app/build.gradle.kts and build.yml), plus the surrounding task and workflow for context.

What's solid: the two-layer design (a task-level timeout backstopped by a job-level timeout) is the right shape for this problem. The timeout guard is correctly checked before the crash/empty-results/failure guards, so a hung-but-partially-passing run is reported as the timeout it was, not a false green. The drain thread is joined before `instrumentOut.toString()` is read, avoiding both a full-pipe-buffer deadlock and a read/write race on the ByteArrayOutputStream. `e2eAdb` is an absolute path, so ProcessBuilder's default working directory doesn't matter. The YAML still parses (`yaml.safe_load`).

One concern I think needs addressing before merge: `destroyForcibly()` only kills the local `adb shell am instrument` client process spawned by ProcessBuilder. Nothing in this change touches the instrumentation actually running on the device/emulator (the `com.gb4pc.test` process under `com.gb4pc`). Killing the local adb client tears down that one shell connection, but nothing here confirms the remote instrumentation stops too, and the PR's own test plan says the timeout-firing path "was verified by code review only" since reproducing a hang wasn't practical in this environment. If the on-device process keeps running past the local timeout, the next `connectedE2EAndroidTest` invocation in the same job (there are several later in build.yml: overlay, gallery, permissions-granted/denied, setup-activity-granted/denied/permission-dialog, partial-access-photo-picker) targets the exact same `com.gb4pc.test/androidx.test.runner.AndroidJUnitRunner`. Depending on how the device handles a second instrumentation request against an already-instrumented target, that could mean either a fast "no results parsed" failure for every remaining step (bounded, if annoying) or another full 10-minute wait per step, well before the new 60-minute job timeout would even get a chance to act as backstop. Either way, this changes what "recovery" means here: instead of one bounded failure, a single device-level hang could still cost most of the job. Please either confirm (with reasoning or evidence) that killing the local adb process also stops the on-device instrumentation, or add explicit device-side cleanup after a timeout (e.g. `adb shell am force-stop com.gb4pc` and `com.gb4pc.test`) so later steps start from a clean device state.

Two smaller things, not blocking:

- `redirectErrorStream(true)` is new. Previously `standardOutput = instrumentOut` captured only stdout, so stderr streamed live to the Gradle console while `am instrument` ran. Now stderr is merged into the buffered `instrumentOut`, so nothing prints until the process finishes or is killed, even in the ordinary (non-hang) case. Given the point of this PR is that a hang is now bounded to 10 minutes, this is a reasonable trade, just flagging that it also removes what little live signal CI logs had during a normal run.
- The build.yml comment states a job timeout "does not by itself skip any if: always() step still queued when it fires." That's true, but only within GitHub's roughly 5-minute post-cancellation grace period before it force-terminates everything regardless; worth a caveat in the comment since this job has a long tail of if: always() upload/cleanup steps after build-and-test's main work.

Given the open question above about whether the timeout actually stops the underlying hang (the one thing this PR most needs to get right, and the one thing its own test plan could not verify), I'm requesting changes.

aunger:
🤖 Author

Good catch, agreed this needed addressing rather than resting on an unverified assumption. Pushed a follow-up commit (eb48abc):

- After a timeout, in addition to `destroyForcibly()` on the local adb client, now explicitly runs `adb shell am force-stop com.gb4pc` and `adb shell am force-stop com.gb4pc.test`. `am force-stop` acts at the ActivityManager level and doesn't wait on app cooperation, so it's not itself a source of a second hang, and it guarantees the device is clean for whichever `connectedE2EAndroidTest` step runs next regardless of whether the local client's death alone would have propagated to the device side.
- Updated the `build.yml` job-timeout comment to caveat the 5-minute cancellation grace period: confirmed against GitHub's own workflow-cancellation docs that `if: always()` steps keep running on cancellation, but only for up to 5 more minutes before everything is force-terminated regardless.

Left the `redirectErrorStream(true)` trade-off as-is since you flagged it as reasonable and non-blocking.

Python/shell unit tests re-run clean (479 passed) and the YAML re-validated with `yaml.safe_load` after these changes.

aunger:
🤖 Reviewer

Re-reviewed after the follow-up commit (eb48abc). This addresses my blocking concern: after `destroyForcibly()` on the local adb client, the task now explicitly runs `adb shell am force-stop com.gb4pc` and `adb shell am force-stop com.gb4pc.test`, both with `isIgnoreExitValue = true`, before the timeout guard throws. `am force-stop` acting at the ActivityManager level without waiting on app cooperation is a correct read of Android's behavior, and it's called only in the `timedOut` branch, so a normal (non-hung) run is unaffected. The ordering checks out too: `destroyForcibly()`, then the two force-stop execs, then `drainThread.join(...)`, all before the buffered output is read, so no new race was introduced.

The build.yml comment update also matches what I found independently: GitHub does re-run `if: always()` steps on a timeout cancellation, but only within its roughly 5-minute grace period, which is now called out explicitly in the comment.

One very minor residual note, not blocking: the two new `am force-stop` `exec {}` calls have no timeout of their own, so if the underlying hang were actually an unresponsive ADB/emulator transport rather than a stuck app process, these could in theory also block. The job-level 60-minute timeout still catches that case as the intended defense-in-depth backstop, so no further change is needed here.

CI on the latest commit was still in progress as of this review; the prior commit's run reported 1118 passing.

LGTM.

aunger:
🤖 Author

Investigated the "1119 fail" CI run (29072233989) on commit eb48abc. `Run SetupActivityDeniedE2ETest` ran from 05:58:03 to 06:08:08 (exactly the new 10-minute timeout plus cleanup) and failed with:

```
E2E: 'am instrument' for com.gb4pc.e2e.SetupActivityDeniedE2ETest did not finish within 10m and was killed (device or test likely hung; check logcat); 0 test(s) completed before the timeout
```

This is the fix working as intended, not a regression from this PR: every other E2E step in that same run (overlay, gallery, permissions granted/denied, setup-activity granted, permission-dialog, partial-access-photo-picker) completed normally with real PASS results in 12 seconds to 2 minutes. Only `SetupActivityDeniedE2ETest` hung, reproducing this exact issue's history (this class's own doc already documents a prior keyguard-related hang from issue #509). Before this PR, that would have blocked the job for up to 6 hours; here it failed loud and clear at the 10-minute mark, and the "Gate on test failures" step correctly treated it as an infrastructure failure (not a silent green), per the existing #309 design this PR's timeout guard was written to respect.

The logcat capture is genuinely interesting: unlike the previously-documented RESUMED-then-PAUSED-45ms-later keyguard race, this time the entire emulator went silent for the full ~10 minutes after `SetupActivity`'s `START` intent, with no `RESUMED`/`Displayed` line at all until my forced cleanup tore it down. Filed #629 with the full timeline and logcat excerpts for whoever picks up the underlying flakiness; it's a pre-existing, intermittent problem in that test suite, out of scope for this PR (which only bounds the hang, not root-causes it).

I don't have permission to re-run the failed job myself (`rerun_failed_jobs` returned a 403 from this integration), so a human (or whoever has access) will need to re-trigger CI on eb48abc, or a fresh push will do it. Given the intermittent nature (this exact suite passes in ~12s most runs), I'd expect a re-run to go green.

aunger:
🤖 Reviewer

Acknowledging the "1119 fail" CI run and issue #629. Confirmed #629 accurately describes what happened: the new 10-minute timeout fired for real against a genuine intermittent hang in `SetupActivityDeniedE2ETest`, while every other E2E step in that same run passed normally. That's the fix doing exactly what it's supposed to (bound the hang, fail loud and fast, keep the rest of the suite informative) rather than a regression, and the underlying test flakiness is correctly filed separately and out of scope here.

This doesn't change my verdict: LGTM stands.

PR #630, opened 2026-07-10
Unify CI artifact/step naming to ci-monitor-feed-/e2e-video-<Class>

### LINKED ISSUE #600: CI: unify per-suite artifact and step naming (ci-monitor-feed-<Class>, e2e-video-<Class>)
Part of #599.
Replaces #595 (names don't match test class) and #596 (rename `testresults-e2e-*` to `ci-monitor-feed-*`).
Those two rename the same `name:` fields on different axes (prefix vs suffix); this change composes both in one pass so the config regex and README are touched once.

## Goal

Given an artifact name in the Actions UI, a reader should be able to name the step and test class that produced it without opening `build.yml`, and the monitor-feed artifacts should say what they are for (feeding `ci_monitor.py`), not read as generic "test results".

## Naming scheme

- Per-suite monitor feeds (the small `.ndjson` marker files consumed by `ci_monitor.py`): `ci-monitor-feed-<TestClassName>`.
  - `testresults-e2e-overlay` -> `ci-monitor-feed-PixelCameraOverlayE2ETest`
  - `testresults-e2e-gallery` -> `ci-monitor-feed-GalleryButtonVisualE2ETest`
  - `testresults-e2e-permissions-granted` -> `ci-monitor-feed-PermissionsGrantedE2ETest`
  - `testresults-e2e-permissions-denied` -> `ci-monitor-feed-PermissionsDeniedE2ETest`
  - `testresults-e2e-setup-activity-granted` -> `ci-monitor-feed-SetupActivityGrantedE2ETest`
  - `testresults-e2e-setup-activity-denied` -> `ci-monitor-feed-SetupActivityDeniedE2ETest`
  - `testresults-e2e-setup-activity-permission-dialog` -> `ci-monitor-feed-SetupActivityPermissionDialogE2ETest`
- Per-suite E2E videos (not consumed by the monitor, so no monitor prefix): `e2e-video-<TestClassName>`.
  - `e2e-overlay-video` -> `e2e-video-PixelCameraOverlayE2ETest`
  - `e2e-gallery-video` -> `e2e-video-GalleryButtonVisualE2ETest`
  - (When #599 child E adds recording to the other five suites, they take the same `e2e-video-<Class>` form.)
- Step names: the "Upload ... markers" and "Upload ... video" steps should name the test class, matching how the `Run <Class>` steps already name themselves.
  e.g. "Upload PixelCameraOverlayE2ETest monitor feed" and "Upload PixelCameraOverlayE2ETest E2E video".

## One decision to make: the unit marker feed

`testresults-unit` (build.yml lines 163/167) is also a marker feed consumed by `ci_monitor.py`, but it spans many unit-test classes rather than one suite.
#596 deliberately left it named `testresults-unit` and proposed the regex `^testresults-unit$|^ci-monitor-feed-`.

- Recommended: rename it to `ci-monitor-feed-unit` too, so the whole monitor-feed family shares one prefix and the regex simplifies to `^ci-monitor-feed-`.
- Conservative fallback: keep `testresults-unit` and use the alternation regex above.

Pick one; the rest of the plan is identical either way.

## Files and changes

- `.github/workflows/build.yml`: the `name:` fields at lines 444, 452, 518, 526, 576, 625, 667, 712, 859 (and 167 if the unit feed is renamed), plus the corresponding `- name:` step titles.
- `scripts/ci_monitor/ci_monitor.config.json`: `artifact_name_regex` (currently `^testresults-`, which also matches the unit feed). Set to `^ci-monitor-feed-` (recommended) or the alternation (fallback).
- `scripts/ci_monitor/README.md`: update the "This repo's committed `ci_monitor.config.json` sets" bullet that lists example artifact names, and the `FAIL`/`SKIP`/`PASS` vocabulary rows that refer to `testresults-<group>`.

## Out of scope

The aggregate artifacts that intentionally span multiple suites keep their own names and are not forced into this scheme: `unit-test-results`, `instrumented-test-results`, `e2e-test-results`, `e2e-screenshots`, and the dev APK (`gb4pc-dev.*`).

## Sequencing and verification

- Do this before #599 child B: B edits the `SetupActivityPermissionDialogE2ETest` step block, whose upload-step name this change touches.
- `ci_monitor.py` reads its config at startup, so the regex and the new artifact names take effect together on the next run; there is no cross-run window where the monitor misses a feed.
- Verify by dispatching the workflow and confirming the Actions artifact list shows the new names, and that a live `ci_monitor.py` run still downloads and parses every feed (no dropped suite).

aunger:
Fixes #600.

Renames the per-suite CI artifacts and their upload steps so an artifact's name in the Actions UI identifies both the step/test class that produced it and what it is for:

- Per-suite monitor feeds (the `.ndjson` marker files `ci_monitor.py` consumes): `testresults-e2e-*` and `testresults-unit` become `ci-monitor-feed-<TestClass>` / `ci-monitor-feed-unit`, per the issue's decision comment to fold the unit feed into the same family. `ci_monitor.config.json`'s `artifact_name_regex` simplifies to `^ci-monitor-feed-`.
- Per-suite E2E videos (not consumed by the monitor): `e2e-overlay-video` / `e2e-gallery-video` become `e2e-video-PixelCameraOverlayE2ETest` / `e2e-video-GalleryButtonVisualE2ETest`.
- The corresponding "Upload ..." step titles now name the test class (e.g. "Upload PixelCameraOverlayE2ETest monitor feed", "Upload PixelCameraOverlayE2ETest E2E video"), matching how the "Run ..." steps already name themselves.

`PartialAccessPhotoPickerE2ETest`'s marker feed isn't listed in the issue's example table (it was likely added after the issue was filed) but follows the same scheme and is renamed too: leaving it as `testresults-*` would have dropped it from the monitor's regex match entirely.

Also touched, as direct consequences of the rename:

- `detect-launch-retry.yml` downloads the overlay suite's marker feed by its literal artifact name; updated to the new name so it keeps finding the logcat to scan.
- `interesting_step_regex` (`Build and run unit tests|E2ETest`) is tightened to `Build and run unit tests|^Run .*E2ETest$`. The renamed "Upload ... monitor feed"/"... E2E video" step titles now also contain each class's `E2ETest` substring, which would otherwise start matching the old regex and add new, undocumented step-success lines to Monitor output beyond the `Run ...E2ETest` steps the README documents.
- `scripts/ci_monitor/README.md`: updated the artifact-name example bullet and the `FAIL`/`SKIP`/`PASS`/mechanism prose that referenced `testresults-`.
- `scripts/test_ci_monitor.py`: updated fixture artifact names and assertions to match, since several tests drive `ci_monitor.main()`, which loads the real committed `ci_monitor.config.json` — the old `testresults-*` fixture names would no longer match the new regex.

Ran the full local suite (`unittest discover`, `test_ci_monitor.py` directly, and all `scripts/test_*.sh`); all green.

Test plan:
- [ ] Dispatch the workflow and confirm the Actions artifact list shows the new names, and that a live `ci_monitor.py` run against a real PR still downloads and parses every feed (no dropped suite), per the issue's verification note.


aunger (commented):
🤖 Reviewer

LGTM. This is a clean, complete rename and I verified it end to end.

What I checked:
- All eight per-suite monitor feeds and the unit feed match the issue's table and the decision comment: `testresults-e2e-*` -> `ci-monitor-feed-<Class>`, `testresults-unit` -> `ci-monitor-feed-unit`. The two videos become `e2e-video-PixelCameraOverlayE2ETest` / `e2e-video-GalleryButtonVisualE2ETest`. Upload-step titles now name the class, matching the `Run <Class>E2ETest` steps.
- No stale consumer remains: grepping the whole tree for the old names turns up only the deliberate in-code default and its historical prose in `ci_monitor.py`/README (the "Default (in code)" row at README.md:56, which correctly still documents `DEFAULT_ARTIFACT_NAME_REGEX = ^testresults-`). `detect-launch-retry.yml` is updated to the overlay suite's new feed name and path, and the overlay feed still ships `e2e-overlay-logcat.txt`, so the launch-retry scan keeps working. The aggregate artifacts (`unit-test-results`, `e2e-test-results`, `instrumented-test-results`, `e2e-screenshots`) are correctly left alone.
- The `interesting_step_regex` tightening from `E2ETest` to `^Run .*E2ETest$` is justified and I confirmed it against build.yml: every `Run ...` E2E step (lines 390-883) ends exactly in `E2ETest`, so all still match; none of the renamed `Upload <Class>E2ETest monitor feed`/`E2E video` steps match, which is the point. Worth noting this also fixes a pre-existing wart, since the old loose `E2ETest` already leaked the `Upload SetupActivity*E2ETest`/`PartialAccess*E2ETest` upload steps into Monitor output; the new anchor matches the behavior the README always documented.
- Renaming `PartialAccessPhotoPickerE2ETest`'s feed (not in the issue table) is correct: leaving it `testresults-*` would have dropped it from the new `^ci-monitor-feed-` regex entirely.
- Ran the suites locally: `scripts/test_ci_monitor.py` is 240 passed, 0 failed, and the fixture/assertion renames are consistent with the new regex.

Two non-blocking observations, neither of which needs to hold up merge:
1. `build.yml:450` and `:524` keep the suite-descriptor style ("Pull and upload overlay/gallery E2E video on failure"). These are the adb-pull `run:` steps, not the artifact-producing `upload-artifact` steps the issue targeted, so they're arguably outside the step-rename scope; flagging only for your judgment on whether a follow-up should align them.
2. The in-code `DEFAULT_ARTIFACT_NAME_REGEX = ^testresults-` and its comment now describe a naming contract this repo no longer emits. The issue's "Files and changes" list scoped `ci_monitor.py` out, and the default vs. committed-config divergence is exactly the config-override design (and README:56 documents it accurately), so this reads as intentional. No change requested; noting it so the choice is explicit.

The manual test-plan item (dispatch + live `ci_monitor.py` run confirming no dropped suite) remains as a before-merge verification step; it's not a code concern.

aunger:
### Verification report (before-merging item #631)

I carried out the live-fire verification called for by #631, using a throwaway test PR (#633, branched directly from this PR's head commit `33041c6`, will be closed after this report).

**Confirmed: the Actions artifact list shows the new names.**
The already-completed `build-and-test` run for this exact commit (https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/29086915037) lists all 9 expected per-suite feeds under the new scheme, matching `^ci-monitor-feed-`:

```
ci-monitor-feed-unit
ci-monitor-feed-PixelCameraOverlayE2ETest
ci-monitor-feed-GalleryButtonVisualE2ETest
ci-monitor-feed-PermissionsGrantedE2ETest
ci-monitor-feed-PermissionsDeniedE2ETest
ci-monitor-feed-SetupActivityGrantedE2ETest
ci-monitor-feed-SetupActivityDeniedE2ETest
ci-monitor-feed-SetupActivityPermissionDialogE2ETest
ci-monitor-feed-PartialAccessPhotoPickerE2ETest
```

No suite is dropped by the new regex. (The `e2e-video-*` artifacts are not present in this run because that run passed; those uploads are gated on E2E failure, which is expected and unrelated to this PR.)

**Result: FAIL, on the "live `ci_monitor.py` run downloads and parses every feed" step.**
Running `python3 scripts/ci_monitor/ci_monitor.py --pr 633` against this commit, every artifact download attempt failed with HTTP 401 from Azure Blob Storage (`InvalidAuthenticationInfo`), for example:

```
request failed (https://api.github.com/repos/aunger/gallery-button-for-pixel-camera/actions/artifacts/8225311039/zip): HTTP 401: Server failed to authenticate the request.
```

I isolated the cause with a minimal reproduction outside `ci_monitor.py`: the `GET .../actions/artifacts/{id}/zip` call 302-redirects to a `*.blob.core.windows.net` URL that is already authorized by a SAS token in its query string. `ci_monitor.py`'s download call (`urllib.request.urlopen` on a `Request` carrying an `Authorization: Bearer <token>` header) follows that redirect and Python's `urllib` does not strip the `Authorization` header when the redirect target is a different host. Azure Blob Storage then rejects the request because it received both a SAS token and a bearer `Authorization` header:

```
$ curl -s -o /dev/null -w "%{http_code}\n" "<signed blob URL>"
200
$ curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $GITHUB_TOKEN" "<same signed blob URL>"
401
```

This reproduces with plain `urllib.request.urlopen` too, independent of any proxying in my sandbox, so it is not an artifact of my environment.

**This looks like a pre-existing bug, not a regression from this PR.** The download/HTTP function in `ci_monitor.py` is untouched by this PR's diff (only `ci_monitor.config.json`, `README.md`, `test_ci_monitor.py`, and the two workflow files changed). The bug blocks the literal verification step in #631 ("confirm it downloads and parses every feed"), so per the verification process I'm recording this as a FAIL rather than closing #631, but the artifact-naming change itself is not implicated: the artifact list evidence above already confirms every renamed feed is present and matches the new regex.

### Summary

| Before-merging item | Result |
|---|---|
| #631: dispatch workflow, confirm renamed artifact list, confirm live `ci_monitor.py` download/parse | **FAIL** (artifact-list naming confirmed; live download/parse blocked by an apparently pre-existing `ci_monitor.py` HTTP-redirect auth bug, unrelated to this PR's diff) |

Leaving #631 open. Test PR #633 will be closed now that evidence has been captured.

aunger:
🤖 Author

Thanks for the live-fire verification. Confirming my read of the result before I decide whether to change this PR:

- The artifact-naming rename itself is fully confirmed: the completed `build-and-test` run for this PR's head commit lists all 9 renamed feeds under `^ci-monitor-feed-`, exactly as this PR intends. No suite was dropped.
- The FAIL is that a live `ci_monitor.py --pr` run can't download any artifact's zip contents at all, on any PR, because `_request()`'s authenticated GET follows `GET .../actions/artifacts/{id}/zip`'s redirect to a SAS-signed Azure Blob Storage URL without stripping the `Authorization: Bearer <token>` header, and Azure rejects the request for carrying both a SAS token and a bearer header. This is in `_request()` (`scripts/ci_monitor/ci_monitor.py:486-514`) and the download call site (lines 713-733), neither of which this PR's diff touches.

Since this bug predates the rename, reproduces for any artifact regardless of name, and sits entirely in code this PR doesn't change, I'm treating it as out of scope for #600 rather than folding a fix into this PR (one topic per PR). I filed #634 with the root cause, evidence, and a suggested fix (a redirect handler that drops `Authorization` on a cross-host redirect). I also left a pointer on #631 so the next verification attempt can find it.

I'm not making code changes to this PR in response to this finding. #631 will need #634 fixed before its "live download and parse" step can pass; the naming change here is otherwise complete and reviewed (LGTM).

aunger (commented):
🤖 Reviewer

Follow-up after the verification round. Verdict unchanged: LGTM.

I re-fetched the issue, the PR, and all comments, including the verification plan (#631 before-merging, #632 follow-on), the verification report (FAIL on #631 due to an Azure Blob Storage 401), issue #634, and the Author's reply declining to fix #634 in this PR.

I independently checked the Author's scope argument rather than taking it at face value:
- `git show 33041c6 --stat` confirms this PR's single commit touches exactly the 5 files in the diff I already reviewed (`build.yml`, `detect-launch-retry.yml`, `ci_monitor.config.json`, README.md, `test_ci_monitor.py`). It does not touch `ci_monitor.py`.
- I read `_request()` (ci_monitor.py:486-514): it builds a plain `urllib.request.Request`, adds a bearer `Authorization` header, and calls `urlopen`, with no redirect handling of any kind. That code is untouched by this PR and would leak the header on any cross-host redirect regardless of which artifact name it was fetching. The bug is generic to `_request`'s raw-download path, not specific to the renamed artifacts.
- Issue #634 documents the same root cause with independent curl evidence (200 without the header, 401 with it), which matches my reading of the code.

So the Author's claim holds up: the download failure is orthogonal to this PR's rename and would reproduce identically under the old `testresults-*` names. The substantive part of #600's own verification note ("confirming the Actions artifact list shows the new names ... no dropped suite") was independently confirmed by the verification report's artifact-list evidence, which lists all 9 renamed feeds matching `^ci-monitor-feed-`. Declining to bundle an unrelated `ci_monitor.py` fix into this naming-only PR, and instead filing #634 with a suggested fix and citing it, is the correct move per this repo's scope-discipline convention, not scope avoidance.

One process note for the record, not a code objection: #634's body states it "is a merge blocker for PR #630's before-merging item #631," and #631 is linked to this PR via the Verification Planner's issue-dependency mechanism. That means the merge gate on #630 stays structurally blocked until #634 is fixed and #631 closes, independent of my verdict here; that's the verification process working as designed, not something my LGTM should try to override.

My assessment of the diff itself is unchanged from my first review: complete, correct, and well-tested (240/240 local). LGTM stands.

PR #636, opened 2026-07-10
Add --sha, --run-id, and --branch identifiers to CI Monitor

### LINKED ISSUE #603: CI Monitor: add --sha, --run-id, and --branch identifier options
Part of #599.
Replaces #580.
The plan below is carried over from that issue's planning session (researched against `ci_monitor.py`, `README.md`, and `scripts/test_ci_monitor.py`).

## Scope: which identifiers to add

`ci_monitor.py` currently accepts only `--pr`.
Add three:

- `--sha <SHA>`: monitor a specific commit's check-runs directly, no PR involved.
- `--run-id <ID>`: monitor one specific Actions run, scoped to that run only.
- `--branch <NAME>`: monitor the head of a branch, re-resolving its SHA each poll.

Rejected: a check-suite-ID option (strictly less useful than `--run-id`, adds a concept for no gain).
A tag resolves to a commit exactly like a branch, so it becomes a one-line addition later if wanted, once `--branch`'s resolver exists.

## CLI shape

Replace `required=True` on `--pr` with a required, mutually exclusive argparse group of `--pr`, `--sha`, `--run-id`, `--branch`.
Exactly one is required; existing `--pr N` invocations are unaffected.

## Why `--pr` mode cannot be reused as-is

`--pr` mode does two PR-only things: it treats a merged/closed PR as an immediate terminal (`Merged`/`Closed`), and it gates the `Clear` terminal on `mergeable_state`.
Neither concept exists for a bare commit, run, or branch.
For the three new modes: no merged/closed short-circuit, and `Clear` fires directly off `parse_check_result(...) == "all_passed"` (skip the `/pulls/{n}` mergeable-state fetch).
This is simpler than the `--pr` path, not a degraded version.

## `--sha` mode

- SHA is fixed; skip the `/pulls/{n}` fetch.
- Each poll: `GET /commits/{sha}/check-runs`, then `parse_check_result` / `parse_check_summary` / `poll_signals` unchanged.
- `Clear` on `all_passed` directly.
- Output prefix `SHA#<short>:` (first 7 chars).

## `--branch` mode

Same as `--sha`, but re-resolve the SHA every poll so a new push is picked up.

- New `parse_commit_sha(commit_json)` reading the top-level `sha` from `GET /commits/{branch}` (shape differs from `/pulls/{n}`, whose SHA is under `head.sha`).
- Generalize `fetch_pr_with_retry` into `fetch_with_retry(url, token, ...)` shared by the PR fetch and the branch-commit fetch.
- Dedup sets (`seen_steps`/`seen_arts`/`seen_fails`) need no special handling on a SHA change; they are already keyed by ids fresh for a new run.
- Output prefix `BRANCH#<name>:`.

## `--run-id` mode

The one with a real choice: resolve verdict and diagnostics from the run object directly, so tracking is scoped to that run and not confused by an unrelated check on the same commit (e.g. `semgrep.yml`).

- `GET /actions/runs/{run_id}` each poll (carries `status`, `conclusion`, `head_sha`).
- New `parse_run_result(run_json)` mirroring `parse_check_result`'s bucketing (`queued`/`in_progress` -> in_progress; `success` -> all_passed; `cancelled`/`timed_out`/`stale`/`startup_failure` -> Infra; `failure`/`action_required` -> Blocked). Hoist the shared conclusion-bucket sets to module level so this third classifier does not re-duplicate them.
- Give `poll_signals` (and `drain_then_print`'s re-poll) an optional explicit-targets path so it can go straight to `run_job_ids = {run_id: None}` and skip `parse_actions_targets`.
- Output prefix `RUN#<id>:`.
- Acceptable first-version fallback if scoping proves invasive: resolve `--run-id` to its `head_sha` and behave like `--sha`. This is a fallback, not the recommendation (it reintroduces the unrelated-check problem).

## Output-prefix note

`agents/dev_orchestration.md`'s Monitor loop reasons only about the terminal keywords (`Clear`/`Blocked`/`Infra`) and never hardcodes the `PR#N:` prefix, so new prefixes are safe there.
`--pr` output must stay byte-for-byte identical under this issue (the test suite and the Orchestrator depend on its exact shape).
Note: #599 child C intentionally reshapes `--pr` output afterward, so land this issue first and let child C update the fixtures.

## Testing

Follow the lettered-case convention in `scripts/test_ci_monitor.py`:

- Parser tests for `parse_run_result` and `parse_commit_sha`.
- Argparse: exactly one of the four is accepted; two or none is rejected.
- One mocked-HTTP `main()` test per new mode covering a `Clear` and a `Blocked` path.
- A regression note confirming `--pr` output is unchanged by the refactor.

## Docs

- `scripts/ci_monitor/README.md`: the three flags, their prefixes, and that `Clear` is not gated on `mergeable_state` outside `--pr` mode.
- The module docstring `Usage:` line.

## Suggested sequencing

`--sha` first (establishes the no-`mergeable_state` terminal path), then `--branch` (adds the resolver), then `--run-id` (needs the targets-override refactor).
One PR covering all three is reasonable; they are also cleanly separable.

aunger:
Fixes #603.

`ci_monitor.py` previously accepted only `--pr`. This adds a required, mutually exclusive group of `--pr`/`--sha`/`--run-id`/`--branch`, so the monitor can track a bare commit, a specific Actions run, or a branch head with no PR involved.

## What changed

- `--sha <SHA>`: monitors a commit's check-runs directly. Output prefix `SHA#<first 7 chars>:`.
- `--branch <BRANCH>`: re-resolves the branch's head SHA every poll via a new `parse_commit_sha`, so a new push is picked up. Output prefix `BRANCH#<name>:`.
- `--run-id <RUN_ID>`: resolves verdict and diagnostics from the Actions run object itself (`GET /actions/runs/{run_id}`) via a new `parse_run_result`, instead of `/commits/{sha}/check-runs`. This scopes tracking to that one run so it can't be confused by an unrelated check on the same commit (e.g. this repo's `semgrep.yml`). Output prefix `RUN#<id>:`.
- `--sha`/`--run-id`/`--branch` have no PR to consult, so their `Clear` terminal fires directly off an all-passed check verdict, with no `mergeable_state` gating (that concept stays `--pr`-only, along with the merged/closed short-circuit).
- `poll_signals`/`drain_then_print` gained an `explicit_targets` path so `--run-id` mode can skip `/commits/{sha}/check-runs` and `parse_actions_targets` entirely, going straight to its one `(run_id, None)` target.
- Generalized `fetch_pr_with_retry` into `fetch_with_retry(url, token, ...)`, shared by the `--pr` and `--branch` fetches; `fetch_pr_with_retry` is now a thin wrapper.
- Hoisted the Infra/Blocked conclusion buckets (`_INFRA_CONCLUSIONS`/`_BLOCKED_CONCLUSIONS`) to module level so the new `parse_run_result` classifier doesn't re-duplicate them.
- `--pr` mode's behavior and output format are unchanged; all pre-existing PR-mode tests still assert exact `PR#N: ...` lines byte-for-byte and continue to pass.

## Tests

Added test cases (am)-(as) to `scripts/test_ci_monitor.py`:

- `parse_run_result` and `parse_commit_sha` classification/parsing.
- Argparse: exactly one of the four identifiers is accepted; two or none is rejected.
- One mocked-HTTP `main()` test per new mode, each covering a `Clear` and a `Blocked` path.
- A regression check that `fetch_pr_with_retry` delegates to `fetch_with_retry` with the same URL, plus a note that the exhaustive pre-existing `--pr`-mode tests are the regression proof that its output is unchanged.

All 287 tests pass (`python3 scripts/test_ci_monitor.py`), and `ruff check` passes clean.

Also updated `scripts/ci_monitor/README.md` (flag/prefix table, and which outcome-vocabulary behaviors are `--pr`-only vs. shared) and the module docstring's `Usage:` section.

## Test plan

- [x] None outside the automated suite; this is a script with no runtime dependency on a live GitHub Actions run to verify manually before merge.

### DIFF
--- a/scripts/ci_monitor/README.md
+++ b/scripts/ci_monitor/README.md
@@ -1,23 +1,41 @@
 # CI Monitor
 
-`ci_monitor.py` polls a pull request's CI and streams a terminal outcome plus per-test signals.
+`ci_monitor.py` polls CI for a pull request, a bare commit, a specific Actions run, or a branch head, and streams a terminal outcome plus per-test signals.
 Each stdout line is the interface: terminal lines (`Clear`/`Blocked`/`Infra`) end the loop, while informational lines (`in_progress` heartbeat, per-step deltas, per-test `FAIL`/`SKIP`/`PASS`) keep it alive.
 
 For how the Orchestrator uses this script as part of the development cycle (the Monitor loop, routing decisions, and the Verification Planner dispatch), see [`agents/dev_orchestration.md`](../../agents/dev_orchestration.md).
 
 ## Running the monitor
 
-Run it from the repo root, passing the PR number via `--pr`:
+Run it from the repo root, passing **exactly one** identifier:
 
 ```bash
 python3 scripts/ci_monitor/ci_monitor.py --pr <PR_NUMBER> [filter flags]
+python3 scripts/ci_monitor/ci_monitor.py --sha <SHA> [filter flags]
+python3 scripts/ci_monitor/ci_monitor.py --run-id <RUN_ID> [filter flags]
+python3 scripts/ci_monitor/ci_monitor.py --branch <BRANCH> [filter flags]
 ```
+
+| Flag | Tracks | Output prefix |
+|---|---|---|
+| `--pr <PR_NUMBER>` | A pull request's head commit. | `PR#<n>:` |
+| `--sha <SHA>` | A specific commit's check-runs directly, no PR involved. | `SHA#<first 7 chars>:` |
+| `--run-id <RUN_ID>` | One specific GitHub Actions run, scoped to that run only--not every check on its commit. | `RUN#<id>:` |
+| `--branch <BRANCH>` | The current head of a branch; its SHA is re-resolved every poll, so a new push is picked up. | `BRANCH#<name>:` |
+
+Supplying zero or more than one of these flags is a usage error (argparse exits with an error before any request is made).
+
+`--pr` mode is the only one with a pull request to consult, so it alone: (1) treats a merged or closed PR as an immediate terminal (`Merged`/`Closed`), and (2) gates the `Clear` terminal on `mergeable_state` (`clean`/`unstable`; `behind`/`dirty` maps to `Blocked`, `blocked` maps to `Infra`).
+`--sha`, `--run-id`, and `--branch` have no such state to consult: `Clear` fires directly off an all-passed check verdict, with no extra `/pulls/{n}` fetch.
+
+`--run-id` additionally does not fetch `/commits/{sha}/check-runs` at all--it resolves its verdict from the run object itself (`GET /actions/runs/{run_id}`, which carries `status`/`conclusion`/`head_sha` directly), and scopes step/artifact diagnostics to that one run's jobs. This keeps `--run-id` immune to an unrelated check on the same commit (e.g. this repo's `semgrep.yml`, which also runs on every commit) confusing its verdict--the problem `--pr`/`--sha`/`--branch` modes can in principle have if unrelated checks land on the same commit.
 
 `OWNER`/`REPO` default to this repo at the top of the script, and it reads `$GITHUB_TOKEN` from the environment (required).
 The script catches transient REST/parse failures per call so they cannot kill the resilient poll loop.
 
-The Monitor discovers which workflow run(s) and job(s) to track from the `/commits/{sha}/check-runs` payload, by parsing each GitHub Actions check run's `details_url` for its `(run_id, job_id)` (gated on `app.slug == "github-actions"`, with a `/actions/runs/` URL-pattern fallback when the `app` block is absent).
+In `--pr`/`--sha`/`--branch` modes, the Monitor discovers which workflow run(s) and job(s) to track from the `/commits/{sha}/check-runs` payload, by parsing each GitHub Actions check run's `details_url` for its `(run_id, job_id)` (gated on `app.slug == "github-actions"`, with a `/actions/runs/` URL-pattern fallback when the `app` block is absent).
 It does not name a workflow or job; the run/job to follow is derived from the same check-runs data that produces the verdict.
+In `--run-id` mode, the run to track is simply the one named on the command line--no check-runs payload is consulted for this purpose.
 
 ## Per-test outcome filters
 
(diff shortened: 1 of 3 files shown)


aunger (commented):
🤖 Reviewer

Thorough work overall. I checked out the branch, ran `python3 scripts/test_ci_monitor.py` (287 passed, 0 failed) and `ruff check` (clean). The implementation follows the issue's plan closely: `--sha`, `--branch`, and `--run-id` are each implemented as described, and `--run-id` goes with the full scoped-diagnostics design rather than falling back to the "resolve to head_sha" shortcut the issue allowed as a first-version fallback. Docs (README flag table, outcome-vocabulary section, module docstring) match the code.

One real bug in the mode-dispatch logic in `main()`:

```python
if args.pr:
    mode = "pr"
elif args.sha:
    mode = "sha"
elif args.run_id:
    mode = "run"
else:
    mode = "branch"
```

This branches on truthiness, not on "was this flag supplied". Argparse's mutually-exclusive group only guarantees exactly one of the four dest names is non-None; it does not stop a user from supplying an empty string as the value. So `--pr ""`, `--sha ""`, or `--run-id ""` all fall through every `elif` (since `""` is falsy) straight to `else: mode = "branch"`, even though `args.branch` is actually `None`.

I reproduced it with `_request`/`fetch_with_retry` mocked out: `--sha ""` produces the tag `BRANCH#None:` and polls `.../repos/.../commits/None` forever (`fetch_with_retry` keeps returning `None`, `parse_commit_sha` keeps returning `""`), silently discarding what the user actually typed instead of surfacing a clear error tied to the flag they used. `--pr ""` reproduces the same way.

Suggest checking `is not None` (argparse's un-supplied default) instead of truthiness when picking `mode`.

Smaller nit: `parse_run_result`'s docstring says it "mirrors parse_check_result's classification" using `queued`/`in_progress`, but the code actually treats any `status != "completed"` as `in_progress`, a superset of what is described (harmless, arguably better, but worth a one-line docstring update so it does not undersell what the function does).

Given the scope of the empty-string dispatch bug (a real logic error that produces silently wrong, endlessly-looping behavior against a nonexistent URL) but a one-line fix, my verdict is Changes requested rather than approving. Everything else here is solid.

aunger:
🤖 Author

Agreed, that was a real bug. Fixed in a6d8f78:

- Extracted the mode/tag selection into a standalone `_select_mode(args)` helper and switched each branch from truthiness (`if args.sha:`) to `is not None`, matching what argparse actually guarantees (an unsupplied dest stays `None`; a supplied dest can still be an empty string). `--sha ""` now correctly selects `sha` mode instead of falling through to `branch` mode with `args.branch=None`.
- Added direct unit tests for `_select_mode` (new case in (ao)) covering an empty-string value for each of the four flags, plus the existing tag-truncation behavior, so this is exercised without needing to run the poll loop.
- Expanded `parse_run_result`'s docstring to note it treats any non-`'completed'` status as `in_progress`, not just `queued`/`in_progress`, per the nit.

292/292 tests pass (`python3 scripts/test_ci_monitor.py`), `ruff check` is clean.

aunger (commented):
🤖 Reviewer

Verified commit a6d8f78: `_select_mode` now branches on `is not None` rather than truthiness, and I re-ran the exact repro (`_select_mode` with `sha=""`) directly, it now correctly returns `("sha", "SHA#")` instead of falling through to `branch` mode. The new tests in (ao) cover the empty-string case for all four flags plus tag truncation. `parse_run_result`'s docstring now also spells out the non-`'completed'` superset behavior.

Re-ran the suite on the updated branch: 292/292 pass, `ruff check` clean.

Both findings from my prior review are resolved. LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #636 can be merged):
None.

None.

Issue #591, opened 2026-07-04
ci_monitor.py's step "success" lines for E2E Run steps do not reflect JUnit test outcomes; real failures surface only at the Gate step


aunger:
## Summary

`ci_monitor.py`'s per-step `step "Run <TestClass>" -> success` lines do not mean the JUnit tests in that class passed.
For every E2E suite step and the general instrumented-test step, the step's own GitHub Actions conclusion is structurally guaranteed to be `success` regardless of whether the contained test failed, by design (issue #309).
The real pass/fail verdict is decided later, by the `Gate on test failures` step, which can be many steps after the misleading `success` line in the same job.
This is confirmed against three real CI runs in this repo's history, below, and the underlying mechanism is unchanged since at least issue #401/#402 despite the fixes those issues produced.

## Mechanism, confirmed from code

`.github/workflows/build.yml`'s nine test-running steps (`Run instrumented tests`, and the eight `Run <TestClass>E2ETest` steps for `PixelCameraOverlayE2ETest`, `GalleryButtonVisualE2ETest`, `PermissionsGrantedE2ETest`, `PermissionsDeniedE2ETest`, `SetupActivityGrantedE2ETest`, `SetupActivityDeniedE2ETest`, `SetupActivityPermissionDialogE2ETest`, `PartialAccessPhotoPickerE2ETest`) all share this shape:

```bash
if ./gradlew connectedE2EAndroidTest -Pe2eClass=com.gb4pc.e2e.<Class> ...; then
  echo "outcome=success" >> "$GITHUB_OUTPUT"
else
  echo "outcome=failure" >> "$GITHUB_OUTPUT"
fi
```

The `if`/`else` always completes and the script always falls off the end with exit status 0, so the step's own GitHub Actions `conclusion` is always `success`, no matter which branch ran.
The real Gradle/JUnit result is captured only in the step `output` variable (`steps.run_X.outputs.outcome`), which is invisible to `parse_steps()` in `ci_monitor.py` (it reads only `conclusion` from `/actions/runs/{id}/jobs`).
This is deliberate: the comment above `Run instrumented tests` in `build.yml` states "We deliberately do NOT use `continue-on-error` here (see issue #309): it swallowed *every* error... Instead we record this step's real exit status as an output and let later steps run; the final `Gate on test failures` step is the single authority on pass/fail."

[...]

Issue #519, opened 2026-06-26
Consider showing step name instead of check-run name in CI Monitor summary output


aunger:
The implementation plan for PR #518 (issue #516, Trade-offs section) noted an open question about whether the per-check summary and `Blocked by:` terminal should display the check-run job name (`No blocking labels`) or the step name (`Check for blocking labels`).

Source: https://github.com/aunger/gallery-button-for-pixel-camera/issues/516 (Trade-offs section, "Label-gate detection keys on the check-run `name`...")

The implementation uses the check-run job name because the GitHub check-runs API exposes it reliably; the step name is not present in that payload and would require a per-run jobs fetch (an additional HTTP request per terminal poll iteration).
The plan explicitly ruled out the extra HTTP call due to cost and the integration-test contract (deque-length assertions).
The issue's example text used the step name, so there is a cosmetic divergence from that mock-up.

This is a follow-on item and does not block merging PR #518.

If the step-name wording is preferred in future, one approach is a separate jobs-API call on terminal (not on every poll), accepted as a one-time cost at conclusion; another is a configurable display-name alias in the config file.

aunger:
Superseded by the harmonized plan in #599.
Merged with #591 into child #602.
The cost objection recorded here (step name needs an extra jobs fetch) is reconciled by #591's finding that the jobs endpoint is already fetched every poll, so the step name is already in hand.
Closing as superseded; the work continues in #602.

Issue #634, opened 2026-07-10
ci_monitor.py: artifact zip download gets HTTP 401 (Authorization header leaks onto redirected Azure Blob Storage URL)


aunger:
## Summary

`ci_monitor.py`'s artifact zip download (`_request(url, token, raw=True)` at `scripts/ci_monitor/ci_monitor.py:722-726`, called from the artifact-download loop at lines 713-733) fails with HTTP 401 whenever it actually tries to download a real artifact's zip contents from `GET /repos/{owner}/{repo}/actions/artifacts/{id}/zip`.

Found while verifying PR #630's before-merging item #631 (`https://github.com/aunger/gallery-button-for-pixel-camera/pull/630#issuecomment-4934609554`); it is a pre-existing bug, not introduced by that PR (PR #630's diff does not touch `_request` or the download loop).

## Root cause

`GET /repos/{owner}/{repo}/actions/artifacts/{id}/zip` responds with a redirect to a `*.blob.core.windows.net` URL that is already fully authorized by a SAS token in its own query string.

`_request()` (`scripts/ci_monitor/ci_monitor.py:486-514`) builds the request with `urllib.request.Request(url)` and adds `Authorization: Bearer <token>` (line 496), then calls `urllib.request.urlopen(req)` (line 500). Python's `urllib` follows the redirect automatically but does not strip the `Authorization` header when the redirect target is a different host. Azure Blob Storage then rejects the request because it receives both a valid SAS token and an unexpected bearer `Authorization` header.

Confirmed independently of `ci_monitor.py`, with plain `curl` against the same redirected, signed blob URL:

```
$ curl -s -o /dev/null -w "%{http_code}\n" "<signed blob URL>"
200
$ curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $GITHUB_TOKEN" "<same signed blob URL>"
401
```

This reproduces with plain `urllib.request.urlopen` too, so it is not an artifact of any sandbox/proxy configuration.

## Impact

[...]

PR #638, opened 2026-07-10
CI Monitor: annotate deferred-verdict steps and attribute the Blocked terminal by step/suite/test

### LINKED ISSUE #602: CI Monitor: name the failing step, suite, and test in output; stop presenting Run steps as verdicts
Part of #599.
Replaces #591 (Run-step `success` lines don't reflect JUnit outcomes) and #519 (show step name instead of check-run name).
Both are about the same thing: the granularity and honesty of `ci_monitor.py`'s streamed and terminal output.

## Background (from #591, confirmed against code and three real runs)

The nine deferred-verdict steps in `build.yml` (`Run instrumented tests` and the eight `Run *E2ETest` steps) wrap the test invocation in an `if`/`else` that records `outcome=success|failure` to `$GITHUB_OUTPUT` but always exits 0 (by design, issue #309).
So each step's own GitHub Actions `conclusion` is always `success`, regardless of whether the contained test failed.
The real verdict is decided later by the `Gate on test failures` step.
`ci_monitor.py`'s `step "Run X" -> success` line therefore carries no test-outcome information, yet reads like a pass.
The only stream signal that both identifies a real failure and can arrive before the terminal line is the per-test `FAIL [...]` line parsed from the monitor-feed artifact (Signal 2).
The terminal line (`Blocked by: build-and-test`) names only the single enclosing job, never the Gate step, the failing suite, or the failing test.

## Reconciliation with #519

#519 deferred showing the step name because it believed the step name needed an extra per-poll jobs fetch.
But `ci_monitor.py` already fetches `/actions/runs/{id}/jobs` (with its per-step list) every poll for `parse_steps()`, so the step name is already in memory.
The cost objection is moot for the step list; #519 becomes a small rider here rather than a separate deferred item.

## Asks

1. Terminal and summary attribution.
   When the terminal is `Blocked` due to a test failure, name the failing suite and test in the summary/terminal using the FAIL markers the monitor already parses (Signal 2), instead of only naming the `build-and-test` job.
   Where a specific step is the blocker (e.g. the Gate step), show the step name, drawn from the already-fetched jobs payload (this is #519's ask, now cheap).
2. Do not present `Run *E2ETest` / `Run instrumented tests` success as a verdict.
   Recommended: annotate these deferred-verdict steps' success lines (for example append a short "(verdict deferred to Gate)" note) rather than suppressing them, so the "this step finished" timing signal is kept.
   Drive the deferred-verdict set from a small config regex (the existing `interesting_step_regex` already enumerates these step names, so reuse or mirror it) rather than hardcoding.
   Suppression is the rejected alternative: it loses useful timing information and hides the step entirely.
3. Keep the existing per-test `FAIL`/`SKIP`/`PASS` behavior; this is surfacing and interpretation, not a new capability. The monitor already downloads and parses these artifacts independent of the overall check conclusion.

## Files

- `scripts/ci_monitor/ci_monitor.py`: `parse_steps()` output for the deferred-verdict steps, and the per-check summary / `Blocked by: ...` terminal construction (including `drain_then_print`'s internal re-poll) so it can attribute a failing suite/test/step.
- `scripts/ci_monitor/README.md`: the outcome-vocabulary rows for `step`, `Blocked by:`, and the note that `Run *E2ETest` steps always conclude `success`; update them to describe the new annotation and attribution.
- `scripts/test_ci_monitor.py`: extend the lettered-case fixtures for the reshaped output.

## Sequencing with #599 child D (`--sha`/`--run-id`/`--branch`)

Both edit `ci_monitor.py`.
Child D's plan assumes `--pr` output stays byte-for-byte identical; this issue deliberately changes that output.
Land child D first (purely additive new modes), then this issue reshapes the shared output path and updates the test fixtures once.
Run them in sequence, not in parallel.

## Note on terminal check-run granularity

`/commits/{sha}/check-runs` reports one row per job, and the whole test phase plus the Gate step is the single `build-and-test` job, so richer attribution here must come from the jobs/step payload and the FAIL markers, not from an additional check-run row.

## Verification

- Unit: fixtures asserting a deferred-verdict `Run` step's line is annotated, and that a Blocked terminal from a test failure names the suite/test (and step where applicable).
- Integration: a mocked-HTTP `main()` run over a Blocked-by-test-failure scenario, plus a live run against a known-red PR to confirm the stream now contradicts the misleading `success` line legibly.

aunger:
🤖 Author

Fixes #602.

The nine deferred-verdict steps in `build.yml` (the eight `Run *E2ETest` steps and `Run instrumented tests`) wrap their test invocation in an `if`/`else` that records `outcome=success|failure` to `$GITHUB_OUTPUT` but always exits 0 (issue #309). So each step's own Actions `conclusion` is always `success`, the real verdict is the later `Gate on test failures` step, and `ci_monitor.py`'s `step "Run X" -> success` line read like a pass while carrying no test-outcome information. The `Blocked` terminal, meanwhile, named only the enclosing `build-and-test` job. This PR fixes both the honesty and the granularity of that output.

## Ask 2: do not present deferred-verdict success as a verdict

- New config key `deferred_verdict_step_regex` drives the deferred-verdict set (this repo: `^Run .*E2ETest$|^Run instrumented tests$`). It is kept **distinct** from `interesting_step_regex`, which also matches the honest `Build and run unit tests` step (that step runs `testDebugUnitTest` under `set -o pipefail`, so its conclusion is a real verdict and must not be annotated).
- `parse_steps` annotates a deferred step's success line `(verdict deferred to Gate)` and surfaces it rather than suppressing it, keeping the "this step finished" timing signal. A deferred step that genuinely *failed* is a real failure: it is reported plain, unannotated.

## Ask 1: attribute the Blocked terminal

- `parse_steps` now collects genuinely-failed `(step name, conclusion)` tuples and `parse_fails` collects failing `[suite] name` identifiers, accumulated across every poll including the drain re-polls (so a late-arriving step/FAIL still feeds attribution).
- `blocking_suffix` enriches the `Blocked by: <check>` terminal to name the specific failing **step(s)** (e.g. `Gate on test failures`, addressing #519) and failing **suite/test(s)** (from the same `FAIL` markers already streamed, addressing #591), instead of only naming the `build-and-test` job. A label-gate-only block keeps its bare ` by: <name> [label gate]` form, so the Orchestrator's label-gate detection is unchanged.

Ask 3 (keep per-test `FAIL`/`SKIP`/`PASS` behavior) needed no capability change; this is surfacing and interpretation only.

## Resulting output for a red E2E run

```
PR#N: step "Build and run unit tests" -> success
PR#N: step "Run instrumented tests" -> success (verdict deferred to Gate)
PR#N: step "Run GalleryButtonVisualE2ETest" -> success (verdict deferred to Gate)
PR#N: step "Gate on test failures" -> failure
PR#N: FAIL [com.gb4pc.e2e.GalleryButtonVisualE2ETest] testButtonGreen: button not green
PR#N: summary
PR#N:   build-and-test .... failure   [BLOCKING]
PR#N: Blocked by: build-and-test (step "Gate on test failures" -> failure; test [com.gb4pc.e2e.GalleryButtonVisualE2ETest] testButtonGreen)
```

## Sequencing

Per the issue, this lands after #603's `--sha`/`--run-id`/`--branch` work (already on `main`), then reshapes the shared `--pr` output path and updates the fixtures once.

## Tests

`scripts/test_ci_monitor.py` (304 checks passing): new cases (at) parse_steps annotation + `failed_steps` collection and (au) `parse_fails` `failed_tests` collection; extended (ah) `blocking_suffix` attribution cases; the (ac) config-shape and (af) step-line fixtures updated for the new key/annotation; and the end-to-end main() Blocked fixtures ((i),(m),(p),(t),(u),(w),(ad)) updated to assert the attributed terminal. `scripts/ci_monitor/README.md` updated (config table, the `step` and `Blocked by:` vocabulary rows, and the deferred-verdict note).

🦀


aunger (commented):
🤖 Reviewer

LGTM. This fully and correctly addresses all three asks of #602, and the reconciliation of #519/#591 is exactly right.

## What I verified

- **Factual basis in `build.yml`.** Confirmed the 9 deferred-verdict steps: 8 `^Run .*E2ETest$` steps plus `Run instrumented tests`, each wrapping its invocation in an `if`/`else` that records `outcome=` and always exits 0 (#309). Confirmed `Build and run unit tests` (line 161) runs `testDebugUnitTest` under `set -o pipefail` with no outcome wrapper, so its conclusion is an honest verdict. This validates the key design decision: keeping `deferred_verdict_step_regex` **distinct** from `interesting_step_regex` rather than reusing it. The issue said "reuse or mirror"; mirroring with a separate key is the more correct choice, since reuse would have wrongly annotated the honest unit-test step. Good call, and well-justified in both the code comment and the config `_comment_deferred_verdict`.
- **Anchoring.** `^Run .*E2ETest$` correctly excludes `Run E2E emulator setup`, the `Upload *E2ETest monitor feed`/video steps (#600), and `Run {Python,shell} script unit tests`; `^Run instrumented tests$` excludes `Copy instrumented test results`.
- **Dedup / accumulation.** `failed_steps` is appended only inside the `seen`-guarded block in `parse_steps`, and `failed_tests` only inside the `seen`-guarded block in `parse_fails`, so both are dedup'd across every poll. The drain re-polls feed the same accumulators via `poll_signals`, and `drain_then_print` reads them for `blocking_suffix`, so a late-arriving step/FAIL still attributes the terminal (#419 lag path). Wiring is correct.
- **Label-gate invariant preserved.** `blocking_suffix` returns early for a label-gate-only block, so `[label gate]` stays the terminal's final token and the Orchestrator's detection is unaffected. A substantive block gets the enriched `(step ...; test ...)` suffix; a non-test block (merge conflict / infra) leaves `detail` empty and keeps the bare `by: <names>` form, so there is no false attribution.
- **Tests + lint.** All 304 checks pass on the branch head. New cases (at)/(au) and the extended (ah) attribution cases are honest and cover the step-only, test-only, and label-gate-ignore paths. `ruff check` is clean.
- **Docs.** README config table (four→five), the `step`/`Blocked by:` vocabulary rows, and the deferred-verdict note are all updated consistently with the code.

## Nit (non-blocking)

- `ci_monitor.py:235`: in the annotated branch `concl` is always `"success"` (guarded by `concl == "success"`), so `% (name, concl)` interpolates a value that is known. Harmless and arguably keeps symmetry with the other branch; leave it if you prefer the symmetry.

Note: `black --check` would reformat `test_ci_monitor.py`, but only on pre-existing lines outside this diff, and black is not run in CI (only ruff is). Not this PR's concern.

Nice work.

aunger:
🤖 Author

Thanks for the thorough review. Two notes:

- The nit on `ci_monitor.py`'s annotated branch (`concl` always being `"success"` there): keeping it as-is for symmetry with the unannotated branch, as you suggested.
- The one failing check, `build-and-test`, is the pre-existing `SetupActivityPermissionDialogE2ETest` flake tracked separately in #581 (confirmed via the job logs and the Gate step's own output: `1 blocking test failure(s) ... SetupActivityPermissionDialogE2ETest#setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`). This PR touches only `scripts/ci_monitor/*` and its README, no app code, so it isn't the cause and isn't in scope to fix here.

aunger:
🤖 Author

Update: `build-and-test` was retried and failed again, this time classified as an infra failure ("step failed without a corresponding test failure") rather than a JUnit assertion — still the same `SetupActivityPermissionDialogE2ETest` step. Issue #581's comment log confirms this is a systemic, ongoing flake: it has failed on `main` directly (twice, most recently commit `6f95ec7`) and on several unrelated branches/PRs (`feature/issue-604-e2e-diagnosability`, `fix/issue-619-...`) in the same window, independent of this PR's Python-only changes. No action needed here; tracked separately in #581.

aunger:
### Verification plan

**Before merging** (must resolve before PR #638 can be merged):
- [ ] #640

None.

aunger:
🤖 Verification Agent

Ran the before-merging item from the [verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/638#issuecomment-4937829239):

| Item | Result |
|---|---|
| #640 — Confirm live `ci_monitor.py` run against a known-red PR | **FAIL** (partial) |

**Details** (full evidence on #640): live-ran the updated `ci_monitor.py` from this branch (`--sha`/`--run-id` against PR #638's own red run `29104792916`, sha `a925b949`). The deferred-verdict annotation (`(verdict deferred to Gate)`) and the step-name half of the `Blocked by:` attribution **are** confirmed live, matching the PR body's example exactly. The suite/test-name half of the attribution (`test [suite] name`) never appeared, because every `ci-monitor-feed-*` artifact download 401'd — root cause is `_request()` forwarding the GitHub `Authorization` header across the redirect to the Azure Blob Storage SAS URL (confirmed independent of this PR's code; `_request()` is unchanged from the pre-PR base commit, so this is a pre-existing bug, not a regression here). This means the artifact-based FAIL/SKIP/PASS signal — and therefore this PR's new test-attribution clause — cannot be demonstrated live until that separate bug is fixed. Also noted (pre-existing, from #603, not a regression): `--run-id` mode never attributes anything (not even the step name) because it hardcodes an empty check-run row set.

Per verification process, this sends the PR back with the diagnosis above; #640 remains open.

aunger:
🤖 Author

Re the Verification Agent's report above: I agree with the diagnosis and am not changing this PR in response.

The two most novel behaviors introduced here, the deferred-verdict annotation and the step-name half of the `Blocked by:` attribution, are both confirmed live, matching the PR body's "Resulting output" example exactly. The suite/test-name half of the attribution is implemented and exhaustively covered by mocked-HTTP unit tests (fixture (au), plus the `main()` cases (i)/(m)/(p)/(t)/(u)/(w)/(ad)), but cannot be demonstrated live today because of a pre-existing, unrelated bug in `_request()`'s artifact-zip download (`Authorization` header forwarded across a cross-host redirect to Azure Blob Storage, already tracked in #634 since PR #630's verification of the same download path). That bug predates this PR and is not touched by this PR's diff, so fixing it here would be out of scope (one topic per PR).

Left a note on #640 leaving it open pending #634's resolution, consistent with how the same situation was handled for PR #630's #631.

PR #644, opened 2026-07-10
CI: synthesize a monitor-feed marker on the permission-dialog teardown path

### LINKED ISSUE #601: CI: synthesize a monitor-feed marker for the SetupActivityPermissionDialogE2ETest teardown path
Part of #599.
Replaces #597.

## Problem

The "Run SetupActivityPermissionDialogE2ETest" step in `.github/workflows/build.yml` (approx. lines 716-861) has a failure-diagnosis branch for when granting the real Photos and Media permission dialog restarts the `com.gb4pc` process mid-test (a scoped-storage remount).
When that happens no in-process `##GB4PC_TEST##` marker is ever emitted for the in-progress method, because the instrumented process died before it could report itself.
The step still uploads `results/e2e-setup-activity-permission-dialog.ndjson`, but for this failure mode the file has no FAIL/PASS/SKIP line at all, so the per-test signal `ci_monitor.py` normally gets for every other suite is absent here.
The build still goes red (the Gate step treats it as an infrastructure failure), but the per-test attribution is missing from this one feed.

## Goal

When the instrumented process is torn down mid-test so no in-process marker is emitted, write an equivalent synthesized marker into the ndjson from the host-side diagnosis branch, so this suite gets the same per-test attribution as the other six.
Synthesize FAIL on the two torn-down failure branches and, for parity, PASS on the benign recovery-confirmed branch.

## Scope

Only `.github/workflows/build.yml`, inside the permission-dialog Run step.
No changes to `app/build.gradle.kts`, `scripts/ci_monitor/*`, `scripts/check_allowed_failures.py`, or the gate wiring.
The file path `results/e2e-setup-activity-permission-dialog.ndjson` is unchanged; only the uploaded artifact name changes, and that is #599 child A's concern, so land A first.

## Fixed marker identity (all synthesized lines)

- `suite` = `com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest`
- `name` = `setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`
- `ms` = `0`

These match the class/method the in-process parser and JUnit XML use, so the `suite#name#outcome` dedup and the `--include-fail`/`--include-pass PATTERN` name filters work unchanged.

## Steps

1. Add a shell helper near the top of the `else` block that appends one JSON marker line, guarded to no-op if a real terminal marker for this method already exists in the ndjson (respect a genuine in-process marker; idempotent).
2. Make FAIL-detection and appends race-free: at the very top of the `else` block, deterministically rebuild the ndjson from the fully-written log with `grep '^##GB4PC_TEST##' /tmp/e2e-permission-dialog.log > results/e2e-setup-activity-permission-dialog.ndjson || true` (the `|| true` is required under `set -o pipefail` when grep finds no matches).
3. Regression branch (`outcome=failure`, recovery relaunch did not land on BATTERY): after the `echo "outcome=failure"`, call the helper with FAIL and a message built only from static ASCII text plus the numeric pids (`PID_BEFORE`, `PID_AFTER`, `RECOVERED_PID`).
4. Final `else` failure branch (non-zero Gradle exit, no in-process FAIL, media not granted): after its `echo "outcome=failure"`, call the helper with FAIL and a similar message.
5. Confirmed-benign branch (`outcome=success`): after its `echo "outcome=success"`, call the helper with PASS. This is suppressed by default in CI Monitor but available to `--include-pass 'SetupActivityPermissionDialog'` task-relevance validation, which otherwise has no marker on this path.
6. Leave the first branch (the genuine in-process FAIL case) untouched; step 2's rebuild guarantees the real marker is present and the helper's name-guard no-ops there.
7. No change to the upload step, the gate wiring, or `ci_monitor.*`.

## Edge cases

- Do not embed the UI Automator dump or grant-line text raw into the JSON (arbitrary XML would break the payload); interpolate only static text and integer pids, and keep `trace` empty, so no general JSON escaping is needed.
- The name-guard prevents a duplicate if a genuine marker exists.
- Process-substitution flush race is eliminated by the log rebuild in step 2.

## Explicitly out of scope

Writing a synthetic JUnit XML `<testcase>` so the gate attributes an allowlist-eligible test failure instead of an infra failure.
Today the two failure branches correctly fail the build as infra failures, and issue-filing keys off JUnit XML, so synthesizing XML would change gate/issue semantics.
Noted as a possible separate follow-up.

## Verification

- Static: confirm the synthesized `printf` output is a single `##GB4PC_TEST## ` line of clean JSON, and run it through `ci_monitor.py`'s `parse_fails` so FAIL surfaces by default and PASS only under `--include-pass`.
- Integration: dispatch the workflow (forcing each branch if needed), confirm the uploaded feed now contains the marker, and that a live monitor run prints the FAIL line for the failure branches and no duplicate on the happy path.

aunger:
🤖 Author

Part of #599. Fixes #601.

## Problem

The "Run SetupActivityPermissionDialogE2ETest" step has a host-side diagnosis branch for when granting the real Photos & Media permission dialog restarts `com.gb4pc` mid-test (a scoped-storage remount). When that happens the instrumented process dies before it can emit an in-process `##GB4PC_TEST##` marker, so `results/e2e-setup-activity-permission-dialog.ndjson` has no FAIL/PASS/SKIP line. This one suite alone then lacks the per-test attribution `ci_monitor.py` gets for every other suite (the build still goes red via the Gate, but the per-test signal is missing from this feed).

## Change

Entirely within the permission-dialog Run step in `.github/workflows/build.yml`:

- **Race-free rebuild.** At the very top of the diagnosis `else` block, deterministically rebuild the ndjson from the fully-written `/tmp/e2e-permission-dialog.log` (`grep '^##GB4PC_TEST##' … > … || true`). The in-line `tee >(grep …)` process substitution can still be flushing when Gradle returns, so this makes the `"outcome":"FAIL"` check and the synthesis below race-free.
- **Idempotent helper.** `synthesize_marker <outcome> <msg>` appends one marker line with the fixed identity `suite=com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest`, `name=setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`, `ms=0` — matching what the in-process parser and JUnit XML use, so ci_monitor's `suite#name#outcome` dedup and name filters work unchanged. A name-guard (`grep -qF`) no-ops if a genuine marker for this method already exists.
- **FAIL** on the two torn-down failure branches (recovery relaunch did not reach BATTERY; non-zero exit with no marker and grant not landed), messages built from static text plus the integer pids.
- **PASS** on the confirmed-benign recovery branch. Suppressed by default in CI Monitor, available to `--include-pass`.
- The genuine in-process FAIL branch is left untouched; the log rebuild guarantees its real marker is present and the helper's name-guard no-ops there.

Only static ASCII text and integer pids are interpolated and `trace` is empty, so the payload needs no JSON escaping. No change to the upload step, gate wiring, or `ci_monitor.*`.

## Verification

- YAML parses; `bash -n` on the extracted 166-line run script is clean.
- Simulated the helper: emits exactly one clean `##GB4PC_TEST##` JSON line; a second call no-ops via the name-guard (idempotent).
- Ran the synthesized FAIL and PASS lines through `ci_monitor.py`'s `parse_fails`: FAIL surfaces by default. PASS is suppressed by default and surfaces under a bare `--include-pass` or any pattern matching a substring of the fixed `name` value (e.g. `setupFlow`).
  - **Correction from an earlier revision of this description:** the issue's illustrative filter, `--include-pass 'SetupActivityPermissionDialog'`, does **not** surface this PASS marker. `parse_fails` matches `--include-pass PATTERN` against the marker's `name` field only, never `suite` (`scripts/ci_monitor/ci_monitor.py`, `parse_fails`), and the fixed `name` (`setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`) does not contain the substring `SetupActivityPermissionDialog`. This appears to originate in the issue text's own example query, not in this PR's marker identity (which is fixed by the issue and matches the real in-process emitter). Making that specific query work would require `ci_monitor.py` to also match `--include-pass` against `suite`, which is out of scope here per the issue ("No changes to ... `scripts/ci_monitor/*`"). Flagged back on #601 as a possible follow-up rather than addressed in this PR.

Integration (dispatching the workflow and forcing each teardown branch on a real emulator) is not runnable from here; the branch logic is exercised only when the dialog grant actually restarts the process.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -938,6 +938,38 @@ jobs:
           else
             # Gradle exited non-zero. Diagnose host-side, immune to com.gb4pc having died.
             echo "=== SetupActivityPermissionDialogE2ETest: non-zero Gradle exit; host-side diagnosis ==="
+            # Deterministically rebuild the monitor feed from the fully-written log before
+            # inspecting it below (issue #601). The in-line `tee >(grep ...)` process
+            # substitution above can still be flushing when Gradle returns, so rebuilding
+            # from /tmp/e2e-permission-dialog.log (which `tee` has finished writing by the
+            # time the pipeline's exit status is known) makes the "outcome":"FAIL" check
+            # and any synthesized marker below race-free. `|| true`: grep exits non-zero
+            # when the log has no marker lines (the torn-down case this step handles),
+            # which is expected and must not abort the step under `set -o pipefail`.
+            grep '^##GB4PC_TEST##' /tmp/e2e-permission-dialog.log > results/e2e-setup-activity-permission-dialog.ndjson || true
+            # Helper: append a synthesized terminal marker for this suite's single test
+            # method when the instrumented process was torn down mid-test so no in-process
+            # ##GB4PC_TEST## line was emitted (issue #601), giving this feed the same
+            # per-test attribution the other six suites get for every run. Idempotent:
+            # no-ops if a genuine marker for this method is already present (the rebuild
+            # above guarantees a real one, if any exists, is there). Only static ASCII text
+            # and integer pids are ever interpolated and `trace` is empty, so the payload
+            # never needs general JSON escaping. $1 is the outcome (FAIL/PASS), $2 the msg.
+            # The suite/name/ms values are fixed to match the class/method the in-process
+            # parser and JUnit XML use, so ci_monitor's suite#name#outcome dedup and its
+            # --include-fail/--include-pass name filters work on these lines unchanged.
+            SYNTH_MARKER_SUITE="com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest"
+            SYNTH_MARKER_NAME="setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances"
+            synthesize_marker() {
+              if grep -q "\"name\":\"$SYNTH_MARKER_NAME\"" results/e2e-setup-activity-permission-dialog.ndjson 2>/dev/null; then
+                echo "  monitor marker for $SYNTH_MARKER_NAME already present; not synthesizing a $1"
+                return 0
+              fi
+              printf '##GB4PC_TEST## {"suite":"%s","name":"%s","outcome":"%s","ms":0,"msg":"%s","trace":""}\n' \
+                "$SYNTH_MARKER_SUITE" "$SYNTH_MARKER_NAME" "$1" "$2" \
+                >> results/e2e-setup-activity-permission-dialog.ndjson
+              echo "  synthesized $1 monitor marker for $SYNTH_MARKER_NAME"
+            }
             PID_AFTER=$("$ADB" shell pidof com.gb4pc | tr -d '\r' || true)
             echo "com.gb4pc pid after dialog grant:  '${PID_AFTER:-<not running>}'"
             GRANT_LINE=$("$ADB" shell dumpsys package com.gb4pc | tr -d '\r' \
@@ -1002,6 +1034,11 @@ jobs:
                 echo "           is not shown again, so the grant survived the restart and no"
                 echo "           setup progress was lost."
                 echo "outcome=success" >> "$GITHUB_OUTPUT"
+                # Synthesize a PASS so this benign torn-down-but-recovered run still gets a
+                # per-test marker (issue #601). CI Monitor suppresses PASS by default, but it
+                # is available to --include-pass 'SetupActivityPermissionDialog' task-relevance
+                # validation, which otherwise has no marker at all on this recovery path.
+                synthesize_marker PASS "host-side diagnosis: the real Photos and Media dialog grant restarted com.gb4pc but the setup flow recovered on the BATTERY step without losing progress (the benign scoped-storage remount)"
               else
                 echo "REGRESSION: relaunching SetupActivity did not land cleanly on the BATTERY"
                 echo "            step after the restart (dump snippet:"
@@ -1009,11 +1046,19 @@ jobs:
                 echo "            the flow reverted to MEDIA or failed to render, so the crash"
                 echo "            was not the benign, expected scoped-storage restart."
                 echo "outcome=failure" >> "$GITHUB_OUTPUT"
+                # Synthesize a FAIL so this torn-down regression gets the same per-test
+                # attribution the other suites emit in-process (issue #601). Message uses
+                # only static text plus the integer pids captured above.
+                synthesize_marker FAIL "host-side diagnosis: the dialog grant tore down com.gb4pc (pid before=${PID_BEFORE:-<none>} after=${PID_AFTER:-<none>}) and the recovery relaunch (pid=${RECOVERED_PID:-<none>}) did not land on the BATTERY step, so the restart was not the benign scoped-storage remount"
               fi
             else
               # Non-zero exit with neither a reported failure nor a confirmed grant: treat as a
               # real failure (e.g. the dialog never appeared, or the grant did not land).
               echo "outcome=failure" >> "$GITHUB_OUTPUT"
+              # Synthesize a FAIL so this torn-down failure gets the same per-test
+              # attribution the other suites emit in-process (issue #601). Message uses
+              # only static text plus the integer pids captured above.
+              synthesize_marker FAIL "host-side diagnosis: non-zero Gradle exit with no in-process FAIL marker and READ_MEDIA_IMAGES not granted (pid before=${PID_BEFORE:-<none>} after=${PID_AFTER:-<none>}); the permission-dialog flow did not complete"
               echo "=== LOGCAT (since test start) ==="
               "$ADB" logcat -d -T "$LOGCAT_SINCE" | bash scripts/filter_logcat.sh
             fi
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -961,7 +961,7 @@ jobs:
             SYNTH_MARKER_SUITE="com.gb4pc.e2e.SetupActivityPermissionDialogE2ETest"
             SYNTH_MARKER_NAME="setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances"
             synthesize_marker() {
-              if grep -q "\"name\":\"$SYNTH_MARKER_NAME\"" results/e2e-setup-activity-permission-dialog.ndjson 2>/dev/null; then
+              if grep -qF "\"name\":\"$SYNTH_MARKER_NAME\"" results/e2e-setup-activity-permission-dialog.ndjson 2>/dev/null; then
                 echo "  monitor marker for $SYNTH_MARKER_NAME already present; not synthesizing a $1"
                 return 0
               fi
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1036,8 +1036,15 @@ jobs:
                 echo "outcome=success" >> "$GITHUB_OUTPUT"
                 # Synthesize a PASS so this benign torn-down-but-recovered run still gets a
                 # per-test marker (issue #601). CI Monitor suppresses PASS by default, but it
-                # is available to --include-pass 'SetupActivityPermissionDialog' task-relevance
-                # validation, which otherwise has no marker at all on this recovery path.
+                # is available to task-relevance validation via --include-pass: a bare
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

Verified structurally: `bash -n` on the extracted 166-line run script is clean, and the diff touches only `.github/workflows/build.yml` (no `app/build.gradle.kts`, `scripts/ci_monitor/*`, `check_allowed_failures.py`, or gate wiring), matching the issue's scope. The synthesized marker's field order/format matches the real emitter in `app/build.gradle.kts` (no spaces after colons), so the pre-existing `"outcome":"FAIL"` grep and the new `"name":"..."` name-guard both work correctly against real markers. I ran the helper standalone: it emits one clean `##GB4PC_TEST##` line and is idempotent on a second call, confirming that part of the PR's stated verification.

One concrete problem, though, with the PASS branch's stated purpose.

The issue's step 5 and the PR body both say the synthesized PASS marker on the confirmed-benign branch is "available to `--include-pass 'SetupActivityPermissionDialog'` task-relevance validation." I ran the actual synthesized PASS line through `ci_monitor.py`'s `parse_fails` with that exact pattern, and it does not surface:

```
name field: "setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances"
pattern:    "SetupActivityPermissionDialog"
```

`parse_fails` matches the `--include-pass` pattern against the marker's `name` field only, never `suite` (see `scripts/ci_monitor/ci_monitor.py` around line 247, and the pattern-matching tests in `scripts/test_ci_monitor.py`). The fixed `name` value the issue specifies, `setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`, does not contain the substring `SetupActivityPermissionDialog` (it has "MediaPermission" and "SystemDialog", not "ActivityPermissionDialog"). So the specific query the issue names as the whole point of the PASS branch silently returns nothing, indistinguishable from the marker not existing at all. Only a bare `--include-pass ''` (or a pattern matching a substring of the actual method name) surfaces it.

This directly contradicts the PR's own Verification section, which states: "PASS is suppressed by default and surfaces under bare `--include-pass` and a name-matching `--include-pass PATTERN`." That second half is not true for the pattern this feature exists to support.

This looks like it originates in the issue text itself (the fixed `name` was specified there, and the example pattern assumes it would match), but since this PR both implements the fixed identity and repeats the incorrect verification claim, this is the place to resolve it, either by correcting the verification claim to a pattern that actually works against the `name` field, or by flagging back on the issue that suite-based matching (a `ci_monitor.py` change) would be needed for the originally intended query, which is out of scope here.

Nit: `synthesize_marker`'s name-guard uses `grep -q` (basic regex) against `$SYNTH_MARKER_NAME` rather than `grep -qF`. Harmless today since the fixed name has no regex metacharacters, but worth the `-F` for robustness given it is guarding a real correctness case (avoiding a duplicate marker).

Everything else lines up well: the log-rebuild race fix, the untouched genuine-FAIL branch, the static-text-only messages that avoid needing JSON escaping, and the shared helper that avoids duplicating logic between the two FAIL branches are all solid.

Decision: Changes requested.

aunger:
🤖 Author

Thanks for catching that. Two changes pushed in 5bdbfed:

- Applied the nit: the name-guard now uses `grep -qF` instead of `grep -q`.
- Corrected the PR description's Verification section. You're right: `parse_fails` matches `--include-pass PATTERN` against the marker's `name` field only, and the issue's own illustrative pattern (`SetupActivityPermissionDialog`) is suite-shaped, not a substring of the fixed `name` (`setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`), so it silently returns nothing. The description now says so explicitly instead of repeating the incorrect claim.

I did not change `ci_monitor.py` to also match `--include-pass` against `suite`, since that's out of scope for this PR per the issue's own scope section ("No changes to ... `scripts/ci_monitor/*`"). Noting it here as a possible follow-up rather than filing a new issue myself, since #601 is still open and can absorb the note if a follow-up issue is wanted.

The marker identity itself (`suite`/`name`/`ms`) is unchanged — it's fixed by the issue and matches the real in-process emitter, so I left it as specified.

aunger (commented):
🤖 Reviewer

Thanks, the `grep -qF` fix and the corrected PR-description Verification section both look right, and I agree with the reasoning: the fixed `name` value is dictated by the issue (and must match the real in-process emitter), and making `--include-pass` match against `suite` would be a `ci_monitor.py` change that's explicitly out of scope here. That's a reasonable place to draw the line, and flagging it back on #601 rather than scope-creeping it into this PR is the right call.

One thing wasn't actually fixed, though: the same incorrect claim is still in the shipped code, not just the PR description. The in-line comment directly above the PASS branch's `synthesize_marker` call (the `outcome=success` branch, currently around line 1034-1037 of `build.yml`) still reads:

```
# Synthesize a PASS so this benign torn-down-but-recovered run still gets a
# per-test marker (issue #601). CI Monitor suppresses PASS by default, but it
# is available to --include-pass 'SetupActivityPermissionDialog' task-relevance
# validation, which otherwise has no marker at all on this recovery path.
```

This is the exact same false claim I flagged, now corrected in the PR body but not here. Since this comment lands in `main` and outlives the PR discussion, whoever reads the workflow file later (without digging through PR #644's comments) will hit the same wrong example and be misled the same way I was. Worth updating this comment to match what the PR description now says (e.g. note that `--include-pass` matches on `name` only, so a bare `--include-pass` or a `name`-substring pattern is what actually surfaces it, not a suite-shaped pattern), so the correction lives where a future reader will actually see it.

Decision: Changes requested.

aunger:
🤖 Author

Good catch, and fair point that the PR description isn't where a future reader will look. Pushed 43c9d43: updated the in-line comment above the PASS branch's `synthesize_marker` call to state what actually surfaces the marker (a bare `--include-pass`, or a pattern matching a substring of the fixed `name`) and to explicitly flag that a suite-shaped pattern like `--include-pass 'SetupActivityPermissionDialog'` does not match, since `parse_fails` only matches `--include-pass` against `name`, never `suite`. Comment-only change, no behavior difference.

aunger (commented):
🤖 Reviewer

Confirmed: the in-line comment above the PASS branch's `synthesize_marker` call now accurately describes the `--include-pass` matching behavior (bare or `name`-substring patterns like `'setupFlow'` surface it; a suite-shaped pattern like `'SetupActivityPermissionDialog'` does not, since `parse_fails` matches only `name`). I independently re-ran the corrected `'setupFlow'` example through `parse_fails` and it surfaces the PASS marker as described.

Both prior concerns are now resolved: the name-guard uses `grep -qF`, and the documentation (PR description and the in-repo comment a future reader will actually see) matches real behavior. Nothing further to flag.

Decision: LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #644 can be merged):
- [ ] #645

- [ ] #646
- [ ] #647

aunger:
🤖 Verification Agent

## Verification summary

Before-merging items from the [Verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/644#issuecomment-4939191323):

| Issue | Result |
|---|---|
| #645: Dispatch workflow and verify synthesized permission-dialog markers on a real run | NOT AUTOMATABLE (left open) |

### #645 detail

Not automatable in this session. The suite's own class doc and this PR's description both record that a prior real `workflow_dispatch` run already settled that the standard `requestPermissions()` dialog flow this suite drives does *not* restart `com.gb4pc` (only an out-of-band `pm grant`/`pm revoke`, a mechanism this suite deliberately avoids, has ever reproduced that, per commit e5d37ed). So a fresh dispatch would almost certainly re-exercise only the untouched happy path, not PR #644's new host-side diagnosis/synthesis branches, and there is no scriptable way to force those branches without either waiting for a currently-unprecedented genuine OS restart or modifying test/workflow source to inject one (out of scope for a Verification Agent). Full reasoning posted on #645.

No test PR was needed since the workflow-dispatch path itself was determined not to exercise the code under test; no repository-changing actions beyond the two comments above were taken.

## Outcome

Verification incomplete: the sole before-merging item is not automatable and remains open (issue #645) for a human, or a future genuine occurrence of the scoped-storage restart, to close out.

aunger:
Not planned; see analysis on #601: https://github.com/aunger/gallery-button-for-pixel-camera/issues/601#issuecomment-4940673140

Issue #640, opened 2026-07-10
(re PR #638) Confirm live ci_monitor.py run against a known-red PR


aunger:
Outstanding verification step from #602's Verification section, not yet confirmed as completed on PR #638:

> Integration: a mocked-HTTP `main()` run over a Blocked-by-test-failure scenario, plus a live run against a known-red PR to confirm the stream now contradicts the misleading `success` line legibly.

Source: https://github.com/aunger/gallery-button-for-pixel-camera/issues/602 (issue description, Verification section)

The mocked-HTTP `main()` integration cases are covered by the new fixtures in `scripts/test_ci_monitor.py` (per the PR body's Tests section). The second half of that verification ask, a **live run** of the updated `ci_monitor.py` against a known-red PR, is not evidenced anywhere in the PR's comments or reviews.

PR #638 itself produced red `build-and-test` runs (e.g. run [29104792916](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/29104792916), attributed by the author to a pre-existing flake tracked in #581), but no PR comment or review demonstrates that the *new* `ci_monitor.py` code (with `deferred_verdict_step_regex` annotation and the `blocking_suffix` step/suite/test attribution) was actually run live against a known-red PR/run and shown to produce the annotated, attributed output described in the PR body's "Resulting output for a red E2E run" example.

Before merging PR #638, run the updated `ci_monitor.py` against a known-red PR/run (for example, PR #638's own red run 29104792916, or another CI run known to fail a deferred-verdict step) and confirm the streamed/terminal output actually shows the deferred-verdict annotation and the step/suite/test attribution as described.

This is a merge blocker for PR #638 (`Blocks PR #638`).

aunger:
🤖 Verification Agent

**Result: FAIL (partial confirmation only)**

I ran the updated `ci_monitor.py` (feature branch `feature/issue-602-ci-monitor-attribution`, commit `745e0dc`, checked out at the pre-existing worktree `/home/user/wt-issue-602`) live against PR #638's own known-red run.

### Confirmed working live

`python3 scripts/ci_monitor/ci_monitor.py --sha a925b949979cedbdebfe309e32c4b91c39731a10 --include-pass ''` (same code path as `--pr`, the tool's documented primary interface) produced:

```
SHA#a925b94: step "Run instrumented tests" -> success (verdict deferred to Gate)
[...]

Issue #631, opened 2026-07-10
(re PR #630) Dispatch workflow and confirm ci_monitor.py parses every renamed artifact feed


aunger:
Manual (unmet by automated tests) verification step called for by the issue and the PR:

> Verify by dispatching the workflow and confirming the Actions artifact list shows the new names, and that a live `ci_monitor.py` run still downloads and parses every feed (no dropped suite).

Source: issue #600's "Sequencing and verification" section, restated as the PR's test-plan checklist item and reiterated in the reviewer's PR comment: https://github.com/aunger/gallery-button-for-pixel-camera/pull/630#pullrequestreview-4670713295

Steps:
- Dispatch the `build.yml` workflow on this branch (or a PR built from it).
- Confirm the Actions UI artifact list shows the new `ci-monitor-feed-*` / `e2e-video-*` names for every per-suite feed and video, including `ci-monitor-feed-unit` and `ci-monitor-feed-PartialAccessPhotoPickerE2ETest`.
- Run `ci_monitor.py` live against that run/PR and confirm it downloads and parses every feed (no suite silently dropped by the new `^ci-monitor-feed-` regex).

This is a merge blocker for PR #630 per the issue's verification requirement.

aunger:
🤖 Author

The "artifact list shows the new names" half of this is confirmed (see PR #630's verification comment). The "live `ci_monitor.py` run downloads and parses every feed" half is blocked by a pre-existing, unrelated bug in `_request()`'s artifact-zip download (an `Authorization` header leaking onto a redirected, already-SAS-signed Azure Blob Storage URL, causing HTTP 401 on every real artifact download, for any PR). Filed as #634 with root cause and a suggested fix. This issue should stay open until #634 is fixed and the live download/parse step can be re-verified.

PR #648, opened 2026-07-10
Strip Authorization header on cross-host artifact-zip redirects

### LINKED ISSUE #634: ci_monitor.py: artifact zip download gets HTTP 401 (Authorization header leaks onto redirected Azure Blob Storage URL)

aunger:
Fixes #634.

`ci_monitor.py`'s artifact-zip download (`_request(url, token, raw=True)`) failed with HTTP 401 whenever it tried to download a real artifact's zip contents from `GET /repos/{owner}/{repo}/actions/artifacts/{id}/zip`.

## Root cause

That endpoint responds with a redirect to a `*.blob.core.windows.net` URL that is already fully authorized by a SAS token in its own query string. `urllib.request.urlopen` follows the redirect automatically but does not strip the `Authorization` header when the redirect target is a different host, so Azure Blob Storage rejects the request because it receives both a valid SAS token and an unexpected bearer `Authorization` header.

## Fix

Adds a `_StripAuthOnCrossHostRedirect` (`urllib.request.HTTPRedirectHandler` subclass) that drops the `Authorization` header only when the redirect target's host differs from the original request's host. `_request(..., raw=True)` now routes through an opener built from this handler; the JSON API path (`raw=False`) is unchanged, since calls to `api.github.com` never redirect off that host.

## Testing

- Added test case (av) to `scripts/test_ci_monitor.py`: unit-tests `_StripAuthOnCrossHostRedirect.redirect_request` directly (strips `Authorization` on a cross-host redirect, keeps it on a same-host redirect, leaves unrelated headers alone), and confirms `_request(raw=True)` routes through the new opener rather than plain `urlopen`.
- Full suite: `python3 scripts/test_ci_monitor.py` — 309 passed, 0 failed.
- Verified live against a real PR artifact (an unexpired `ci-monitor-feed-*` artifact from a recent workflow run): a plain `urlopen` call reproduces the HTTP 401, while the patched `_request(url, token, raw=True)` downloads the zip and `extract_ndjson_lines` parses a real `PASS` marker, with no `request failed (...): HTTP 401` line to stderr.

This unblocks PR #630's before-merging item #631 and PR #638's before-merging item #640, both of which were waiting on this fix (see issue #634 for cross-references).

### DIFF
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -38,6 +38,7 @@ import re
 import sys
 import time
 import urllib.error
+import urllib.parse
 import urllib.request
 import zipfile
 
@@ -605,6 +606,36 @@ def _retry_after_seconds(e, now):
     return None
 
 
+class _StripAuthOnCrossHostRedirect(urllib.request.HTTPRedirectHandler):
+    """Redirect handler that drops the Authorization header on a cross-host redirect.
+
+    `GET /repos/{owner}/{repo}/actions/artifacts/{id}/zip` (the `raw=True` path)
+    redirects to a `*.blob.core.windows.net` URL that is already fully
+    authorized by a SAS token in its own query string. urllib's default
+    redirect handling forwards the original request's `Authorization` header
+    to the redirect target regardless of host, and Azure Blob Storage then
+    rejects the request with HTTP 401 because it receives both a valid SAS
+    token and an unexpected bearer token (issue #634). Stripping the header
+    only when the redirect target's host differs from the original request's
+    host keeps same-host redirects (if any) unaffected.
+    """
+
+    def redirect_request(self, req, fp, code, msg, headers, newurl):
+        new_req = super().redirect_request(req, fp, code, msg, headers, newurl)
+        if new_req is None:
+            return None
+        old_host = urllib.parse.urlparse(req.full_url).hostname
+        new_host = urllib.parse.urlparse(newurl).hostname
+        if new_host != old_host:
+            new_req.remove_header("Authorization")
+        return new_req
+
+
+# Reused across raw=True (artifact-zip) requests: the handler itself is
+# stateless, so a single opener built once suffices (issue #634).
+_RAW_OPENER = urllib.request.build_opener(_StripAuthOnCrossHostRedirect)
+
+
 def _request(url, token, raw=False):
     """Perform an authenticated GET. Returns parsed JSON (or raw bytes if raw).
 
@@ -612,6 +643,11 @@ def _request(url, token, raw=False):
     On an HTTP error the originating HTTPError is recorded on the function as
     `_request.last_error` so callers that retry (e.g. the SHA fetch) can inspect
     rate-limit headers; it is cleared to None on a successful request.
+
+    The `raw=True` path (artifact-zip downloads) uses an opener whose redirect
+    handler strips the `Authorization` header on a cross-host redirect (see
+    `_StripAuthOnCrossHostRedirect`); the JSON API path never redirects off
+    `api.github.com`, so it keeps the plain `urllib.request.urlopen` call.
     """
     _request.last_error = None
     req = urllib.request.Request(url)
@@ -619,8 +655,12 @@ def _request(url, token, raw=False):
     req.add_header("Accept", "application/vnd.github+json")
     try:
         # URL is built from a GitHub API constant; the file:// risk does not apply.
-        with urllib.request.urlopen(req) as resp:  # nosemgrep
-            data = resp.read()
+        if raw:
+            with _RAW_OPENER.open(req) as resp:  # nosemgrep
+                data = resp.read()
+        else:
+            with urllib.request.urlopen(req) as resp:  # nosemgrep
+                data = resp.read()
     except (urllib.error.URLError, OSError) as e:
         _request.last_error = e
         sys.stderr.write("request failed (%s): %s\n" % (url, _err_detail(e)))
--- a/scripts/test_ci_monitor.py
+++ b/scripts/test_ci_monitor.py
@@ -58,6 +58,9 @@ Covers:
        into failed_steps
   (au) #602 parse_fails: failed_tests collects only FAIL "[suite] name"
        identifiers (not PASS/SKIP), deduped via the shared seen set
+  (av) #634 _StripAuthOnCrossHostRedirect strips Authorization on a cross-host
+       redirect (the artifact-zip -> Azure Blob Storage case) but keeps it on
+       a same-host redirect; _request(raw=True) routes through this handler
 
 No network calls required; no GITHUB_TOKEN needed.
 Run this file directly to execute the suite: exits 0 on success, non-zero on failure.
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

Verified the fix directly (Python 3.11 `inspect.getsource`): `HTTPRedirectHandler.redirect_request` indeed forwards `Authorization` unconditionally to the redirect target regardless of host, confirming the root cause. The subclass approach here is correct and minimal:

- `redirect_request` correctly extracts host from `req.full_url` (original) vs. `newurl` (already resolved to absolute by `http_error_302`'s `urljoin` before `redirect_request` is called), so the host comparison is sound.
- `remove_header("Authorization")` correctly matches the capitalization `add_header` uses (`"Authorization".capitalize() == "Authorization"`), so the header actually gets dropped rather than silently no-op'ing on a case mismatch.
- Confirmed `build_opener(_StripAuthOnCrossHostRedirect)` swaps out only the default `HTTPRedirectHandler` (verified: the handler list contains the custom class but not a duplicate stock one, plus `HTTPSHandler`/`HTTPErrorProcessor`/etc. from the defaults), so https support and error processing are unaffected — this opener is a faithful drop-in for `urlopen`'s default behavior other than the one deliberate change.
- All four redirect codes (301/302/303/307/308) route through `redirect_request` since they alias to the same handler method, so the fix isn't accidentally scoped to just 302.
- Ran the full suite in the PR's worktree: 309 passed, 0 failed, including the five new (av) checks.
- Confirmed the `raw=True` path is the only call site of `_request(..., raw=True)` in the file, so the fix is applied everywhere it's needed and nowhere it isn't (the `api.github.com` JSON path is untouched, matching the PR's stated scoping rationale).

One minor nit, not blocking: the issue's suggested fix mentioned stripping "the `Authorization` header (and any other GitHub-specific headers)" on the cross-host hop, but this only strips `Authorization`; `Accept: application/vnd.github+json` still rides along to `*.blob.core.windows.net`. The issue's own `curl` repro only showed `Authorization` causing the 401, so this is very likely harmless in practice (Azure Blob Storage presumably ignores an unrecognized `Accept` header), and the PR's live-verification claim (real artifact downloaded and a `PASS` marker parsed with no stderr 401) is consistent with that. Worth a one-line acknowledgment in the PR description or a follow-up issue if the Author wants to be extra safe, but I would not hold up this PR on it.

LGTM. Nice, tight fix — good instinct to verify the root cause independently via `curl` before touching code, and the new tests exercise both the handler in isolation and its wiring into `_request`.

aunger:
🤖 Reviewer

CI has since reported: `build-and-test` failed on a blocking (non-allowlisted) E2E failure — `GalleryButtonVisualE2ETest#test4a_secureCameraLockedEmptyGalleryNoGreen`. This is unrelated to this PR's diff (Python-only: `scripts/ci_monitor/ci_monitor.py`, `scripts/test_ci_monitor.py`) and is a pre-existing flake already tracked in issue #403 (auto-filed/reopened by CI at this same run). Not a reflection on this fix.

My verdict is unchanged: LGTM.

aunger:
### Verification summary

Before-merging items from the [verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/648#issuecomment-4939421854):

- #649 (independently confirm live artifact-zip download avoids HTTP 401): **PASS**. Re-ran `python3 scripts/ci_monitor/ci_monitor.py --pr 648 --include-pass ''` from PR #648's own head commit against PR #648's own real workflow run (12 unexpired `ci-monitor-feed-*` artifacts, each redirecting through a SAS-signed `*.blob.core.windows.net` URL). stderr was empty (no `HTTP 401`), and stdout carried 298 PASS + 2 FAIL markers across 39 test classes from both `ci-monitor-feed-unit` and every `ci-monitor-feed-<E2ETest>` artifact. Full unit suite also re-run: 309 passed, 0 failed. Details and evidence on the issue: https://github.com/aunger/gallery-button-for-pixel-camera/issues/649#issuecomment-4939445088. Issue closed.

(#650 is a follow-on item and does not gate this merge; left untouched.)

All before-merging items pass. No test PR was created for this verification; PR #648 itself supplied the real artifacts needed.

PR #651, opened 2026-07-11
Capture UI Automator dump and full logcat on E2E instrumentation timeout

### LINKED ISSUE #629: Flaky SetupActivityDeniedE2ETest: hung again in CI
# `SetupActivityDeniedE2ETest` hung again in CI, this time with no keyguard PAUSED/RESUMED cycling

While validating PR #628 (issue #611: bound `connectedE2EAndroidTest`'s `am instrument` call with a 10-minute timeout so hangs can't block CI for hours), the new timeout actually fired for real: `Run SetupActivityDeniedE2ETest` hung again in CI on that PR's own branch.

Run: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/29072233989/job/86295958564
Step "Run SetupActivityDeniedE2ETest" ran from 05:58:03 to 06:08:08 (exactly the new 10-minute timeout plus a few seconds of cleanup), then failed with:

```
E2E: 'am instrument' for com.gb4pc.e2e.SetupActivityDeniedE2ETest did not finish within 10m and was killed (device or test likely hung; check logcat); 0 test(s) completed before the timeout
```

This is not a new bug from PR #628: every other E2E step in that same run (overlay, gallery, permissions granted/denied, setup-activity granted, permission-dialog, partial-access-photo-picker) completed normally in 12 seconds to 2 minutes with real PASS results. Only this one suite hung, matching this exact issue's history (this class's own KDoc already documents a prior keyguard/PIN-related hang from issue #509 / PR #564).

### What's different this time

The captured logcat shows a different symptom from the previously-documented one. Previously (per the class doc): `SetupActivity` reached `RESUMED` then `PAUSED` again ~45ms later, consistent with the keyguard re-engaging right as the activity launched.

This time: `SetupActivity`'s `START` intent fires at device time 05:58:10, and then the entire emulator goes essentially silent — no `RESUMED`, `PAUSED`, or `Displayed` line for that activity at all, just two harmless package-manager/memory-management blips (05:58:24 and 06:03:16) — until the new timeout's forced cleanup (`am force-stop com.gb4pc`) tears the activity down at 06:08:07, logged as `ActivityManager: Crash of app com.gb4pc running instrumentation ... Process crashed.` So this time it doesn't look like a fast keyguard re-engagement race; it looks like the activity launch (or Compose's test-synchronization wait inside `createAndroidComposeRule<SetupActivity>()`) got stuck before ever becoming visible, with no further system activity for the full ~10 minutes.

### Suggested next step

Whoever picks this up should decide whether this is the same root cause as issue #509 recurring intermittently (the PIN-secured keyguard `scripts/setup-e2e-emulator.sh` sets up, dismissed via `E2EFixture.dismissSecureKeyguard()`) or a distinct new failure mode, ideally with a full instrumented-test logcat capture (this run's was filtered through `scripts/filter_logcat.sh`, which may drop lines relevant here) and possibly a UI Automator dump captured on timeout to see what's actually on screen when this happens.

This is out of scope for PR #628, which only bounds the hang so it fails fast and loud instead of blocking CI for hours; it does not fix the underlying flakiness in this test.

aunger:
Fixes #629, as far as an automated change reasonably can.

### Why not a "fix" for the hang itself

Issue #629 documents `SetupActivityDeniedE2ETest` hanging in CI a second time, with a *different* symptom from the first occurrence (issue #509/PR #564): no `RESUMED`/`PAUSED` keyguard flicker this time, just total silence after `SetupActivity`'s `START` intent until the issue #611/#628 timeout force-killed it. The issue's own "Suggested next step" says whoever picks it up should decide whether this is the same root cause recurring or a distinct failure mode, "ideally with a full instrumented-test logcat capture ... and possibly a UI Automator dump captured on timeout to see what's actually on screen when this happens." Guessing at a code fix for an unconfirmed root cause isn't warranted; this PR implements exactly the diagnostic capture the issue asked for, so the *next* occurrence (of this or a similar hang, in any E2E suite) has real evidence to diagnose from.

### What changed

- **`app/build.gradle.kts`**: when `connectedE2EAndroidTest`'s own 10-minute `am instrument` timeout fires, capture a UI Automator screen dump and the full, unfiltered `adb logcat -d` buffer *before* the existing `am force-stop` cleanup runs — so the captured state reflects the hang itself, not a freshly reset device. This runs for whichever suite times out, not just `SetupActivityDeniedE2ETest`, since the timeout mechanism itself is generic across all `connectedE2EAndroidTest` invocations.
  - The logcat capture is deliberately unfiltered: `scripts/filter_logcat.sh`'s tag allowlist is exactly what the issue flagged as a likely reason the last recurrence's actual symptom (silence, no matching tag at all) went uncaptured.
  - Both are best-effort (`isIgnoreExitValue = true`) so a device unresponsive enough to hang a whole suite can't also mask the timeout `GradleException` that already fails the build.
- **`.github/workflows/build.yml`**: one new "Upload E2E timeout diagnostics" step uploads `app/build/outputs/e2e-diagnostics/` as the `e2e-timeout-diagnostics` artifact, `if-no-files-found: ignore` since the directory is only populated on an actual timeout.

### Test plan

- [x] `python3 -m unittest discover -s scripts -p "test_*.py"`, `scripts/test_ci_monitor.py`, and `scripts/test_*.sh` all still pass (unaffected by this change, run as a general health check).
- [x] `.github/workflows/build.yml` validated with `yaml.safe_load`.
- [ ] This PR's own CI run exercises the modified `connectedE2EAndroidTest` code path on every "Run *E2ETest" step in the normal (non-timeout) case. I could not run a full local Gradle build in this environment (no Android SDK, and the Gradle-distribution download is blocked here too), and deliberately reproducing a 10-minute hang isn't practical to do here; the new capture logic was verified by code review only. If a hang recurs after this merges, the `e2e-timeout-diagnostics` artifact should contain the UI Automator dump and full logcat for that run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1167,6 +1167,23 @@ jobs:
           $ANDROID_HOME/platform-tools/adb emu kill 2>/dev/null || true
           [[ -n "${EMULATOR_PID:-}" ]] && kill "$EMULATOR_PID" 2>/dev/null || true
 
+      - name: Upload E2E timeout diagnostics
+        # Issue #629: connectedE2EAndroidTest (app/build.gradle.kts) writes a UI Automator
+        # screen dump and the full, unfiltered logcat buffer into
+        # app/build/outputs/e2e-diagnostics/ only when its own 10-minute `am instrument`
+        # timeout (issue #611/#628) fires, captured before force-stop tears the hung
+        # activity down. Whichever E2E suite step (if any) hit that timeout is the only one
+        # that would have populated the directory, so this single step runs unconditionally
+        # after every suite rather than being duplicated per-suite; if-no-files-found:
+        # ignore keeps the common (no timeout) case from warning.
+        if: always() && steps.diff_check.outputs.needs_full_build == 'true'
+        uses: actions/upload-artifact@v7
+        with:
+          name: e2e-timeout-diagnostics
+          path: app/build/outputs/e2e-diagnostics/
+          retention-days: 14
+          if-no-files-found: ignore
+
       - name: Upload instrumented test results
         if: always() && steps.diff_check.outputs.needs_full_build == 'true'
         uses: actions/upload-artifact@v7
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -245,6 +245,43 @@ val e2eXmlDir =
     layout.buildDirectory
         .dir("outputs/androidTest-results/connected/debug")
 
+/**
+ * Runs [command] via a plain ProcessBuilder, bounded by [timeoutSeconds]: `Process.waitFor(timeout,
+ * unit)` + `destroyForcibly()` on expiry, the same pattern `e2eInstrumentTimeoutMinutes` uses below
+ * for `am instrument` -- not the `timeout` coreutil `build.yml` uses for its own adb waits, to keep
+ * this task portable to local dev on any OS (see that call site's own rationale).
+ *
+ * Returns combined stdout/stderr, or "" on any failure (non-zero exit, thrown exception, or the
+ * timeout itself, which is logged as a warning rather than thrown). Callers use this only for
+ * best-effort diagnostic capture (issue #629): a command that hangs or errors here must never
+ * block the build or mask the failure that triggered the capture in the first place -- exactly the
+ * class of hang this whole task's timeout guard exists to eliminate.
+ */
+fun runBestEffort(
+    command: List<String>,
+    timeoutSeconds: Long,
+): String =
+    try {
+        val process = ProcessBuilder(command).redirectErrorStream(true).start()
+        val out = ByteArrayOutputStream()
+        val drainThread =
+            Thread {
+                process.inputStream.copyTo(out)
+            }.apply {
+                isDaemon = true
+                start()
+            }
+        if (!process.waitFor(timeoutSeconds, TimeUnit.SECONDS)) {
+            process.destroyForcibly()
+            println("WARNING: '${command.joinToString(" ")}' did not finish within ${timeoutSeconds}s; killed")
+        }
+        drainThread.join(TimeUnit.SECONDS.toMillis(5))
+        out.toString()
+    } catch (e: Exception) {
+        println("WARNING: '${command.joinToString(" ")}' failed: ${e.message}")
+        ""
+    }
+
 tasks.register("connectedE2EAndroidTest") {
     group = "verification"
     description = "Runs E2E instrumented tests (requires device/emulator with Pixel Camera installed)."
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

Overall shape is right: scoping this to exactly the diagnostic capture the issue asked for (rather than guessing at a fix for an unconfirmed root cause) is the correct call, and the placement/ordering (capture before `force-stop`, single unconditional upload step after "Kill emulator", `if-no-files-found: ignore`, filenames keyed by `xmlSuiteName` so concurrent-suite timeouts in one job don't clobber each other) all check out against the surrounding code.

**Blocking concern: the new diagnostic capture has no timeout, so it can reintroduce the exact hang this issue chain exists to eliminate.**

`app/build.gradle.kts`'s three new `exec {}` calls (`uiautomator dump`, `adb pull`, `adb logcat -d`) only guard against the *target commands failing fast* via `isIgnoreExitValue = true`. None of them guard against the commands themselves hanging. Gradle's `exec {}` has no built-in timeout — it blocks until the child process exits, exactly like the plain `am instrument` call this same PR's parent issue (#611/#628) had to move off of `exec {}` and onto a hand-rolled `ProcessBuilder` + `waitFor(timeout)` + `destroyForcibly()` specifically because `exec {}` can't be bounded.

This matters here because the *specific* symptom issue #629 documents is "the entire emulator goes essentially silent" — no ActivityManager/Compose activity for 10 minutes straight. That's precisely the kind of device-unresponsive state where `adb shell uiautomator dump` (which has to walk the live accessibility tree of whatever's in the foreground) or even `adb logcat -d` could themselves hang rather than exit non-zero. If that happens, this task has no fallback: there's no workflow-level `timeout-minutes` set on this job either (confirmed — grepped `build.yml`, none present), so a hang in the new diagnostic code falls through to GitHub's bare 360-minute job default. That's exactly the multi-hour CI block described in issue #611 as the original problem, just moved three `exec{}` calls later instead of eliminated. The fix that was supposed to make hangs "fail fast and loud instead of blocking CI for hours" would, in the one scenario most likely to trigger it (a totally unresponsive device), stop bounding the total time again.

The repo already has the idiom to fix this cheaply: `build.yml` line 228 already wraps a similarly hang-prone adb wait with the coreutils `timeout` command (`timeout 1200 "$ADB" wait-for-device`), and this file already has the `ProcessBuilder`/`waitFor(timeout)`/`destroyForcibly()` pattern a few lines above this exact change for `am instrument`. Either bounding each of the three new `exec {}` calls with a short timeout (say 30–60s, since these are diagnostic best-effort captures, not the main test run) would close the gap without needing new abstractions.

Nits (non-blocking):
- The `println` after a successful capture doesn't mention that the dump/logcat may legitimately be empty/partial for the same "device unresponsive" reason being diagnosed — minor, since the artifact itself would show that.
- Test plan is honest about not exercising the timeout branch live; understandable given the environment constraints, but it means the un-bounded-hang risk above wasn't caught by any run, only by re-reading the code against the parent issue's own rationale.

Requesting changes on the timeout gap; everything else here is solid.

aunger:
🤖 Author

Good catch, agreed in full. Fixed in 1abb8f5: added `runBestEffort()`, using the same `ProcessBuilder` + `waitFor(timeout, unit)` + `destroyForcibly()` pattern this file already uses for `am instrument` (not the `timeout` coreutil, to keep the task portable to local dev on any OS, matching that call site's own rationale), and routed all three new diagnostic commands (`uiautomator dump`, `adb pull`, `adb logcat -d`) through it with a 30s bound each. Each still swallows failures (including its own timeout) rather than throwing, so a diagnostic capture that itself hangs now gets killed and logged as a warning instead of blocking the build, and still can't mask the real timeout `GradleException` thrown afterward.

aunger (commented):
🤖 Reviewer

Verified commit 1abb8f5. `runBestEffort()` closes the gap I flagged: all three diagnostic commands (`uiautomator dump`, `adb pull`, `adb logcat -d`) now go through `ProcessBuilder` + `waitFor(30, SECONDS)` + `destroyForcibly()`, matching the existing `am instrument` pattern in this same file, and the whole thing is wrapped in a try/catch so a thrown exception (e.g. command not found) can't escape either. Worst case now is a bounded ~90s added to the timeout path (3 × 30s) instead of an unbounded hang — the exact risk I raised is gone. I also checked that the function doesn't check exit codes, which correctly preserves the original "ignore failure, best-effort" semantics for commands like `pull` that would legitimately fail if the preceding `dump` timed out before writing its output file.

No other concerns. The two nits from my last round were non-blocking and don't need action.

LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #651 can be merged):
None.

- [ ] #652

PR #653, opened 2026-07-11
Fail loudly when dismissSecureKeyguard() times out without dismissing the keyguard

### LINKED ISSUE #642: Fail loudly (not silently) when dismissSecureKeyguard() times out without dismissing the keyguard
Source: reviewer suggestion on PR #637 (https://github.com/aunger/gallery-button-for-pixel-camera/pull/637#issuecomment-4937345586), acknowledged by the author (https://github.com/aunger/gallery-button-for-pixel-camera/pull/637#issuecomment-4937508500).

`dismissSecureKeyguard()` in `app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt` now polls `KeyguardManager.isKeyguardLocked` and retries its dismissal sequence until unlocked or `DISMISS_KEYGUARD_TIMEOUT_MS` (10s) elapses, but on timeout it silently returns even if the keyguard is still locked, rather than failing loudly (e.g. with `Log.w` or an explicit `fail(...)`) the way its sibling `lockScreen()` does on its own timeout path.

Given this PR (#637) is specifically about diagnosability, and this exact method was already implicated in a confusing intermittent CI failure during this PR's review, adding an explicit warning/failure on the timeout path would make the *next* occurrence of this race much faster to diagnose than another silent race into a downstream assertion failure.

This is a follow-on item and does not block merging PR #637. The reviewer explicitly did not request it as a blocking change, and the author deferred it to a follow-up rather than folding it into PR #637.

aunger:
Fixes #642.

`dismissSecureKeyguard()` (`app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt`) polls `KeyguardManager.isKeyguardLocked` and retries its wake/dismiss/PIN/enter sequence until unlocked or `DISMISS_KEYGUARD_TIMEOUT_MS` (10s) elapses, but previously returned silently on timeout even if the keyguard was still locked. This meant a timeout would surface later as a confusing downstream assertion failure (e.g. an activity bouncing `RESUMED` then `PAUSED`) instead of pointing at the actual cause — exactly the kind of race that made the CI failure during PR #637's review hard to diagnose.

This PR adds an explicit check after the retry loop: if the keyguard is still locked once the budget elapses, the method now fails immediately with a descriptive `fail(...)` message, the same way its sibling `lockScreen()` already fails loudly on its own timeout path.

## Test plan

- The four existing callers of `dismissSecureKeyguard()` (`SetupActivityGrantedE2ETest`, `SetupActivityDeniedE2ETest`, `SetupActivityPermissionDialogE2ETest`, `PartialAccessPhotoPickerE2ETest`) already exercise this method on every CI run; no new test class is needed since this only changes what happens on the (already-covered) timeout path, matching the existing precedent of `lockScreen()`'s own fail-on-timeout path having no dedicated test.
- This sandbox has no Android SDK and no network access to provision one (the Gradle wrapper distribution download is blocked), so `./gradlew` could not be run locally, consistent with the note on PR #637. The change is a small, mechanical addition (one `if` block) reviewed manually against the existing `lockScreen()` pattern in the same file.

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -270,6 +270,19 @@ class E2EFixture(
      * [KeyguardManager.isKeyguardLocked] and retrying the whole dismissal sequence until it
      * reports unlocked (or [DISMISS_KEYGUARD_TIMEOUT_MS] elapses) makes this method robust to that
      * variance instead of assuming a fixed wall-clock budget is always enough.
+     *
+     * ### Fails loudly on timeout instead of racing a downstream assertion (issue #642)
+     *
+     * If the keyguard is still locked when [DISMISS_KEYGUARD_TIMEOUT_MS] elapses, this method
+     * fails immediately with [fail], the same way [lockScreen] fails loudly on its own timeout
+     * path, instead of silently returning and letting the caller's next assertion fail with a
+     * symptom (e.g. an activity bouncing `RESUMED` then `PAUSED`) that does not point back to the
+     * keyguard as the actual cause. This exact silent-return path was implicated in a confusing
+     * intermittent CI failure (PR #637) before the retry loop above was added; failing loudly here
+     * makes any future recurrence of that race immediately diagnosable instead of requiring the
+     * same investigation to be repeated.
+     *
+     * @throws AssertionError if the keyguard is still locked after [DISMISS_KEYGUARD_TIMEOUT_MS].
      */
     fun dismissSecureKeyguard() {
         val keyguardManager = context.getSystemService(Context.KEYGUARD_SERVICE) as KeyguardManager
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Reviewer

Reviewed the diff against issue #642 and the `lockScreen()` precedent it's modeled on.

The fix is correct and minimal: after the existing poll/retry `do-while` loop exits, it checks `keyguardManager.isKeyguardLocked` once more and calls `fail(...)` with a descriptive message if the keyguard is still locked, mirroring `lockScreen()`'s own fail-on-timeout path exactly (same `KeyguardManager` check, same `fail()` import already in use, same message style). No new import was needed since `org.junit.Assert.fail` was already imported for `lockScreen()` and `launchSecureCamera()`.

I checked all four current callers of `dismissSecureKeyguard()` (`SetupActivityGrantedE2ETest`, `SetupActivityDeniedE2ETest`, `SetupActivityPermissionDialogE2ETest`, `PartialAccessPhotoPickerE2ETest`). Each calls it as a fire-and-forget setup step in a rule chain or `@Before`, none rely on silent continuation past a timeout, so failing loudly here is safe and consistent with the fixture's stated philosophy elsewhere in the file ("Silent skip would let an unconfigured emulator hide real regressions").

The KDoc addition is thorough and correctly cross-references the issue, the sibling `lockScreen()` method, and the historical PR #637 CI incident. The `@throws AssertionError` annotation is accurate.

No dedicated test was added, but the PR's stated rationale (this only changes the untested timeout path of an already-exercised method, matching the existing precedent that `lockScreen()`'s own fail-on-timeout path also has no dedicated test) is reasonable and consistent with prior practice in this file.

I don't see any correctness, scope, or consistency issues. LGTM.

PR #655, opened 2026-07-11
Strip Accept header, not just Authorization, on cross-host artifact-zip redirects

### LINKED ISSUE #650: Strip Accept (and other GitHub-specific) headers, not just Authorization, on cross-host artifact-zip redirects
From the review of PR #648 (https://github.com/aunger/gallery-button-for-pixel-camera/pull/648#pullrequestreview-4674776479):

> One minor nit, not blocking: the issue's suggested fix mentioned stripping "the `Authorization` header (and any other GitHub-specific headers)" on the cross-host hop, but this only strips `Authorization`; `Accept: application/vnd.github+json` still rides along to `*.blob.core.windows.net`. The issue's own `curl` repro only showed `Authorization` causing the 401, so this is very likely harmless in practice ... Worth a one-line acknowledgment in the PR description or a follow-up issue if the Author wants to be extra safe, but I would not hold up this PR on it.

`_StripAuthOnCrossHostRedirect` (added in PR #648, `scripts/ci_monitor/ci_monitor.py`) only removes the `Authorization` header on a cross-host redirect; the `Accept: application/vnd.github+json` header set by `_request()` still forwards to the redirected `*.blob.core.windows.net` URL. Consider stripping it too (or any other GitHub-specific headers) for extra safety, even though it is not currently known to cause a failure.

This is a follow-on item and does not block merging PR #648.

aunger:
Fixes #650.

`_StripAuthOnCrossHostRedirect` (added in PR #648 for issue #634, in `scripts/ci_monitor/ci_monitor.py`) only removed the `Authorization` header when a redirect crossed hosts (the artifact-zip download hopping from `api.github.com` to a SAS-signed `*.blob.core.windows.net` URL). The `Accept: application/vnd.github+json` header set by `_request()` is just as GitHub-specific and had no reason to keep riding along to a non-GitHub host, even though it is not currently known to cause a failure. This PR strips both headers on a cross-host redirect, while leaving genuinely unrelated headers untouched.

Changes:
- `_StripAuthOnCrossHostRedirect.redirect_request` now strips `Authorization` and `Accept` (via a small `_GITHUB_SPECIFIC_HEADERS` tuple) instead of only `Authorization`, when the redirect target's host differs from the original request's host.
- Updated docstrings on the class and on `_request()` to describe the broadened stripping.

Test changes (`scripts/test_ci_monitor.py`):
- Modified test `cross-host redirect keeps unrelated headers (e.g. Accept)` to instead require that `Accept` is stripped on a cross-host redirect, since that is the behavior this PR intentionally changes.
- Added an `X-Unrelated` header to the same test case so it still proves that headers that are *not* GitHub-specific survive the cross-host redirect untouched.

All 310 checks in `scripts/test_ci_monitor.py` pass locally.

### DIFF
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -607,18 +607,26 @@ def _retry_after_seconds(e, now):
 
 
 class _StripAuthOnCrossHostRedirect(urllib.request.HTTPRedirectHandler):
-    """Redirect handler that drops the Authorization header on a cross-host redirect.
+    """Redirect handler that drops GitHub-specific headers on a cross-host redirect.
 
     `GET /repos/{owner}/{repo}/actions/artifacts/{id}/zip` (the `raw=True` path)
     redirects to a `*.blob.core.windows.net` URL that is already fully
     authorized by a SAS token in its own query string. urllib's default
-    redirect handling forwards the original request's `Authorization` header
-    to the redirect target regardless of host, and Azure Blob Storage then
-    rejects the request with HTTP 401 because it receives both a valid SAS
-    token and an unexpected bearer token (issue #634). Stripping the header
-    only when the redirect target's host differs from the original request's
-    host keeps same-host redirects (if any) unaffected.
+    redirect handling forwards the original request's headers to the redirect
+    target regardless of host. The `Authorization` header causes Azure Blob
+    Storage to reject the request with HTTP 401 because it receives both a
+    valid SAS token and an unexpected bearer token (issue #634). The `Accept:
+    application/vnd.github+json` header set by `_request()` is GitHub-specific
+    too and has no business riding along to a non-GitHub host, so it is
+    stripped alongside `Authorization` for the same reason, even though it is
+    not currently known to cause a failure (issue #650). Stripping only
+    applies when the redirect target's host differs from the original
+    request's host, so same-host redirects (if any) are unaffected.
     """
+
+    # Headers added by `_request()` that are meaningful only to api.github.com
+    # and should not be forwarded to a cross-host redirect target.
+    _GITHUB_SPECIFIC_HEADERS = ("Authorization", "Accept")
 
     def redirect_request(self, req, fp, code, msg, headers, newurl):
         new_req = super().redirect_request(req, fp, code, msg, headers, newurl)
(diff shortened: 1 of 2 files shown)


aunger (commented):
🤖 Reviewer

This does exactly what issue #650 asked for, cleanly.

What I checked:
- `_request()` only ever sets two headers, `Authorization` and `Accept` (verified by grepping for `add_header` in `ci_monitor.py`), so `_GITHUB_SPECIFIC_HEADERS = ("Authorization", "Accept")` is complete; there is no third GitHub-specific header being missed.
- The `remove_header` call sites use the same capitalization (`"Authorization"`, `"Accept"`) that `add_header`'s `.capitalize()` produces, so both strips actually take effect (not a no-op from a case mismatch). Same reasoning applies to the test's `"X-Unrelated"` / `get_header("X-unrelated")` pair; both normalize to the same key, so that assertion is meaningful, not a false pass.
- Checked out the branch and ran `scripts/test_ci_monitor.py` directly: 310 passed, 0 failed, matching the PR description's claim.
- The test change is a legitimate tightening (Accept goes from "must survive" to "must be stripped"), not a loosening, and the PR description documents the modified test on its own line per `code_edit.md`. The new `X-Unrelated` header preserves the original test's coverage of "genuinely unrelated headers survive."
- `py_compile` on both touched files succeeds; no stray em dashes/smart quotes introduced in the docstring prose.

One nit, not blocking: the module docstring's test index now lists `(aw) #650 ...` as though it's its own section (mirroring how `(av) #634` and every other lettered entry has a matching `print("=== ... ===")` header in the test body). But the new #650 checks were appended inside the existing `(av)` section's block instead of getting their own `print` header, so there's no `"=== (aw) #650 ..."` in the actual test output. That is a small self-inconsistency in this file's own indexing convention (every other letter maps to a real printed section), not a functional problem, so I would not hold up this PR on it.

LGTM.

aunger:
Owner wants the nit addressed

aunger:
🤖 Author

Addressed the nit: moved the new `(aw) #650` checks (Accept-strip and unrelated-header-survives) out of the `(av) #634` print block into their own `=== (aw) #650 ... ===` section, printed after the `(av)` checks. The module docstring's test index already listed `(aw)` as its own entry, and now the printed test output matches it, consistent with every other lettered section in this file.

Ran `scripts/test_ci_monitor.py` again: 310 passed, 0 failed.

aunger (commented):
🤖 Reviewer

Confirmed the follow-up commit (a737159) addresses the nit: only `scripts/test_ci_monitor.py` changed (`ci_monitor.py` is untouched from the round I already reviewed), the new `(aw) #650` checks now have their own `print("=== (aw) #650 ... ===")` header instead of piggybacking on `(av) #634`, and they correctly reuse the `cross_host_redirected` variable from the `(av)` section above (no reassignment in between, so the reuse is valid). Re-ran the full suite on this branch: 310 passed, 0 failed. Docstring index now matches the actual printed sections for every lettered entry.

LGTM.

PR #654, opened 2026-07-11
DO NOT MERGE: throwaway CI run to verify e2e-video-* artifacts (issue #641)

aunger:
Disposable verification PR for issue #641 (unautomated verification step tracked from PR #637 / issue #604).

This PR deliberately breaks one test method in each of the five E2E suites that have never failed in real CI (`GalleryButtonVisualE2ETest`, `PixelCameraOverlayE2ETest`, `PermissionsGrantedE2ETest`, `PermissionsDeniedE2ETest`, `SetupActivityGrantedE2ETest`), so this run's CI actually reaches the "pull and upload e2e-video-*" step for each of them. No product code is touched--only a `fail()` call appended after each test's real body.

Purpose: download and inspect the resulting `e2e-video-*` artifacts to confirm taps/toast/valid-video content, closing out the last unconfirmed acceptance criteria on issue #641.

**Do not merge.** This PR will be closed (not merged) once the artifacts have been inspected, and the throwaway branch will be deleted.

PR #658, opened 2026-07-11
Add delete_gh_comment.sh for removing GitHub issue/PR comments

aunger:
## Summary

- Adds `scripts/delete_gh_comment.sh`, a small REST wrapper for `DELETE /repos/{owner}/{repo}/issues/comments/{comment_id}`.
- Adds `scripts/test_delete_gh_comment.sh` with coverage for argument validation, missing `GITHUB_TOKEN`, and the 204/404/403/401/unexpected-status response paths, using a fake `curl` on `PATH`.

Neither the GitHub MCP server nor the `gh` CLI expose a way to delete an issue/PR comment. This came up removing PR #654's auto-filed throwaway test-failure comments from issues #657, #656, #581, #571, #241, and #233--there was no tool available to do it, so this script fills that gap for future cleanup.

## Test plan

- [x] `bash scripts/test_delete_gh_comment.sh` passes locally (24/24 assertions).

### DIFF
--- a/scripts/delete_gh_comment.sh
+++ b/scripts/delete_gh_comment.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+# delete_gh_comment.sh: delete a single GitHub issue/PR comment via the REST API.
+#
+# Background: cleaning up auto-filed test-failure comments (e.g. the throwaway
+# CI-artifact-verification comments left on issues #657, #656, #581, #571,
+# #241, and #233 by PR #654) has no dedicated GitHub MCP tool, and no `gh` CLI
+# is available in some environments this repo's agents run in. This script
+# fills that gap with a plain REST call.
+#
+# Deletes ISSUE comments (the /issues/comments/{id} endpoint), which also
+# covers PR conversation comments (PRs are issues under the hood). It does
+# NOT delete PR review comments (inline diff comments); those live under
+# /pulls/comments/{id} instead.
+#
+# Usage:
+#   scripts/delete_gh_comment.sh <owner> <repo> <comment_id>
+#
+# Example:
+#   scripts/delete_gh_comment.sh aunger gallery-button-for-pixel-camera 4940940270
+#
+# Required environment variables:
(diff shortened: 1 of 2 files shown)


aunger:
🤖 Reviewer

## Overview

Adds `scripts/delete_gh_comment.sh`, a small `curl`-based wrapper around `DELETE /repos/{owner}/{repo}/issues/comments/{comment_id}`, plus `scripts/test_delete_gh_comment.sh` with a fake-`curl`-on-`PATH` test suite covering argument validation and the 204/404/403/401/unexpected-status response paths.

## Code quality and style

- Matches this repo's shell-script conventions well: header comment block with Background/Usage/Example/required-env-vars, `-h`/`--help` self-documenting via `grep '^#' "$0"`, `set -euo pipefail`.
- `COMMENT_ID` is validated against `^[0-9]+$` before being interpolated into the URL, so no injection risk there.
- The `curl -sS -o "$BODY_FILE" -w '%{http_code}'` + case-statement pattern is a reasonable choice over this repo's more common `-fsSL` style (e.g. `detect_launch_retry.sh`), since `-f` would suppress the body and collapse the 404/403/401 distinction the script deliberately preserves.
- Test suite is thorough and follows the same fake-binary-on-`PATH` pattern as `test_detect_launch_retry.sh`.

## Nits (non-blocking)

- `OWNER` and `REPO` aren't validated the way `COMMENT_ID` is. A typo'd argument just becomes a different (still well-formed) URL, so the blast radius is low, but a `^[A-Za-z0-9._-]+$`-style check would catch fat-fingering earlier with a clearer error than whatever GitHub returns.
- A genuine network failure (DNS/connection error, not an HTTP status) will abort via bare `set -e` with `curl`'s own stderr text rather than one of the script's own error messages. Fine for an interactively-run admin script; just flagging it as the untested path.
- The fake `curl` in the test suite doesn't capture headers, so nothing asserts the `Authorization`/`Accept` headers are actually sent. Low value to add given how static that part of the code is, but worth knowing it's not covered.

## Verdict

No correctness issues found, and the change is well-scoped and well-tested.

`LGTM`

PR #659, opened 2026-07-11
Add test-failure as a mutually exclusive label prefix group

aunger:
## Summary

- Adds `"test-failure"` to `MUTUALLY_EXCLUSIVE_PREFIXES` in `scripts/enforce_mutually_exclusive_labels.py`, so `test-failure` and `test-failure-archive` are enforced as mutually exclusive the same way the existing `c-a-*`/`c-r-*` prefix groups are.
- `test-failure-archive` already replaces `test-failure` when `archive_stale_test_failures.py` runs (see that script's own label swap), but nothing previously stopped the two from coexisting if `test-failure-archive` were added by hand without also removing `test-failure`--which is exactly what happened when issues #657, #656, #571, and #233 were manually archived. This closes that gap for future manual label changes.
- Adds test coverage in `scripts/test_enforce_mutually_exclusive_labels.py`: `find_conflicting_prefix`/`labels_to_remove_by_prefix` unit tests for both directions (`test-failure-archive` added removes `test-failure`, and vice versa), a check that the unrelated `testing` label is untouched, and `main()` integration tests mirroring the existing prefix-group coverage.

## Test plan

- [x] `python3 -m unittest scripts.test_enforce_mutually_exclusive_labels -v` passes locally (76/76 tests).

### DIFF
--- a/scripts/enforce_mutually_exclusive_labels.py
+++ b/scripts/enforce_mutually_exclusive_labels.py
@@ -12,8 +12,11 @@ Mutually exclusive sets (fixed):
     [orchestrate, orchestrating]
 
 Mutually exclusive prefix groups (any label sharing a prefix is exclusive):
-    c-a-*   (author model, e.g. c-a-haiku, c-a-sonnet, c-a-opus)
-    c-r-*   (reviewer model, e.g. c-r-haiku, c-r-sonnet, c-r-opus)
+    c-a-*          (author model, e.g. c-a-haiku, c-a-sonnet, c-a-opus)
+    c-r-*          (reviewer model, e.g. c-r-haiku, c-r-sonnet, c-r-opus)
+    test-failure   (test-failure vs. test-failure-archive; an issue should
+                    carry at most one, since test-failure-archive replaces
+                    test-failure once the archive window has passed)
 
 Usage:
     python3 scripts/enforce_mutually_exclusive_labels.py
@@ -58,6 +61,7 @@ MUTUALLY_EXCLUSIVE_SETS: list[frozenset[str]] = [
 MUTUALLY_EXCLUSIVE_PREFIXES: list[str] = [
     "c-a-",
     "c-r-",
+    "test-failure",
 ]
 
 
--- a/scripts/test_enforce_mutually_exclusive_labels.py
+++ b/scripts/test_enforce_mutually_exclusive_labels.py
@@ -128,10 +128,20 @@ class TestFindConflictingPrefix(unittest.TestCase):
     def test_case_insensitive_c_r(self):
         self.assertEqual(emxl.find_conflicting_prefix("C-R-Haiku"), "c-r-")
 
+    def test_test_failure_label_returns_prefix(self):
+        self.assertEqual(emxl.find_conflicting_prefix("test-failure"), "test-failure")
+
+    def test_test_failure_archive_label_returns_prefix(self):
+        self.assertEqual(emxl.find_conflicting_prefix("test-failure-archive"), "test-failure")
+
+    def test_case_insensitive_test_failure(self):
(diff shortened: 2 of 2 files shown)


aunger:
🤖 Reviewer

## Overview

Adds `"test-failure"` to `MUTUALLY_EXCLUSIVE_PREFIXES` so `test-failure`/`test-failure-archive` are enforced as mutually exclusive, plus matching test coverage.

## Issue found

`scripts/test_enforce_mutually_exclusive_labels.py`, in `TestLabelsByPrefix`: the new `test_test_failure_prefix_does_not_touch_testing_label` method absorbed a trailing statement that belongs to the *previous*, pre-existing test:

```python
    def test_does_not_remove_labels_from_other_prefix(self):
        """c-r-* labels must not be removed when enforcing c-a-*."""
        result = emxl.labels_to_remove_by_prefix(
            "c-a-sonnet", ["c-a-haiku", "c-r-opus", "ci"], "c-a-"
        )
        self.assertIn("c-a-haiku", result)
        self.assertNotIn("c-r-opus", result)

    def test_removes_test_failure_when_archive_added(self):
        ...

    def test_removes_test_failure_archive_when_test_failure_added(self):
        ...

    def test_test_failure_prefix_does_not_touch_testing_label(self):
        """The 'testing' label shares no real prefix with 'test-failure' and must survive."""
        result = emxl.labels_to_remove_by_prefix(
            "test-failure", ["testing", "ci"], "test-failure"
        )
        self.assertEqual(result, [])
        self.assertNotIn("ci", result)   # <-- this line belongs to test_does_not_remove_labels_from_other_prefix
```

`test_does_not_remove_labels_from_other_prefix` originally asserted three things about enforcing `c-a-*`: `c-a-haiku` is removed, `c-r-opus` survives, and `ci` survives. The third assertion (`self.assertNotIn("ci", result)`) is now missing from that test and instead sits at the end of the new, unrelated `test_test_failure_prefix_does_not_touch_testing_label`, where it's a no-op (`result` is `[]` there, so `"ci" not in []` is trivially true regardless of correctness).

Net effect: `test_does_not_remove_labels_from_other_prefix` silently lost one of its three assertions, and the new test carries a dead, misleading line. Tests still pass (nothing here fails), but coverage for the `c-a-*` group quietly regressed. Per this repo's own `code_edit.md`, a loosened test's requirements must be declared explicitly in the PR text--this one wasn't, because it looks like an artifact of how the new methods were inserted (a `Read` that stopped one line short of this trailing assertion) rather than a deliberate edit.

Fix: move `self.assertNotIn("ci", result)` back into `test_does_not_remove_labels_from_other_prefix`, and drop the now-redundant `self.assertNotIn("ci", result)` from the new test (its `self.assertEqual(result, [])` already covers that case, since `[]` implies `"ci" not in []`).

## Everything else

- The core change (`MUTUALLY_EXCLUSIVE_PREFIXES` addition + docstring update) is correct: traced through `find_conflicting_prefix`/`labels_to_remove_by_prefix` by hand for both `test-failure` and `test-failure-archive` as the added label, and the no-trailing-dash prefix does what's intended (symmetric mutual exclusion between exactly those two labels) without matching unrelated labels like `testing`.
- The rest of the new test coverage (`find_conflicting_prefix` cases, the two `labels_to_remove_by_prefix` direction tests, and the three `main()` integration tests) is well-targeted and mirrors the existing `c-a-*`/`c-r-*` coverage style.

## Verdict

`Changes requested`

aunger:
🤖 Author

Confirmed and fixed in ebe8e8d: restored `self.assertNotIn("ci", result)` to `test_does_not_remove_labels_from_other_prefix` and dropped the redundant copy from `test_test_failure_prefix_does_not_touch_testing_label`. All 76 tests still pass locally.

PR #666, opened 2026-07-12
Pin third-party CI/pre-commit deps to commit SHAs; require SHA pinning in code_edit.md

aunger:
## Summary

- `android-actions/setup-android` is a third-party action (not GitHub- or Microsoft-authored) and was pinned only to the `v4` tag in all four places it's used (`build.yml` x2, `codeql.yml`, `release.yml`), unlike `softprops/action-gh-release`, the repo's other third-party action, which is already pinned to a full commit SHA.
- Resolves `v4` to its current commit (`40fd30fb8d7440372e1316f5d1809ec01dcd3699`) and pins each usage there, with a trailing `# v4` comment matching the existing `softprops/action-gh-release` style.
- Adds a "Dependency versions" section to `agents/code_edit.md` requiring dependency versions be pinned to a commit SHA, to a recent version, with the latest patch build. Explicitly drops "tag" as a valid pin and explains why: a tag can be repointed to a different commit after the fact, so it isn't truly immutable.
- Applies the same fix to `.pre-commit-config.yaml`: the three hosted hook repos (`pre-commit-hooks`, `ruff-pre-commit`, `markdownlint-cli2`) were pinned by tag (`rev: vX.Y.Z`); resolved each to its commit SHA, keeping the version as a trailing comment.

## Test plan

- [ ] Confirm CI (build, codeql, release workflows) runs successfully against the pinned SHA.
- [x] Verified `pre-commit run` resolves and installs each hook correctly from its pinned SHA.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,7 +106,7 @@ jobs:
 
       - name: Set up Android SDK
         if: steps.diff_check.outputs.needs_full_build == 'true'
-        uses: android-actions/setup-android@v4
+        uses: android-actions/setup-android@40fd30fb8d7440372e1316f5d1809ec01dcd3699 # v4
 
       - name: Enable KVM
         if: steps.diff_check.outputs.needs_full_build == 'true'
@@ -1403,7 +1403,7 @@ jobs:
           distribution: 'temurin'
 
       - name: Set up Android SDK
-        uses: android-actions/setup-android@v4
+        uses: android-actions/setup-android@40fd30fb8d7440372e1316f5d1809ec01dcd3699 # v4
 
       - name: Cache Gradle packages
         uses: actions/cache@v5
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -69,7 +69,7 @@ jobs:
           distribution: 'temurin'
 
       - name: Set up Android SDK
-        uses: android-actions/setup-android@v4
+        uses: android-actions/setup-android@40fd30fb8d7440372e1316f5d1809ec01dcd3699 # v4
 
       - name: Cache Gradle packages
         uses: actions/cache@v5
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
           distribution: 'temurin'
 
       - name: Set up Android SDK
-        uses: android-actions/setup-android@v4
+        uses: android-actions/setup-android@40fd30fb8d7440372e1316f5d1809ec01dcd3699 # v4
 
       - name: Cache Gradle packages
         uses: actions/cache@v5
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v6.0.0
+    rev: 3e8a8703264a2f4a69428a0aa4dcb512790b2c8c # v6.0.0
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
@@ -10,14 +10,14 @@ repos:
       - id: check-added-large-files
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.8.4
+    rev: f0b5944bef86f50d875305821a0ab0d8c601e465 # v0.8.4
     hooks:
       - id: ruff
         args: [--fix]
       - id: ruff-format
 
   - repo: https://github.com/DavidAnson/markdownlint-cli2
-    rev: v0.14.0
+    rev: da711fa8b30421506086eb2dc7ea0a6c3cb27b9f # v0.14.0
     hooks:
       - id: markdownlint-cli2
 
--- a/agents/code_edit.md
+++ b/agents/code_edit.md
@@ -14,9 +14,9 @@
 
 ## Dependency versions
 
-- Every dependency version must be pinned (exact version or commit SHA; never a range, a tag, or an unpinned "latest"). A tag is not a valid pin: unlike a commit SHA, a tag can be moved to point at a different commit after the fact, so it is not truly immutable.
-- Pin to a recent version, not an old one you happen to already know; check what is current before picking a version.
-- Always pin the latest patch build of whichever minor/major version you choose.
+- This section applies to third-party dependencies: anything not published directly by GitHub itself (the `actions/` and `github/` GitHub Actions namespaces are first-party and exempt). It covers third-party GitHub Actions (for example `android-actions/setup-android`, `softprops/action-gh-release`) and pre-commit hook repos (for example `astral-sh/ruff-pre-commit`), regardless of how well-known or trusted the publisher is.
+- Every such dependency version must be pinned to a commit SHA (never a range, a tag, or an unpinned "latest"). A tag is not a valid pin; unlike a commit SHA, a tag can be moved to point at a different commit after the fact, so it is not truly immutable.
+- When you newly introduce or update a pin, pin to a recent version, not an old one you happen to already know. Check what is current before picking a version, and pin the latest patch build of whichever minor/major version you choose. This does not obligate you to bump every existing pin in a file just because you touched it; only the ones you are adding or intentionally updating.
 
 ## Test coverage
 
--- a/agents/code_edit.md
+++ b/agents/code_edit.md
@@ -12,6 +12,12 @@
   - **Do not** organize your commits by file rather than by concern (e.g. "all changes to Foo.kt in one commit, Bar.kt in another").
 - When you refactor existing code, **always** commit the pure refactoring (which does not change behavior or output) separately. **Do not** combine that refactoring with a change in functionality or behavior.
 
+## Dependency versions
+
+- Every dependency version must be pinned (exact version, tag, or commit SHA; never a range or an unpinned "latest").
+- Pin to a recent version, not an old one you happen to already know; check what is current before picking a version.
+- Always pin the latest patch build of whichever minor/major version you choose.
+
 ## Test coverage
 
 - New behavior must be accompanied by automated tests.
--- a/agents/code_edit.md
+++ b/agents/code_edit.md
@@ -14,7 +14,7 @@
 
 ## Dependency versions
 
-- Every dependency version must be pinned (exact version, tag, or commit SHA; never a range or an unpinned "latest").
+- Every dependency version must be pinned (exact version or commit SHA; never a range, a tag, or an unpinned "latest"). A tag is not a valid pin: unlike a commit SHA, a tag can be moved to point at a different commit after the fact, so it is not truly immutable.
 - Pin to a recent version, not an old one you happen to already know; check what is current before picking a version.
 - Always pin the latest patch build of whichever minor/major version you choose.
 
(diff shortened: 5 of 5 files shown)


aunger:
🤖 Reviewer

I verified all four SHAs by resolving each tag with `git ls-remote`: `40fd30fb...` correctly matches `android-actions/setup-android@v4`, `3e8a8703...` matches `pre-commit-hooks@v6.0.0`, `f0b5944b...` matches `ruff-pre-commit@v0.8.4`, and `da711fa8...` matches `markdownlint-cli2@v0.14.0`. No SHA-spoofing errors — good.

Two problems, both in the newly-added "Dependency versions" section of `agents/code_edit.md`, which is the part of this PR that will actually govern future agent behavior:

**1. The rule as written is unscoped, and is already violated by this repo.** The new text says "Every dependency version must be pinned... A tag is not a valid pin," with no carve-out. But the PR body itself justifies pinning `android-actions/setup-android` specifically *because* it's third-party ("not GitHub- or Microsoft-authored"), while leaving `actions/checkout@v6`, `actions/setup-java@v5`, `actions/cache@v5`, `actions/upload-artifact@v7`, `actions/download-artifact@v7`, `actions/github-script@v9`, and `github/codeql-action/*@v4` all pinned by tag only in the very same workflow files. That distinction lives only in the PR description — it's not encoded anywhere in the rule text that actually lands in the repo. A future agent reading `code_edit.md` literally has two bad options: treat the rule as already broken and ignore it, or "fix" every first-party GitHub Action pin in an unrelated PR because the doc gave no exception. Please add the third-party/vendor-scoping language to the rule itself (or explain why it's meant to be universal, in which case this PR is incomplete).

**2. The PR contradicts its own new "pin to a recent version" requirement.** The rule also says: "Pin to a recent version, not an old one you happen to already know; check what is current before picking a version" and "Always pin the latest patch build." But this PR resolves `ruff-pre-commit` from `v0.8.4` to its SHA without bumping it — current is `v0.15.21`, about 20 releases newer. Same for `markdownlint-cli2`: pinned at `v0.14.0`, current is `v0.23.0`. (For contrast, `pre-commit-hooks` at `v6.0.0` and `setup-android` at `v4`/`v4.0.1` do happen to already be current, so those two are fine as-is.) Introducing a "must be recent" rule in the same commit that pins two dependencies to versions many releases stale is a direct self-contradiction. Either bump `ruff-pre-commit` and `markdownlint-cli2` to current in this PR, or narrow the rule's wording to apply only when a pin is newly introduced/updated (not retroactively to every existing pin), so the doc doesn't misstate what this PR actually did.

Commit organization (setup-android pin / doc rule / pre-commit pins split into three commits) otherwise matches the by-concern guidance in `code_edit.md`.

CI is still pending (no checks reported yet as of this review).

aunger:
🤖 Author

Thanks for the review. Both points addressed in de57eee:

**1. Scoping.** Added third-party scoping language to the rule itself: it now applies to anything not published directly by GitHub (the `actions/` and `github/` namespaces are exempt), covering third-party GitHub Actions and pre-commit hook repos regardless of how trusted the publisher is. This matches the rationale already in the PR body and does not require touching the first-party `actions/*`/`github/*` pins elsewhere in the workflows, which stay out of scope for this PR.

**2. Recency self-contradiction.** I looked into actually bumping `ruff-pre-commit` and `markdownlint-cli2` to current, to see which resolution made more sense:

- `ruff-pre-commit` v0.8.4 -> v0.15.21 was clean: only 3 files reformatted, all mechanical (line-wrap changes), no lint failures.
- `markdownlint-cli2` v0.14.0 -> v0.23.0 was not clean: it introduces a new rule (MD060/table-column-style) that fails on tables in several files this PR doesn't touch (`agents/dev_orchestration.md`, `scripts/ci_monitor/README.md`, `.claude/environment.md`, `agents/task_complexity.md`). Fixing those would pull unrelated table-formatting changes into a PR about dependency pinning.

Bumping one and not the other for consistency's sake felt arbitrary, so instead I narrowed the rule's wording per your second option: recency now applies when a pin is newly introduced or intentionally updated, not retroactively when an existing pin's version is merely being re-expressed as a SHA. That's also just a more accurate description of what this PR does (convert tag pins to SHA pins at their current versions), and leaves a version bump for `ruff-pre-commit`/`markdownlint-cli2` as a follow-up if wanted.

Separately, while testing this I found `scripts/file_test_failure_issues.py`, `scripts/test_ci_monitor.py`, and `scripts/test_enforce_mutually_exclusive_labels.py` don't actually match the formatting the pinned `ruff-format` v0.8.4 produces (confirmed this predates this PR, on `main` too). Left untouched here since it's unrelated to dependency pinning.

CI (build/codeql/release) still needs to confirm the SHA-pinned `setup-android` action resolves correctly at runtime, per the test plan checkbox.

aunger:
🤖 Reviewer

Re-verified the two SHAs discussed last round (`f0b5944b...` = `ruff-pre-commit@v0.8.4`, `da711fa8...` = `markdownlint-cli2@v0.14.0`) — both still correct.

**Point 1 (scoping) is resolved.** The new "This section applies to third-party dependencies... (the `actions/` and `github/` GitHub Actions namespaces are first-party and exempt)" language matches the rationale already given in the PR body and doesn't require touching the unrelated first-party pins. Good.

**Point 2 (recency self-contradiction) is not actually resolved — only explained away in the PR comment, not in the committed text.**

Your PR comment says the new rule clarifies that recency applies "not retroactively when an existing pin's version is merely being re-expressed as a SHA." But that specific distinction — *format conversion of an existing pin* vs. *a version update* — isn't in `code_edit.md` itself. The committed text reads:

> When you newly introduce or update a pin, pin to a recent version... This does not obligate you to bump every existing pin in a file just because you touched it; only the ones you are adding or intentionally updating.

Read literally, the `ruff-pre-commit` and `markdownlint-cli2` lines in `.pre-commit-config.yaml` are exactly "ones you are... intentionally updating" in this PR — you edited both `rev:` lines by hand, in a commit whose own message says "Pin pre-commit hook revs to commit SHAs." Nothing in the rule text carves out "changing the pin's format without changing its version" as distinct from "updating a pin." A future agent reading only the file (not this PR's comment thread, which won't travel with it) has no textual basis for your interpretation — they'd read this PR's own diff as violating the rule it introduces, same as round 1.

Please fold the actual distinction into the rule text itself — e.g., something like "Converting an existing pin from a tag to the equivalent commit SHA, without changing which version it points to, is not itself an 'update' for this purpose; only pin an unfamiliar/newer version when you are intentionally changing which version is used." Otherwise the doc still won't match what this PR did, for the same reason flagged last round.

Secondary point, not blocking by itself: you confirmed the `ruff-pre-commit` bump to v0.15.21 was clean (3 files, mechanical, no lint failures) but chose not to take it, citing inconsistency with leaving `markdownlint-cli2` behind. Since one is a confirmed free win and the other is blocked by an unrelated MD060 fallout, bumping the clean one and filing a follow-up issue for the blocked one would have been more thorough than declining both and narrowing the rule to avoid the question — but this is a judgment call, not something I'm blocking on. The wording gap above is the actual blocker.

CI: `build-and-test` was still in progress as of this review; the three "No blocking labels" failures are expected during an active review cycle per `pr_participation.md` and not a concern.

aunger:
🤖 Author

Point 2 addressed for real this time, in b185291. Folded the actual distinction into the rule text itself, using your suggested wording:

> Converting an existing pin from a tag to the equivalent commit SHA, without changing which version it points to, is not itself an "update" for this purpose. Only pin an unfamiliar or newer version when you are intentionally changing which version is used.

This now travels with the file, so a future agent (or reviewer) reading only `code_edit.md` has a textual basis for why this PR's `.pre-commit-config.yaml` changes don't violate the recency rule they introduce.

On the secondary point (bumping `ruff-pre-commit` to v0.15.21): leaving that as a judgment call per your framing, since you said it isn't blocking. Happy to file a follow-up issue for it plus the `markdownlint-cli2`/MD060 bump if you'd like one on record, but didn't want to open it unprompted since this PR is about format conversion, not version bumps.

aunger:
🤖 Reviewer

Re-verified all four SHAs independently via `git ls-remote --tags` against upstream (not just against the prior round's claims): `40fd30fb...` = `android-actions/setup-android@v4` (and confirmed `v4` still equals the latest patch `v4.0.1`), `3e8a8703...` = `pre-commit-hooks@v6.0.0`, `f0b5944b...` = `ruff-pre-commit@v0.8.4`, `da711fa8...` = `markdownlint-cli2@v0.14.0`. All correct, no spoofing. Confirmed all four `setup-android` call sites (`build.yml` x2, `codeql.yml`, `release.yml`) and all three pre-commit `rev:` lines were converted.

Both points from prior rounds are now genuinely resolved in the committed text of `agents/code_edit.md`, not just explained in PR comments:

- Point 1 (scoping): the third-party/first-party (`actions/`, `github/` exempt) distinction is in the rule itself.
- Point 2 (recency self-contradiction): the new line, "Converting an existing pin from a tag to the equivalent commit SHA, without changing which version it points to, is not itself an 'update' for this purpose," now travels with the file and gives a future reader a textual basis for why this PR's tag-to-SHA conversions of `ruff-pre-commit`/`markdownlint-cli2` don't violate the recency rule the same PR introduces. Good, this is what I asked for.

No new issues found in this round. Prose in the new section is clean (checked for smart quotes/em-dashes per `.claude/rules/prose-style.md`, none present).

Not blocking, just recorded for the record since you offered it: I'd take you up on filing a follow-up issue for the `ruff-pre-commit` v0.15.21 bump (confirmed clean, 3 mechanical files) and the `markdownlint-cli2` v0.23.0 bump (blocked on new MD060 rule), so the "one is a free win" observation from last round doesn't just evaporate. Your call whether to file it before or after merge.

CI: `build-and-test` still in progress as of this review, CodeQL/Semgrep clean, the three "No blocking labels" failures are expected during an active review per `pr_participation.md`. Not blocking on CI per the same doc.

LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #666 can be merged):
- [ ] #672

- [ ] #673
- [ ] #674
- [ ] #675

aunger:
🤖 Verification Agent

Processed the before-merging list from the Verification Planner comment (#4951769345).

| Item | Result |
|---|---|
| #672 — Confirm `release.yml` resolves the pinned `setup-android` SHA at runtime | **Not automatable** in this environment |

**#672 detail:** attempted to trigger `release.yml` via `workflow_dispatch` on `ci/pin-setup-android-action`; GitHub rejected it with `403 Resource not accessible by integration` (confirmed via both the MCP tool and a raw authenticated REST call — the session's token/App installation lacks `actions: write`/`contents: write`). Did not fall back to pushing a `v*` tag, since that would both fail for the same permission reason and, if it somehow succeeded, would run a real signed release build against production secrets and publish an actual GitHub Release under a fabricated tag with no safe way to stop it after the SDK-setup step — outside this agent's authorized side effects. No local substitute (`act`) is available either (Docker daemon not running, `act` not installed). Full detail and the exact failing calls are on the issue. Left #672 open for a human with elevated credentials to run the `workflow_dispatch` (or to confirm at the next real release cut, per the issue's second suggested path).

No FAIL results. `build-and-test` and `codeql` already exercised the pinned SHA successfully as part of this PR's own CI (per #672's body), so only the `release.yml`-specific confirmation remains outstanding.

Issue #551, opened 2026-07-01
Sandbox git-proxy blocks pre-commit's first-run hook bootstrap (403 on third-party hook repos)


aunger:
## What happened

In this remote/sandboxed session, `git commit` reliably fails on the first commit of every fresh session because `pre-commit`'s hook-environment bootstrap tries to `git clone` each configured hook's own repository (e.g. `pre-commit/pre-commit-hooks`, `astral-sh/ruff-pre-commit`, `DavidAnson/markdownlint-cli2`) to resolve its manifest, and this sandbox's git proxy is scoped to only this session's own repository. Any other GitHub repo returns a 403.

This happened independently to at least two different Programmer sub-agents in the same session, on unrelated issues:
- Issue #540 / PR #548 (`agents/dev_orchestration.md` prose change)
- Issue #545 / PR #550 (`scripts/dismiss_anr.sh` / `scripts/test_dismiss_anr.sh` change), on two separate commits

## Why `SKIP=<hook-ids>` doesn't help

One Programmer investigated a narrower workaround than `--no-verify`: exempting just the blocked hooks via `SKIP=`. This does not work. Per `pre-commit` 4.6.0's source (`repository.py` / `store.py`), hook *environment installation* (the clone step that 403s) runs unconditionally for every repo listed in `.pre-commit-config.yaml`, before `SKIP` is ever consulted; `SKIP` only filters which already-installed hooks get *run*. So the same 403 occurs whether or not the blocked hooks are in `SKIP`.

## Impact

Every fresh sandbox session hits this on its first commit. The only working avoidance is `git commit --no-verify`, which per this repo's own git safety protocol requires the user's own explicit, direct authorization, not a coordinator/Orchestrator relay. In practice this means:
- Programmer sub-agents stall mid-task waiting for a human round-trip on routine commits.
- The local pre-commit gate (whitespace, merge-conflict markers, ruff, markdownlint, ktlint, etc.) is skipped entirely for the first commit of every session, relying on manual verification and downstream CI instead.

## Possible directions (not prescribing a fix)

- Allow the sandbox's git-proxy egress policy to reach the specific hook repos this project's `.pre-commit-config.yaml` depends on.
- Pre-warm/vendor the pre-commit hook environments into the base container image so bootstrap never needs network access.
- Pin/cache hook environments some other way that avoids a live clone on first use.

[...]

PR #677, opened 2026-07-13
Replace markdownlint-cli2 with flowmark for Markdown formatting

aunger:
## Markdown-lint options compared

This grid compares the four Markdown-lint candidates evaluated in this bake-off; each cell links to the GitHub diff between the two branches.

| | f | p | r | m |
|---:|---|---|---|---|
| [Flowmark](https://github.com/aunger/gallery-button-for-pixel-camera/pull/677) **f** | | [f : p](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/panache-markdown-formatter) | [f : r](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/rumdl-markdown-formatter) | [f : m](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/mdformat-markdown-formatter) |
| [Panache](https://github.com/aunger/gallery-button-for-pixel-camera/pull/678) **p** | [p : f](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/panache-markdown-formatter...claude/flowmark-markdown-formatter) | | [p : r](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/panache-markdown-formatter...claude/rumdl-markdown-formatter) | [p : m](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/panache-markdown-formatter...claude/mdformat-markdown-formatter) |
| [rumdl](https://github.com/aunger/gallery-button-for-pixel-camera/pull/679) **r** | [r : f](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/rumdl-markdown-formatter...claude/flowmark-markdown-formatter) | [r : p](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/rumdl-markdown-formatter...claude/panache-markdown-formatter) | | [r : m](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/rumdl-markdown-formatter...claude/mdformat-markdown-formatter) |
| [mdformat](https://github.com/aunger/gallery-button-for-pixel-camera/pull/680) **m** | [m : f](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/mdformat-markdown-formatter...claude/flowmark-markdown-formatter) | [m : p](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/mdformat-markdown-formatter...claude/panache-markdown-formatter) | [m : r](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/mdformat-markdown-formatter...claude/rumdl-markdown-formatter) | |

---

## Summary

- Replace the `markdownlint-cli2` pre-commit hook with `flowmark`, a Markdown auto-formatter installed from PyPI (`pip install flowmark`) and wired in as a `local`/`language: system` hook, the same pattern the existing `ktlint` hook already uses. The old hook cloned `DavidAnson/markdownlint-cli2` directly, which this sandbox's git proxy blocks for any repo other than this one (see issue #551).
- Add `.flowmark.toml`, enabling `--semantic` (sentence-boundary line wrap) and `--cleanups` (safe structural fixes like unbolding all-bold headings), and setting `width` to a very large number (`100_000`) rather than leaving the 88-column default, so `--semantic` is the sole determinant of line breaks. `--smartquotes` and `--ellipses` are deliberately left off, and `--auto` is avoided entirely, since both would curl straight quotes and a literal "..." into typographic characters, the opposite of this repo's ASCII-only punctuation rule.
- Install flowmark automatically at session start (pinned to 0.7.2), following the same present-check/`--force-reinstall` pattern already used for `pre-commit`. Updated `.claude/environment.md`'s tool/hook tables to match.
- Ran the new hook across the repo's existing Markdown and committed the resulting reformat (mechanical rewrap only, no prose content changes).
- Removed `.markdownlint.yaml` since nothing reads it anymore.
- Verified `pre-commit run --all-files` passes locally, aside from one pre-existing, unrelated `ruff-format` finding on three Python scripts already present on `origin/main` and out of scope for this PR.

## Known limitations

- **Cannot reverse pre-existing curly quotes, em-dashes, or ellipsis characters back to ASCII.** Verified directly against Flowmark's source (`smartquotes.py`/`ellipses.py` are one-directional: they only ever add typographic characters, never remove them), its release notes (no version from v0.5.8 through the current v0.7.2 mentions a reverse-normalization feature), and a live test (feeding real curly-quote/em-dash/ellipsis characters through the installed binary with `--smartquotes`/`--ellipses`/`--auto` all off passes them through byte-for-byte unchanged). Panache's `smart` extension actively normalizes existing curly punctuation back to ASCII, a real capability gap, not just a stylistic wash: Flowmark cannot fix legacy curly punctuation if any gets pasted in, Panache can. This repo's Markdown is already all-ASCII, so this is not an active problem today.
- **Does not enforce "no spaces around dashes or double-hyphens" at all.** A source-level check (no dash/hyphen handling anywhere in the formatter) confirms Flowmark will not flag or fix `foo -- bar` style spacing. This is the one prose-style.md rule Flowmark cannot enforce; no workaround hook was added for it.
- **A single very long sentence produces a correspondingly long line.** `width` is set to a very large number (see `.flowmark.toml`) specifically so `--semantic` is the sole determinant of line breaks: a sentence only ever breaks at its end, never mid-sentence at a width limit. This matches Panache's `wrap = "sentence"` mode, which has the identical trade-off. No sentence in this repo's current Markdown is long enough for this to be visually disruptive, but an exceptionally long one would produce a correspondingly long line with no fallback wrap.
- **Blank lines inside a multi-paragraph blockquote always render with a trailing space.** Flowmark renders such a line as `"> "` (a trailing space after the `>`), which is how it keeps the line a valid blockquote continuation. `trailing-whitespace` (from `pre-commit-hooks`) would otherwise strip that space back out on every run, so `.pre-commit-config.yaml` excludes Markdown from that hook to prevent the two hooks fighting forever (confirmed on `SPEC.md`'s `> [!NOTE]` callout).

## Parallel experiment

This is one of two parallel, independent experiments replacing the `markdownlint-cli2` hook, evaluated against this same repo's `.claude/rules/prose-style.md` rules: this PR (flowmark) and a sibling PR using Panache. Both are meant for the user to compare and pick one (or neither). A cross-review from the Panache engineer will follow once both PRs are ready.

Compare the two branches directly: https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/panache-markdown-formatter


aunger:
🤖 Reviewer

Cross-review from the Panache PR (#678), comparing both against the same three `.claude/rules/prose-style.md` rules. Disclosure up front so you're comparing against my actual state, not an idealized one: my own PR has a known gap where Panache normalizes a curly em dash to `---` (Pandoc's triple-hyphen convention) rather than this repo's preferred `--`, and it does not enforce "no spaces around dashes" at all.

## Rule 2 (ASCII punctuation): the reversibility claim checks out, but verify it the same way for both bullets

I installed `flowmark==0.7.2` (the pinned version) and read the actual installed source, not just `--help`, to check both "Known limitations" bullets independently.

- **Confirmed accurate:** `flowmark/typography/smartquotes.py` and `ellipses.py` are both one-directional. `smart_quotes()` only ever replaces straight quotes with curly ones; `ellipses()` only ever replaces `...` with `…`. Neither module contains, or could easily be flagged to contain, a reverse path. So "no mechanism to reverse ... back to ASCII" is correct, not just an inference from the flags being off.
- **Confirmed accurate, same as your dash bullet:** grepping the whole package for dash/hyphen handling turns up nothing except table-separator-row normalization (`block_heuristics.py`, purely structural) and em dash used only as a boundary character inside the smartquotes/ellipses regexes (deciding where a quote or ellipsis may start/end, not transforming a dash itself). There is no dash normalization or spacing logic anywhere.
- **The asymmetry worth naming:** your dash bullet says "a source-level check confirms," but the quotes/ellipsis bullet doesn't say how it was verified, it reads like it was inferred from the flags being off rather than checked the same way. Both bullets turn out to be true, but only one says how you know. Worth making that explicit, since it's the more consequential of the two claims: Panache's `smart` extension actively normalizes existing curly punctuation back to ASCII (I verified this empirically in my own PR), so this is a real capability gap between the two tools, not a stylistic wash. Flowmark cannot fix legacy curly punctuation if any gets pasted in; Panache can. Worth stating that contrast plainly in the limitations bullet rather than just "it only avoids introducing new ones," so a reader comparing the two PRs sees the asymmetry without having to check both PRs' source themselves.

## Rule 1 (sentence-boundary wrap): concrete gap, not just a nit

`.flowmark.toml` sets `semantic = true` but does not set `width = 0`, and `--width`/`-w` defaults to 88 independently of `--semantic` (confirmed via `flowmark --help`). I verified empirically that this means a single sentence longer than 88 columns still gets wrapped mid-sentence, at a non-sentence boundary. Minimal repro:

```
$ flowmark --semantic --cleanups -o - sample.md
This is one single very long sentence with no internal punctuation that just keeps going
and going past the eighty eight column default width limit to see what happens.
Short one.
```

This isn't synthetic, it already happened in your own reformatting. From `.claude/agents/NOTES.md`, one sentence became two lines broken mid-sentence:

```
+Effort modulates how much the model reasons (thinking-token spend) and, secondarily, how
+persistently it works and how long its output runs.
```

The break lands after "how", not at a sentence boundary. Panache's `wrap = "sentence"` ignores line-width entirely by design (verified: the same sentence stays on one line no matter how long), which is what actually delivers the rule's stated "diff-stability" benefit: editing a word near the start of a long sentence won't reflow the rest of it under Panache, but it will under your current config, since the sentence still has an internal wrap point subject to width-based refill. If you want `--semantic` alone to be the deciding factor (matching the rule's literal "split at sentence breaks", not "split at sentence breaks or 88 columns, whichever comes first"), consider adding `width = 0` to `.flowmark.toml`. If you'd rather keep the 88-column cap intentionally, that's a legitimate call, but then it should be named as a limitation next to the other two, since it's the same category of gap (partial, not full, rule compliance).

## Structural comparison (environment.md / session-start.sh / Known limitations)

- `session-start.sh`: you inserted flowmark's install as step 3b and renumbered the existing pre-commit (3b to 3c) and hook-wiring (3c to 3d) steps to make room. My PR instead appended panache as a new 3d, leaving 3a to 3c untouched. Renumbering unrelated, already-working steps just to slot a new one in alphabetically-before-pre-commit adds avoidable diff churn and risk (anything that ever cites "step 3b" or "step 3c" by name silently means something different after this PR). Consider switching to append-only, unless there's a real ordering dependency I'm missing (I didn't see one: the pip installs for flowmark and pre-commit look independent of each other).
- `environment.md`, "Tools installed at session start" table: the `flowmark` row's Location cell adds a `(pip)` annotation ("`~/.local/bin/flowmark` (pip)") that no other row in the same table uses, including your own `pre-commit` row, which is also pip-installed but has no such tag. Minor, but it's an inconsistency within your own table.
- "Known limitations": your two bullets are plain sentences; mine use a bolded lead-in phrase per bullet. Not a correctness issue, but the two PRs will be read side by side, and matching that formatting would make the comparison easier to skim. Your "Notes" section (the blockquote/trailing-whitespace fight with `trailing-whitespace`) is a good catch and clearly explained, better than anything I had to deal with on the Panache side since it never produced that artifact. I'd suggest either bringing that into "Known limitations" (it's a real, permanent constraint of using flowmark here, not just a note) or, if you keep it separate, add an equivalent "Notes" section to my PR for the reader's sake. I'll leave that decision to whoever finalizes both PRs.

## Small, non-blocking observation

Not a request, just interesting: I hit a genuine, reproducible Panache bug on `SPEC.md`'s `**EC-09**` bullet (a bold span immediately followed by a code span came out with corrupted, unbalanced asterisks). I checked how flowmark handles the identical line, it comes out clean:

```
- **EC-09** **USM query returns stale data:** `UsageStatsManager` event data can lag by
  a few hundred milliseconds.
```

Not something to act on, just confirms it's a Panache-specific defect rather than a generally hard case.

## Summary

Both PRs are close to comparable quality. The one thing I'd actually ask you to change before this is ready to put in front of the user: either disable width-based wrapping in `.flowmark.toml` or explicitly document the mid-sentence-wrap gap as a limitation, since right now the PR claims "semantic (sentence-boundary line wrap)" without qualifying that it is capped at 88 columns and will still break sentences mid-stream past that. Everything else here is a nit-level convergence request (rigor-of-claim wording, table consistency, step renumbering).

aunger:
🤖 Author

Thanks, all four points landed. Pushed two follow-up commits.

**Rule 1 width gap (the main ask):** fixed, but not with `width = 0` as literally suggested, I checked and that would have made things worse. Flowmark's own source (`linewrapping/line_wrappers.py`) special-cases `width <= 0` to collapse an entire paragraph onto a single line *before* it ever calls the sentence splitter, so `width = 0` would have thrown away sentence-per-line wrapping entirely (verified: tested both `--width 0` and a large `--width` value directly against the CLI, the `0` case merges every sentence in a paragraph onto one line). Set `width = 100_000` instead, confirmed empirically that `--semantic` is now the sole line-break authority (re-tested the exact NOTES.md sentence you quoted, it's back to one line), and re-ran the hook across the repo. Documented the resulting trade-off (a single very long sentence now produces a correspondingly long line, matching Panache's own sentence/semantic mode) as a new Known limitations bullet, worded to match the parity you flagged.

**Rule 2 verification-method asymmetry:** agreed, fixed. The "cannot reverse" bullet now states how it was verified (source, full release history, live byte-level test) and states the Panache contrast plainly, matching how the dash bullet was already written.

**session-start.sh renumbering:** agreed, no real ordering dependency, fixed. Flowmark's install is now appended as step 3d, steps 3a-3c are untouched, matching #678's own append-only approach.

**`(pip)` annotation:** agreed, inconsistent for no reason, removed.

On folding the blockquote/trailing-whitespace item into Known limitations rather than a separate Notes section: took that suggestion too, it reads better grouped with the other three limitations.

PR body's Known limitations section and Summary are both updated to reflect all of this.

PR #678, opened 2026-07-13
Replace markdownlint-cli2 with a Panache formatting hook

aunger:
## Markdown-lint options compared

This grid compares the four Markdown-lint candidates evaluated in this bake-off; each cell links to the GitHub diff between the two branches.

| | f | p | r | m |
|---:|---|---|---|---|
| [Flowmark](https://github.com/aunger/gallery-button-for-pixel-camera/pull/677) **f** | | [f : p](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/panache-markdown-formatter) | [f : r](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/rumdl-markdown-formatter) | [f : m](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/mdformat-markdown-formatter) |
| [Panache](https://github.com/aunger/gallery-button-for-pixel-camera/pull/678) **p** | [p : f](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/panache-markdown-formatter...claude/flowmark-markdown-formatter) | | [p : r](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/panache-markdown-formatter...claude/rumdl-markdown-formatter) | [p : m](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/panache-markdown-formatter...claude/mdformat-markdown-formatter) |
| [rumdl](https://github.com/aunger/gallery-button-for-pixel-camera/pull/679) **r** | [r : f](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/rumdl-markdown-formatter...claude/flowmark-markdown-formatter) | [r : p](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/rumdl-markdown-formatter...claude/panache-markdown-formatter) | | [r : m](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/rumdl-markdown-formatter...claude/mdformat-markdown-formatter) |
| [mdformat](https://github.com/aunger/gallery-button-for-pixel-camera/pull/680) **m** | [m : f](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/mdformat-markdown-formatter...claude/flowmark-markdown-formatter) | [m : p](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/mdformat-markdown-formatter...claude/panache-markdown-formatter) | [m : r](https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/mdformat-markdown-formatter...claude/rumdl-markdown-formatter) | |

---

## Summary

Replaces the `markdownlint-cli2` pre-commit hook with [Panache](https://panache.bz/), a Rust Markdown/Quarto/Pandoc formatter, configured to enforce this repo's `.claude/rules/prose-style.md` conventions as closely as the tool allows.

- `panache.toml` (new): `flavor = "gfm"`, `wrap = "sentence"` (one sentence per line), and `[extensions] smart = true` (normalizes any curly quotes, en/em dashes, or a Unicode ellipsis back to ASCII, rather than the reverse).
- `.pre-commit-config.yaml`: drops the `DavidAnson/markdownlint-cli2` repo entry (it required cloning a third-party pre-commit hook repo, which this sandbox's git-proxy does not allow) and adds a `local`/`language: system` `panache` hook, matching the existing `ktlint` hook's pattern.
- `.markdownlint.yaml` removed (no longer used by any hook).
- `.claude/hooks/session-start.sh` and `.claude/environment.md` updated so `panache` installs automatically in a fresh session, via `cargo install` (a curl'd GitHub release binary was tried first but rejected by this session's git-proxy for this specific repo; crates.io is reachable directly). Tool installs (ktlint, panache, pre-commit) are grouped before the pre-commit git-hook-wiring step.
- All existing Markdown reformatted with `panache format` and reviewed for sanity (commits: hook wiring, the mechanical reformat, environment docs/automation, then a step-ordering follow-up, so the auto-formatting diff is easy to review on its own).
- `pre-commit run --all-files` passes cleanly locally, aside from one pre-existing, unrelated `ruff-format` finding on three Python scripts that already reproduces on a clean `origin/main` checkout with none of this PR's changes, and is out of scope here.

## Known limitations

- **Spaces around dashes are not enforced.** prose-style.md says to omit the spaces around a double-hyphen dash (`like this--not this`), but Panache does not touch whitespace around dashes at all, whether they are already ASCII (`--`) or converted from a curly en/em dash. It leaves `word -- word` and `word --- word` exactly as found. No lint rule covers this either (the built-in lint catalog is structural: headings, cross-references, footnotes, YAML frontmatter, and similar; nothing about prose typography). This is the one prose-style.md rule Panache cannot enforce; no workaround hook was added for it.
- **Em dashes normalize to `---`, not this repo's preferred `--`.** When Panache's `smart` extension does normalize a curly em dash back to ASCII, it maps it to Pandoc's own triple-hyphen convention (`---`) rather than this repo's double-hyphen style. Confirmed hardcoded, not a config gap: `panache-formatter` 0.19.0's `src/formatter/smart.rs` maps `'\u{2013}'` (en dash) to `"--"` and `'\u{2014}'` (em dash) to `"---"` directly in code, and `FormatterExtensions` in `config.rs` has no dash-style field to override it. It still lands on plain ASCII (satisfying the "avoid non-ASCII punctuation" half of the rule), just not always the exact hyphen count the style guide prefers. This repo had no curly dashes to begin with, so it did not come up in practice, but would if one were ever introduced.
- **`wrap = "sentence"` (and `"semantic"`) apply no width limit at all, so an unusually long sentence becomes one very long line.** Checked for a config combination that would let `wrap = "sentence"` fall back to a width-based break for overlong sentences; there isn't one, and it's hardcoded the same way the em-dash target is. In `panache-formatter` 0.19.0's `src/formatter/inline_layout.rs`, the width-limit check inside `StreamingCoreSink::consume` is gated behind `if !self.sentence_mode`, and both `sentence_lines_for_paragraph` and `semantic_lines_for_paragraph` call `wrapped_lines_for_node` with an empty widths slice (`&[]`); `config.line_width` (default 80) is never threaded into either path. This is a real, present-day gap, not just a theoretical one: `.github/workflows/CLAUDE.md` already has several single-sentence lines around 130-190 characters (for example line 9, 180+ characters) that Panache leaves completely untouched, which is why that file does not appear in this PR's reformatting diff at all, every sentence there was already one per line. This partially undercuts the word-wrap rule's stated "moderate line lengths" rationale for any sentence that runs long.

## Context: parallel experiment

This is one of two independent experiments evaluating a `markdownlint-cli2` replacement against this repo's `.claude/rules/prose-style.md` rules, both replacing the same hook slot in mutually exclusive ways so they can be compared directly. The other evaluates Flowmark on a separate branch/PR (#677). A cross-review between the two has happened; both known-limitations lists above reflect points raised and independently verified during that exchange.

Compare the two branches directly: https://github.com/aunger/gallery-button-for-pixel-camera/compare/claude/flowmark-markdown-formatter...claude/panache-markdown-formatter

## Test plan

- [ ] Confirm Panache's sentence-per-line wrapping reads acceptably across the reformatted docs (a matter of taste beyond what automated checks can confirm).


aunger:
🤖 Reviewer

Cross-review from the Flowmark PR (#677), comparing both against `.claude/rules/prose-style.md` rule by rule. I installed `panache` locally (cargo, same 2.61.0 pin) and read the `panache-formatter` 0.19.0 source directly rather than taking either PR's claims at face value, so the notes below are source- and test-verified, not just doc-read.

## Rule 2 ("avoid angle quote, em-dash, and ellipsis characters")

Point 1 asked me to double check my own PR's claim that Flowmark cannot reverse existing curly punctuation to ASCII, since this PR's write-up reports the opposite for Panache. I re-verified this three ways, not just by reading Flowmark's `--help` text:

- Read the actual formatter source (`smartquotes.py`, `doc_cleanups.py`): no code path anywhere converts non-ASCII punctuation to ASCII; `--smartquotes`/`--ellipses` only ever add typographic characters, never remove them.
- Checked every Flowmark release's notes from v0.5.8 through the current v0.7.2: no release ever mentions a reverse-normalization feature.
- Fed real U+2014 em dash, U+201C/U+201D curly quotes, and a U+2026 ellipsis character through the installed `flowmark` binary with only `--semantic --cleanups` (no `--smartquotes`/`--ellipses`/`--auto`): all four passed through byte-for-byte unchanged.

So this holds up: Panache's `smart` extension normalizing existing curly punctuation back to ASCII is a real, verified capability advantage over Flowmark for this rule. Credit where due, that's a genuinely better fit for "avoid angle quote, em-dash, and ellipsis characters" than what I shipped.

On point 2 (whether the `---` em-dash target is a fixable near-miss): I checked, and it is not fixable via config. `panache-formatter-0.19.0/src/formatter/smart.rs` hardcodes the mapping:

```rust
'\u{2013}' if smart_enabled => out.push_str("--"),   // en dash -> "--"
'\u{2014}' if smart_enabled => out.push_str("---"),  // em dash -> "---"
```

`FormatterExtensions` in `config.rs` only has plain `smart: bool` / `smart_quotes: bool` fields; there is no dash-style/dash-length field anywhere in the schema, and the CLI's only escape hatch (`-o extensions.<name>=<bool>`) only toggles booleans. So there's no config knob to push a follow-up commit for here, current Panache just does not expose this. One thing worth adding to the PR's limitation note since it's a small extra wrinkle: an en dash (`–`, U+2013) already converts to this repo's preferred `--`, it's specifically the em dash that lands on `---` instead. Not actionable, just worth being precise about in the writeup.

## Rule 1 ("word wrap... balances line-diff stability with moderate line lengths")

This one I don't think either PR's write-up mentions, but it's a real difference worth surfacing. `wrap = "sentence"` (and `"semantic"`, which is defined in the source as literally `sentence` plus `preserve_newlines: true`) never applies any width limit at all. In `inline_layout.rs`:

```rust
if !self.sentence_mode {
    // ...width_limit check and break...
}
```

The width-based break only fires when NOT in sentence mode. So a single long sentence becomes one long line, full stop, no fallback wrap. This isn't a synthetic worry, it already shows up in this repo: `.github/workflows/CLAUDE.md` doesn't appear in this PR's diff at all, and I confirmed why, every sentence in that file was already one-per-line, so Panache found nothing to touch, even though several of those lines are 130+ and 180+ characters long (measured directly). Flowmark's `--semantic` still falls back to width-wrapping inside an overlong sentence (that file's equivalent lines got wrapped around 88 cols in #677). Neither approach is "wrong" per a literal reading of "split prose at sentence breaks", but the rule's own stated rationale is "moderate line lengths", and Panache's sentence/semantic modes don't deliver that for long sentences. Worth a line in this PR's known limitations, I don't think it's fixable by config (same hardcoded-in-source situation as the dash target), just worth naming.

## Rule 3 (dash/double-hyphen spacing)

Both PRs land in the same place: neither tool has any concept of this, and both PRs correctly declined to bolt on a custom workaround hook for it. No note needed here, just confirming parity.

## One thing Panache does better than my PR

I tested this directly since it seemed like a fair thing to check both ways: my PR had to exclude Markdown from the `trailing-whitespace` hook because Flowmark always renders a blank line inside a multi-paragraph blockquote as `"> "` (trailing space), which `trailing-whitespace` then strips, causing the two hooks to fight forever. I built the same test case (a `> [!NOTE]` blockquote with two paragraphs) and ran it through `panache format`: it renders the blank continuation line as a bare `>`, no trailing space, stable and idempotent. So this PR didn't need (and doesn't have) an equivalent `trailing-whitespace` exclude, and won't hit that failure mode. That's a real point in this PR's favor, no action needed, just noting it since point 3 asked for a side-by-side.

## Nits on session-start.sh / environment.md thoroughness

- Minor ordering nit: step 3d (panache install) is added after step 3c (wiring pre-commit into git hooks), so the script wires the hook before installing the tool the hook depends on. Harmless in practice (pre-commit doesn't need the binary until the first real commit touches Markdown), but #677 groups all three tool installs before the hook-wiring step, which reads slightly cleaner. Not blocking, take it or leave it.
- The PR body doesn't state whether `pre-commit run --all-files` passes locally. #677's writeup has an explicit line confirming this (and calling out one pre-existing, unrelated `ruff-format` finding it left alone). Since these two PRs are meant to be compared side by side, it'd help the reader if this PR stated the same thing explicitly, even if the answer is just "yes, clean."
- I also tried to independently check the session-start.sh comment's claim that curl'd GitHub release binaries are blocked the same way `git clone` is (not just crate/pip installs). I couldn't confirm or refute it cleanly: I tried curling this repo's own already-merged `ktlint` release-binary download (the exact URL in `session-start.sh` step 3a, which is presumably known-working since it's already on `main`) and got a 403 in my current session too. So the proxy's behavior here may be more session-dependent or broader than the task background assumed, this isn't a knock against choosing `cargo install`, if anything it may support it, just flagging that I can't fully verify the stated reasoning either way.

## Verdict

No blocking correctness problems found, the formatter-bug writeup (SPEC.md corruption + exclude workaround) reads as an honest, well-documented, non-hacky catch. The one thing I'd actually ask for before calling this done is adding the Rule 1 (unbounded sentence length) limitation to the PR's Known limitations, since it's real, verified, and the kind of thing the user comparing both PRs would want called out explicitly the same way the dash-target one already is. The `pre-commit run --all-files` confirmation and the session-start.sh step-ordering nit are smaller asks, take or leave.

Issue #682, opened 2026-07-15
Automated Test: Increase toast message time to 3 seconds


aunger:
The automated tests show one-second toasts before each test case. Increase the delay cancel time to three seconds. Also increase the toast duration from SHORT to LONG.

PR #683, opened 2026-07-15
Rename "testing" label to "automated tests" in title-matching automation

### LINKED ISSUE #643: Rename “testing” label to “automated tests”
Watch out for how it’s used in existing automation.

aunger:
Fixes #643 (the code-automation portion; see below).

`automated tests` and `testing` are two distinct label objects in this repo, not one label that was renamed in place. This PR retargets the automation so it applies `automated tests` going forward:

- `scripts/label_by_title.py`: the `LABEL_PATTERNS` dict key `"testing"` is now `"automated tests"`.
- `.github/release.yml`: the `changelog.exclude.labels` entry `testing` is now `automated tests`.
- Updated the corresponding unit tests in `scripts/test_label_by_title.py`, `scripts/test_backfill_labels_by_title.py`, and `scripts/test_enforce_mutually_exclusive_labels.py` (including renaming test methods/classes that referenced the old label name).

`scripts/backfill_labels_by_title.py`'s `TRACKED_LABELS` derives from `label_by_title.LABEL_PATTERNS`, so it picks up the new name automatically. `scripts/test_label_existence_integration.py` derives its required-labels set from `.github/release.yml`, so it now checks for "automated tests" instead of "testing" against the real repo (which already has that label, applied independently of this PR).

### What this PR does not do

This PR does not delete or rename the `testing` label object, and does not retroactively relabel the 83 issues/PRs (mostly closed) that still carry it, including #682 (open, carrying both labels). Both are out-of-repo GitHub-administration actions that no script in this repo can perform. Tracked as a follow-up: #684.

### Test plan
- [x] `python3 -m unittest scripts/test_label_by_title.py scripts/test_backfill_labels_by_title.py scripts/test_enforce_mutually_exclusive_labels.py` — all pass.
- [ ] Delete the stale `testing` label from the repository (tracked in #684, not part of this PR).

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -2,5 +2,5 @@ changelog:
   exclude:
     labels:
       - ci
-      - testing
+      - automated tests
       - agents
--- a/scripts/label_by_title.py
+++ b/scripts/label_by_title.py
@@ -74,7 +74,7 @@ LABEL_PATTERNS: dict[str, re.Pattern[str]] = {
         rf"|{_B}sub[\W_]?agent\w*",
         re.IGNORECASE,
     ),
-    "testing": re.compile(
+    "automated tests": re.compile(
         rf"{_B}e2e{_B}"
         rf"|{_B}unit{_B}"
         rf"|{_B}test\w*"
--- a/scripts/test_backfill_labels_by_title.py
+++ b/scripts/test_backfill_labels_by_title.py
@@ -25,13 +25,13 @@ class TestTrackedLabels(unittest.TestCase):
     def test_includes_every_rule_label(self):
         self.assertIn("ci", blt.TRACKED_LABELS)
         self.assertIn("agents", blt.TRACKED_LABELS)
-        self.assertIn("testing", blt.TRACKED_LABELS)
+        self.assertIn("automated tests", blt.TRACKED_LABELS)
 
     def test_has_no_extra_labels(self):
         # A label with no title-matching rule (e.g. "product") must not be
         # tracked: it could never appear in "matched", so it would sit in
         # "miss" forever, on every run, for every issue/PR that carries it.
-        self.assertEqual(blt.TRACKED_LABELS, frozenset({"ci", "agents", "testing"}))
+        self.assertEqual(blt.TRACKED_LABELS, frozenset({"ci", "agents", "automated tests"}))
 
 
 # ---------------------------------------------------------------------------
@@ -132,7 +132,7 @@ class TestBuildReport(unittest.TestCase):
 
 class TestApplyLabels(unittest.TestCase):
     def test_posts_once_per_issue(self):
-        to_apply = {1: ["ci"], 2: ["agents", "testing"]}
+        to_apply = {1: ["ci"], 2: ["agents", "automated tests"]}
         with patch.object(blt.label_by_title, "gh_api") as mock_api:
             result = blt.apply_labels(to_apply, "owner/repo", "tok")
         self.assertEqual(mock_api.call_count, 2)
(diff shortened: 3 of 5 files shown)


aunger (commented):
🤖 Reviewer

The code changes themselves are correct and complete. I checked out the branch and ran the three named test modules (196 tests, all pass), and did a repo-wide search for other hardcoded references to the old label name; none remain outside these files. The `LABEL_PATTERNS` key, `.github/release.yml`, and the three test files are all consistently updated to "automated tests", and the renamed test methods/classes read fine.

One factual problem with the PR description, though. It states: "The GitHub label was renamed from 'testing' to 'automated tests'" and that the repo "already has the renamed label." I checked the live repo state via the GitHub API and that is not what happened:

- `automated tests` (id `LA_kwDORy4CL88AAAACeI1WyA`) and `testing` (id `LA_kwDORy4CL88AAAACrxeOHQ`) are two distinct label objects, not one label with a new name. A true GitHub label rename keeps the same id.
- The old `testing` label still exists and is still attached to at least issue #682 (alongside the new `automated tests` label, which was applied there by title-matching against the *old*, still-live `main` code).
- `scripts/label_by_title.py`'s own docstring says "Labels are only ever added, never removed." So merging this PR will not clean up `testing` from #682 or any other issue that still carries it; it just stops the automation from ever applying `testing` again going forward.

So the label wasn't actually renamed; a new label was created alongside the old one, and the old one is now orphaned (no longer referenced by any automation after this merges, but still sitting on at least one open issue). That's a meaningfully different state than the PR describes, and issue #643's title ("Rename 'testing' label to 'automated tests'") isn't fully satisfied by a code-only change if the actual label object was never renamed/consolidated.

Please either:
1. Correct the PR description to accurately reflect that a new label was created (not a rename), and confirm whether cleaning up the stale `testing` label (deleting it, or removing it from issues like #682) is in scope here or should be tracked as a separate follow-up, or
2. If completing the actual rename/cleanup is achievable from here, do that so the issue is genuinely resolved as titled.

Everything else in the diff looks good.

aunger:
🤖 Author

You're right, and I verified it independently via the API: `testing` (`LA_kwDORy4CL88AAAACrxeOHQ`) and `automated tests` (`LA_kwDORy4CL88AAAACeI1WyA`) are separate label objects, and `testing` is still attached to 83 issues/PRs, #682 among the open ones.

I've corrected the PR description to stop claiming this was an in-place rename, and to spell out what this PR does and doesn't do.

On the cleanup question: I don't have a tool available that can rename or delete a GitHub label (checked -- the GitHub MCP surface here has `get_label` but no create/update/delete for labels), so consolidating or removing the stale `testing` label isn't something I can do from this PR. Deleting the label is a one-shot admin action (it strips the label from every issue/PR at once, so no per-issue relabeling pass is needed first), which makes it squarely an out-of-repo action rather than a code change. I filed #684 to track it separately, per the scope-creep guidance in `pr_participation.md`.

Given that constraint, I'd call the code-automation portion of #643 complete as of this PR, with #684 as the remaining out-of-repo piece.

aunger (commented):
🤖 Reviewer

Confirmed: the GitHub MCP surface available here only exposes `get_label`, no create/update/delete for labels, so the Author's claim that deleting/consolidating the `testing` label isn't achievable from a code PR checks out.

The corrected PR description now accurately describes the situation (two distinct label objects, not an in-place rename), and #684 properly tracks the out-of-repo cleanup as a follow-up. That resolves my concern. Combined with the earlier verification (196 tests pass, no other stale references to the old label name anywhere in the repo), this is good to merge.

LGTM.

PR #685, opened 2026-07-15
Match ci_monitor --include-* patterns against suite as well as name

### LINKED ISSUE #646: Match ci_monitor.py --include-pass against suite as well as name
Follow-on item. This is a follow-on item and does not block merging PR #644.

Issue #601 illustrated the intended validation query for the synthesized PASS marker as `--include-pass 'SetupActivityPermissionDialog'` (a suite-shaped pattern). During review of PR #644 it was established that `ci_monitor.py`'s `parse_fails` matches `--include-pass PATTERN` only against the marker's `name` field, never `suite`, so that illustrative query silently matches nothing against the fixed `name` (`setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`).

Source (reviewer's original finding): https://github.com/aunger/gallery-button-for-pixel-camera/pull/644#pullrequestreview-4674545383

Source (author's follow-up note, explicitly deferring a `ci_monitor.py` change as out of scope for PR #644): https://github.com/aunger/gallery-button-for-pixel-camera/pull/644#issuecomment-4939009887

> I did not change `ci_monitor.py` to also match `--include-pass` against `suite`, since that's out of scope for this PR per the issue's own scope section ("No changes to ... `scripts/ci_monitor/*`"). Noting it here as a possible follow-up rather than filing a new issue myself, since #601 is still open and can absorb the note if a follow-up issue is wanted.

## Proposal

Extend `scripts/ci_monitor/ci_monitor.py`'s `parse_fails` so that `--include-pass PATTERN` (and, for consistency, `--include-fail PATTERN` if applicable) can also match against the marker's `suite` field, not only `name`. This would let suite-shaped queries like `--include-pass 'SetupActivityPermissionDialog'` surface the synthesized PASS marker added in PR #644, matching the validation workflow originally envisioned in issue #601.

aunger:
## Summary

`ci_monitor.py`'s `parse_fails` matched an `--include-<outcome> PATTERN` regex only against a marker's `name` field, never its `suite`. As a result the suite-shaped validation query illustrated in issue #601 (`--include-pass 'SetupActivityPermissionDialog'`) silently matched nothing against the fixed `name` (`setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`), so the synthesized PASS marker added in PR #644 was never surfaced.

This PR extends the filter so a non-`None` pattern selects a marker when it matches **either** `name` **or** `suite`. A suite-shaped query now surfaces a marker whose distinguishing text lives in `suite`, matching the validation workflow envisioned in #601.

The change lives in the single shared filter path, so it applies uniformly to every include filter (`--include-fail`, `--include-skip`, `--include-pass`), as the issue requested for consistency.

## Changes

- `scripts/ci_monitor/ci_monitor.py`: `parse_fails` now matches the pattern against `name` or `suite`; docstring and `--include-*` argparse help updated.
- `scripts/ci_monitor/README.md`: the per-test-filter table and prose now document name-or-suite matching, with the #601 example.
- `scripts/test_ci_monitor.py`: new group `(ax)` covering suite matching (via `suite`), unchanged name matching, non-matching suppression, and the same behavior for `--include-fail`.

## Testing

- `python3 scripts/test_ci_monitor.py` -> 315 passed, 0 failed.

Fixes #646

### DIFF
--- a/scripts/ci_monitor/README.md
+++ b/scripts/ci_monitor/README.md
@@ -44,14 +44,15 @@ Independent filter flags narrow or expand which per-test outcomes are streamed:
 
 | Flag | Effect |
 |---|---|
-| `--include-fail [PATTERN]` | Report FAIL markers (default); optionally restrict to those whose `name` matches PATTERN. |
+| `--include-fail [PATTERN]` | Report FAIL markers (default); optionally restrict to those whose `name` or `suite` matches PATTERN. |
 | `--no-include-fail` | Suppress all FAIL markers. |
-| `--include-skip [PATTERN]` | Report SKIP markers (default); optionally restrict to those whose `name` matches PATTERN. |
+| `--include-skip [PATTERN]` | Report SKIP markers (default); optionally restrict to those whose `name` or `suite` matches PATTERN. |
 | `--no-include-skip` | Suppress all SKIP markers. |
-| `--include-pass [PATTERN]` | Report PASS markers (not the default); optionally restrict to those whose `name` matches PATTERN. |
+| `--include-pass [PATTERN]` | Report PASS markers (not the default); optionally restrict to those whose `name` or `suite` matches PATTERN. |
 | `--no-include-pass` | Suppress all PASS markers (explicit form of the default). |
 
-Each `--include-*` flag takes an **optional regex** matched against the marker's `name` field.
+Each `--include-*` flag takes an **optional regex** matched against the marker's `name` field or its `suite` field: a marker is selected when the pattern matches *either* (issue #646).
+This lets a suite-shaped query like `--include-pass 'SetupActivityPermissionDialog'` surface a marker whose distinguishing text lives in `suite` rather than `name`.
 Supplied without a pattern it includes *all* markers of that outcome.
 The three outcomes keep their distinct labels in output: `--include-pass` never relabels a SKIP as PASS.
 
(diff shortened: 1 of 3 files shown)


aunger:
🤖 Reviewer

Verified this against issue #646 and PR #644's original finding.

- The fix is in the right place: the single shared filter block in `parse_fails` now does `re.search(pattern, name) or re.search(pattern, suite)`, so all three `--include-*` flags (`fail`/`skip`/`pass`) get suite-matching uniformly, as the issue asked for consistency's sake. No per-flag duplication.
- Docstring, README (per-flag table + prose), and argparse `help=` text were all updated to say "name or suite" — no stale copy left describing the old name-only behavior.
- New test group `(ax)` is well targeted: it reconstructs the exact scenario from #601/#644 (suite `SetupActivityPermissionDialogE2ETest`, fixed name `setupFlow_grantsMediaPermissionViaSystemDialog_andAdvances`), confirms the suite-shaped query now surfaces the PASS marker, confirms name-matching still works (additive, not replaced), confirms a pattern in neither field still suppresses, and confirms `--include-fail` gets the same suite fallback.
- Cloned the branch and ran `python3 scripts/test_ci_monitor.py` locally: 315 passed, 0 failed.
- Scope is tight: only the three files the PR claims to touch, no drive-by changes elsewhere.

No correctness issues found. LGTM.

PR #686, opened 2026-07-15
Bump ruff-pre-commit to v0.15.21 and reformat tree

aunger:
Fixes #673.

Bumps `astral-sh/ruff-pre-commit` in `.pre-commit-config.yaml` from `v0.8.4` (pinned SHA `f0b5944bef86f50d875305821a0ab0d8c601e465`) to `v0.15.21` (pinned SHA `01a675ea018f2fb714478a5ffb83fcea8374bb06`), the current tag as of this PR. This was confirmed clean during review of PR #666: `ruff format` under the new version produces only mechanical line-wrap changes, and `ruff check` reports no lint failures.

Re-confirmed at implementation time: `ruff format` across the whole tree reformats 4 files (not the 3 originally observed during PR #666's review; the tree has changed since then). All changes are mechanical line-wrap adjustments with no behavior change:

- `scripts/ci_monitor/ci_monitor.py`
- `scripts/file_test_failure_issues.py`
- `scripts/test_ci_monitor.py`
- `scripts/test_label_by_title.py`

`ruff check .` passes with no findings.

## Coordination with #675

Of the three scripts #675 names as pre-existing format drift, this PR's tree-wide format already cleans two of them: `scripts/file_test_failure_issues.py` and `scripts/test_ci_monitor.py`. The third, `scripts/test_enforce_mutually_exclusive_labels.py`, is left unchanged by `ruff format` under `v0.15.21`, meaning it is already compliant at the new version. #675 can therefore reduce to its Markdown-only part (the `agents/pr_participation.md` fenced-code-block language tag).

## Test plan

- [x] `ruff format .` and `ruff check .` run locally under `ruff 0.15.21` (matching the new pin); only mechanical changes, no lint failures.
- [x] `python3 scripts/test_ci_monitor.py`, `scripts/test_label_by_title.py`, and `scripts/test_file_test_failure_issues.py` all pass locally.

Issue #673, opened 2026-07-12
Bump ruff version


aunger:
During review of PR #666, the Author confirmed that bumping `ruff-pre-commit` from `v0.8.4` to `v0.15.21` (current at the time) is a clean change: only 3 files reformatted, all mechanical (line-wrap changes), no lint failures.

PR #666 itself only converted the existing `v0.8.4` tag pin to the equivalent commit SHA without bumping the version, to keep the PR scoped to format conversion rather than version bumps. The Reviewer asked for this bump to be tracked as a follow-up so the "confirmed clean, free win" observation isn't lost.

Source: https://github.com/aunger/gallery-button-for-pixel-camera/pull/666#issuecomment-4951728832 (and https://github.com/aunger/gallery-button-for-pixel-camera/pull/666#issuecomment-4951710988)

This is a follow-on item and does not block merging PR #666.

---

## Sequencing (added by the #676 plan)

This is Step 1 of #676; do it first, before #667 removes the `pre-commit` framework.

- Target while `pre-commit` still exists: bump the `astral-sh/ruff-pre-commit` `rev:` in `.pre-commit-config.yaml` from the `v0.8.4` SHA to the `v0.15.21` SHA, keeping the `# vX.Y.Z` trailing-comment convention already used in that file (per the repo's SHA-pinning rule in `agents/code_edit.md`).
- Run `ruff format` across the whole tree in the same PR so the reformat lands once at the final version rather than incrementally.
- Coordinate with #675 (Step 2), which also runs `ruff format` and therefore follows this issue.
  If this PR's tree-wide format already cleans the three scripts #675 names, say so in the PR, so #675 can reduce to its Markdown-only part.
- After #667 lands, the pinned ruff version lives in `scripts/requirements-lint.txt` instead; doing this bump first lets #667 pin `v0.15.21` directly, with no intermediate old-version pin.

Re-confirm the "3 files, mechanical only" observation still holds at implementation time before committing.

Issue #681, opened 2026-07-14
Switch to mdformat from markdownlint-cli2


aunger:
Switch from `markdownlint-cli2` to [mdformat](https://github.com/executablebooks/mdformat) for Markdown formatting.

- Proof of concept: #680
- Do not attempt to use the semantic break plugin.

This is Step 3 of the #676 plan.
It is done while `pre-commit` still exists (as the #680 proof of concept did), so #667 later migrates a settled local hook rather than inventing one.

## What to do

- Add `mdformat` as a `repo: local`, `language: system` hook in `.pre-commit-config.yaml`, replacing the `markdownlint-cli2` hook and mirroring the existing `ktlint` local hook (`files: '\.md$'`, `entry: mdformat`, plus the args below).
- Install `mdformat` and its plugins from PyPI at session start (STEP 3 of `.claude/hooks/session-start.sh`), pinned.
- Plugins: `mdformat-tables` (GFM pipe tables, needed by the many doc tables) and `mdformat-frontmatter` (preserve YAML frontmatter, for example in `.claude/agents/deep-planner.md`).
  Do not install `mdformat-sentencebreak`.
- Hook args: `--wrap keep --number`.
  `--wrap keep` preserves this repo's intentional one-sentence-per-line prose (see `.claude/rules/prose-style.md`) instead of rewrapping it.
- Run the hook across the whole Markdown tree and commit the resulting reformat.
- Remove `.markdownlint.yaml`; nothing reads it after this switch.

## Why not the semantic break plugin

`mdformat-sentencebreak` was in the #680 proof of concept but is withdrawn (owner instruction on this issue).
The #680 evaluation found it corrupts inline code spans that contain a period (splitting `` `pr_creation.md` `` mid-token), breaks at commas rather than only at sentence ends, is unmaintained (a single 2022 release), and pins the toolchain to `mdformat<0.8.0`.
Dropping it removes all four problems and lifts the version ceiling, so `mdformat` can be pinned to a current release rather than `0.7.22`.

## Version pins (decision for the Author)

Without `mdformat-sentencebreak` the `<0.8.0` ceiling is gone.
Pin `mdformat`, `mdformat-tables`, and `mdformat-frontmatter` to specific current versions and record them in the session-start install (and, after #667, in `scripts/requirements-lint.txt`).
Re-run across the tree and commit; confirm a second pass is a no-op (idempotent).

## Enforcement gaps to note (not blockers)

[...]

PR #687, opened 2026-07-15
Tag Review pattern code fence as `text` in pr_participation.md

### LINKED ISSUE #675: Fix pre-existing format drift in three test scripts
Also do one markdown file that is missing the code block language notation.

---

While testing PR #666, the Author found that `scripts/file_test_failure_issues.py`, `scripts/test_ci_monitor.py`, and `scripts/test_enforce_mutually_exclusive_labels.py` don't actually match the formatting the pinned `ruff-format` (v0.8.4) produces. This was confirmed to predate PR #666 (also present on `main`), and is unrelated to dependency pinning, so it was left untouched there.

Source: https://github.com/aunger/gallery-button-for-pixel-camera/pull/666#issuecomment-4951710988

This is a follow-on item and does not block merging PR #666.

---

## Sequencing and specifics (added by the #676 plan)

This is Step 2 of #676; do it after #673 (the ruff bump) and before #681 (the mdformat switch).

The Markdown file is `agents/pr_participation.md`.
The fenced block under the "Review pattern:" heading (around line 26) opens with a bare ``` and no language.
Tag it `text` (it is pseudo-code, not a real language).
Do this before #681, while `markdownlint-cli2` is still active, so its MD040 rule validates the fix; mdformat does not enforce MD040, so after #681 this would not be caught.

The ruff drift: run `ruff format` on the three named scripts under the ruff version from #673 (`v0.15.21`), not `v0.8.4`, so the tree ends up clean at the final version.
If #673's tree-wide format already reformatted these three files, this part is a no-op; confirm and note it in the PR (the issue may then reduce to the Markdown fix).

aunger:
Fixes #675.

## Changes

- `agents/pr_participation.md`: the fenced block under the "Review pattern:" heading opened with a bare triple-backtick and no language, which fails markdownlint's MD040 rule. Tagged it `text` since it is pseudo-code, not a real language. Verified with `markdownlint-cli2 v0.14.0` (the version currently pinned in `.pre-commit-config.yaml`) that this file now lints clean.

## The ruff-format part of #675 is a no-op

#675 also asked to run `ruff format` on `scripts/file_test_failure_issues.py`, `scripts/test_ci_monitor.py`, and `scripts/test_enforce_mutually_exclusive_labels.py` under ruff v0.15.21 (the version from #673), since they didn't match the older pinned `ruff-format` (v0.8.4) that was in place when the drift was first noticed.

By the time this PR was started, #673 had already merged as PR #686 ("Bump ruff-pre-commit to v0.15.21 and reformat tree"), which reformatted the whole tree, including these three files, under ruff-format v0.15.21. Running `ruff format --diff` against the pinned v0.15.21 on these three scripts now reports "3 files already formatted" with no changes, and `ruff format --diff .` / `ruff check .` report the whole tree clean. So no ruff-format changes were needed here; this PR reduces to the Markdown fix, as anticipated by the issue.

## Test plan

- [x] `python3 -m pytest scripts/test_ci_monitor.py scripts/test_enforce_mutually_exclusive_labels.py scripts/test_file_test_failure_issues.py` (154 passed)
- [x] `ruff format --diff .` and `ruff check .` (both clean, ruff 0.15.21)
- [x] `npx markdownlint-cli2@0.14.0 agents/pr_participation.md` (0 errors)

### DIFF
--- a/agents/pr_participation.md
+++ b/agents/pr_participation.md
@@ -23,7 +23,7 @@
 
   (This is the PR-path mechanism. If the Author opened no PR, see "Reviewing an Author who declined to open a PR" below.)
 
-  ```
+  ```text
   (ReviewText, IsReviewApproval) := <review the diff; form your verdict and review text>
   post review: mcp__github__pull_request_review_write(ReviewText, IsReviewApproval)
   ```


aunger (commented):
🤖 Reviewer

Verified everything in this PR independently:

- The diff is exactly the one-line fix the issue's sequencing plan specified: the bare fence under "Review pattern:" (line 26) is now tagged `` ```text ``. Nothing else changed (1 file, +1/-1).
- `npx markdownlint-cli2@0.14.0 agents/pr_participation.md` passes with 0 errors, confirmed locally.
- Confirmed #681 (the mdformat switch) has not merged into main yet (`430fe39` is not an ancestor of `origin/main`), so doing this fix now, while markdownlint-cli2's MD040 rule is still active, is correctly sequenced per the issue.
- Re the ruff-format no-op claim: confirmed `ruff format --diff` (pinned v0.15.21) reports the three named scripts, and the whole tree, already formatted, and `ruff check .` is clean. Also confirmed PR #686 (the v0.15.21 bump and tree reformat) is an ancestor of this branch, which explains why the drift is already gone. The PR's reasoning here is accurate, not just asserted.
- Ran the named test suite locally: 154 passed, matching the PR's test plan.

Nothing to add. Small, well-scoped, and the claims in the description are verifiably true rather than just plausible-sounding. LGTM.

PR #702, opened 2026-07-16
Always dispatch Programmers in worktrees, not only when parallel

### LINKED ISSUE #701: Orchestration: always use worktrees
Currently, worktrees are advised for parallel orchestrations but parallel conditions are often unnoticed.

Just always use worktrees so agents are less likely to misunderstand or forget.

aunger:
Fixes #701

## Problem

Worktree isolation was advised only for parallel orchestrations (`dev_orchestration.md` Delegation rules tied "their own branch and worktree" to the parallel-dispatch case).
But parallel conditions are often unnoticed, so Orchestrators forgot to isolate a sub-agent's checkout when they did not recognize the dispatch as parallel.
The issue is worded generically about agents, and the risk is not limited to Programmers: the Verification Agent (`pr_verify.md`) also performs repository-changing git operations (creates a test branch and pushes it) that can collide with in-flight work on a shared checkout.

## Change

Make worktree isolation unconditional for every sub-agent the Orchestrator dispatches:

- `agents/dev_orchestration.md`
  - Delegation rules: replaced the parallel-only worktree advice with an **"Always use worktrees"** rule that dispatches **every sub-agent** in its own git worktree (Agent tool `isolation: "worktree"`), naming the Programmer, Reviewer, Verification Planner, and Verification Agent explicitly and identifying which of them mutate a checkout (Programmer commits, Verification Agent test-branch push). Rationale: unconditional application is safer than detecting the parallel case, which is easy to miss. The parallel-dispatch bullet now covers only the per-issue branch requirement.
  - Assigning a Programmer: added a step to always dispatch the Programmer in its own worktree, pointing to the Delegation rules statement.
- `agents/inaugurate.md`
  - Dispatching each Programmer: added that each Programmer is dispatched in its own worktree per the "Always use worktrees" rule, applying to every dispatch rather than only parallel ones.
- `agents/pr_verify.md`
  - Boundaries: added that the Verification Agent runs in its own worktree, so its enumerated repository-changing actions never disturb a shared checkout or another agent's in-flight work, with a cross-reference to the rule.

This is a documentation change to the `agents/` process instructions; there is no runtime behavior to add automated tests for. `scripts/lint.sh` passes on all three edited files.

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -191,11 +191,10 @@ Routing on the Verification Agent's signal:
 ## Assigning a Programmer
 
 - Create a sub-agent at the Author model (see Model selection above)
-- Always dispatch the Programmer in its own git worktree (Agent tool `isolation: "worktree"`), whether or not this dispatch is parallel.
-  See "Always use worktrees" under Delegation rules below.
 - Create a dedicated per-issue branch for the Programmer to use.
   Branch names should follow the pattern `fix/issue-N-short-description` for bug fixes or `feature/issue-N-short-description` for new features.
   Never direct two Programmers for unrelated issues to the same branch.
+- Always dispatch the Programmer in its own git worktree (Agent tool `isolation: "worktree"`), whether or not this dispatch is parallel.
 - Dispatch using the dispatch template above, with the Programmer role assignment statement and the literal issue number and branch name tokens.
 
 When the Author returns, apply this transition.
@@ -370,10 +369,8 @@ Orchestrator-specific notes:
 
 ## Delegation rules
 
-- **Always use worktrees.** Dispatch every sub-agent in its own git worktree (Agent tool `isolation: "worktree"`), not only Programmers and not only when dispatching in parallel.
-  This covers the Programmer, Reviewer, Verification Planner, and Verification Agent alike.
-  Several of them mutate a checkout (the Programmer's commits, the Verification Agent's test-branch creation and push), and parallel conditions often go unnoticed, so applying worktrees unconditionally is safer than deciding case by case and keeps each task's checkout isolated.
 - If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch.
+- **Always use worktrees.** Dispatch every sub-agent in its own git worktree (Agent tool `isolation: "worktree"`), not only Programmers and not only when dispatching in parallel.
 - **One branch per ticket.** Each issue gets its own dedicated branch.
 - **Separate subagents per ticket.** Each issue or PR gets its own independent Author and Reviewer agents.
 - **Report subagent timing.** Use the Bash tool to run `date -u` immediately before dispatching each subagent, and again immediately after it returns. Report both times to the user.
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -370,8 +370,9 @@ Orchestrator-specific notes:
 
 ## Delegation rules
 
-- **Always use worktrees.** Dispatch every Programmer in its own git worktree (Agent tool `isolation: "worktree"`), not only when dispatching in parallel.
-  Parallel conditions often go unnoticed, so applying worktrees unconditionally is safer than deciding case by case, and it keeps each task's checkout isolated.
+- **Always use worktrees.** Dispatch every sub-agent in its own git worktree (Agent tool `isolation: "worktree"`), not only Programmers and not only when dispatching in parallel.
+  This covers the Programmer, Reviewer, Verification Planner, and Verification Agent alike.
+  Several of them mutate a checkout (the Programmer's commits, the Verification Agent's test-branch creation and push), and parallel conditions often go unnoticed, so applying worktrees unconditionally is safer than deciding case by case and keeps each task's checkout isolated.
 - If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch.
 - **One branch per ticket.** Each issue gets its own dedicated branch.
 - **Separate subagents per ticket.** Each issue or PR gets its own independent Author and Reviewer agents.
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -191,6 +191,8 @@ Routing on the Verification Agent's signal:
 ## Assigning a Programmer
 
 - Create a sub-agent at the Author model (see Model selection above)
+- Always dispatch the Programmer in its own git worktree (Agent tool `isolation: "worktree"`), whether or not this dispatch is parallel.
+  See "Always use worktrees" under Delegation rules below.
 - Create a dedicated per-issue branch for the Programmer to use.
   Branch names should follow the pattern `fix/issue-N-short-description` for bug fixes or `feature/issue-N-short-description` for new features.
   Never direct two Programmers for unrelated issues to the same branch.
@@ -368,7 +370,9 @@ Orchestrator-specific notes:
 
 ## Delegation rules
 
-- If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch and worktree.
+- **Always use worktrees.** Dispatch every Programmer in its own git worktree (Agent tool `isolation: "worktree"`), not only when dispatching in parallel.
+  Parallel conditions often go unnoticed, so applying worktrees unconditionally is safer than deciding case by case, and it keeps each task's checkout isolated.
+- If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch.
 - **One branch per ticket.** Each issue gets its own dedicated branch.
 - **Separate subagents per ticket.** Each issue or PR gets its own independent Author and Reviewer agents.
 - **Report subagent timing.** Use the Bash tool to run `date -u` immediately before dispatching each subagent, and again immediately after it returns. Report both times to the user.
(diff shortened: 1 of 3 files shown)


aunger (commented):
🤖 Reviewer

The core change is sound: it makes worktree isolation for Programmer dispatch unconditional instead of tied to the parallel case, which directly addresses the "parallel conditions are often unnoticed" problem in #701. The wording is clear, the cross-reference between "Assigning a Programmer" and the "Always use worktrees" bullet in Delegation rules is consistent, and `scripts/lint.sh` does pass clean on both files (verified locally in a worktree checkout of this branch — no diff after running it).

One gap I think needs addressing before merge:

**Scope narrower than the issue.** #701 says "Just always use worktrees so agents are less likely to misunderstand or forget," worded generically about agents, not "Programmers" specifically. This PR only makes worktree isolation unconditional for Programmer dispatch. But `pr_verify.md` shows the Verification Agent also performs real git operations on the shared checkout outside a worktree: it fetches the feature branch, creates a test branch (`git checkout -b test/verify-pr-{N}-{short-description} origin/{feature-branch}`), and pushes it. That's exactly the kind of operation that can collide with other in-flight work on the same checkout if it isn't isolated — the same risk class #701 is about, just unnoticed here the same way parallel dispatch was unnoticed before this PR. Neither `dev_orchestration.md`'s "Dispatch a Verification Agent" step nor `pr_verify.md` mentions worktree isolation at all.

I'm not asking you to design the fix, but the PR should either extend "Always use worktrees" to cover Verification Agent dispatch too, or explain in the PR description why Verification Agent's git operations are safe to leave unisolated (e.g., if there's a reason its test-branch flow can never race with a Programmer's checkout). Right now it reads like an oversight rather than a deliberate scope decision, given the issue text is unscoped and doesn't say "Programmer" specifically.

Minor nit: the Delegation rules bullet is titled "**Always use worktrees**" as a general-sounding rule name, then immediately scoped down to "every Programmer" in the body. Given the point of #701 is to stop rules from being silently under-applied, consider naming it something like "Always use worktrees for Programmers" so the header itself doesn't overstate the rule's actual reach.

aunger:
🤖 Author

Agreed on the blocking point, and I've broadened the fix rather than justifying the narrower scope. You're right that #701 is worded generically ("agents... less likely to misunderstand or forget"), and that `pr_verify.md` has the Verification Agent perform real repository-changing git operations (`git checkout -b test/verify-pr-...`, then push) that can race with in-flight work on a shared checkout, the same unnoticed-risk class the issue is about.

Changes in `5b63f21`:

- **`dev_orchestration.md` Delegation rules**: generalized "Always use worktrees" from "every Programmer" to "every sub-agent," naming the Programmer, Reviewer, Verification Planner, and Verification Agent explicitly, and calling out which of them mutate a checkout (Programmer commits, Verification Agent test-branch push).
- **`pr_verify.md` Boundaries**: added that the Verification Agent runs in its own worktree, so its enumerated repo-changing actions never disturb a shared checkout, with a cross-reference to the rule. This closes the "neither doc mentions worktree isolation" gap you flagged.

On the nit (header "Always use worktrees" overstating): I resolved it in the opposite direction from your suggestion, by widening the body to match the header instead of narrowing the header to "for Programmers." Now that the rule genuinely applies to every dispatched sub-agent, the general-sounding header is accurate, which also better matches the issue's generic intent. If you'd still prefer a scoped header, say so and I'll reconsider, but I think the broad header is now correct rather than overstated.

`scripts/lint.sh` passes clean on all three edited files.

aunger (commented):
🤖 Reviewer

Commit `5b63f21` resolves the scope gap I raised: the Delegation rules bullet now names Programmer, Reviewer, Verification Planner, and Verification Agent explicitly, correctly identifies which of them mutate a checkout (Programmer commits, Verification Agent test-branch push), and `pr_verify.md`'s new Boundaries note closes the "neither doc mentions worktree isolation" gap. On the header nit, widening the body to match the general header instead of narrowing the header is a reasonable call — the rule genuinely applies broadly now, so I have no objection. I re-verified `scripts/lint.sh` passes clean on all three edited files (`dev_orchestration.md`, `inaugurate.md`, `pr_verify.md`) in a fresh worktree checkout of `5b63f21`.

One loose end before merge: the PR description itself is stale. It still says the change is "unconditional for every **Programmer** dispatch," lists only `dev_orchestration.md` and `inaugurate.md` as edited files, and doesn't mention `pr_verify.md`. The actual shipped change is broader (every sub-agent) and touches three files. The full explanation lives in your issue-comment reply, but the top-level PR description is what persists in the merged history, so please update it to match what's actually in the diff now.

Modulo that description update, this is good to merge.

aunger:
🤖 Author

Done. Updated the top-level PR description to match the shipped diff: it now states the change is unconditional for **every sub-agent** (not just Programmers), explains the Verification Agent motivation, and lists all three edited files (`dev_orchestration.md`, `inaugurate.md`, `pr_verify.md`). No code changes in this round, so `5b63f21` remains the tip.

aunger (commented):
🤖 Reviewer

The PR description now matches the shipped diff: it states the broadened "every sub-agent" scope, explains the Verification Agent motivation, and lists all three edited files. That closes the last item. No code changes needed this round, and `5b63f21` was already lint-clean and technically sound from my prior pass.

LGTM.

agents/dev_orchestration.md
```diff
@@ -368,7 +370,10 @@ Orchestrator-specific notes:
 
 ## Delegation rules
 
-- If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch and worktree.
+- **Always use worktrees.** Dispatch every sub-agent in its own git worktree (Agent tool `isolation: "worktree"`), not only Programmers and not only when dispatching in parallel.
+  This covers the Programmer, Reviewer, Verification Planner, and Verification Agent alike.
+  Several of them mutate a checkout (the Programmer's commits, the Verification Agent's test-branch creation and push), and parallel conditions often go unnoticed, so applying worktrees unconditionally is safer than deciding case by case and keeps each task's checkout isolated.
+- If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch.
```
aunger, line 372: Move this line to the top of the list

agents/dev_orchestration.md
```diff
@@ -191,6 +191,8 @@ Routing on the Verification Agent's signal:
 ## Assigning a Programmer
 
 - Create a sub-agent at the Author model (see Model selection above)
+- Always dispatch the Programmer in its own git worktree (Agent tool `isolation: "worktree"`), whether or not this dispatch is parallel.
+  See "Always use worktrees" under Delegation rules below.
```
aunger: Remove this line
```suggestion
```

agents/dev_orchestration.md
```diff
@@ -191,6 +191,8 @@ Routing on the Verification Agent's signal:
 ## Assigning a Programmer
 
 - Create a sub-agent at the Author model (see Model selection above)
+- Always dispatch the Programmer in its own git worktree (Agent tool `isolation: "worktree"`), whether or not this dispatch is parallel.
```
aunger: Move this down in the list, just below the branch bullet point

agents/dev_orchestration.md
```diff
@@ -368,7 +370,10 @@ Orchestrator-specific notes:
 
 ## Delegation rules
 
-- If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch and worktree.
+- **Always use worktrees.** Dispatch every sub-agent in its own git worktree (Agent tool `isolation: "worktree"`), not only Programmers and not only when dispatching in parallel.
+  This covers the Programmer, Reviewer, Verification Planner, and Verification Agent alike.
+  Several of them mutate a checkout (the Programmer's commits, the Verification Agent's test-branch creation and push), and parallel conditions often go unnoticed, so applying worktrees unconditionally is safer than deciding case by case and keeps each task's checkout isolated.
```
aunger: Remove
```suggestion
```

agents/dev_orchestration.md
```diff
@@ -368,7 +370,10 @@ Orchestrator-specific notes:
 
 ## Delegation rules
 
-- If requested by the user, **dispatch in parallel** for independent issues. Parallel issues must each have their own branch and worktree.
+- **Always use worktrees.** Dispatch every sub-agent in its own git worktree (Agent tool `isolation: "worktree"`), not only Programmers and not only when dispatching in parallel.
+  This covers the Programmer, Reviewer, Verification Planner, and Verification Agent alike.
```
aunger: Remove
```suggestion
```

agents/dev_orchestration.md
aunger: 🤖 Author

Done in `a38a38b`: moved the parallel-dispatch bullet to the top of the Delegation rules list.

agents/dev_orchestration.md

Done in `a38a38b`: removed the cross-reference sub-line.

agents/dev_orchestration.md

Done in `a38a38b`: moved the worktree bullet down to just below the branch bullet.

agents/dev_orchestration.md

Done in `a38a38b`: removed this rationale line, condensing the bullet to a single line.

agents/dev_orchestration.md

Done in `a38a38b`: removed this line. The single-line "every sub-agent" rule still preserves the broadened scope, and `pr_verify.md`'s Boundaries note retains the Verification Agent specifics.

aunger (commented):
🤖 Reviewer

Commit `a38a38b` only reorders/condenses wording per the inline suggestions (parallel-dispatch bullet back to the top, worktree bullet moved and shortened, cross-reference sub-line dropped). The substantive scope — every sub-agent, not just Programmers — is still intact in the condensed single-line rule, and `pr_verify.md`'s Boundaries note is untouched. Re-verified `scripts/lint.sh` passes clean on all three files at this commit.

Still LGTM.

PR #703, opened 2026-07-16
Auto-remove label-gate blocks instead of stopping (resolve Step 8a contradiction)

### LINKED ISSUE #671: Remove dev_orchestration.md:255 ("no auto-remove blocking label") — contradicts Step 8a's own permitted-actions note
## What line 255 says today

```
// The Orchestrator does not auto-remove the blocking label (issue #516, Step 8a): it is a user-controlled process label needing human judgment.
```

This forces the Orchestrator to stop and ask a human before clearing a label-gate-only CI block (e.g. its own `orchestrating` label tripping the "No blocking labels" check), rather than resolving it autonomously.

## Where this came from

Traced via `git blame` + `git log` back through PR #518 ("Add per-check status summary and attributed terminal to CI Monitor", implementing the plan in issue #516):

1. Issue #516's plan, Step 8a, proposed the Orchestrator resolve a label-gate-only block itself:
   > if the attributed terminal names only a label-gate check (`[label gate]`), the Orchestrator removes the offending process label rather than routing `newAuthor`, then re-launches the Monitor. **Keep this within the Orchestrator's permitted actions (it already manages labels per "Starting to orchestrate a PR"). Reference the existing label-transition tables for the mechanics.**

   That last sentence explicitly anticipates the objection that this might be a new, out-of-scope capability, and pre-empts it: label management is *already* a normal, permitted Orchestrator action — `dev_orchestration.md`'s own "Starting to orchestrate a PR" section has the Orchestrator add `orchestrating`, and "Concluding PR orchestration" has it remove `orchestrating`. Step 8a's proposal was to reuse that exact, already-sanctioned mechanism, not invent a new one.

2. The initial implementation (commit `7857846`) landed the label-gate branch but dropped Step 8a's auto-removal behavior in favor of stop-and-inform, without explaining why.

3. Self-review on PR #518 flagged this as a "meaningful behavioral difference" and asked the author to confirm it was intentional and document the reasoning (review comment [#pullrequestreview-4579239915](https://github.com/aunger/gallery-button-for-pixel-camera/pull/518#pullrequestreview-4579239915)).

4. The author's reply ([#issuecomment-4809437070](https://github.com/aunger/gallery-button-for-pixel-camera/pull/518#issuecomment-4809437070)) gave two justifications:
   - The label is "a user-controlled process label that the Orchestrator does not own; its removal requires human judgment about whether the process condition is actually satisfied" (this became line 255's wording, via commit `5320427`, then two later prose-trimming passes: `0cfb1e82` and `350f39b6`).
   - "The Orchestrator's permitted actions table ('May not: Edit or write files') also makes autonomous label removal technically out of scope for the Orchestrator as currently defined--removing a label via MCP would be an unlisted 'May' action."

## The contradiction

That second justification is the one Step 8a's own text already addressed and rejected as a concern: the plan explicitly says label removal is "within the Orchestrator's permitted actions" because "it already manages labels per 'Starting to orchestrate a PR.'" The author's response doesn't engage with that sentence at all, and asserts the opposite (label removal is "technically out of scope," "an unlisted May action") without acknowledging the plan had already called this out by name.

Nothing in `dev_orchestration.md`'s "May"/"May not" lists actually distinguishes label-writes from the label-writes the Orchestrator already performs constantly (adding/removing `orchestrating`, `verified`, `changes requested`, `changes done`, `verification needed` per the various transition tables throughout the same document). The "May not: Edit or write files" line is about source files, not GitHub labels — labels are managed through a separate mechanism (the MCP issue-write tool) that the document's own transition tables already direct the Orchestrator to use unconditionally, with no "ask a human first" caveat anywhere else.

The "user-controlled process label" framing is also questionable for this specific label: `orchestrating` is not a label a human applies to signal a hold requiring judgment (like some hypothetical `needs-legal-review` label would be) — it's the Orchestrator's *own* bookkeeping label, added and removed by the Orchestrator itself as a matter of routine per "Starting to orchestrate a PR" and "Concluding PR orchestration." Treating its removal as requiring special human sign-off, while every other label transition in the document does not, is an inconsistency the review discussion doesn't seem to have caught.

## Ask

Remove line 255 (and reconsider whether the label-gate branch it justifies should route back to Step 8a's original design: the Orchestrator removes the label-gate-only blocker itself and re-launches the Monitor, using the same label-transition mechanism it already uses everywhere else in the document).

This surfaced today because an Orchestrator run had to stop mid-cycle on four straightforward, low-risk PRs and ask a human to authorize removing its own `orchestrating` label before each one could merge, purely because of this line--exactly the friction Step 8a's plan tried to avoid.

aunger:
Fixes #671.

## What changes

`dev_orchestration.md:255` asserted that the Orchestrator does not auto-remove the blocking label because it is "a user-controlled process label needing human judgment." That directly contradicted Step 8a's own text (issue #516), which explicitly called out that label management is already a normal, permitted Orchestrator action (it already adds/removes `orchestrating` per "Starting to orchestrate a PR" and "Concluding PR orchestration"), so reusing that mechanism for the label-gate case was not a new capability.

This PR removes the contradictory comment and implements Step 8a's original design:

- On a label-gate-only `Blocked` terminal (the `by: ...` suffix names only label-gate checks, e.g. the Orchestrator's own `orchestrating` label tripping the "No blocking labels" check), the Orchestrator now removes the blocking label itself and re-launches the Monitor, instead of stopping to ask a human.
- This is implemented as a new `labelGateBlock:` routing target that removes the label via the existing label-transition mechanism, informs the user, and resumes the Monitor loop from the fresh invocation.
- Since the label-gate branch no longer stops for a human, `surfaceBeforeMergingRequirements` no longer needs its "Clear path or label-gate path" special-casing (the label-gate path now only ever reaches that step via the ordinary Clear path, after the Monitor confirms CI is clear).
- Added an explicit "May" entry ("Add or remove GitHub labels per the transition tables in this document") to the Orchestrator's permitted-actions list, resolving the ambiguity the contradiction traded on (the prior text argued label removal was "technically out of scope" as an "unlisted May action," even though the document's transition tables direct exactly this action elsewhere, unconditionally).

## Test plan

This is process documentation for the Orchestrator role; there is no automated test that exercises the routing logic itself.

- [ ] Manual: next time an Orchestrator run hits a label-gate-only CI block, confirm it removes the label and re-launches the Monitor rather than stopping to ask the user.

### DIFF
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -279,19 +279,36 @@ labelGateBlock:
   // normal, permitted Orchestrator action--this branch reuses the same label-transition
   // mechanism used throughout this document (e.g. "Starting to orchestrate a PR" and
   // "Concluding PR orchestration"), rather than inventing a new one.
-  // At this point in the routing (the Reviewer already gave LGTM, so `changes requested` and
-  // `changes done` are not applied, and `verification needed` has not yet been added), the
-  // only blocking label still applied to the PR is normally `orchestrating`, the
-  // Orchestrator's own bookkeeping label. Remove whichever blocking label is actually applied.
-  Apply this transition to the PR:
+  //
+  // The Monitor's attributed terminal names the failing check ("No blocking labels"), not the
+  // specific label causing it (the check-runs API does not expose that). The routing above
+  // guarantees which blocking label this branch can find applied: `changes requested` and
+  // `changes done` were already cleared when the Reviewer returned (see "CI checking after a
+  // Reviewer exits"), and `verification needed` is not applied yet--that only happens in
+  // `surfaceBeforeMergingRequirements`, downstream of this branch. So `orchestrating` is the
+  // only blocking label this branch expects to find.
+  // `verification needed` is deliberately never auto-removed by this branch, even if some
+  // future routing change let it coexist here: unlike `orchestrating`, it represents real
+  // outstanding process state (before-merging items pending the Verification Agent), and
+  // clearing it here would defeat that gate instead of clearing stale bookkeeping.
+  //
+  // This branch removes `orchestrating` well before "Concluding PR orchestration" normally
+  // would. That is fine: nothing in this document treats `orchestrating` as a concurrency
+  // guard or precondition, only as a human-visible "automation is mid-cycle" marker and as
+  // the mechanism gating this check. The Orchestrator's own session state, not the label,
+  // is what tracks progress through the rest of the cycle, and the later unconditional
+  // removal at "Concluding PR orchestration" is a no-op once this branch has already
+  // removed it.
+  If `orchestrating` is the only blocking label currently applied to the PR (the expected case): apply this transition to the PR:
 
-  | Remove label |
-  |---|
-  | `orchestrating` |
+    | Remove label    |
+    | --------------- |
+    | `orchestrating` |
 
-  Inform the user that a process-label gate blocked the merge (code is review-approved) and that the Orchestrator removed the blocking label automatically.
-  Re-launch the Monitor tool call (same command as the original, fresh invocation).
-  Resume the routing above from "Act only on the terminal lines..." with the fresh invocation.
+    Inform the user that a process-label gate blocked the merge (code is review-approved) and that the Orchestrator removed the blocking label automatically.
+    Re-launch the Monitor tool call (same command as the original, fresh invocation).
+    Resume the routing above from "Act only on the terminal lines..." with the fresh invocation.
+  Otherwise (`verification needed`, `changes requested`, or `changes done` is applied instead of, or alongside, `orchestrating`): this is an unexpected state the routing above should not produce. Do not remove any label; escalate to the user; stop.
 
 surfaceBeforeMergingRequirements:
   // Surfaces outstanding before-merging requirements (unautomated verification steps,
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -54,7 +54,7 @@ The Orchestrator is not a Reviewer or a Programmer.
 **May:**
 
 - Create local Git branches to keep tasks separate
-- Add or remove GitHub labels per the transition tables in this document (already exercised throughout, e.g., "Starting to orchestrate a PR" and "Concluding PR orchestration")
+- Add or remove GitHub labels per the transition tables in this document
 - Read project instructions (AGENTS.md and the files it references)
 - Dispatch and communicate with subagents
   - Replace subagents, reluctantly and when necessary, to complete a workflow
@@ -272,33 +272,12 @@ PR routing (Monitor loop):
       goto surfaceBeforeMergingRequirements (entered on the Clear path)
 
 labelGateBlock:
-  // Step 8a's original design (issue #516): a label-gate-only Blocked terminal names a
-  // process label the Orchestrator itself controls (see the blocking-label set in
-  // `.github/workflows/block-merge-on-blocking-labels.yml`), not a code or test failure, so
-  // routing a new Author round would accomplish nothing. Label management is already a
-  // normal, permitted Orchestrator action--this branch reuses the same label-transition
-  // mechanism used throughout this document (e.g. "Starting to orchestrate a PR" and
-  // "Concluding PR orchestration"), rather than inventing a new one.
-  //
-  // The Monitor's attributed terminal names the failing check ("No blocking labels"), not the
-  // specific label causing it (the check-runs API does not expose that). The routing above
-  // guarantees which blocking label this branch can find applied: `changes requested` and
-  // `changes done` were already cleared when the Reviewer returned (see "CI checking after a
-  // Reviewer exits"), and `verification needed` is not applied yet--that only happens in
-  // `surfaceBeforeMergingRequirements`, downstream of this branch. So `orchestrating` is the
-  // only blocking label this branch expects to find.
-  // `verification needed` is deliberately never auto-removed by this branch, even if some
-  // future routing change let it coexist here: unlike `orchestrating`, it represents real
-  // outstanding process state (before-merging items pending the Verification Agent), and
-  // clearing it here would defeat that gate instead of clearing stale bookkeeping.
-  //
-  // This branch removes `orchestrating` well before "Concluding PR orchestration" normally
-  // would. That is fine: nothing in this document treats `orchestrating` as a concurrency
-  // guard or precondition, only as a human-visible "automation is mid-cycle" marker and as
-  // the mechanism gating this check. The Orchestrator's own session state, not the label,
-  // is what tracks progress through the rest of the cycle, and the later unconditional
-  // removal at "Concluding PR orchestration" is a no-op once this branch has already
-  // removed it.
+  // Step 8a's original design (issue #516). `orchestrating` is the only blocking label
+  // expected here: `changes requested`/`changes done` clear when the Reviewer returns, and
+  // `verification needed` isn't applied until `surfaceBeforeMergingRequirements`, downstream
+  // of this branch. Never auto-remove `verification needed`--it is real outstanding process
+  // state, not bookkeeping. Removing `orchestrating` early is fine; nothing in this document
+  // treats it as a concurrency guard.
   If `orchestrating` is the only blocking label currently applied to the PR (the expected case): apply this transition to the PR:
 
     | Remove label    |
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

The core fix (removing the contradictory line 255 comment, adding the explicit "May: add or remove GitHub labels..." entry, and simplifying `surfaceBeforeMergingRequirements`/the Verification Agent routing back to a single unconditional path) is a correct, well-reasoned resolution of issue #671. The new `labelGateBlock` rationale comment correctly grounds itself in Step 8a and the existing label-transition mechanism. Good use of a direct reference to `.github/workflows/block-merge-on-blocking-labels.yml` for the blocking-label set.

However, `labelGateBlock` itself contains a new internal contradiction, the same class of bug this PR exists to remove:

```
// ... The
// only blocking label still applied to the PR is normally `orchestrating`, the
// Orchestrator's own bookkeeping label. Remove whichever blocking label is actually applied.
Apply this transition to the PR:

| Remove label |
|---|
| `orchestrating` |
```

The comment's operative sentence says "Remove whichever blocking label is actually applied," but the table that immediately follows it is not parameterized on anything--it unconditionally removes `orchestrating` and only `orchestrating`. There are four blocking labels per `block-merge-on-blocking-labels.yml` (`verification needed`, `changes requested`, `changes done`, `orchestrating`), and the Monitor's attributed `Blocked by: <name>` form (referenced two lines above in the same branch, and again in the "Orchestrator-specific notes" at the bottom of the file) already names which check/label is actually blocking. If a label-gate terminal is ever reached where the sole remaining blocking label is not `orchestrating` (stale `changes requested`/`changes done`/`verification needed` from an earlier round, a re-entrant label-gate hit after a prior `labelGateBlock` pass, etc.), this rule removes the wrong label, CI stays blocked, the re-launched Monitor reports the same `[label gate]` terminal, and `labelGateBlock` loops forever removing a label that was never the problem. That is exactly the stuck-Orchestrator failure mode issue #671 was filed to eliminate, just relocated one level down.

Given the document's own framing ("RULES for the Orchestrator, not suggestions"), the operative table needs to match the operative sentence: either name the actual blocking label(s) from the Monitor's `Blocked by:` attribution and remove those, or state plainly that this branch only handles the `orchestrating` case and define what happens when a different label is the sole blocker (even a fallback to `newAuthor` or an escalation would be internally consistent; silently no-op'ing while claiming "whichever ... is actually applied" is not).

Separate, smaller observation worth a reply (not necessarily a blocker, but worth addressing or explicitly deciding not to): in the normal Clear path, `orchestrating` stays applied for the entire orchestration lifecycle and is removed only once, at "Concluding PR orchestration," at the very end. `labelGateBlock` now removes it much earlier, while the Orchestrator still has real work left to do (surfacing before-merging requirements, dispatching a Verification Planner/Agent, applying `verified`). For that remaining window, the PR shows no label indicating an Orchestrator is actively working it. Nothing in the doc states `orchestrating` is a concurrency guard, but the symmetric add-at-start/remove-at-end bracket strongly implies that's part of its purpose, and this PR unbrackets it early for label-gate PRs only. Worth a sentence in the rationale comment addressing why that's fine, if it's fine.

Verdict: Changes requested.

aunger:
🤖 Author

Both review points addressed in the latest commit.

## `labelGateBlock`'s operative-sentence/table mismatch

Agreed, this was a real bug and the same class of contradiction the PR is meant to remove. Fixed:

- The branch now reasons explicitly from the routing above about which blocking label it can expect: `changes requested`/`changes done` are already cleared by the time the Reviewer returns (see "CI checking after a Reviewer exits"), and `verification needed` is not applied until `surfaceBeforeMergingRequirements`, downstream of this branch. So `orchestrating` is the only label this branch expects to find, and the table now only fires in that explicit case.
- `verification needed` is now called out by name as never auto-removed here, even hypothetically: it represents real outstanding process state (before-merging items pending the Verification Agent), not bookkeeping, so clearing it would defeat that gate rather than clean up stale state.
- If some other combination is ever found (a routing bug elsewhere), the branch now escalates to the user instead of silently no-op'ing or looping, closing the infinite-loop failure mode you described.

## Removing `orchestrating` ahead of "Concluding PR orchestration"

Added a rationale paragraph: nothing in this document treats `orchestrating` as a concurrency guard or precondition, only as a human-visible "mid-cycle" marker and as the mechanism gating the "No blocking labels" check. The Orchestrator's own session state (not the label) is what tracks progress through the rest of the cycle, so removing it early for label-gate PRs is safe, and the later unconditional removal at "Concluding PR orchestration" becomes a no-op once this branch has already removed it.

aunger (commented):
🤖 Reviewer

Both points addressed satisfactorily.

On the `labelGateBlock` operative-sentence/table mismatch: the new conditional ("If `orchestrating` is the only blocking label currently applied... Otherwise ...escalate to the user; stop") closes the gap. I confirmed against `scripts/ci_monitor/README.md` and `ci_monitor.py` (`blocking_suffix`, `label_gate_check_regex`) that the `Blocked by: <name>` attribution genuinely names the check-run ("No blocking labels"), not the specific label, so there was no attribution string available to parse in the first place; reasoning from the routing's own guaranteed label state (as the new comment does, citing "CI checking after a Reviewer exits" and the deferred timing of `verification needed`) is the correct approach here, and the escalate-on-unexpected-state fallback closes the infinite-loop failure mode rather than leaving it latent.

On the early `orchestrating` removal: the added rationale (nothing in the document treats the label as a concurrency guard, only as a human-visible marker and the check-gating mechanism, with the Orchestrator's own session state tracking cycle progress) is a reasonable, internally consistent position and directly engages the concern. No factual objection to it.

Verdict: LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #703 can be merged):
- [ ] #704

None.

aunger:
### Verification summary

Before-merging items from the [Verification Planner report](https://github.com/aunger/gallery-button-for-pixel-camera/pull/703#issuecomment-4990119422):

- #704 (Manually confirm Orchestrator auto-removes label-gate block on next real run): **NOT AUTOMATABLE**. See explanation on the tracking issue: https://github.com/aunger/gallery-button-for-pixel-camera/issues/704#issuecomment-4990143967. Static review of `agents/dev_orchestration.md` on this branch (`f1fd66b`) confirms the `labelGateBlock` branch is present and internally consistent, but confirming actual Orchestrator behavior requires a real Orchestrator run reaching this state, which cannot be scripted in a verification session. Left open for a human or a future live run.

No item failed. One item remains open as not automatable, so this PR is not yet clear to merge on the automated path.

agents/dev_orchestration.md
```diff
@@ -271,29 +271,65 @@ PR routing (Monitor loop):
       clear the silentVanish re-launch flag
       goto surfaceBeforeMergingRequirements (entered on the Clear path)
 
+labelGateBlock:
+  // Step 8a's original design (issue #516): a label-gate-only Blocked terminal names a
```
aunger: Cool story, bro. Are the next 25 lines useful, though?

agents/dev_orchestration.md
```diff
@@ -54,6 +54,7 @@ The Orchestrator is not a Reviewer or a Programmer.
 **May:**
 
 - Create local Git branches to keep tasks separate
+- Add or remove GitHub labels per the transition tables in this document (already exercised throughout, e.g., "Starting to orchestrate a PR" and "Concluding PR orchestration")
```
aunger:
```suggestion
- Add or remove GitHub labels per the transition tables in this document
```

agents/dev_orchestration.md

Fair. Trimmed the whole rationale block down to the essentials (which label to expect and why, plus the one non-negotiable: never auto-remove `verification needed`). Dropped everything else it was restating.

agents/dev_orchestration.md

Applied as suggested.

aunger (commented):
🤖 Reviewer

Checked the third commit (`9daa38a`, "Trim over-explained comments per feedback") against the version I approved (`f1fd66b`). It's a pure comment-trimming pass in response to the inline nit on this thread ("Cool story, bro. Are the next 25 lines useful, though?") — the operative routing (the `If orchestrating is the only blocking label... / Otherwise ... escalate` conditional and its table) is byte-for-byte unchanged; only the surrounding `//` rationale and the "May" list entry lost their extra explanatory clauses. The trimmed comment still retains every fact load-bearing for correctness: which label is expected and why (`changes requested`/`changes done` clear on Reviewer return, `verification needed` is added downstream), the non-negotiable "never auto-remove `verification needed`," and the "not a concurrency guard" note. Nothing that was cut was necessary to keep the operative table from contradicting its own rationale, so this doesn't reopen the issue from my first round.

Verdict: LGTM (unchanged from my prior review of this branch).

PR #706, opened 2026-07-16
Add a CI workflow enforcing lint.sh, with a --check/--only mode

### LINKED ISSUE #698: Add a CI running lint.sh
## Implementation plan

Add CI enforcement of the repo's linters so formatting and hygiene violations are caught even when the commit-time git hook did not run.
The hook is wired only by `.claude/hooks/session-start.sh`, and only in remote (`CLAUDE_CODE_REMOTE=true`) sessions, so a local clone has no hook, and `git commit --no-verify` bypasses it everywhere.
CI is the only enforcement point that cannot be locally skipped.

The linters run as four independent jobs by tool family (python-ruff, markdown-mdformat, kotlin-ktlint, hygiene), in parallel, each passing or failing on its own so the checks show red or green individually.
Each job invokes `scripts/lint.sh` restricted to one tool family through a new `--only` selector, so the git hook and CI share one definition of how each tool runs and cannot drift apart.

### Design constraints

- CI checks, it does not fix.
  The `lint.sh` stack is auto-fixers (`ruff --fix`, `ruff format`, `mdformat`, `ktlint --format`, and the two whitespace fixers), so CI runs each tool in its check-only invocation: `ruff check` (no `--fix`), `ruff format --check`, `mdformat --check`, and `ktlint` with no `--format`.
  `check-yaml`, `check-toml`, and `check-merge-conflict` are already read-only.
  `trailing-whitespace-fixer` and `end-of-file-fixer` have no check mode, so those run and then fail on a non-empty `git diff` (their exit status already reflects a change, and the diff prints to the log).
- `check-added-large-files` needs `--enforce-all` on the whole-tree run.
  It intersects its file-list argument with the staged index unless `--enforce-all` is passed, so over a whole-tree run, where nothing is staged, it checks nothing without that flag (verified: an unstaged 600 KB file passes without it at exit 0, and is flagged with it at exit 1).
  The flag applies only to the whole-tree (`--all`) path; the staged hook path must not pass it, or it would flag pre-existing large files the commit is not adding.
- The lint checks live in their own workflow, not in `build.yml`.
  `build.yml`'s `build-and-test` job is gated on `needs_full_build` and is skipped on docs-only PRs, but Markdown and hygiene linting must run precisely on docs-only PRs.
  A standalone workflow also keeps fast lint feedback off the slow emulator job.
- The tree is green under every linter today (ruff, mdformat, ktlint, and all six hygiene checks, verified in planning), so no reformat-to-green commit is required first; the implementation branch should be re-verified green before merge.

### Steps

1. Add a check mode and a tool-family selector to `scripts/lint.sh`.
   Add `--check` (run each tool in its check-only invocation; for the two fixers with no check mode, run then fail on a dirty tree) and `--only ruff|mdformat|ktlint|hygiene` (run one family).
   Leave the existing default behavior (fix mode over a file list, or `--all`) byte-for-byte unchanged so the pre-commit hook is unaffected.
   In the same area, fix the large-files handling: pass `check-added-large-files --enforce-all` when the target set is `--all`, staged-only otherwise.
   Per `code_edit.md`, keep the behavior-preserving refactor, the one-line large-files fix, and the CI wiring as separate commits.

2. Extract ktlint provisioning into `scripts/install-ktlint.sh`.
   `KTLINT_VERSION`, `KTLINT_SHA256`, and the download-and-verify logic currently live only in `session-start.sh` STEP 3a.
   Move them into the shared script and have `session-start.sh` call it, so the CI ktlint job runs the identical, SHA-verified install rather than duplicating the checksum step.

3. Add `.github/workflows/lint.yml` with four jobs.
   Triggers: `push` to `main`, `pull_request` to `main`, and `workflow_dispatch`; `permissions: contents: read`.
   All jobs `runs-on: ubuntu-latest`, each starting with `actions/checkout@v6`:
   - python-ruff: `actions/setup-python@v6`, `pip install -r scripts/requirements-lint.txt`, then `scripts/lint.sh --check --only ruff --all`.
   - markdown-mdformat: same Python setup, then `scripts/lint.sh --check --only mdformat --all`.
   - hygiene: same Python setup, then `scripts/lint.sh --check --only hygiene --all` (the six pre-commit-hooks checks, large-files enforced by step 1).
   - kotlin-ktlint: `actions/setup-java@v5` (temurin 17, matching `build.yml`), `scripts/install-ktlint.sh`, then `scripts/lint.sh --check --only ktlint --all`.
   Reuse `requirements-lint.txt` for the Python pins so CI and the hook never diverge on versions.
   Pin any third-party action to a commit SHA per `code_edit.md`; `actions/` and `github/` first-party actions are exempt and should match the versions already used in this repo.

4. Extend `scripts/test_lint_hook.sh`.
   Its harness already builds throwaway repos and skips cleanly when the lint stack is absent. Add cases:
   - `--check` passes on a clean fixture and fails on a lint-dirty `.py`, a format-dirty `.md`, and a format-dirty `.kt` (the last gated on ktlint availability, as the suite already does).
   - `--check` leaves the working tree unmodified (fixture byte-identical after the run).
   - `--only <family>` runs only that family (a dirty `.md` passes `--only ruff` but fails `--only mdformat`).
   - a tracked file over 500 KB fails `--all` while the staged hook path is unaffected.
   These run in CI already via `build.yml`'s "Run shell script unit tests" step, so no new test wiring is needed.

### Files involved

- `scripts/lint.sh`: add `--check` and `--only`; enforce `--enforce-all` on the `--all` large-files check. Single definition of how each tool runs.
- `.claude/hooks/session-start.sh`: STEP 3a calls the new shared ktlint installer.
- `scripts/install-ktlint.sh` (new): pinned, SHA-verified ktlint install shared by session-start and CI.
- `.github/workflows/lint.yml` (new): the four tool-family jobs.
- `scripts/requirements-lint.txt`: unchanged, reused by the Python jobs for pinned versions.
- `scripts/test_lint_hook.sh`: new cases for check mode, tool selection, non-mutation, and the large-files rule.
- `.github/workflows/build.yml`: no change; its shell-test step already exercises the extended `test_lint_hook.sh`.

### Out of scope and deferred

- Making the lint checks block merge is deferred: building the workflow precedes configuring PR checks to gate on it, and that gating is branch-protection configuration in GitHub settings, handled separately later.
- CI does not auto-fix or push; it reports red and leaves fixing to the contributor, whose hook already fixes locally. Auto-commit from CI would need write access to PR branches and would fight the working tree.
- Whole-tree linting every run, rather than changed-files-only. The safety net stays simple; a base-ref diff optimization can be added later if wall-clock becomes a concern.

### Risks

- Drift between what the hook and CI enforce: mitigated by both going through `scripts/lint.sh`, and by CI reusing `requirements-lint.txt` and the shared ktlint installer.
- A later ruff or ktlint bump with new rules can turn `main` red; that is the intended signal, and the bump PR must reformat the tree in the same change, as issues #673, #681, and #686 did.
- ktlint job cost is a JVM cold start plus JAR download per run; acceptable and isolated to the one Kotlin job.

### Verification

- Run each job's exact command on a clean checkout and confirm exit 0 (confirmed for all linters in planning), then confirm each fails on a deliberately dirtied fixture of the matching type.
- The extended `scripts/test_lint_hook.sh` asserts check mode, tool selection, non-mutation, and the large-files rule, and runs in CI.
- `lint.yml` is itself YAML and is validated by the hygiene job's `check-yaml`.
- On a PR, confirm the four checks appear and that dirtying one file type shows exactly that job red while the others stay green.

## See also (original follow-up notes)

Source: PR #696 description, "Notes" section:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/696

> No CI safety net change: these linters run only at commit time, as before. A CI job running `scripts/lint.sh --all` would close that gap and reuse the same entry point; recommended follow-up, out of scope.

The same point is made in issue #667's "Risks and open decisions" section:
https://github.com/aunger/gallery-button-for-pixel-camera/issues/667

> No CI safety net: these linters run only at commit time, so a commit made where `core.hooksPath` was not set escapes them (already true of `pre-commit` today). A CI job running `scripts/lint.sh --all` would close the gap and reuse the same entry point; recommended follow-up, out of scope.

When implementing this, note a caveat raised in PR review:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/696#pullrequestreview-4710095934

> `check-added-large-files` in `lint.sh` runs on the passed file list, but the hook (without `--enforce-all`) only flags files staged in the index. Under `--all` (the future CI use), nothing is staged, so it effectively checks nothing.

This is a follow-on item and does not block merging PR #696.

aunger:
Fixes #698.

Adds CI enforcement of the repo's linters so formatting and hygiene violations are caught even when the commit-time git hook did not run.
A local clone has no hook, and `git commit --no-verify` bypasses it everywhere, so CI is the one enforcement point that cannot be locally skipped.

The linters run as four independent jobs by tool family (python-ruff, markdown-mdformat, hygiene, kotlin-ktlint), in parallel, each passing or failing on its own so the checks show red or green individually.
Each job invokes `scripts/lint.sh --check --only <family> --all`, so the git hook and CI share one definition of how each tool runs and cannot drift apart.

## How it works

- `scripts/lint.sh` gains `--check` (run each tool in its check-only invocation: `ruff check` with no `--fix`, `ruff format --check`, `mdformat --check`, `ktlint` with no `--format`) and `--only ruff|mdformat|ktlint|hygiene` (run one family).
  The read-only hygiene checks are unchanged; the two whitespace fixers have no check-only mode, so they run as usual and fail the run via their exit status if they change anything.
  On a clean tree, check mode writes nothing.
  The existing default (fix-mode) behavior is unchanged, so the pre-commit hook is unaffected.
- The `check-added-large-files` handling is fixed to pass `--enforce-all` on the whole-tree (`--all`) run.
  Without it, nothing is staged on an `--all` run, so the size check checked nothing; the flag is omitted on the file-list (hook) path, or it would flag pre-existing large files a commit is not adding.
- ktlint provisioning (version, published SHA-256, download-and-verify) is extracted from `session-start.sh` into `scripts/install-ktlint.sh`, which both the hook and the CI ktlint job now run, so the SHA-verified install is defined once.
- `.github/workflows/lint.yml` runs the four jobs on `push`/`pull_request` to `main` and `workflow_dispatch`, with `permissions: contents: read`.
  The Python jobs reuse `scripts/requirements-lint.txt` for pinned versions; the ktlint job runs `scripts/install-ktlint.sh` after `actions/setup-java`.
  It lives in its own workflow rather than `build.yml`, whose `build-and-test` job is skipped on docs-only PRs where Markdown and hygiene linting must still run.

CI checks only; it never auto-fixes or pushes.

## Commits

1. Add `--check` and `--only` to `scripts/lint.sh` (+ tests).
2. Enforce `check-added-large-files` over the whole tree under `--all` (+ test).
3. Extract ktlint provisioning into `scripts/install-ktlint.sh` (behavior-preserving).
4. Add `.github/workflows/lint.yml`.

## Testing

`scripts/test_lint_hook.sh` is extended with cases for check mode, tool-family selection, check-mode non-mutation, and the large-files rule; it runs in CI via `build.yml`'s existing "Run shell script unit tests" step.
Locally the full suite passes (18/18), and each of the four CI commands (`scripts/lint.sh --check --only <family> --all`) exits 0 against a clean checkout.

## Out of scope

- Making these checks block merge (branch-protection configuration in GitHub settings) is deferred, per the issue.

🦀

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -172,49 +172,14 @@ fi
 
 # STEP 3a: ktlint (Kotlin formatter), from Maven Central.
 #
-# ktlint is fetched as the ktlint-cli fat JAR from Maven Central--the same
-# trusted, immutable, checksummed host Gradle already resolves this build
-# from--rather than from GitHub Releases, which this sandbox blocks with a 403
-# (issue #667).  The published SHA-256 is verified before the JAR is used, so a
-# proxy error page fails loudly here instead of masquerading as a corrupt
-# "binary" that only breaks later with an opaque Exec format error.  A small
-# wrapper on PATH runs the JAR via `java -jar`.
-#
-# When bumping KTLINT_VERSION, update KTLINT_SHA256 to the new release's
-# published checksum from
-# https://repo1.maven.org/maven2/com/pinterest/ktlint/ktlint-cli/<version>/ktlint-cli-<version>-all.jar.sha256
-KTLINT_VERSION="1.8.0"
-KTLINT_SHA256="369ad2b789f95a011f807e1fcb690ccef80bd7cd014fd139e73ae82dcc0baeab"
-KTLINT_JAR_DIR="$HOME/.local/lib/ktlint"
-KTLINT_JAR="$KTLINT_JAR_DIR/ktlint-cli-$KTLINT_VERSION-all.jar"
-KTLINT_BIN="$LOCAL_BIN/ktlint"
-if [[ -x "$KTLINT_BIN" && -f "$KTLINT_JAR" ]]; then
-    echo "[session-start] Step 3a: ktlint present--skip"
-else
-    echo "[session-start] Step 3a: installing ktlint $KTLINT_VERSION from Maven Central..."
-    mkdir -p "$KTLINT_JAR_DIR"
-    KTLINT_URL="https://repo1.maven.org/maven2/com/pinterest/ktlint/ktlint-cli/$KTLINT_VERSION/ktlint-cli-$KTLINT_VERSION-all.jar"
-    TMP_JAR=$(mktemp "$KTLINT_JAR_DIR/.download-XXXXXX")
-    trap 'rm -f "$TMP_JAR"' EXIT
-    # -f: fail on an HTTP error status instead of writing the error response
-    # body to the JAR as if it were the artifact (issue #667).
-    curl -fsSL "$KTLINT_URL" -o "$TMP_JAR"
-    ACTUAL_SHA=$(sha256sum "$TMP_JAR" | cut -d' ' -f1)
-    if [[ "$ACTUAL_SHA" != "$KTLINT_SHA256" ]]; then
-        echo "[session-start] Step 3a: ERROR: ktlint SHA-256 mismatch (refusing to install)" >&2
-        echo "[session-start]   expected $KTLINT_SHA256" >&2
-        echo "[session-start]   actual   $ACTUAL_SHA" >&2
-        exit 1
-    fi
-    mv "$TMP_JAR" "$KTLINT_JAR"
-    trap - EXIT
-    cat > "$KTLINT_BIN" << EOF
-#!/usr/bin/env bash
-exec java -jar "$KTLINT_JAR" "\$@"
-EOF
-    chmod +x "$KTLINT_BIN"
-    echo "[session-start] Step 3a: ktlint installed and SHA-256 verified"
-fi
+# The pinned, SHA-256-verified install lives in scripts/install-ktlint.sh, which
+# the CI ktlint lint job (.github/workflows/lint.yml) also runs, so both paths
+# provision the identical ktlint from one definition.  It installs the wrapper
+# into $HOME/.local/bin (the $LOCAL_BIN this step already put on PATH) and is
+# idempotent, so re-running is a fast no-op.  See that script for the rationale
+# on fetching the fat JAR from Maven Central rather than GitHub Releases.
+echo "[session-start] Step 3a: ensuring ktlint is installed..."
+"$REPO_ROOT/scripts/install-ktlint.sh"
 
 # STEP 3b: Python lint tools (ruff, pre-commit-hooks checks, mdformat), from PyPI.
 #
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,85 @@
+name: Lint
+
+# CI enforcement of the repo's linters, so formatting and hygiene violations are
+# caught even when the commit-time git hook did not run (a local clone has no
+# hook, and `git commit --no-verify` bypasses it everywhere). CI is the one
+# enforcement point that cannot be locally skipped.
+#
+# Four independent jobs, one per tool family, run in parallel and pass or fail on
+# their own so each shows red or green individually. Each invokes scripts/lint.sh
+# --check --only <family> --all, so the git hook and CI share one definition of
+# how every tool runs. CI checks only; it never auto-fixes or pushes.
+#
+# This lives in its own workflow rather than build.yml: build.yml's build-and-test
+# job is skipped on docs-only PRs, but Markdown and hygiene linting must run
+# precisely there.
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  python-ruff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.x'
+      - name: Install Python lint tools
+        run: |
+          pip install -r scripts/requirements-lint.txt
+          echo "LINT_BIN_DIR=$(python -c 'import sysconfig; print(sysconfig.get_path("scripts"))')" >> "$GITHUB_ENV"
+      - name: Ruff check
+        run: scripts/lint.sh --check --only ruff --all
+
+  markdown-mdformat:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.x'
+      - name: Install Python lint tools
+        run: |
+          pip install -r scripts/requirements-lint.txt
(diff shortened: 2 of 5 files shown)


aunger:
🤖 Reviewer

LGTM. This faithfully implements the plan in #698 and I found nothing blocking.

What I verified locally against the PR head (76ec7a9):

- `scripts/test_lint_hook.sh` passes 18/18, including the five new groups (h-l): check mode on a clean tree, check mode flagging dirty `.py`/`.md`/`.kt`, non-mutation, `--only` isolation, and the large-files `--enforce-all` vs. hook-path distinction.
- Each of the four exact CI commands (`scripts/lint.sh --check --only <family> --all`) exits 0 on a clean checkout.
- Largest tracked file is ~202 KB (`scripts/test_ci_monitor.py`), comfortably under the 500 KB `check-added-large-files` limit, so the hygiene job's new `--enforce-all` will not turn `main` red on merge.

Design and correctness notes:

- The single-source-of-truth goal holds: hook and CI both route through `lint.sh`, and the `--enforce-all` flag is correctly scoped to the `--all` path only (line 146), leaving the hook's staged-only large-files behavior byte-for-byte unchanged.
- The ktlint extraction is genuinely behavior-preserving: same pinned 1.8.0 + SHA-256, same `~/.local/bin` wrapper and `~/.local/lib/ktlint` JAR, same idempotent skip check. `REPO_ROOT` is defined (session-start.sh:160) before the STEP 3a call site.
- No false-green paths: a missing tool resolves to a non-zero `run`, and the whitespace fixers surface a dirty tree via their exit status. The header's non-mutation claim is correctly scoped to "on a clean tree" (the fixers still have no true check mode, which is documented).
- Action versions are consistent: `setup-python@v6` matches `semgrep.yml`; `setup-java@v5` and `checkout@v6` match `build.yml`. All actions are first-party `actions/*`, so no third-party SHA pin is owed; ktlint is version+SHA-256 pinned.
- Commits are cleanly split by concern (feature+tests, large-files fix+test, behavior-preserving refactor, workflow), and no existing test requirements were loosened.

Non-blocking observation (no change required): the three Python jobs duplicate the "Install Python lint tools" step verbatim. A matrix over the python families would remove the repetition while keeping each family as its own red/green check, but the current explicit form is clear and the four-independent-jobs split is the stated design intent.

CI had not reported results at review time; not blocking, per review protocol.

PR #708, opened 2026-07-16
Remove stale ktlint JARs on a KTLINT_VERSION bump

### LINKED ISSUE #700: Clean up stale ktlint JAR versions left behind on a KTLINT_VERSION bump
**Blocked by: #698** (implementation-sequencing only; this does not change the merge-blocking status noted below).
Reason: both change the ktlint provisioning that currently lives in `.claude/hooks/session-start.sh` STEP 3a.
#698 extracts that logic into a new shared `scripts/install-ktlint.sh` used by both the git hook and the new CI ktlint job, so doing #698 first lets this stale-JAR cleanup be added once in `install-ktlint.sh` (covering the hook and CI together) and avoids a merge conflict from both changes editing the same install block.
(Added 2026-07-16 by cross-issue dependency-sequencing analysis.)

---

Source: PR #696 review comments:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/696#pullrequestreview-4710095934
https://github.com/aunger/gallery-button-for-pixel-camera/pull/696#pullrequestreview-4710131054

> On a `KTLINT_VERSION` bump, the version-suffixed old JAR is left behind in `~/.local/lib/ktlint/` (the new one installs under a new name and the wrapper is regenerated correctly). Harmless disk residue, not a correctness issue.

Noted as a non-blocking observation in both the initial review and the re-review; no changes were requested for it in PR #696.

This is a follow-on item and does not block merging PR #696: https://github.com/aunger/gallery-button-for-pixel-camera/pull/696

aunger:
Fixes #700.

`scripts/install-ktlint.sh` (extracted from `.claude/hooks/session-start.sh` STEP 3a by #698) names its JAR with a version suffix, `ktlint-cli-$KTLINT_VERSION-all.jar`. On a `KTLINT_VERSION` bump, the new JAR installs under its own new name and the `ktlint` wrapper is correctly regenerated to point at it, but the previous version's JAR is never removed, so it accumulates as disk residue in `$KTLINT_JAR_DIR` (`~/.local/lib/ktlint/` by default) indefinitely.

This adds `cleanup_stale_jars()` to `install-ktlint.sh`, which removes every `ktlint-cli-*-all.jar` in that directory other than the one the script's current `KTLINT_VERSION` names. It runs only after the current version's JAR is confirmed present, whether this run just downloaded and SHA-256-verified it or found it already installed, so a failed download never costs the previously working install. Since both the git hook and the CI `kotlin-ktlint` job call this one shared script, the cleanup applies uniformly to both.

New coverage in `scripts/test_install_ktlint.sh` exercises the network-free "already installed" fast path: a lone stale JAR removed, several stale JARs all removed, a no-op when there is nothing stale, and an unrelated file in the same directory left alone. The real download-and-verify path continues to run for real on every `session-start.sh` invocation and in the `kotlin-ktlint` CI job, so it did not need a separate network-dependent test.

## Test plan

- [x] `scripts/test_install_ktlint.sh` passes locally.
- [x] `scripts/test_lint_hook.sh` still passes (unaffected by this change).
- [x] `scripts/lint.sh --check scripts/install-ktlint.sh scripts/test_install_ktlint.sh` is clean.
- [x] Ran `scripts/install-ktlint.sh` for real against an already-installed ktlint; confirmed it still exits via the fast path with no unwanted deletions.

### DIFF
--- a/scripts/install-ktlint.sh
+++ b/scripts/install-ktlint.sh
@@ -17,7 +17,16 @@
 # Install locations (override via the environment if needed):
 #   $KTLINT_BIN_DIR/ktlint            wrapper script on PATH (default ~/.local/bin)
 #   $KTLINT_JAR_DIR/ktlint-cli-*.jar  the verified JAR (default ~/.local/lib/ktlint)
-# Idempotent: if both already exist, it does nothing.
+# Idempotent: if both already exist, it does nothing (beyond the stale-JAR
+# cleanup below).
+#
+# On a KTLINT_VERSION bump, the previous version's JAR is not overwritten (the
+# filename is version-suffixed) and the wrapper is simply repointed at the new
+# one, so the old JAR would otherwise be left behind in $KTLINT_JAR_DIR forever
+# (issue #700). cleanup_stale_jars() removes every ktlint-cli-*-all.jar in that
+# directory other than the one this script's KTLINT_VERSION currently names,
+# once that current JAR is confirmed present, whether this run just verified
+# and installed it or found it already there.
 #
 # When bumping KTLINT_VERSION, update KTLINT_SHA256 to the new release's
 # published checksum from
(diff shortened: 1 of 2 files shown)


aunger:
🤖 Reviewer

(Posted as an ordinary comment rather than via the PR review tool: this account cannot approve its own PR, since it's also the PR author's account.)

Verified against issue #700 and its blocker #698 (closed, so `scripts/install-ktlint.sh` already exists on `main` as this diff assumes; no conflict).

Checked:
- `cleanup_stale_jars()` is called from both the "already installed" fast path and after a fresh verified install, never before the current version's JAR is confirmed on disk, so a failed download or SHA mismatch can't cost the previously working install. Matches the stated design.
- The download's temp file (`.download-XXXXXX`) doesn't match the `ktlint-cli-*-all.jar` glob, so it's never a cleanup target; irrelevant anyway since it's cleaned by the existing `trap`.
- Both call sites that provision ktlint (`.claude/hooks/session-start.sh` STEP 3a and the `kotlin-ktlint` job in `.github/workflows/lint.yml`) go through this one script, so the cleanup applies uniformly as claimed.
- Ran `scripts/test_install_ktlint.sh` locally: all 9 assertions pass (stale-JAR removal, multi-stale removal, no-op when nothing stale, unrelated file untouched).
- Ran `scripts/lint.sh --check scripts/install-ktlint.sh scripts/test_install_ktlint.sh`: clean.
- `build.yml`'s shell-test step is `for test_script in scripts/test_*.sh`, so the new test file is picked up automatically; no CI wiring was needed and none was added. The CI summary comment on this PR already shows `build-and-test` passing.

Nit (non-blocking): in test case (d), `run_case "$CASE" "ktlint-cli-1.6.0-all.jar"` runs install-ktlint.sh once (exercising the same stale-removal already covered by case (a)) before `notes.txt` is added and the script is invoked a second time by hand to check the unrelated file. The first invocation isn't needed to prove the point of case (d); building the fixture directly would be a little more direct. Not asking for a change, just flagging it.

No correctness issues found. LGTM.

PR #688, opened 2026-07-15
Switch Markdown formatting from markdownlint-cli2 to mdformat

aunger:
Fixes #681.

## What changed

- Replaced the `markdownlint-cli2` pre-commit hook with a `repo: local`, `language: system` `mdformat` hook (`.pre-commit-config.yaml`), mirroring the existing `ktlint` hook. Args: `--wrap keep --number`. Files pattern: `'\.md$'`, matching the issue's literal instruction.
- Added STEP 3d to `.claude/hooks/session-start.sh`, installing `mdformat==1.0.0`, `mdformat-gfm==1.0.0`, and `mdformat-frontmatter==2.1.2` from PyPI, pinned.
- Does **not** install `mdformat-sentencebreak`, per the issue's instruction: it corrupts inline code spans containing a period, breaks at commas rather than only sentence ends, and is unmaintained (single 2022 release).
- Removed `.markdownlint.yaml`; nothing reads it after this switch.
- Reformatted the whole Markdown tree with the new hook and committed the result (mechanical: table padding, list blank-line spacing, thematic-break normalization, ordered-list renumbering; no prose content changes).

## Table plugin: mdformat-gfm, not mdformat-tables

The issue named `mdformat-tables` for GFM pipe table support. That plugin is deprecated: its GitHub repo is archived, and its own README says "Mdformat-tables has moved to mdformat-gfm. You should use the GFM plugin for tables support from now on." It also never lifted its `mdformat<0.8.0,>=0.7.5` dependency ceiling after that move.

`mdformat-gfm` is the maintained successor. It declares `mdformat>=0.7.5` with no upper bound, so it does not reintroduce a version ceiling. `mdformat` is therefore pinned to the current `1.0.0`, as the issue intended once `mdformat-sentencebreak` is dropped: `mdformat-gfm` also brings GFM task lists, strikethrough, and autolink support in addition to tables, none of which changed any output on this repo's current Markdown tree (verified: reformatting the whole tree with `mdformat-gfm` produces byte-identical output to the table-only formatting this repo already used).

## Enforcement gaps (called out per the issue, not blockers)

- MD040 (fenced code blocks must name a language) is no longer enforced now that `markdownlint-cli2` is gone. Checked the whole tree for opening fences with no language tag: none found, so nothing regresses today, but new offenders will not be caught going forward.
[...]

PR #709, opened 2026-07-17
Enforce MD040 (fenced code blocks must name a language) via scripts/lint.sh

### LINKED ISSUE #689: Enforce MD040 (fenced code blocks must specify a language) now that markdownlint-cli2 is gone
**Blocked by: #698** (implementation-sequencing only; this does not change the merge-blocking status noted below).
Reason: #689 exists to catch new MD040 offenders going forward, which requires an enforcement point that cannot be locally skipped.
Today the only lint enforcement is the commit-time git hook, which is absent on a fresh clone and bypassable with `git commit --no-verify`.
#698 adds `.github/workflows/lint.yml`, the first un-skippable enforcement point, plus the `scripts/lint.sh --check --only mdformat` selector and a markdown job that an MD040 check (a small script or an mdformat plugin) should route through so it runs identically in the hook and in CI.
Implement #698 first so #689's check is enforced where it matters and wired in one place, rather than added to the bypassable local hook now and re-plumbed into CI later.
(Added 2026-07-16 by cross-issue dependency-sequencing analysis.)

---

This is a follow-on item and does not block merging PR #688.

Source: PR #688 description, "Enforcement gaps" section: https://github.com/aunger/gallery-button-for-pixel-camera/pull/688
Also noted in issue #681's "Enforcement gaps to note (not blockers)" section: https://github.com/aunger/gallery-button-for-pixel-camera/issues/681
And in the "Cross-cutting risks" section of the umbrella issue #676: https://github.com/aunger/gallery-button-for-pixel-camera/issues/676

After PR #688 switches Markdown formatting from `markdownlint-cli2` to `mdformat`, MD040 (fenced code blocks must name a language) is no longer enforced.
PR #688 checked the tree at merge time and found no offenders, but new offenders will not be caught going forward.
Consider adding an automated check (for example, a small script or a first-party mdformat plugin) that enforces this rule.

Reason not a merge blocker: explicitly called out as "not a blocker" in issue #681 and PR #688, and as an accepted, documented consequence of the formatter switch in the umbrella issue #676.

aunger:
Fixes #689.

markdownlint-cli2 enforced MD040 (fenced code blocks must name a language) until PR #688 replaced it with mdformat, which reformats a fence's contents but does not check whether it declares a language. This left new MD040 offenders uncaught going forward.

## Changes

- `scripts/check_md040.py` (new): scans given Markdown files for fenced code blocks with no language, using a stdlib-only, CommonMark-approximating fence scanner. Handles both backtick and tilde fences, and correctly ignores a bare fence that appears as literal content nested inside a longer, labeled outer fence.
- `scripts/lint.sh`: wires the new check into the mdformat family, in both fix and check mode (there is nothing to auto-fix, so it always reports). This runs it identically at commit time (the git pre-commit hook) and in CI (`.github/workflows/lint.yml`'s `markdown-mdformat` job, added by #698/#706), per the sequencing note on the issue.
- `scripts/test_lint_hook.sh`: new end-to-end cases -- an unlabeled fence blocks a commit, `--check` flags it without modifying the file, and a labeled fence passes.
- `scripts/test_check_md040.py` (new): unit tests for the fence scanner (labeled/unlabeled, backtick/tilde, multiple violations, nested-as-content, indented code blocks, `main()`'s exit codes).

The tree has no existing MD040 offenders (confirmed by running the new check over every tracked `.md` file), so no reformatting was needed.

## Test plan

- [x] `scripts/lint.sh --check --all` passes on the full tree.
- [x] `bash scripts/test_lint_hook.sh` passes (22 cases, including the 3 new MD040 cases).
- [x] `python3 -m unittest scripts.test_check_md040` passes (12 cases).
- [x] `python3 -m unittest discover -s scripts -p "test_*.py"` shows no new failures (the 4 pre-existing errors are unrelated missing-dependency/environment issues in this sandbox, not caused by this change).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -51,6 +51,8 @@ jobs:
           pip install -r scripts/requirements-lint.txt
           echo "LINT_BIN_DIR=$(python -c 'import sysconfig; print(sysconfig.get_path("scripts"))')" >> "$GITHUB_ENV"
       - name: mdformat check
+        # Also runs scripts/check_md040.py (issue #689): MD040 (a fenced code
+        # block must name a language) is not something mdformat itself checks.
         run: scripts/lint.sh --check --only mdformat --all
 
   hygiene:
--- a/scripts/check_md040.py
+++ b/scripts/check_md040.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+"""Check Markdown files for fenced code blocks with no declared language.
+
+MD040 (markdownlint's rule of the same name: "fenced code blocks should have a
+language specified") was enforced by markdownlint-cli2 until issue #688
+replaced it with mdformat. mdformat reformats a fence's contents but does not
+check whether the fence declares a language, so a new offender (a bare
+```` ``` ```` with nothing after it) would go uncaught. This script closes that
+gap: it runs as part of the mdformat family in scripts/lint.sh (the git
+pre-commit hook) and in .github/workflows/lint.yml's markdown-mdformat job, so
+the hook and CI share one definition of the check.
+
+Only the Python standard library is used, so this runs under the ambient
+`python3` with no dependency install step, the same way scripts/lint.sh invokes
+other first-party scripts (see e.g. scripts/check_non_docs_changes.sh). This
+also sidesteps scripts/lint.sh's $LINT_BIN_DIR indirection, which exists to
+resolve third-party tools installed from a trusted registry (issue #667) and
+does not apply to a first-party script checked out with the repo itself.
+
+The fence scanner is a lightweight, line-based approximation of CommonMark's
+fenced-code-block rules (opening delimiter of 3+ backticks or tildes, optional
+up to 3 spaces of indentation, closed only by a same-character delimiter at
+least as long): the mdformat and mdformat-gfm output this repo produces is
+plain, unindented fences at column 0 (`--wrap keep` normalizes prose but not
+fence indentation), so full blockquote/list-nesting fidelity is not needed.
+
+Usage:
+    scripts/check_md040.py FILE [FILE ...]
+
+Exit code:
+    0  no fenced code block is missing a language in any given file.
+    1  at least one fenced code block is missing a language; each is printed
+       to stderr as "path:line: fenced code block has no language (MD040)".
+    2  usage error (no files given).
+"""
+
+import re
+import sys
+
+# A fence opener/closer: up to 3 spaces of indentation, then 3+ of the same
+# fence character, then (for an opener) an optional info string.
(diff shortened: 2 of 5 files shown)


aunger (commented):
placeholder

aunger (commented):
🤖 Reviewer

(Disregard my previous review on this PR, body "placeholder" — that was posted by mistake, before I'd finished forming my review. This one supersedes it.)

Verified the sequencing precondition from the issue: #698/#706 are already merged into `main` (`.github/workflows/lint.yml` and the `--check --only mdformat --all` selector both exist there), so this PR is correctly building on top of the un-skippable enforcement point rather than re-plumbing later.

I checked out the branch and ran the full suite:

- `python3 -m unittest scripts.test_check_md040` — 12/12 pass.
- `bash scripts/test_lint_hook.sh` — 22/22 pass, including the 3 new MD040 cases.
- `scripts/lint.sh --check --all` — clean on the whole tree.

I also hand-tested the fence scanner against several CommonMark edge cases not covered by the existing test suite, to check the "lightweight, line-based approximation" claim in the docstring: an unclosed fence at EOF, a backtick-fence opener whose info string itself contains a backtick (invalid per spec, correctly not treated as an opener), a closing delimiter shorter than the opener (correctly treated as content, not a close), a tilde closer with trailing text after it (correctly rejected as a closer, per CommonMark's "nothing else but trailing space" rule), and a tab-indented fence (correctly not recognized as a fence, since a tab isn't 0-3 spaces). All behaved correctly. I also grepped the tree for any indented fences nested inside list items (the one documented gap versus true CommonMark fidelity) and found none, so the gap has no live impact today.

One nit, non-blocking: the docstring in `scripts/check_md040.py` justifies invoking ambient `python3` (bypassing `$LINT_BIN_DIR`) by saying this is "the same way scripts/lint.sh invokes other first-party scripts (see e.g. scripts/check_non_docs_changes.sh)". That comparison doesn't hold up: `lint.sh` doesn't invoke `check_non_docs_changes.sh` at all — that script is invoked separately from `.github/workflows/build.yml` for an unrelated purpose (deciding whether to skip the full build), and it's a bash script, not Python. Grepping `lint.sh` confirms `check_md040.py` is actually the *first and only* first-party script it invokes directly — there's no existing precedent, this PR sets one. The reasoning that stands on its own (first-party script, not a registry-installed tool, so `$LINT_BIN_DIR` doesn't apply) is fine; it's just the "see e.g." citation that's misleading to a future reader trying to find the precedent.

Everything else — scope, integration point (single added line in `lint.sh`'s mdformat branch, comment-only change to `lint.yml`), and test coverage — looks tight and correct.

LGTM.

PR #712, opened 2026-07-17
Rename lint.sh's tool families to language names

### LINKED ISSUE #711: Lint.py should rename its “families” of linters
The lint.py CLI is supposed to accept the name of a formatter/linter *family*, but the families are too specific. 

Since #709 merged, asking for the `mdformat` family runs more than just mdformat.

I suggest the groups should be named after their main languages, rather than their main tools:
- kotlin
- markdown
- python
- hygiene

Is there a reason not to do this?

aunger:
Fixes #711.

`scripts/lint.sh --only` grouped its checks under values named after the primary tool in each group (`ruff`, `mdformat`, `ktlint`, `hygiene`). Since #709, the `mdformat` family also runs `scripts/check_md040.py`, so asking for that family now runs more than just mdformat. A tool-named family is misleading once it can contain more than one tool.

This renames the families to their main language instead, per the issue's suggestion:

- `ruff` -> `python`
- `mdformat` -> `markdown`
- `ktlint` -> `kotlin`
- `hygiene` -> `hygiene` (unchanged; it is not language-specific)

Updated:

- `scripts/lint.sh`: the `--only` argument parsing, usage text, and the `want <family>` calls/comments for each block.
- `.github/workflows/lint.yml`: the `--only` flag passed by each per-family CI job (job IDs are left as-is, e.g. `python-ruff`, since they already lead with the language name).
- `scripts/test_lint_hook.sh`: the `--only` values used in the test fixtures that exercise family isolation.
- `scripts/check_md040.py`: a comment referencing the old family name.

## Test plan

Covered by the existing automated suite (`scripts/test_lint_hook.sh`, exercised via the git hook and CI); no manual verification is needed. Ran the suite locally with the lint tools provisioned into a throwaway venv (plus the already-installed `ktlint`): all 22 assertions pass. Also ran `scripts/lint.sh --check --only <family> --all` for each of the four new family names directly against the tree, and confirmed an old/unknown family name (e.g. `ruff`) is now rejected by the usage error.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,7 +37,7 @@ jobs:
           pip install -r scripts/requirements-lint.txt
           echo "LINT_BIN_DIR=$(python -c 'import sysconfig; print(sysconfig.get_path("scripts"))')" >> "$GITHUB_ENV"
       - name: Ruff check
-        run: scripts/lint.sh --check --only ruff --all
+        run: scripts/lint.sh --check --only python --all
 
   markdown-mdformat:
     runs-on: ubuntu-latest
@@ -53,7 +53,7 @@ jobs:
       - name: mdformat check
         # Also runs scripts/check_md040.py (issue #689): MD040 (a fenced code
         # block must name a language) is not something mdformat itself checks.
-        run: scripts/lint.sh --check --only mdformat --all
+        run: scripts/lint.sh --check --only markdown --all
 
   hygiene:
     runs-on: ubuntu-latest
@@ -84,4 +84,4 @@ jobs:
         # needed for the check step below.
         run: scripts/install-ktlint.sh
       - name: ktlint check
-        run: scripts/lint.sh --check --only ktlint --all
+        run: scripts/lint.sh --check --only kotlin --all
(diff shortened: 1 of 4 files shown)


aunger:
🤖 Reviewer

Verified this independently rather than just reading the diff:

- Checked out the branch and grepped the whole tree (not just the changed files) for lingering `--only ruff|mdformat|ktlint` usages or other stale family-name references. Found none outside the four files the PR touches; `.claude/environment.md`, `.claude/hooks/session-start.sh`, `AGENTS.md`, `scripts/git-hooks/pre-commit`, etc. only reference the underlying tool binaries (`ruff`, `mdformat`, `ktlint`), which correctly keep their names since only the `--only` family values are being renamed.
- Ran `scripts/test_lint_hook.sh` for real (tools were provisioned): 22/22 assertions pass, matching the PR's claim.
- Ran `scripts/lint.sh --check --only ruff --all` directly: correctly rejected with the updated usage message (old family names are gone).
- Ran `scripts/lint.sh --check --only <family> --all` for each of `python`, `markdown`, `kotlin`, `hygiene`: all exit 0 against the clean tree, matching the PR's claim.

This is a clean, minimal, purely mechanical rename. It matches the issue's exact suggested names (kotlin/markdown/python/hygiene), the CI job IDs are sensibly left alone with a stated rationale (they already lead with the language name, e.g. `kotlin-ktlint`), and the test fixtures/variable names (`RC_RUFF` -> `RC_PY`) and comments were updated consistently rather than just the string literals. No behavior change is bundled in, consistent with a pure-rename single commit.

No issues found. LGTM.

PR #713, opened 2026-07-17
Propagate issue labels to a linked PR on a Fixes relationship

### LINKED ISSUE #621: Propagate issue labels to a linked PR on a Fixes relationship
When a PR closes an issue via a "Fixes #N" (or other GitHub closing-keyword) relationship, automatically copy the issue's labels onto the PR.

## Requirements

- Detect the Fixes/Closes relationship between a PR and an issue (e.g. via the PR's closing references, not just regex-parsing the PR body, since GitHub recognizes several keyword variants and cross-references).
- Apply each of the issue's labels to the linked PR.
- **Exception:** skip applying an issue label if doing so would cause `scripts/enforce_mutually_exclusive_labels.py`'s mutual-exclusion enforcement to remove a label the PR already carries. In other words, don't let this automation indirectly strip an existing PR label via the mutual-exclusion side effect; existing PR labels win over a conflicting label being copied in from the issue.

## Context

This came up because issue-to-PR label propagation doesn't currently exist: labels applied to an issue (by title-matching automation, by hand, or otherwise) don't carry over to the PR that fixes it, so the PR starts unlabeled or only gets labels from its own title.

See `scripts/enforce_mutually_exclusive_labels.py` and `.github/workflows/enforce-mutually-exclusive-labels.yml` for the existing mutual-exclusion sets/prefix groups (`[p1, p2, p3]`, `[verification needed, verified]`, `[changes requested, changes done]`, `[orchestrate, orchestrating]`, `c-a-*`, `c-r-*`) this automation must not fight with.

aunger:
Fixes #621.

Adds `scripts/propagate_issue_labels.py`, run by a new
`.github/workflows/propagate-issue-labels.yml` workflow on `pull_request`
`opened` (and `edited` when the body changed), that copies a linked issue's
labels onto a PR that closes it.

## How this addresses the issue

- **Detecting the Fixes/Closes relationship**: uses the GitHub GraphQL API's
  `closingIssuesReferences` field on the PR, which is GitHub's own
  resolution of every closing-keyword variant ("Fixes", "Closes",
  "Resolves", etc.) and manually linked issues, rather than a regex scan of
  the PR body. Cross-repository closing references are skipped, matching
  the rest of this repo's label automation, which is scoped to a single
  repository.
- **Applying labels**: every label carried by a closing issue is applied to
  the PR, unless it is already present.
- **The mutual-exclusion exception**: before applying a candidate label,
  `labels_to_propagate()` checks whether doing so would give
  `scripts/enforce_mutually_exclusive_labels.py` grounds to remove a label
  the PR already carries (or one accepted earlier in the same run, so
  multiple linked issues can't fight each other either). If so, the
  candidate is skipped instead, so an existing PR label always wins over a
  conflicting one being copied in from a linked issue.

Labels are only ever added, never removed, consistent with `label_by_title.py`
and `backfill_labels_by_title.py`.

## Test plan
- [x] Unit tests in `scripts/test_propagate_issue_labels.py` cover the
  GraphQL query helper, closing-issue label extraction (including the
  cross-repo filter), the propagate/skip partitioning logic (fixed-set and
  prefix-group conflicts, same-run conflicts between two linked issues,
  case-insensitive dedup), and `main()`'s error and success paths.
- [x] `scripts/lint.sh` (ruff + hygiene checks) passes on the new files.
- [x] Manually confirm the workflow fires correctly on a real PR in this
  repo that closes a labeled issue (GitHub Actions trigger behavior can't be
  exercised by the unit tests above).

### DIFF
--- a/.github/workflows/propagate-issue-labels.yml
+++ b/.github/workflows/propagate-issue-labels.yml
@@ -1,0 +1,36 @@
+name: Propagate issue labels to linked PR
+
+# When a pull request closes an issue via a closing-keyword relationship
+# ("Fixes #N", "Closes #N", the other variants GitHub recognizes, or a
+# manually linked issue), copies that issue's labels onto the pull request.
+# See scripts/propagate_issue_labels.py for the exact rules, including the
+# exception that protects an existing PR label from an indirect
+# mutual-exclusion removal.
+#
+# Runs on "opened" (the closing relationship may already be declared in the
+# initial body) or "edited" where the body itself changed
+# (github.event.changes.body is only present when the body was part of the
+# edit), so unrelated edits (title, labels, base branch) do not re-trigger it.
+
+on:
+  pull_request:
+    types: [opened, edited]
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  propagate-issue-labels:
+    runs-on: ubuntu-latest
+    if: github.event.action == 'opened' || github.event.changes.body
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Propagate linked issue labels
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: python3 scripts/propagate_issue_labels.py
--- a/.github/workflows/propagate-issue-labels.yml
+++ b/.github/workflows/propagate-issue-labels.yml
@@ -30,7 +30,7 @@ name: Propagate issue labels to linked PR
 
 on:
   pull_request:
-    types: [opened, edited, reopened, synchronize, ready_for_review, review_requested]
+    types: [opened, edited, reopened]
 
 permissions:
   issues: write
--- a/.github/workflows/propagate-issue-labels.yml
+++ b/.github/workflows/propagate-issue-labels.yml
@@ -7,14 +7,30 @@ name: Propagate issue labels to linked PR
 # exception that protects an existing PR label from an indirect
 # mutual-exclusion removal.
 #
-# Runs on "opened" (the closing relationship may already be declared in the
-# initial body) or "edited" where the body itself changed
-# (github.event.changes.body is only present when the body was part of the
-# edit), so unrelated edits (title, labels, base branch) do not re-trigger it.
+# GitHub fires no webhook event for linking a PR to an issue via the
+# "Development" sidebar after the PR is already open, so a closing
+# relationship established that way, with no body edit, cannot be caught by a
+# single trigger type. Rather than run on a perpetual schedule to cover that
+# gap, this runs on every pull_request event type that commonly recurs during
+# a PR's normal review lifecycle in this repo (a fresh push, a reopen, a
+# ready-for-review flip, a review request), so most sidebar-linked PRs get
+# reconciled by their own ordinary activity. "labeled"/"unlabeled" are
+# deliberately excluded: re-running on "unlabeled" would re-apply a label a
+# human just removed on purpose, clobbering their edit.
+#
+# "edited" only re-runs when the body itself changed
+# (github.event.changes.body is only present then), so unrelated edits
+# (title, base branch) do not re-trigger it. Every other listed type always
+# runs, since none of them carry a "changes" object to gate on.
+#
+# A PR that gets sidebar-linked and then sees none of these events before
+# merge is not covered automatically; scripts/reconcile_issue_labels.py is
+# available to reconcile every open PR on demand (run locally, or dispatch
+# it in Actions) for that residual case.
 
 on:
   pull_request:
-    types: [opened, edited]
+    types: [opened, edited, reopened, synchronize, ready_for_review, review_requested]
 
 permissions:
   issues: write
@@ -23,7 +39,7 @@ permissions:
 jobs:
   propagate-issue-labels:
     runs-on: ubuntu-latest
-    if: github.event.action == 'opened' || github.event.changes.body
+    if: github.event.action != 'edited' || github.event.changes.body
     steps:
       - name: Check out repository
         uses: actions/checkout@v6
--- a/.github/workflows/reconcile-issue-labels.yml
+++ b/.github/workflows/reconcile-issue-labels.yml
@@ -1,0 +1,29 @@
+name: Reconcile issue labels on open PRs
+
+# Periodically re-applies scripts/propagate_issue_labels.py's label
+# propagation rule to every open pull request. This exists because GitHub
+# fires no webhook event for linking a PR to an issue via the "Development"
+# sidebar after the PR is already open, so the pull_request-triggered
+# propagate-issue-labels workflow can miss that case. See
+# scripts/reconcile_issue_labels.py for the full rationale.
+
+on:
+  schedule:
+    - cron: '0 */6 * * *'   # every 6 hours
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  reconcile:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Reconcile issue labels across open PRs
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+        run: python3 scripts/reconcile_issue_labels.py
--- a/.github/workflows/reconcile-issue-labels.yml
+++ b/.github/workflows/reconcile-issue-labels.yml
@@ -1,15 +1,21 @@
 name: Reconcile issue labels on open PRs
 
-# Periodically re-applies scripts/propagate_issue_labels.py's label
-# propagation rule to every open pull request. This exists because GitHub
-# fires no webhook event for linking a PR to an issue via the "Development"
-# sidebar after the PR is already open, so the pull_request-triggered
-# propagate-issue-labels workflow can miss that case. See
-# scripts/reconcile_issue_labels.py for the full rationale.
+# On-demand re-application of scripts/propagate_issue_labels.py's label
+# propagation rule to every open pull request. propagate-issue-labels.yml's
+# pull_request triggers cover a sidebar-linked PR whenever it sees further
+# ordinary activity (a push, a reopen, and so on); this workflow exists for
+# the residual case of a PR that gets sidebar-linked and then sees none of
+# those events before merge.
+#
+# Deliberately workflow_dispatch-only rather than a schedule: a periodic
+# cron would poll every open PR indefinitely even when nothing needs
+# reconciling, for a gap that is otherwise already closed by the pokes in
+# propagate-issue-labels.yml. Run this manually (from the Actions tab, or
+# locally with `python3 scripts/reconcile_issue_labels.py`) if you suspect a
+# PR was sidebar-linked without triggering that workflow.
 
 on:
-  schedule:
-    - cron: '0 */6 * * *'   # every 6 hours
+  workflow_dispatch: {}
 
 permissions:
   issues: write
--- a/scripts/propagate_issue_labels.py
+++ b/scripts/propagate_issue_labels.py
@@ -16,7 +16,17 @@ is left alone. A candidate label is also skipped, rather than applied, when
 applying it would cause scripts/enforce_mutually_exclusive_labels.py's
 mutual-exclusion enforcement to remove a label the PR already carries--an
 existing PR label always wins over a conflicting label being copied in from a
-linked issue. See labels_to_propagate() for the exact rule.
+linked issue.
+
+An issue's own orchestration-cycle state labels (PROCESS_STATE_LABELS, e.g.
+`orchestrating`, `verification needed`, `changes done`) are never propagated
+at all, even onto a PR that carries none of their conflicting siblings.
+agents/dev_orchestration.md deliberately lets an issue and its PR diverge on
+these mid-cycle (for example, `orchestrating` is removed from the PR alone,
+not the issue, to clear the "No blocking labels" merge gate while
+orchestration is still active), so blindly copying them from issue to PR
+fights that state machine instead of just adding a convenience label. See
+labels_to_propagate() for the exact rule.
 
 The pull_request trigger for this module's main() (see
 .github/workflows/propagate-issue-labels.yml) covers most sidebar-linked PRs
@@ -33,8 +43,8 @@ Usage:
 Exit code:
     0  the PR has no closing issue references, its closing issues carry no
        labels, every eligible label was applied successfully, or every
-       candidate was skipped due to a mutual-exclusion conflict (skipping is
-       not itself a failure).
+       candidate was skipped due to a mutual-exclusion conflict or excluded
+       as orchestration-cycle state (neither is itself a failure).
     1  required configuration is missing/invalid, or fetching or applying
        labels failed.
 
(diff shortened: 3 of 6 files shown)


aunger (commented):
🤖 Reviewer

Changes requested.

I checked out the branch, ran the new unit tests (all 29 pass) and `scripts/lint.sh --check` on the new files (clean), and read `scripts/enforce_mutually_exclusive_labels.py` to confirm the helper signatures (`find_conflicting_set`, `find_conflicting_prefix`, `labels_to_remove`, `labels_to_remove_by_prefix`, `gh_api`) are used correctly. The core logic is sound: `labels_to_propagate()`'s staged-list approach correctly makes an already-real PR label win over any conflicting candidate, and correctly prevents two candidates from the same run conflicting with each other, which is exactly what the issue asked for. Nice touch that this PR's own `propagate-issue-labels` check run against itself and actually applied `P3`/`orchestrating` from #621 to this PR (see the job log: "Applied labels ['P3', 'orchestrating'] to PR #713") — that's real evidence the happy path works, not just a claim.

One substantive gap, though. The issue's stated reason for using `closingIssuesReferences` instead of regex is explicitly "GitHub recognizes several keyword variants **and cross-references**" (i.e., manually linked issues, not just "Fixes #N" in the body). But the workflow only runs `on: pull_request: types: [opened, edited]`, gated further by `github.event.action == 'opened' || github.event.changes.body`. GitHub does not fire any `pull_request` webhook event when someone links an issue after the fact via the "Development" sidebar — that's a documented platform gap (no `linked_issue`-style event exists today). So for a PR that gets manually linked to an issue *after* it's opened, with no subsequent body edit, this workflow will never re-run, and the issue's labels will silently never propagate — precisely the scenario the GraphQL-based detection was chosen to handle. The detection logic itself is correct once it runs; it's the trigger surface that misses this case.

Please address this (e.g. also trigger on `labeled`/`unlabeled`/`synchronize` as a poke, add a periodic reconciliation, or something else of your choosing), or, if you believe it's out of scope/acceptable, say so explicitly and I'll consider it.

Minor nit (non-blocking): the module docstring's "Exit code" section says `0` covers "the PR has no closing issue references, its closing issues carry no labels, or every eligible label was applied successfully" — it doesn't mention the case where every candidate is skipped due to a conflict (also exit 0, per `test_skip_only_case_returns_0_without_posting`). Worth a phrase for completeness.

FYI, not blocking: `github-advanced-security[bot]`'s Semgrep review comment on line 85 (`dynamic-urllib-use-detected`) is already suppressed with `# nosemgrep` and matches the exact same pattern already in use in `enforce_mutually_exclusive_labels.py`'s `gh_api()`, so it's consistent with existing convention, not a new issue.

aunger:
🤖 Author

Agreed on the substantive gap, and fixed rather than declared out of scope: GitHub genuinely has no webhook event for linking a PR to an issue via the "Development" sidebar, so a `pull_request`-only trigger can permanently miss that case, which is exactly the scenario the GraphQL-based detection was chosen to handle.

Rather than adding fragile "poke" triggers (`labeled`/`unlabeled`/`synchronize` don't actually correlate with the linking action, so they'd only improve the odds, not close the gap), I added a scheduled reconciliation pass:

- Extracted `propagate_to_pr()` out of `propagate_issue_labels.py`'s `main()` so the single-PR propagation logic is reusable.
- Added `scripts/reconcile_issue_labels.py`, which walks every open PR and applies that same function, mirroring this repo's existing `archive_stale_test_failures.py` / `archive-stale-test-failures.yml` scheduled-job pattern.
- Added `.github/workflows/reconcile-issue-labels.yml`, running every 6 hours.
- Added `scripts/test_reconcile_issue_labels.py` (11 tests: pagination, no-op, per-PR delegation, one-failure-doesn't-stop-the-rest) and split the existing `propagate_issue_labels` tests so `TestPropagateToPr` covers the behavioral cases directly instead of only through `main()`.

This means a sidebar-linked PR gets reconciled within 6 hours even with zero further activity, instead of never.

Also addressed the docstring nit: the "Exit code: 0" section now explicitly calls out the skip-only case.

All 43 tests across the two propagation-related test files pass, and `scripts/lint.sh` is clean. Pushed as b68d315.

aunger (commented):
🤖 Reviewer

LGTM. Verified b68d315 rather than trusting the description:

- Checked out the branch and ran the full test suite: `python3 -m unittest scripts.test_propagate_issue_labels scripts.test_reconcile_issue_labels -v` → 43 tests, all pass, matching the claim.
- `scripts/lint.sh --check` on all six changed/added files → clean.
- Confirmed the `propagate_to_pr()` extraction is a clean, behavior-preserving refactor: `main()` now just parses env vars and delegates, and the old `main()`-level tests were correctly moved down to exercise `propagate_to_pr()` directly (`TestPropagateToPr`), which is actually a better unit boundary than before.
- Confirmed `scripts/reconcile_issue_labels.py`'s pagination (`fetch_open_pr_numbers`, looping `state=open&per_page=100&page=N` until an empty page) is correct, and that a failure on one PR doesn't stop reconciliation of the rest (`test_one_failure_does_not_stop_remaining_prs`), which is the right behavior for a periodic sweep.
- Compared `.github/workflows/reconcile-issue-labels.yml` against the existing `archive-stale-test-failures.yml` it's modeled on — same shape (`schedule`/`cron`, `permissions`, checkout + single run step). Consistent with repo convention as claimed.
- Re-pulled PR comments/check-runs on the new commit: no new Semgrep/CodeQL findings, same pre-existing suppressed `dynamic-urllib-use-detected` note (now at line 95, same `# nosemgrep` pattern). `No blocking labels` still fails only because of `orchestrating`/`changes done`, which is expected mid-cycle.
- Docstring nit is fixed: `propagate_issue_labels.py`'s exit-code section now explicitly covers the skip-only case.

This closes the trigger-timing gap I raised: the schedule can't miss the sidebar-link case since it doesn't depend on any webhook event at all, and the explanation for why a "poke" trigger (`labeled`/`synchronize`) wouldn't actually close the gap is correct. Nothing further from me.

aunger:
### Verification plan

**Before merging** (must resolve before PR #713 can be merged):
- [x] #721
- [ ] #724

- None.

aunger:
Find a way to do this without perpetual cron.

aunger:
### Verification summary

Processed the before-merging list from the verification plan (comment 4999377954):

- #721 (Manually verify label-propagation workflows fire correctly on a real PR/issue): **NOT AUTOMATABLE (partial)**.
  Sub-item 1, the `pull_request`-triggered path (`propagate-issue-labels.yml`), is confirmed PASS: independently re-fetched job 87804736365's log from PR #713's first commit, showing `Applied labels ['P3', 'orchestrating'] to PR #713.`
  Sub-item 2, a live out-of-band run of `scripts/reconcile_issue_labels.py`, could not be completed from this verification session: the real GraphQL `closingIssuesReferences` query it depends on returns HTTP 403 through this sandbox's own egress proxy ("only the pinned set of PR-review operations is served"), confirmed with a minimal probe query.
  This is an environment restriction on this verification session, not a defect in the PR; the same GraphQL query already ran successfully on GitHub's own infrastructure for PR #713 (sub-item 1's evidence), and `reconcile_issue_labels.py` calls the identical `propagate_to_pr()` function.
  Real pagination (`fetch_open_pr_numbers`) was confirmed live: it returned the 6 real open PR numbers.
  Full details and the harness script are on #721.
  Leaving #721 open; recommend watching the first scheduled run of `reconcile-issue-labels.yml` after merge, as the issue itself offers as an alternative.

No FAIL results; no bugs found.

aunger:
🤖 Author

Removed the perpetual schedule (75bce6e). Two changes:

1. `propagate-issue-labels.yml` now also runs on `reopened`, `synchronize`, `ready_for_review`, and `review_requested`, in addition to `opened` and body-`edited`. None of these events are actually caused by a sidebar link (GitHub doesn't expose that as an event under any name), but this repo's PRs go through repeated pushes and review rounds as a matter of course, so a sidebar-linked PR gets reconciled the next time any of that ordinary activity happens, with no schedule involved at all. `labeled`/`unlabeled` are deliberately left out: re-running on `unlabeled` would immediately re-apply a label a human had just removed on purpose, which is a real clobbering risk, not just a theoretical one.

2. `reconcile-issue-labels.yml` (the full-sweep script) is now `workflow_dispatch`-only, no `schedule:`. It's for the residual case for a PR that is sidebar-linked and then sees literally none of the events above before merge. That's now available on demand (Actions tab, or `python3 scripts/reconcile_issue_labels.py` locally), the same way `scripts/backfill_labels_by_title.py` already exists in this repo as a manually-run reconciliation tool with no workflow of its own.

Net effect: no indefinite polling, and the practical gap that motivated the reconciliation script in the first place is now closed by real PR activity instead of a timer.

All 43 tests still pass; `scripts/lint.sh` is clean.

aunger (commented):
🤖 Reviewer

Changes requested. This supersedes my prior LGTM: the redesign in 75bce6e surfaced a real problem I hadn't caught before, with live evidence from this very PR.

First, the good part, verified rather than trusted: checked out 75bce6e, ran the full suite (43 tests pass), `scripts/lint.sh --check` on all six files (clean), and confirmed `analyze-actions` passed on the widened workflow YAML. The `if: github.event.action != 'edited' || github.event.changes.body` logic is a correct generalization: it runs unconditionally for the five new non-"edited" types and preserves the body-changed gate for "edited". Dropping the perpetual cron per the owner's comment, in favor of piggybacking on ordinary PR activity plus an on-demand `workflow_dispatch` fallback mirroring `backfill_labels_by_title.py`, is a reasonable trade-off.

Now the problem. `propagate_issue_labels.py` copies every label an issue carries onto its closing PR, with no distinction between ordinary content/category labels (`p1`-`p3`, `ci`, `bug`, ...) and this repo's own process/workflow-state labels: `orchestrate`/`orchestrating`, `verification needed`/`verified`, `changes requested`/`changes done`. Per `agents/dev_orchestration.md`, `orchestrating` is explicitly applied to **both** the issue and the PR when orchestration starts, but is sometimes removed from the **PR alone** mid-cycle (see `labelGateBlock`: "Removing `orchestrating` early is fine... apply this transition to the PR: Remove label: orchestrating") specifically to clear `block-merge-on-blocking-labels.yml`'s "No blocking labels" gate, while the issue keeps carrying it for the rest of the active orchestration. That means the issue and the PR are *supposed* to diverge on this label mid-cycle, by design, not by accident.

I watched this collide, live, on this PR:
- After my round-2 LGTM, PR #713 carried `["P3", "orchestrating", "changes done"]`.
- Before this push, it had become `["P3", "changes done"]` ("orchestrating" removed from the PR, matching the `labelGateBlock` pattern above; issue #621 still carries `orchestrating`, confirmed via `issue_read get_labels`).
- This push (75bce6e) is an ordinary `synchronize` event, now one of the widened trigger types. The `propagate-issue-labels` job ran and its log (run 29559097278, job 87817516274) reads: `Applied labels ['ci', 'orchestrating'] to PR #713.` It put `orchestrating` right back.
- Moments later, `pull_request_read get` shows the PR back down to `["P3", "changes done"]` again, i.e. something removed `orchestrating`/`ci` right back off.

That's an active oscillation between this automation and the rest of the repo's own label-based state machine, not a hypothetical. Since `orchestrating` is one of `block-merge-on-blocking-labels.yml`'s `BLOCKING_LABELS`, this can directly re-obstruct the merge gate the Orchestrator just worked to clear, for any PR whose linked issue is still under active orchestration, which by definition is the entire active lifetime of the issue.

This is exactly the "real clobbering risk, not just theoretical" you called out when excluding `labeled`/`unlabeled` from the trigger types, but `synchronize` (an ordinary push, unavoidable in this repo's own multi-round dev cycle, as this PR itself demonstrates) reintroduces the identical risk for any label an artifact's own workflow deliberately keeps asymmetric between issue and PR.

Please address this. I won't design the fix for you, but at minimum it needs to stop `orchestrating` (and probably its `enforce_mutually_exclusive_labels.py`-defined siblings, since they're documented as the same kind of process-state label) from being blindly copied from issue to PR.

aunger:
🤖 Author

Good catch, and thanks for chasing it down to a live log rather than just theorizing. Fixed in 144f4dd.

Added `PROCESS_STATE_LABELS` (`orchestrate`/`orchestrating`, `verification needed`/`verified`, `changes requested`/`changes done`) and excluded all of them from propagation entirely, not just when the PR happens to already carry a conflicting sibling. That's the actual bug: the mutual-exclusion guard only protects a label the PR *already has*; `orchestrate`/`orchestrating` was absent from the PR outright (correctly removed by `labelGateBlock`), so there was nothing for that guard to catch. `p1`-`p3`, `ci`, `bug`, and other ordinary content labels are unaffected and still propagate exactly as before, including "P3" from #621 onto this PR, which is still expected.

`labels_to_propagate()` now returns `(to_add, skipped, excluded)`, with `excluded` covering this new category and logged with its own distinct reason in `propagate_to_pr()`. Added a dedicated test, `test_does_not_reapply_orchestrating_the_orchestrator_just_removed`, that reproduces the exact scenario you observed live (issue carries `orchestrating`, PR carries none of it, PR still gets other eligible labels), plus coverage for every label in `PROCESS_STATE_LABELS`, case-insensitivity, and that exclusion and skip-due-to-conflict are reported distinctly rather than conflated.

50 tests pass (7 new); `scripts/lint.sh` is clean. Left the widened `synchronize`/etc. trigger surface as-is, since you confirmed that part is a reasonable trade-off and the bug was in the copying logic, not the trigger surface.

aunger (commented):
🤖 Reviewer

LGTM. Verified 144f4dd rather than trusting the description, and specifically re-tested the exact live scenario I flagged:

- Checked out 144f4dd, ran the full suite: 50 tests pass (confirmed the +7 claim), including `test_does_not_reapply_orchestrating_the_orchestrator_just_removed`, which reproduces my exact observation (issue carries `orchestrating`, PR carries none of it, PR still gets other eligible labels like `p1`).
- `scripts/lint.sh --check` on all six files: clean.
- Read the diff: `PROCESS_STATE_LABELS` correctly covers all three `enforce_mutually_exclusive_labels.py` process pairs (`orchestrate`/`orchestrating`, `verification needed`/`verified`, `changes requested`/`changes done`) while correctly leaving `p1`-`p3` untouched, since those are legitimate shared triage labels, not asymmetric per-artifact state. The exclusion check runs before the mutual-exclusion logic and is reported in its own `excluded` list, distinct from `skipped`, so the two reasons aren't conflated in the log or the return value.
- Best confirmation: re-pulled the live check runs on this commit. Job 87818739498 (`propagate-issue-labels`, triggered by this very push against this very PR) logs: `Excluding orchestration-cycle labels ['orchestrating'] on PR #713: the issue and its PR are allowed to carry these differently while orchestration is active (see agents/dev_orchestration.md).` followed by `Applied labels ['ci'] to PR #713.` That's the exact bug scenario, fixed, observed live on the same PR that exposed it. I also re-checked `No blocking labels` (job 87818803182): it now fails only on `changes done` (`LABELS: ["P3", "changes done"]`), not `orchestrating`, confirming the merge-gate fight is gone.

Nothing further from me.

aunger:
### Verification summary (2nd pass)

Processed the before-merging list from the verification plan (comment 4999377954):

- #721 — already closed prior to this run (previous verification pass: `pull_request` trigger path confirmed PASS live; the reconcile-script sub-item was NOT AUTOMATABLE from that session and was closed on the recommendation to watch the (then-existing) scheduled run). No further action taken here since it's resolved and the checkbox already reflects that.
- #724 (Manually verify `scripts/reconcile_issue_labels.py` against the live repo now that its scheduled trigger was removed): **NOT AUTOMATABLE**.
  Directly re-ran the script's real functions (`fetch_open_pr_numbers`, `propagate_to_pr`) against the live repo with a real `GITHUB_TOKEN`: REST pagination works (`[713, 654]`, the real open PRs), but the GraphQL `closingIssuesReferences` query returns HTTP 403 from this session's own egress proxy for every PR — the identical restriction #721 hit.
  Also tried the issue's suggested alternative, dispatching `reconcile-issue-labels.yml` via the Actions API — blocked with `403 Resource not accessible by integration`. A control-test dispatch of an unrelated, already-`main`-branch `workflow_dispatch` workflow (`simulate-preflight-failure.yml`) got the identical error, confirming this session's credentials aren't scoped for Actions dispatch at all (not a branch-presence issue specific to the new workflow).
  Full evidence and a recommendation (run locally with a personal token, or dispatch from the Actions tab in the browser) are on #724, which is left open.
  Supporting, non-live check: re-ran both test files at the PR's current head (207df40) — 50 tests pass.

No FAIL results; no bugs found in this pass.

Issue #714, opened 2026-07-17
Harden the Gradle build with dependency verification metadata


aunger:
Split out from #699, which bundled two independent hardening steps: pip `--require-hashes` for the Python lint tools, and Gradle dependency verification for the Android build.

The pip half is handled in the PR for #699.
This issue tracks the remaining half.

## Goal

Add Gradle dependency verification (`gradle/verification-metadata.xml`) so every plugin, wrapper, and dependency artifact resolved by the build is checked against a pinned SHA-256, matching the &#34;verify what you download&#34; posture already applied to ktlint (SHA-256 verified) and now the Python lint tools (pip `--require-hashes`).

## Why it was not done in the #699 PR

`gradle/verification-metadata.xml` is high blast radius: once present, *every* Gradle invocation fails if any resolved artifact is missing from or mismatched against the file.
Generating a complete, correct file requires the full Android toolchain (Android SDK, AGP, Kotlin, Compose, AndroidX, and the instrumented-test dependency graph) and a run that resolves every configuration; it must then be validated by running the real CI build end to end.
The #699 working environment had only a JDK and no Android SDK, so the file could neither be generated completely nor verified, and shipping a partial one would break all builds.

## Suggested approach

- Generate the metadata in an environment with the full Android toolchain, resolving all configurations (for example `./gradlew --write-verification-metadata sha256` against tasks that resolve the app, both `e2e-mock-*` modules, and the instrumented-test classpaths), not just `help`.
- Decide the verification scope (`sha256` checksums; optionally `pgp` for signed artifacts) and whether to verify the wrapper JAR too.
- Confirm the complete CI build (`build.yml`, including instrumented/E2E steps) passes with the metadata in place before merging, since that is the only way to catch a configuration whose dependencies were missed during generation.
- Document how to regenerate the file when dependencies change, so a version bump does not leave the build failing verification.

---

### Recommendation: proceed, with a tighter scope than the issue text implies

The change is worth doing.
It is consistent with the repo's existing, deliberate posture (ktlint SHA-256 in `scripts/install-ktlint.sh`, pip `--require-hashes` from #699), it is a first-party Gradle feature, and the maintenance churn is low here because there is no Dependabot/Renovate in this repo (`.github/` has neither), so the dependency graph only moves on manual, reviewed version bumps.
Two things in the issue's suggested approach should be scoped down or corrected before implementing:

[...]

Issue #699, opened 2026-07-16
Harden lint tool installs with pip --require-hashes and Gradle dependency verification


aunger:
**Blocked by: #698** (implementation-sequencing only; this does not change the merge-blocking status noted below).
Reason (pip half): #698 adds new `pip install -r scripts/requirements-lint.txt` sites in the new `.github/workflows/lint.yml` (its python-ruff, markdown-mdformat, and hygiene jobs), which are the highest-value place to enforce `--require-hashes`.
Those CI install sites do not exist until #698 lands, and once #699 adds hashes to `requirements-lint.txt` pip switches every `pip install -r` into hash-checking mode, so #698's CI installs must already exist and be authored against a hash-pinned file.
Implement #698 first so #699 can harden the complete install surface (`.claude/hooks/session-start.sh` STEP 3b plus every `lint.yml` job) in one coherent pass.
The Gradle dependency-verification half of #699 is independent of the other issues in this set.
(Added 2026-07-16 by cross-issue dependency-sequencing analysis.)

---

Source: issue #667 description, "Risks and open decisions" section:
https://github.com/aunger/gallery-button-for-pixel-camera/issues/667

> Version pinning policy: pins match the "pin the rev" spirit already applied in #664/#666; ruff tracks #673. `pip install --require-hashes` and Gradle dependency verification are available hardening steps, left as optional follow-ups to keep this change focused.

This is a follow-on item and does not block merging PR #696: https://github.com/aunger/gallery-button-for-pixel-camera/pull/696

aunger:
🤖 Author

This issue bundled two independent hardening steps from #667's "Risks and open decisions": pip `--require-hashes` for the Python lint tools, and Gradle dependency verification for the Android build.

The **pip `--require-hashes` half** is addressed in PR #715. That PR splits `scripts/requirements-lint.txt` into a readable source of top-level pins (`scripts/requirements-lint.in`) and a fully resolved, hash-pinned lock, then adds `--require-hashes` to every `pip install -r scripts/requirements-lint.txt` site (`.claude/hooks/session-start.sh` STEP 3b, the three `lint.yml` jobs, and the `build.yml` shell-test lint venv).

[...]

PR #715, opened 2026-07-17
Install the Python lint tools from a hash-pinned lock with pip --require-hashes

aunger:
Addresses the pip half of #699.

Issue #699 bundled two independent hardening steps from #667's "Risks and open decisions": pip `--require-hashes` for the Python lint tools, and Gradle dependency verification for the Android build.
This PR does the pip half.
The Gradle half is split out into #714 (see "Scope" below).

## What changed

- Split `scripts/requirements-lint.txt` into a human-readable source of top-level pins (`scripts/requirements-lint.in`) and a fully resolved lock (`scripts/requirements-lint.txt`) in which every package, top-level and transitive, is pinned to an exact version and a SHA-256 hash.
  The lock is generated with `uv pip compile --universal --python-version 3.10 --generate-hashes`, so it is a superset covering every interpreter the install sites use (system `python3` in the session-start hook, and setup-python's `3.x` in CI).
  Marker-gated entries such as `tomli` on Python < 3.11 resolve correctly wherever pip runs it.
- Added `--require-hashes` to every `pip install -r scripts/requirements-lint.txt` site:
  - `.claude/hooks/session-start.sh` STEP 3b
  - the `python-ruff`, `markdown-mdformat`, and `hygiene` jobs in `.github/workflows/lint.yml`
  - the "Install lint tools for shell tests" step in `.github/workflows/build.yml`
- Updated `.claude/environment.md` and the hook/lock comments to describe the lock, the `--require-hashes` install, and how to regenerate the lock.

With `--require-hashes`, pip refuses any wheel whose downloaded artifact does not match a hash in the lock, so a compromised or substituted package on PyPI, or an intercepted download, can no longer slip into the lint toolchain.
This extends the "verify what you download" posture already applied to ktlint (SHA-256 verified).

## Verification

- Installed the exact lock into a fresh venv with `pip install --require-hashes -r scripts/requirements-lint.txt` (exit 0); all eight console scripts (ruff, mdformat, and the six pre-commit-hooks checks) are present.
- Confirmed `tomli` is marker-gated (`python_full_version < '3.11'`) and skipped on Python 3.11.
- Ran the repo's own linters over the whole tree via `scripts/lint.sh --check --only {ruff,mdformat,hygiene} --all`: all pass.
- Ran `scripts/test_lint_hook.sh`: 20 passed, 0 failed (the ktlint case skips cleanly, no JDK/ktlint in the sandbox).
- `bash -n` on the edited `session-start.sh`.

## Scope

The Gradle dependency-verification half of #699 is tracked separately in #714.
It is high blast radius (once `gradle/verification-metadata.xml` exists, every Gradle invocation fails on any missing or mismatched artifact) and needs the full Android toolchain to generate completely and the real CI build to validate.
This PR's environment had only a JDK and no Android SDK, so that half could not be generated or verified here; shipping a partial metadata file would break all builds.
Because #699 itself notes the Gradle half is independent, it is deferred to #714 rather than shipped unverified.


aunger (commented):
🤖 Reviewer

Verified this thoroughly rather than taking the PR description's verification section on faith, since a hash-pinning PR is exactly the kind of change where a wrong hash silently defeats the point.

What I checked:

- **Every hash in `scripts/requirements-lint.txt` against PyPI's real JSON API metadata**, programmatically, package by package (`markdown-it-py`, `mdformat`, `mdformat-frontmatter`, `mdformat-gfm`, `mdit-py-plugins`, `mdurl`, `pre-commit-hooks`, `ruamel-yaml`, `ruff`, `tomli`, `wcwidth`). All 11 packages: lock hash count == PyPI hash count, zero missing, zero extra. Every single SHA-256 is genuine.
- **Dependency closure completeness**: walked `requires_dist` for the 5 top-level tools and their transitive deps by hand against PyPI metadata. The 11-package lock is exactly the correct closure — nothing missing, nothing extraneous.
- **Actual install**: ran `pip install --require-hashes -r scripts/requirements-lint.txt` in a clean venv (Python 3.11). Succeeded, all 8 console scripts present (ruff, mdformat, and the 6 pre-commit-hooks checks). `tomli` correctly skipped via its `python_full_version < '3.11'` marker.
- **Reproduced the PR's own verification claims** on a checked-out worktree of this branch: `scripts/lint.sh --check --only {ruff,mdformat,hygiene} --all` all exit 0; `scripts/test_lint_hook.sh` → 20 passed, 0 failed (ktlint case skips cleanly, matching the PR description exactly); `bash -n` on `session-start.sh` passes; both edited workflow YAML files parse.
- **Coverage of install sites**: grepped for every `pip install ... requirements-lint.txt` occurrence — `session-start.sh` STEP 3b, all three `lint.yml` jobs (`python-ruff`, `markdown-mdformat`, `hygiene`), and `build.yml`'s "Install lint tools for shell tests" step. All five got `--require-hashes`. None missed.
- **Scope split vs #714**: confirmed #698 (which #699 said this PR was sequenced behind) is closed, and `lint.yml` already exists on `main` with the exact job names this PR touches, so the sequencing precondition is genuinely satisfied. #714 (Gradle half) is a real, well-justified issue, not just a scope dodge.

Everything I could independently verify checks out exactly as claimed. I found no correctness problems.

Two non-blocking nits:

1. `scripts/requirements-lint.in`'s regeneration comment and `.claude/environment.md` both tell a future contributor to run `uv pip compile ...`, but `uv` isn't installed anywhere in this repo's dev environment (session-start.sh, environment.md's tool table) and there's no mention of how to get it. Whoever next edits a pin will hit a bare "command not found" with no pointer. Worth a one-line "install with `pip install uv`" note wherever this is documented.
2. Not a defect, just noting scope for the record: `scripts/requirements.txt` (defusedxml/requests/PyYAML, installed unhashed in `build.yml` at two sites) and `semgrep.yml`'s unversioned `pip install semgrep` remain unhardened. That's legitimately outside #699's stated scope ("lint tool installs"), so I'm not asking for it here — flagging only in case a follow-up issue is wanted for full pip-install coverage.

LGTM.

aunger:
Verification planner: file issues for the "Two non-blocking nits"

aunger:
### Verification plan

**Before merging** (must resolve before PR #715 can be merged):
None.

- [ ] #722
- [ ] #723

PR #717, opened 2026-07-17
Add update_gh_labels.sh: delta GitHub label add/remove

### LINKED ISSUE #710: Orchestrator clobbers labels
To  add and remove labels, the orchestrator uses `mcp__github__issue_write`, which only accepts a *replacement set* of labels.

If the orchestrator thinks it knows the issue’s current labels, but another agent, automation, or human has updated them, the orchestrator will unknowingly overwrite the others’ update.

When running on Claude Code for Web, this is the orchestrator’s **only tool** for changing GitHub labels, because recently and infuriatingly, Anthropic forbade those agents from using the REST API.

An imperfect workaround would be to instruct Orchestrator to read before write. This would certainly improve the situation, but only narrows the race condition’s window.

Another potential solution is to provide another tool to Claude, one that accepts label deltas (add, remove). This can wrap the GitHub REST API, which I believe has such an operation, and follow this pattern:
- #658

aunger:
Fixes #710.

`mcp__github__issue_write`'s `labels` field is a replacement set, not a delta: it overwrites the issue's or PR's entire label list. When the Orchestrator (or any agent) uses it to move an issue or PR through a label transition, it silently discards any label change made by another agent, a workflow, or a human since the labels were last read. This is the only label-write tool available to agents on Claude Code for Web, which cannot call the GitHub REST API directly.

This PR adds `scripts/update_gh_labels.sh`, which calls GitHub's native delta label endpoints via `curl` instead:

- `POST /issues/{n}/labels` adds labels without touching the rest of the set.
- `DELETE /issues/{n}/labels/{name}` removes one label without touching the rest of the set (a 404, meaning the label was already absent, is treated as success, not a failure).

It follows the REST-via-curl pattern established by `scripts/delete_gh_comment.sh` (#658), which the issue pointed to as a model.

Adding the tool alone would not fix the clobbering, since the Orchestrator's process document never named a tool for its label transitions and would keep defaulting to `mcp__github__issue_write`. `agents/dev_orchestration.md` now has an "Applying label transitions" section directing every "Remove label" / "Add label" transition table in the document to `scripts/update_gh_labels.sh` instead.

Test plan: `scripts/test_update_gh_labels.sh` covers the script end-to-end (argument validation, add, remove, combined add+remove ordering, percent-encoding of label names, idempotent 404-on-remove, and the 401/403/404/500 error paths) against a fake `curl`, and runs automatically in CI via the `scripts/test_*.sh` sweep in `build.yml`. No manual verification steps remain.

### DIFF
--- a/.claude/environment.md
+++ b/.claude/environment.md
@@ -108,9 +108,31 @@ so the `labels` field is never sent and all existing labels remain.
 
 **Implication:** There is no way to remove *all* labels from an issue using
 `mcp__github__issue_write` alone. To drop N−1 labels, set `labels` to the one label
-you want to keep. The last remaining label cannot be removed via this tool, and the
-`GITHUB_TOKEN` environment variable also cannot help; it is read-only for the Issues
-API (write attempts return 403).
+you want to keep. The last remaining label cannot be removed via this tool, but
+`GITHUB_TOKEN` can remove it directly against the labels sub-resource endpoint; see
+"`GITHUB_TOKEN` can write labels" below, which supersedes the read-only finding this
+section used to state here.
+
+### `GITHUB_TOKEN` can write labels (supersedes the 2026-05 finding above)
+
+The 2026-05 empirical test above found that `GITHUB_TOKEN` returned 403 on `DELETE`
+and `PATCH` against the Issues API, and concluded the token was read-only for
+issue/PR writes. That conclusion does not hold for the labels sub-resource
+endpoints.
+
+**Re-verified 2026-07 (issue #710 / PR #717):** `POST /issues/{n}/labels` (add) and
+`DELETE /issues/{n}/labels/{name}` (remove) both succeed with `GITHUB_TOKEN`. A
+round trip on issue #710's own `P3` label (`scripts/update_gh_labels.sh ... --remove P3`, then `--add P3`) was confirmed by a follow-up `get_labels` read after each
+call. `scripts/update_gh_labels.sh` (issue #710) relies on this, and is the
+supported way to add or remove specific labels without the replace-all behavior of
+`mcp__github__issue_write` (see "Applying label transitions" in
+`dev_orchestration.md`).
+
+Why the 2026-05 finding differed is unconfirmed: either the fine-grained PAT's
+permissions changed since then, or the original 403 was specific to `PATCH /issues/{n}` (updating the issue resource itself) rather than the labels
+sub-resource endpoints this script uses. Treat the write-permission boundary as the
+labels sub-resource only; do not assume other Issues/PR API writes succeed with
+`GITHUB_TOKEN` without testing them individually.
 
 ### Always verify writes with a follow-up read
 
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -94,6 +94,10 @@ scripts/update_gh_labels.sh {owner} {repo} {issue-or-PR number} --remove orchest
 Run it once per artifact a transition's note tells you to apply to (issue, PR, or both).
 See the script's own `--help` text for full usage and the required `GITHUB_TOKEN` environment variable.
 
+If `scripts/update_gh_labels.sh` exits non-zero, the transition did not fully apply--do not treat it as done.
+Retry the same call once, since a non-2xx GitHub response can be transient.
+If it still fails, do not proceed with the transition; escalate using the Orchestrator escalation/abort line below, with the script's error output as the reason token.
+
 ## Starting to orchestrate a PR
 
 When you begin orchestrating a PR (the first thing you do once you have entered the Orchestrator role for a given issue and its PR), apply this transition to **both the issue and the PR**:
--- a/agents/dev_orchestration.md
+++ b/agents/dev_orchestration.md
@@ -78,6 +78,22 @@ The Orchestrator is not a Reviewer or a Programmer.
 
 - See `inaugurate.md` for the full protocol when starting fresh work.
 
+## Applying label transitions
+
+Apply every "Remove label" / "Add label" transition table in this document with `scripts/update_gh_labels.sh`, not `mcp__github__issue_write`.
+That MCP tool's `labels` field is a replacement set: it overwrites the issue's or PR's entire label list, so it silently discards any label another agent, a workflow, or a human applied since you last read the labels (issue #710).
+`scripts/update_gh_labels.sh` instead calls GitHub's delta label endpoints, adding and removing only the specific labels you name, so a transition never touches any label outside its own row.
+
+Run one call per transition row, passing every "Remove label" entry as a `--remove` flag and every "Add label" entry as a `--add` flag.
+For example, the "Starting to orchestrate a PR" transition below becomes:
+
+```text
+scripts/update_gh_labels.sh {owner} {repo} {issue-or-PR number} --remove orchestrate --add orchestrating
+```
+
+Run it once per artifact a transition's note tells you to apply to (issue, PR, or both).
+See the script's own `--help` text for full usage and the required `GITHUB_TOKEN` environment variable.
+
 ## Starting to orchestrate a PR
 
 When you begin orchestrating a PR (the first thing you do once you have entered the Orchestrator role for a given issue and its PR), apply this transition to **both the issue and the PR**:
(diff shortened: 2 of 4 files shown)


aunger (commented):
🤖 Reviewer

## Overview

Adds `scripts/update_gh_labels.sh` (delta add/remove of GitHub issue/PR labels via the REST API), `scripts/test_update_gh_labels.sh` (a fake-curl unit suite), and a new "Applying label transitions" section in `agents/dev_orchestration.md` directing the Orchestrator to use the script instead of `mcp__github__issue_write` for every transition table. This follows the `delete_gh_comment.sh` (#658) precedent closely and correctly.

I ran `bash scripts/test_update_gh_labels.sh` locally: 48/48 assertions pass. I also checked the documented HTTP status codes for both endpoints against GitHub's REST docs; the script's 200/404/403/401/default handling matches the real API (`DELETE .../labels/{name}` returns 200 on success and 404 if the label is already absent; `POST .../labels` returns 200 on success), so the status-code branching is correct, not just internally consistent with the mocks.

## A documentation contradiction worth resolving before merge

This whole PR rests on the premise that `GITHUB_TOKEN` in the container can write to the GitHub Issues API via `curl`. But `.claude/environment.md`'s "GitHub MCP tool quirks" section currently says:

> The last remaining label cannot be removed via this tool, and the `GITHUB_TOKEN` environment variable also cannot help; it is read-only for the Issues API (write attempts return 403).

That's not stale filler text - it was added deliberately in commit `49634e7` ("Refine MCP quirks docs: GITHUB_TOKEN is read-only for Issues API"), whose message says: "Add finding from empirical test: DELETE and PATCH on the Issues API return 403 with GITHUB_TOKEN, confirming it has no write access to issue labels." That is a direct, specific contradiction of this PR's entire mechanism (`update_gh_labels.sh` does exactly a DELETE and a POST against `/issues/{n}/labels...` using `GITHUB_TOKEN`), and neither the PR description's test plan ("No manual verification steps remain") nor the diff addresses it.

So I tested it live rather than just flagging the discrepancy on paper. Using the actual script from this PR's branch and this container's real `$GITHUB_TOKEN`, I round-tripped issue #710's own `P3` label:

- `scripts/update_gh_labels.sh aunger gallery-button-for-pixel-camera 710 --remove P3` → `Removed label 'P3' ...`, exit 0. Confirmed via a follow-up `get_labels` read: `P3` was actually gone.
- `scripts/update_gh_labels.sh aunger gallery-button-for-pixel-camera 710 --add P3` → `Added label(s) [P3] ...`, exit 0. Confirmed restored via another `get_labels` read.

So, good news: the mechanism does work today, live, in this environment. It looks like the fine-grained PAT's permissions have changed since the May 30 finding (or the original test's 403 was specific to `PATCH /issues/{n}` rather than the labels sub-resource endpoints this script uses) - I can't tell which from here, and it doesn't really matter for this PR.

What does matter: `.claude/environment.md` still asserts the opposite, and this PR adds instructions to `dev_orchestration.md` that only make sense if that assertion is wrong. Left as-is, a future agent reading `environment.md` gets directly contradictory guidance about whether `GITHUB_TOKEN` can write labels - exactly the kind of "small error that causes misunderstandings down the road" this process is supposed to catch, and ironically in the same trust-in-tooling vein as the clobbering bug this issue is about. Please update the "Implication" paragraph in `environment.md`'s MCP quirks section to reflect the current, re-verified behavior (labels sub-resource writes via `GITHUB_TOKEN` succeed) rather than leaving the stale 403 claim standing uncontested next to new instructions that depend on it being false.

## Smaller points

- `agents/dev_orchestration.md`'s new section and the rest of the Monitor-loop routing are otherwise very prescriptive about failure handling (retries, escalation branches, etc.), but nothing says what the Orchestrator should do if `update_gh_labels.sh` itself exits non-zero (a transient 500, a real 403 if permissions regress, etc.). Previously this was moot because the MCP tool always returned a success-looking response regardless of outcome (per `environment.md`); now that label writes can genuinely and audibly fail, silence on this is a real gap, not a hypothetical one. Worth at least one sentence.
- Everything else - argument validation, percent-encoding of label names, ordering (remove-before-add), the deliberate no-early-abort-on-partial-failure behavior, idempotent 404-on-remove - is well-reasoned and well-tested. No correctness issues found there.

## Verdict

`Changes requested` - specifically for the `environment.md` contradiction. The script itself works (I verified this live, not just via the mocked test suite), so this shouldn't take another full round; it's a documentation fix plus, ideally, a line on Orchestrator behavior when the script fails.

aunger:
🤖 Author

Thanks for testing this live rather than just flagging the discrepancy on paper. Convinced; pushed both changes:

- `.claude/environment.md`: replaced the stale "read-only for the Issues API" claim in the labels-implication note with a new "`GITHUB_TOKEN` can write labels" subsection recording your re-verified finding (the P3 round trip, with the caveat that the write-permission boundary is the labels sub-resource specifically, not a blanket claim about the Issues/PR API).
- `agents/dev_orchestration.md`: added guidance for a non-zero exit from `update_gh_labels.sh` (retry once, then escalate via the existing Orchestrator escalation/abort line rather than silently treating the transition as applied).

Both commits are on the branch; lint and the 48-case test suite still pass.

aunger (commented):
🤖 Reviewer

Both points from my last review are addressed in `7b21f66`:

- `.claude/environment.md` now correctly documents that `GITHUB_TOKEN` can write to the labels sub-resource endpoints, superseding the stale 2026-05 finding, and is appropriately careful not to overclaim beyond that sub-resource. No more contradiction with `dev_orchestration.md`'s new instructions.
- `dev_orchestration.md` now tells the Orchestrator what to do on a non-zero exit from `update_gh_labels.sh` (retry once, then escalate via the existing escalation line). I checked this against the script's actual behavior: because remove-404 and add-of-already-present-label are both treated as success, retrying the *same full command* is safe regardless of which leg failed the first time, so this guidance is not just plausible but correct.

Re-ran `bash scripts/test_update_gh_labels.sh` against the latest commit: 48/48 pass. CI is green. No new issues found.

`LGTM`

aunger:
If the script isn't working, it's better to fall back to MCP. (honestly, labels aren't even worth escalating for; it's better to get the work finished)

aunger:
🤖 Author

Good point, fixed in `4e60202`. The failure-handling guidance in `agents/dev_orchestration.md` now falls back to `mcp__github__issue_write` (read the current labels, apply the transition's Remove/Add columns locally, write the result) instead of escalating when `update_gh_labels.sh` fails twice. Escalating the whole cycle over a label was disproportionate; this treats the replace-all race as an acceptable one-off cost only in that rare failure path, not the normal one.

aunger (commented):
🤖 Reviewer

Re-reviewed after `4e60202`, which implements the owner's direct feedback (fall back to MCP rather than escalate on a label-script failure).

Checked the new `agents/dev_orchestration.md` text for correctness and consistency, not just intent:

- The escalation path it replaces (`If it still fails, do not proceed with the transition; escalate...`) is fully removed, no stale reference to it remains, and no other part of the document (the `labelGateBlock` "unexpected state" escalation at line ~312, the `Cannot work`/Monitor-timeout/silentVanish escalations, "When to abort") depends on or contradicts the new "never escalate solely over a label transition" rule. Those are all orthogonal failure modes (Reviewer verdicts, CI infra, state-machine invariants), not `update_gh_labels.sh` failures.
- The fallback procedure itself is correctly specified: read current labels via `issue_read`/`get_labels`, apply the transition's Remove/Add columns locally, write the result via `mcp__github__issue_write`. That's the right sequence, and it happens to match the documented "read before write" workaround for the MCP tool's "token expired" quirk in `environment.md`, for whatever that's worth.
- Retry-once-then-fallback still composes safely with the script's idempotent design (double-checked this last round): a retry of the full command after a partial failure never double-applies anything.
- No prose-style regressions (re-checked for em/en-dashes and curly quotes; none introduced).
- Scripts are unchanged in this commit; re-ran `bash scripts/test_update_gh_labels.sh` on the latest branch head anyway as a sanity check: 48/48 still pass.

This is a product-direction call from the repo owner (accept the old replace-all race as a rare, bounded fallback rather than blocking automation over a label), not something for me to re-litigate; my job here was to confirm the text implementing it is accurate and doesn't leave the document in an inconsistent state, and it doesn't.

`LGTM`

Issue #707, opened 2026-07-16
ci_monitor.py latches onto a stale check-run when a named check re-runs multiple times on the same commit


aunger:
## Summary

While orchestrating PR #706 (issue #698), `scripts/ci_monitor/ci_monitor.py --pr 706` repeatedly reported a `Blocked` terminal citing the `No blocking labels` check-run, even after the blocking label had been removed and a later run of the same check-run name had succeeded.

## Background

PR #706's labels were changed twice during the review cycle: `changes done` was added then removed, and `orchestrating` was removed from the linked issue #698. Each PR-side label add/remove re-triggers `.github/workflows/block-merge-on-blocking-labels.yml` (it listens for `labeled`/`unlabeled` among other `pull_request` events), so the same check-run name, `No blocking labels`, ended up with three separate check-run entries against the same head commit (`76ec7a90bc96731a63ac2fb09001e715ba390469`):

```text
87686988990 No blocking labels completed success 2026-07-16T16:57:43Z 2026-07-16T16:57:47Z
87687158072 No blocking labels completed failure 2026-07-16T16:58:25Z 2026-07-16T16:58:29Z
87688242514 No blocking labels completed success 2026-07-16T17:02:58Z 2026-07-16T17:03:01Z
```

(Queried directly against `/repos/aunger/gallery-button-for-pixel-camera/commits/{sha}/check-runs`.)

The middle one (`failure`, from when `changes done` was briefly applied) is stale; the most recent one (`success`, after all blocking labels were removed) is authoritative. GitHub's own `mergeable_state` for the PR agreed it was authoritative, reporting `clean`.

## What ci_monitor.py reported

Three separate invocations of `python3 scripts/ci_monitor/ci_monitor.py --pr 706`, run several minutes apart (the last one run well after the final, successful check-run completed), all produced the same `Blocked` terminal citing the stale `failure` entry:

**Invocation 1** (right after the `orchestrating` label was removed from issue #698):

```text
PR#706: summary
PR#706:   dev-build ..................... skipped
PR#706:   file-issues ................... success
PR#706:   link-pr-to-ci-summary ......... success
PR#706:   No blocking labels ............ success
PR#706:   report-skipped ................ success
PR#706:   enforce-exclusive-labels ...... success
PR#706:   No blocking labels ............ failure   [BLOCKING] [label gate]
PR#706:   CodeQL ........................ success
PR#706:   Semgrep OSS ................... success
[...]

PR #716, opened 2026-07-17
Collapse same-named check runs to the latest in ci_monitor

aunger:
Fixes #707.

## Problem

A single named check can accumulate several check-run entries against one commit when its workflow re-runs. In issue #707, each PR-side label add/remove re-triggered `block-merge-on-blocking-labels.yml`, so the `No blocking labels` gate ended up with three runs on the same head commit: `success`, then a stale `failure` (from a blocking label that was briefly applied), then `success` again.

`ci_monitor.py` iterated *every* check run, so the stale `failure` outvoted the authoritative later `success`: the tool reported `Blocked by: No blocking labels [label gate]` even though GitHub's own `mergeable_state` for the same commit was `clean`. The per-check summary also listed the name three times, and the stale run's job was tracked for diagnostics.

## Fix

Add `latest_check_runs()`, which collapses the `/commits/{sha}/check-runs` payload so each distinct check-run *name* keeps only its most recent run. Recency is judged by `started_at` (ISO 8601, lexically sortable), then the numeric check-run `id` as a monotonic tiebreak; the surviving run keeps the position of that name's first appearance, so the summary's row order is otherwise unchanged.

The collapsed payload is then used for the verdict (`parse_check_result`), the summary (`parse_check_summary`), and the Actions run/job target discovery (`parse_actions_targets` via `poll_signals`), keeping all three consistent with each other and with `mergeable_state`. Runs with no usable `name` pass through unchanged (each kept), so the unnamed-Actions-run fixtures other paths rely on are untouched, and GitHub always names its own check runs anyway. `--run-id` mode reads a single run object and is unaffected.

This mirrors what GitHub's `mergeable_state` does: judge a required check by its latest run per name. It also stops the stale run's job from being polled for step/FAIL diagnostics.

The issue also floated surfacing Run IDs / Run Attempts and grouping output by Run ID. Collapsing to the latest run per name addresses the reported defect (the wrong verdict) directly and matches GitHub's own dedup semantics, rather than leaving the stale run visible for a human to disambiguate; I went with the collapse.

## Tests

Added to `scripts/test_ci_monitor.py`:

- `(ay)` unit coverage of `latest_check_runs`: the exact three-run shape from the issue reads `Blocked` raw and `all_passed` after collapse; `started_at` is primary and `id` is the tiebreak; distinct names and unnamed runs are preserved in order; only the latest run is a diagnostic target; empty payloads are handled.
[...]

PR #726, opened 2026-07-17
Rank a queued re-run above an older same-named completion in ci_monitor

### LINKED ISSUE #719: Harden ci_monitor non-PR modes against a premature Clear when a re-run has null started_at
## Background

Follow-on from the review of PR #716 (which fixes #707).
Source: reviewer comment https://github.com/aunger/gallery-button-for-pixel-camera/pull/716#issuecomment-4999016257 (non-blocking observation 1).

## Problem

PR #716 adds `latest_check_runs()`, collapsing same-named check runs to the most recent per name.
`_check_run_recency_key` maps a missing `started_at` to `""`, which sorts oldest.
A required check that has an old `completed`/`success` run plus a freshly-queued re-run (GitHub leaves `started_at: null` until a run starts) would collapse to the old `success`, so `parse_check_result` reads `all_passed` instead of `in_progress` during that window.

In `--pr` mode the `mergeable_state` gate is a backstop (a pending required check reports non-`clean`), but `--sha`/`--branch` modes have no such backstop and could emit a premature `Clear`.

## Approach (supersedes the closed PR #726)

PR #726 fixed this with a leading `pending` discriminator: `_check_run_recency_key` → `(status != "completed", started_at, id)`. It works, but it was closed because the discussion behind it showed the recency key is aimed at the wrong axis, and the `pending` flag papers over that instead of removing it.

Two problems with `pending`-primary:

1. **It conflates *selection* with *verdict*.** `latest_check_runs`'s only job is to pick the single most-recent run per name; `parse_check_result` then reads that run's `status`/`conclusion`. `pending` biases the *selection* toward runs whose status is one of the values we are about to *report* — it sorts by the answer. The pick should be "most recent," full stop.
2. **It's internally incoherent.** Within the not-completed group the key still tiebreaks on `started_at`, and a *queued* run has `started_at == ""`, which sorts oldest. So given a slow `in_progress` original plus a fresh `queued` re-run of the same name, `(pending, started_at, id)` picks the **older in_progress** run over the **newer queued** one — the same inversion the fix was meant to prevent, moved one case over.

Both trace to one root cause: **`started_at` is the wrong primary signal.** It is null until a run starts (the #719 bug), and it is not even monotonic with attempt order — a slow-to-schedule original can start *after* a faster re-run. The `pending` flag is a patch on top of a signal that shouldn't be primary.

### The simplification: sort by check-run `id` alone

A re-run always creates a **new** check-run row with a **higher `id`** — the current code already documents this ("a re-run always gets a higher id"), it just uses `id` only as a tiebreak. `id` is:

- **Never null** — assigned at creation, *before* the run starts, so a freshly-queued re-run already carries its (higher) id. This is exactly what #719 needs.
- **Monotonic with creation order** — the newest check-run row is the latest attempt, regardless of `started_at` scheduling jitter.
- **Already present** — no new field, no extra request.

So collapse to the highest-`id` run per name and read the verdict off that winner:

```python
def _check_run_recency_key(run):
    """Return a sort key ordering check runs oldest-to-newest by check-run id.

    GitHub attaches several check runs with the same `name` to one commit when a
    workflow re-runs (e.g. each label add/remove re-triggers the label gate);
    only the most recent is authoritative (issue #707). A re-run always creates a
    new check-run row with a higher `id`, and -- unlike `started_at`, which
    GitHub leaves null until a run actually starts (issue #719) -- the `id` is
    assigned at creation, so a freshly-queued re-run already outranks the
    completed run it supersedes. Sorting by `id` alone keeps the latest attempt
    for both completed re-runs (#707) and not-yet-started ones (#719), with no
    null-handling and no status special-casing. A run with an unusable id sorts
    oldest.
    """
    try:
        return int(run.get("id"))
    except (TypeError, ValueError):
        return -1
```

`latest_check_runs` is unchanged apart from its docstring: its `>=` winner test already works on a scalar key, and `id` uniqueness makes the list-order tiebreak moot.

This is a **net deletion** — the `pending` flag, the `started_at` read, the `""`/null sentinel, and the multi-element tuple all go away. Selection becomes pure recency; the verdict is read off the winner, which is what the objection to #726 argued for.

### Why this preserves #707

Verified against the real #707 fixture (`LABEL_GATE_3` in `test_ci_monitor.py`): its three same-named runs have ids `87686988990 < 87687158072 < 87688242514`, and the intended winner is the highest, `87688242514` (the authoritative `success`) — which is also the latest `started_at`. On real GitHub payloads `id` order and `started_at` order agree, because a re-run is both created later and starts later. `id`-only reproduces the #707 collapse exactly.

### What changes: the synthetic `STARTED_AT_WINS` case

The one existing assertion that flips is `STARTED_AT_WINS` (id 999 `failure`, `started_at` 2026-01-01 vs id 1 `success`, `started_at` 2026-01-02 — it asserts the low-id success wins); `id`-only picks id 999.

That fixture is **synthetic and not grounded in how GitHub assigns ids.** For the authoritative-latest run to have a *lower* check-run id than a stale one, GitHub would have had to create the newer attempt's row *before* the older attempt's — the reverse of how re-runs work. It encodes a belief ("`started_at` beats id") that the #716 author asserted but that no real payload exhibits; `LABEL_GATE_3`, the actual observed bug, has id and `started_at` in agreement. Replace `STARTED_AT_WINS` with a realistic same-name pair where the newer attempt has both the higher id and the later `started_at` (agreement case), or drop it — it was locking in a tiebreak direction that no longer exists.

This trades one empirical GitHub assumption for a stronger one: we stop assuming "`started_at` is populated at queue time" (demonstrably false — that *is* #719) and instead rely on "a re-run gets a higher check-run id" (which the code already relies on as its tiebreak, and which holds for every real re-run). Net, we remove reliance on a false assumption.

### Resulting behavior

| Scenario | Winner (highest id) | Verdict |
|---|---|---|
| #719: old completed `success` (id 1) + queued re-run (id 2) | queued re-run | `in_progress` (keeps polling; no premature `Clear`) |
| #719 symmetric: old completed `failure` (id 10) + queued re-run (id 11) | queued re-run | `in_progress` (no spurious `Blocked`) |
| #707: `success`, stale `failure`, `success` (ascending ids) | latest `success` | `all_passed` |
| slow `in_progress` original (id 1) + queued re-run (id 2) | queued re-run | `in_progress` |

The defect is fixed in every non-PR mode, and the fix no longer depends on `started_at` being present or on a status-based override.

### Sequencing

`latest_check_runs`/`_check_run_recency_key` are on `main` (PR #716 merged). Base this work on `main`; no cross-branch dependency remains.

### Tests (`scripts/test_ci_monitor.py`)

- **#719 regression (unit).** Old `{"status":"completed","conclusion":"success","started_at":"2026-01-01T00:00:00Z","id":1}` + queued `{"status":"queued","id":2}` (no `started_at`). Assert `latest_check_runs` keeps id 2 and that `parse_check_result` on the collapsed payload returns `in_progress` (not `all_passed`).
- **Symmetric variant.** Stale completed `failure` + queued re-run → assert `in_progress`, not `Blocked`.
- **#719 end-to-end (`--sha` `main()`).** Head commit carries the old completed `success` plus a queued re-run of the same required check; assert the monitor does **not** terminate `Clear` on the pending poll and only Clears once the re-run completes. This is the only test that exercises the real user-visible defect (premature loop termination) — keep it.
- **#707 preserved.** Re-assert `LABEL_GATE_3` collapses to id `87688242514`/`all_passed`; `ID_TIEBREAK` and `MIXED` unchanged.
- **Replace `STARTED_AT_WINS`** with a realistic "newer attempt has both higher id and later `started_at`" pair, or remove it (see above). Do not keep the id-999-beats-id-1 assertion.

Verification: `python3 scripts/test_ci_monitor.py` (0 failed) and `scripts/lint.sh --check` on the changed files.

### Scope / non-goals

- Localized to `_check_run_recency_key` (sole caller `latest_check_runs`, `ci_monitor.py:391`) plus two docstrings and the tests above. Outcome vocabulary is unchanged, so `scripts/ci_monitor/README.md` needs no update.
- Do **not** expand into Run ID / Run Attempt surfacing or a GraphQL join to workflow-run metadata. Those were considered and rejected: `run_attempt` is not on the check-run payload (it lives on the workflow-run object, and GraphQL doesn't expose it either), and it isn't needed — the check-run `id` already provides a monotonic, never-null, always-present per-attempt key.

aunger:
Fixes #719.

## Problem

PR #716's `latest_check_runs` collapses same-named check runs to the most recent per name, keyed by `_check_run_recency_key` = `(started_at, id)`.
GitHub leaves `started_at` null until a run actually starts, so a freshly-queued re-run maps to `("", id)`, which sorts *oldest*.
When a required check has an old completed run plus a freshly-queued re-run of the same name, the queued run is collapsed away and only the stale completion survives.
`parse_check_result` then reads that completion (`all_passed`, or `Blocked` in the symmetric stale-`failure` case) instead of `in_progress`.

In `--pr` mode the `mergeable_state` gate is a backstop, but `--sha`/`--branch` modes have no such backstop, so this window could fire a premature terminal `Clear` (a wrong terminal that ends the monitor while a required re-run is still pending).

## Fix

Add a leading `pending` discriminator to `_check_run_recency_key`: a run whose `status` is not `completed` (queued/in_progress) now sorts above every completed run of the same name, regardless of `started_at`.
`True` sorts above `False`, so a pending re-run wins the collapse and the verdict correctly reads `in_progress`.

Among completed runs `pending` is uniformly `False`, so the existing `started_at`-primary / `id`-tiebreak ordering is preserved verbatim; issue #707's scenarios collapse exactly as before.
The residual error direction is fail-safe: at worst the monitor waits (`in_progress`) rather than terminating early on a wrong `Clear`.

Docstrings on `_check_run_recency_key` and `latest_check_runs` are updated to state the new pending-supersedes-completed rule.

## Tests

- `(ba)` unit cases: a same-named old-`success` + queued re-run and the symmetric old-`failure` + queued re-run both collapse to keep the queued run, and the collapsed payload reads `in_progress` (not `all_passed`/`Blocked`). The #707 fixtures (three-completed-run collapse, `STARTED_AT_WINS`, `ID_TIEBREAK`, `MIXED`) are re-asserted unchanged.
- `(bb)` end-to-end `--sha` `main()` case: a first poll with the queued re-run must not terminate `Clear`; the monitor keeps polling and only Clears once the re-run completes. This exercises the actual user-visible defect (premature `Clear` in a non-PR mode), not just the verdict token.

Verified locally: `python3 scripts/test_ci_monitor.py` reports 343 passed, 0 failed, and `scripts/lint.sh --check` passes on both changed files.

No functional change to the outcome vocabulary, so `scripts/ci_monitor/README.md` needs no update.

### DIFF
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -346,17 +346,29 @@ def _check_run_recency_key(run):
     label-gate workflow, so that gate's check-run name accumulates several
     entries against the same head commit, of which only the most recent
     conclusion is authoritative (the earlier ones are stale).
-    Recency is judged by `started_at` (ISO 8601, so lexically sortable), then by
-    the numeric check-run `id` as a monotonic tiebreak (a re-run always gets a
-    higher id). A run missing both fields sorts oldest, so a run carrying real
-    recency data always outranks one that does not.
+
+    A not-yet-completed run (`status` other than `completed`, i.e. `queued` or
+    `in_progress`) is a re-run currently superseding any completed run of the
+    same name, so it sorts above every completed run regardless of timestamps
+    (issue #719). GitHub leaves `started_at` null until a run actually starts, so
+    without this a freshly-queued re-run would sort oldest by `started_at` and be
+    collapsed away in favor of the stale completed run it supersedes; the verdict
+    would then read that stale completion instead of `in_progress`. `True` sorts
+    above `False`, so the `pending` flag leads the key.
+    Among runs sharing a completion state, recency is judged by `started_at`
+    (ISO 8601, so lexically sortable), then by the numeric check-run `id` as a
+    monotonic tiebreak (a re-run always gets a higher id). Among completed runs
+    the `pending` flag is uniformly `False`, so their order is unchanged: a run
+    missing both timestamp and id sorts oldest, and a run carrying real recency
+    data always outranks one that does not.
     """
     started = run.get("started_at") or ""
     try:
         run_id = int(run.get("id"))
     except (TypeError, ValueError):
         run_id = -1
-    return (started, run_id)
+    pending = run.get("status") != "completed"
+    return (pending, started, run_id)
 
 
 def latest_check_runs(check_json):
@@ -366,6 +378,11 @@ def latest_check_runs(check_json):
     each distinct non-empty check-run `name`, only the most recent run (by
     `_check_run_recency_key`); the surviving run sits at the position of that
     name's first appearance, so the summary's row order is otherwise preserved.
+    A pending (queued/in_progress) re-run is never collapsed away in favor of an
+    older completed run of the same name, even before GitHub populates its
+    `started_at` (issue #719), so the verdict keeps reading `in_progress` while a
+    required check's re-run is still pending rather than latching onto the stale
+    completion it supersedes.
     Runs with no usable `name` are passed through unchanged (each kept), since
     name-based identity does not apply to them and GitHub always names its own
     check runs; this also leaves the unnamed-Actions-run fixtures other tests
--- a/scripts/test_ci_monitor.py
+++ b/scripts/test_ci_monitor.py
@@ -76,6 +76,13 @@ Covers:
        does not outvote the authoritative latest success -- the monitor
        terminates Clear (mergeable_state=clean), lists the gate once, and polls
        only the latest run's jobs/artifacts
+  (ba) #719 latest_check_runs: a queued/in_progress re-run (started_at null)
+       outranks an older completed run of the same name, so the collapsed
+       payload reads in_progress (not all_passed / not Blocked) while the re-run
+       is pending; the #707 all-completed fixtures still collapse unchanged
+  (bb) #719 main(): in --sha mode (no mergeable_state backstop) a queued re-run
+       of a required check does not fire a premature Clear -- the monitor keeps
+       polling and terminates Clear only once the re-run completes
 
 No network calls required; no GITHUB_TOKEN needed.
 Run this file directly to execute the suite: exits 0 on success, non-zero on failure.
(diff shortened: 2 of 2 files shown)


aunger:
🤖 Reviewer

LGTM. This faithfully implements the deep-planner plan on issue #719 and I found no blocking issues.

## Correctness

- The one-line change adds a leading `pending = run.get("status") != "completed"` discriminator to `_check_run_recency_key`, returning `(pending, started, run_id)`. Because `True` sorts above `False`, a queued/in_progress re-run now outranks any completed run of the same name in `latest_check_runs`'s `>= ` winner selection, so the queued run survives the collapse and `parse_check_result` reads `in_progress` (its `any(s in ("in_progress", "queued") ...)` branch) rather than latching onto the stale completion. This is exactly the gap described in the issue for `--sha`/`--branch` modes, which lack the `--pr` `mergeable_state` backstop.
- The #707 behavior is preserved verbatim: among completed runs `pending` is uniformly `False`, so ordering collapses back to the existing `(started_at, id)` key. I confirmed the re-asserted #707 fixtures (three-completed collapse, `STARTED_AT_WINS`, `ID_TIEBREAK`, `MIXED`) still pass unchanged.
- The tuple keys are shape-consistent (`bool, str, int` in every key), so element-wise comparison never crosses types.
- The residual-error direction is fail-safe, as the plan argues: a lingering pending run at worst keeps the monitor polling (`in_progress`) rather than firing a wrong terminal `Clear`. I could not construct a realistic GitHub payload where pending-supersedes-completed causes a wrong *terminal* outcome (a superseded run transitions to `completed/cancelled`, not a stuck `queued`).

## Tests

- Verified locally on the PR branch: `python3 scripts/test_ci_monitor.py` reports 343 passed, 0 failed, and `scripts/lint.sh --check` passes on both changed files.
- The `(ba)` unit cases cover both the old-`success` and symmetric old-`failure` variants (started_at absent and explicitly `None`), asserting both the surviving id and the collapsed verdict.
- The `(bb)` end-to-end `--sha` `main()` case is the valuable one: it exercises the actual user-visible defect (loop termination), asserting the monitor does not `Clear` on the pending first poll and only Clears once the re-run completes. The exhausted-deque assertion (`len(side_effects_bb) == 0`) neatly pins the exact request count.

## Docstrings / scope

- Both `_check_run_recency_key` and `latest_check_runs` docstrings are updated accurately to state the pending-supersedes-completed rule. The change stays a tight hardening tweak and does not drift into the Run ID / Run Attempt surfacing that #707 floated. README needs no update since the outcome vocabulary is unchanged.

## CI note (non-blocking)

At review time most checks are green; `build-and-test` is still `in_progress`, and the `No blocking labels` failure is the expected development-cycle label gate (it has a paired `success` run and fails by design during agentic review). Not blocking on CI; the Orchestrator runs the Monitor after I exit.

aunger:
🤖 Author

Note on the red `build-and-test`: the one blocking failure is `[E2E Tests] com.gb4pc.e2e.GalleryButtonVisualE2ETest#test4a_secureCameraLockedEmptyGalleryNoGreen`, an Android instrumented visual E2E test run on the emulator.

This PR's diff is Python-only (`scripts/ci_monitor/ci_monitor.py` and `scripts/test_ci_monitor.py`); it does not touch any app/Kotlin/E2E code, and `detect-changes` skipped `analyze-kotlin` accordingly. A change to the CI-monitor script cannot affect an emulator visual-diff test, so this failure is unrelated to the change (a flaky/environmental E2E result). A re-run of `build-and-test` should clear it.

The Python suite this PR actually exercises passed locally (343 passed, 0 failed) and `analyze-python`/`python-ruff` are green in CI.

aunger:
🤖 Reviewer

Re-reviewed after the Author's CI note. The code is unchanged since my review (head `63201a2`, no new commits, no open review threads), so my verdict stands: LGTM.

I concur with the Author on the red `build-and-test`: the failing check is `GalleryButtonVisualE2ETest#test4a_secureCameraLockedEmptyGalleryNoGreen`, an Android emulator visual-diff E2E test. This PR's diff is Python-only (`scripts/ci_monitor/ci_monitor.py`, `scripts/test_ci_monitor.py`), `detect-changes` skipped `analyze-kotlin`, and a ci_monitor script change cannot influence an emulator visual test — so this is an unrelated flaky/environmental failure, not a regression from the change. The Python paths this PR exercises are green (`analyze-python`/`python-ruff`) and the suite passes locally (343 passed, 0 failed). Not blocking on CI.

aunger:
Wrong direction. Will rewrite the issue.

PR #730, opened 2026-07-17
Poll for the green feed to vanish after the empty-roll overlay taps (test2a/test4a)

### LINKED ISSUE #241: [GalleryButtonVisualE2ETest] test2a_emptyGalleryNoGreenAfterTap
## Test Failure

| Field | Value |
|-------|-------|
| Branch | `fix/issue-237-auto-filed-issue-format` |
| Commit | `907f257` |
| Detected at | 2026-05-26 04:59 UTC |

### Failure message

java.lang.AssertionError: test2a_emptyGalleryNoGreenAfterTap: GREEN coverage after tap is 87.320335% — expected &lt; 10%. The gallery opened showing unexpected green content with an empty roll.

### Stack trace

```
java.lang.AssertionError: test2a_emptyGalleryNoGreenAfterTap: GREEN coverage after tap is 87.320335% — expected &lt; 10%. The gallery opened showing unexpected green content with an empty roll.
	at com.gb4pc.e2e.GalleryButtonVisualE2ETest.test2a_emptyGalleryNoGreenAfterTap(GalleryButtonVisualE2ETest.kt:230)
```

### Links

- [CI run](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/26432954222)
- [Test artifact: e2e-test-results](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/26432954222#artifacts)
- [PR](https://github.com/aunger/gallery-button-for-pixel-camera/pull/238)

---
_Filed automatically by CI on failure of `com.gb4pc.e2e.GalleryButtonVisualE2ETest.test2a_emptyGalleryNoGreenAfterTap`._

## Possible connection

This may share a root cause with #705 (same "unexpected GREEN coverage" symptom, filed after #403 was reopened for the identical failure) and possibly #86 (FG detector misidentifying the foreground app). Speculative — based on matching symptoms, not a confirmed cause.

### LINKED ISSUE #705: [GalleryButtonVisualE2ETest] test4a_secureCameraLockedEmptyGalleryNoGreen fails with unexpected GREEN coverage, not the original clearCameraRoll() precondition
## Summary

`test4a_secureCameraLockedEmptyGalleryNoGreen` is failing again on `main`, but with a different assertion than the one originally tracked in #403.

#403 tracked a `clearCameraRoll()` precondition failure (`MediaStore should be empty ... expected:<0> but was:<1>`, `E2EFixture.kt:276`), which was fixed by PR #400 (merged as `f2fab4e`) and confirmed green in #403's closing comment on 2026-06-15.

Starting 2026-06-27, CI began reopening #403 for a completely different failure in the same test method:

```
java.lang.AssertionError: test4a_secureCameraLockedEmptyGalleryNoGreen: GREEN coverage after tap is 87.45232%--expected < 10%. The gallery opened showing unexpected green content with an empty roll.
```

This is the identical failure message and symptom as the still-open #241 (`test2a_emptyGalleryNoGreenAfterTap`), and both plausibly trace back to #86 ("FG detector is wrong," still open, P1). As of 2026-07-15 this failure is still reproducing on `main` (commit `cb89df0`).

## Why a new issue

#403's title and history are specific to the `clearCameraRoll()` bug, which is fixed. The auto-filer keeps reopening it under the original title for an unrelated, still-live bug, which makes the tracking history for both bugs confusing. This issue tracks the current, live "unexpected GREEN coverage" failure in `test4a` on its own.

## Recent failure history

- 2026-06-27 19:53 UTC (`f9ad5d8`, main)
- 2026-06-29 03:54 UTC (`7f2265e`, main)
- 2026-07-10 18:32 UTC (`3b3d76e`, main)
- 2026-07-12 12:55 UTC (`3857e46`, main)
- 2026-07-15 08:31 UTC (`cb89df0`, main)

See #403 for the full comment history through 2026-07-15.

## Suggested next step

Investigate together with #241 and #86, since all three describe the same "unexpected green content after tap" symptom and may share a single root cause in the foreground/overlay detection path.

aunger:
Fixes #705. Fixes #241.

Both issues are symptoms of the same technical problem, so they are resolved together here (per `agents/pr_creation.md`, one topic per PR).

## Diagnosis

`test4a_secureCameraLockedEmptyGalleryNoGreen` (#705) and `test2a_emptyGalleryNoGreenAfterTap` (#241) fail deterministically with ~87% GREEN coverage after the overlay tap. The failing screenshot is the *pre-tap* frame: the full-screen green MockCameraActivity, captured before the tapped activity finished its cold start.

Evidence that the taps themselves work and the foreground/overlay path is not at fault:

- Both tests screenshot a fixed 1 s after `tapOverlay()`. test3a's own comment documents the mock gallery `LastPhotoActivity` cold start at ~1.4 s on the CI emulator (process spawn + MediaStore query), i.e. longer than the pause.
- test3a exercises the identical unlocked tap path as test2a and passes, because it polls up to 15 s for its expected post-tap state instead of pausing 1 s.
- test5a exercises the identical locked tap path as test4a. Its recent failure records on #243 measure **0 green pixels** after its 15 s post-tap poll, which means the full-green camera was replaced by SecureViewerActivity's black background: the locked tap opens the viewer. (test5a's own failures are at later stages: empty render / captureOnePhoto timeouts, tracked separately in #243.)
- The near-identical coverage values across runs and commits (87.32% / 87.45% / 88.39%) are what a screenshot of the same static pre-tap frame produces, not what a varying post-tap screen would.

So #705 is not an instance of the FG-detector bug (#86): the overlay activates (both tests' launch helpers gate on `OverlayService.isOverlayActive`), renders at the configured position (test1a-1c pass), and its tap launches the target activity. The tests simply measured too early.

## Change

- Add `E2EFixture.waitForGreenCoverageBelow(maxCoverage, timeoutMs, intervalMs)`: the inverse companion to `waitForGreenCoverage`, polling full-screen GREEN coverage until it drops below the threshold or the budget elapses.
- test4a and test2a: replace the fixed `pause(1000)` after `tapOverlay()` with a bounded 15 s `waitForGreenCoverageBelow(0.10f)` poll, mirroring the post-tap polls that already keep test3a and test5a's wait logic honest.

The assertions are unchanged, as test4a's kdoc requires. The red-light property of both tests is preserved: a genuinely blocked or missing tap leaves the green camera on screen, the poll never satisfies its threshold, and the unchanged `< 10%` assertion still fails at ~87% after the 15 s budget.

Note on test-requirement strictness: this replaces a fixed 1 s post-tap wait with an up-to-15 s bounded poll in `test2a_emptyGalleryNoGreenAfterTap` and `test4a_secureCameraLockedEmptyGalleryNoGreen`, giving the tapped activity more time to render before the measurement. The tests never specified a latency requirement; the same 15 s budget is already the established pattern in test3a and test5a.

## Verification

- `:app:compileDebugAndroidTestKotlin` and `testDebugUnitTest` pass locally; ktlint clean on both edited files.
- The E2E suite itself needs the CI emulator; this PR's CI run is the authoritative check that test2a and test4a go green.

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -818,17 +818,9 @@ class E2EFixture(
      * returns the screenshot it last captured -- so the caller asserts against the exact same
      * image that proved (or failed to prove) the color's presence, not a separately-captured one.
      *
-     * Issue #556: [waitForOverlayActive] only observes [OverlayService.isOverlayActive], an
-     * internal UsageStats-based flag that can flip true before the corresponding frame (camera
-     * feed plus overlay) is actually composited by the WindowManager -- especially on a
-     * cold-started activity, which briefly shows Android's mandatory splash-screen frame first.
-     * A fixed post-activation pause is a guess at how long compositing takes and silently stops
-     * being enough as CI load changes (e.g. issue #520's concurrent `screenrecord`). Polling for
-     * the actual pixel evidence removes that guess: `ScreenshotTestRule.failed()` proved the
-     * content was correct just moments after a fixed-pause capture went stale (it takes its own
-     * screenshot after the assertion throws and consistently shows the overlay rendered exactly
-     * at the configured position), so the content was never wrong -- only the single early
-     * capture was.
+     * Issue #556: [waitForOverlayActive] only observes [OverlayService.isOverlayActive], which
+     * can flip true before the corresponding frame is composited (a cold-started activity
+     * briefly shows the splash frame first), so callers poll for pixel evidence instead.
      *
      * @param color      The [Rgb] color to poll for.
      * @param timeoutMs  Maximum wait time in milliseconds.
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -767,25 +767,18 @@ class E2EFixture(
         timeoutMs: Long = 15_000L,
         intervalMs: Long = 500L,
     ): Float {
-        val deadline = System.currentTimeMillis() + timeoutMs
         var lastCoverage = 0f
-        while (System.currentTimeMillis() < deadline) {
-            val screen = Screenshot.captureScreen()
-            val w = screen.width
-            val h = screen.height
+        captureScreenUntil(timeoutMs, intervalMs) { screen ->
             val margin = (1f - 0.60f) / 2f
             val region =
                 android.graphics.Rect(
-                    (w * margin).toInt(),
-                    (h * margin).toInt(),
-                    (w * (1f - margin)).toInt(),
-                    (h * (1f - margin)).toInt(),
+                    (screen.width * margin).toInt(),
+                    (screen.height * margin).toInt(),
+                    (screen.width * (1f - margin)).toInt(),
+                    (screen.height * (1f - margin)).toInt(),
                 )
-            val greenMask = ColorMatch.mask(screen, Rgb.GREEN)
-            val coverage = ColorMatch.coverageFraction(greenMask, region)
-            lastCoverage = coverage
-            if (coverage >= minCoverage) return coverage
-            Thread.sleep(intervalMs)
+            lastCoverage = ColorMatch.coverageFraction(ColorMatch.mask(screen, Rgb.GREEN), region)
+            lastCoverage >= minCoverage
         }
         return lastCoverage
     }
@@ -845,16 +838,7 @@ class E2EFixture(
         color: Rgb,
         timeoutMs: Long = 5_000L,
         intervalMs: Long = 200L,
-    ): Bitmap {
-        val deadline = System.currentTimeMillis() + timeoutMs
-        while (true) {
-            val screen = Screenshot.captureScreen()
-            if (ColorMatch.mask(screen, color).pixelCount > 0 || System.currentTimeMillis() >= deadline) {
-                return screen
-            }
-            Thread.sleep(intervalMs)
-        }
-    }
+    ): Bitmap = captureScreenUntil(timeoutMs, intervalMs) { ColorMatch.mask(it, color).pixelCount > 0 }
 
     // ── Private helpers ───────────────────────────────────────────────────────
 
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -775,21 +775,9 @@ class E2EFixture(
      * threshold was reached.
      *
      * Inverse companion to [waitForGreenCoverage], for tests that assert the green mock-camera
-     * feed *disappears* after an overlay tap (issue #705). The screen the tap replaces is the
-     * full-screen green MockCameraActivity, and the expected post-tap screen (the mock gallery's
-     * or SecureViewerActivity's black empty state) only appears after the tapped activity's cold
-     * start (process spawn + first frame), which the CI logcat shows can take ~1.4 s and more
-     * under CI load. A fixed 1 s post-tap pause therefore deterministically screenshots the
-     * still-green pre-tap frame (the ~87% GREEN failures tracked in #705 and #241). Polling for
-     * the green feed to actually vanish removes that guess, exactly as test3a/test5a already
-     * poll for their expected post-tap state.
-     *
-     * The red-light property of the calling tests is preserved: a no-op tap (overlay blocked or
-     * not tappable) leaves the green camera on screen, this poll never satisfies its threshold,
-     * and the caller's own assertion still fails on the unchanged green screen after [timeoutMs].
-     *
-     * Measures full-screen coverage, not [waitForGreenCoverage]'s central 60% region, because
-     * the full screen is what the callers' assertions measure.
+     * feed *disappears* after an overlay tap (issue #705). Measures full-screen coverage, not
+     * [waitForGreenCoverage]'s central 60% region, because the full screen is what the callers'
+     * assertions measure.
      *
      * @param maxCoverage Full-screen GREEN fraction below which the poll stops.
      * @param timeoutMs   Maximum wait time in milliseconds.
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -612,8 +612,8 @@ class E2EFixture(
     }
 
     /**
-     * Taps the overlay, then polls screenshots until [expected] accepts [stableSamples]
-     * consecutive captures, returning the last capture (see [captureScreenUntil]).
+     * Taps the overlay, then polls screenshots until [expected] holds continuously for
+     * [stableForMs], returning the last capture (see [captureScreenUntil]).
      *
      * [tapOverlay] is silent when the overlay is not on screen, so every tap test must await
      * its expected post-tap screen; a fixed pause races the tapped activity's cold start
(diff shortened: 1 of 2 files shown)


aunger:
Remove the yammering about 1s delays and other stuff we don't do anymore. That's really NOT worth documenting.

aunger:
🤖 Review team

Review findings on this PR (adversarial pass; 6 confirmed, 3 plausible), most severe first.
Verdict: **Changes requested**.

**Correctness**

1. **Poll exits on a transient sub-threshold frame, masking the guarded regression** (CONFIRMED; `E2EFixture.kt:809`).
   `waitForGreenCoverageBelow` gates on the *absence* of green with no stability window and no positive evidence of the expected screen, so it returns on the first frame below 10% green, including an app-transition starting window or pre-decode black frame.
   In the regression case test2a guards (empty roll but the gallery renders GREEN content after its ~1.4 s cold start), the poll can exit inside the splash window and the assertion samples that same window, passing before the green frame renders.
   Same mechanism applies to a wrongly populated secure session in test4a.
   The sibling `waitForGreenBand` deliberately requires positive letterbox geometry for exactly this reason.

2. **Environmental non-green states can fake a red-light pass in test4a** (PLAUSIBLE; `GalleryButtonVisualE2ETest.kt:376`).
   If the display re-sleeps, the keyguard surface shows, or a blank screenshot frame occurs during the 15 s window, coverage drops below 10% for a reason unrelated to the tap and the test passes with the overlay regression (#81) live.
   CI's `stay_on_while_plugged_in 7` protects against timeout re-sleep in CI, but local runs and transient blank frames are unprotected; the old 1 s pause had a 15x smaller exposure window.

3. **The assertion measures a different frame than the poll accepted** (CONFIRMED; `GalleryButtonVisualE2ETest.kt:243` and `:378`).
   The poll's qualifying screenshot is discarded and `s2` is captured fresh at the identical 10% boundary, so poll and assertion can disagree.
   `captureScreenUntilColorVisible`'s kdoc (issue #556) names this exact anti-pattern and returns the measured bitmap to avoid it.

4. **No pre-tap green precondition; vacuous instant pass possible** (PLAUSIBLE; `GalleryButtonVisualE2ETest.kt:336`).
   Neither test asserts `s1` is actually green.
   If the secure launch fails after retries (`launchAndAwaitOverlay` only logs) the keyguard check still passes, the pre-tap screen is non-green, and the poll returns instantly, certifying "no green after tap" for a tap never exercised against a green camera.
   Inherited from the old `pause(1000)` version, but the poll makes the vacuous path instant.

**Cleanup / altitude**

5. **Per-test wait patches are one level too shallow** (CONFIRMED; `E2EFixture.kt:602`).
   `tapOverlay()` is silent on a miss, and the fixed-pause bug class has now recurred across rounds (#556, then #241/#705).
   A fixture-level `tapOverlayAndAwait(timeoutMs, expected: (Bitmap) -> Boolean): Bitmap` would subsume all four tap tests' waits and prevent a third recurrence.

6. **Fourth bespoke deadline loop; duplicates `waitForCondition`** (CONFIRMED; `E2EFixture.kt:798`).
   Both call sites discard the Float return, so the wait is exactly a boolean condition expressible via the existing generic `waitForCondition`, or one shared `captureScreenUntil(predicate)` poller.

7. **Dead Float return; the `1f` seed fabricates a measurement** (CONFIRMED; `E2EFixture.kt:804`).
   No caller consumes the return; with `timeoutMs <= 0` the helper reports 100% coverage without ever capturing a screen, contradicting its kdoc; the loop also lacks `waitForCondition`'s final post-deadline re-check.

8. **Rationale duplicated five times, with a contradiction** (CONFIRMED; `E2EFixture.kt:782`).
   The helper kdoc and test4a texts say a 1 s pause "deterministically" screenshots the pre-tap frame while also calling it a race against a variable ~1.4 s cold start; test2a's copy says "races" only.
   The story should live once in the helper kdoc with one-line call-site pointers, and "deterministically" should be corrected.

9. **Full mask per iteration where a pixel count would do** (PLAUSIBLE; `E2EFixture.kt:807`).
   Each iteration builds a `BooleanArray(w*h)` plus bbox/centroid via ~2.6M per-pixel `getPixel` calls though only `pixelCount` is used, coarsening the effective cadence on a loaded emulator.
   Pattern is inherited from `ColorMatch.mask`; the right fix is bulk `getPixels` inside `ColorMatch`, benefiting all callers.

Findings 1-3 share one fix: require the below-threshold reading on consecutive samples and have the helper return the qualifying bitmap for the assertion.

aunger:
🤖 Author

Pushed four commits addressing the review findings and finishing the comment cleanup requested above.

**Findings 1-4 (correctness), commit 7b356f5.**
`waitForGreenCoverageBelow` is gone, replaced by a generic `captureScreenUntil(timeoutMs, intervalMs, stableSamples, predicate): Bitmap` and a `tapOverlayAndAwait` fixture step.
test2a/test4a now:

- require 3 consecutive qualifying captures, so a transient app-transition frame cannot end the wait (finding 1);
- require the expected package (mock gallery / SecureViewer) to be foregrounded, both in the poll predicate and as a final assertion, so a screen that goes non-green for an unrelated reason (screen off, keyguard) fails loudly instead of passing (finding 2);
- assert on the exact frame the poll measured (finding 3);
- gate on the green camera actually being on screen before tapping, failing loudly otherwise (finding 4).

The unchanged `< 10%` GREEN assertion still runs first, so the red-light no-op-tap case keeps its canonical failure message.

**Findings 5-7 (altitude, reuse, dead return), commits 7b356f5 + d6a7e7a + 5b4995e.**
All four tap tests now go through `tapOverlayAndAwait`; test5a's band poll uses a shared `GreenBandMetrics` so its poll and assertion cannot diverge (it previously polled only 2 of the 4 asserted bounds); `waitForGreenBand` is deleted; `waitForGreenCoverage` and `captureScreenUntilColorVisible` are reimplemented on `captureScreenUntil`, leaving one screenshot loop implementation.
The dead Float return and `1f` sentinel are gone; the poller always captures at least once and measures a final capture at the deadline edge.

**Finding 8 + the storytelling cleanup, commit 2d8c6c4** (plus the trims in c186b71/91832df).
The fixed-pause rationale now lives in exactly one place, `tapOverlayAndAwait`'s kdoc, with one-line pointers at call sites; the "deterministically" overclaim is deleted with the rest.

**Finding 9 (ColorMatch bulk `getPixels`): declined here, filed as #731.**
The cost profile is `ColorMatch.mask`'s and is shared by all 12 call sites; hand-optimizing only this PR's call site would diverge from the uniform mask-based approach, so the fix belongs inside `ColorMatch`.

Compile and ktlint are clean locally; the E2E suite needs the CI emulator, so this PR's CI run is the check that the hardened polls (notably the new foreground-package conditions) behave on the real device.

aunger:
🤖 Reviewer 


1. **stableSamples = 3 at the default 500 ms interval is a tight margin against the 1.4 s figure finding 1 (of the original review round) was about** (PLAUSIBLE; E2EFixture.kt:799-802).
   Three consecutive qualifying samples take roughly 1 to 1.5 seconds of poll time before screenshot-capture latency is even counted. This PR's own diagnosis puts the gallery's cold start at about 1.4 seconds. If a regression made the empty-state gallery show erroneous green only once that cold start completes, the stability window and the cold-start estimate are close enough that I would not call finding 1 conclusively closed. Worth either a wider margin or a comment on why 3 x 500 ms is known to be enough headroom.
2. E2EFixture.pause(ms) (E2EFixture.kt:705) is now dead code. Its only two callers (test2a's and test4a's pause(1000)) were removed by this PR, and nothing else in the tree calls it. Delete it, or note why it is being kept.

aunger:
🤖 Author

Both points accepted; two commits pushed.

**Point 1 (tight stability margin): agreed, and fixed at the mechanism level rather than by widening the sample count** (`eda6943`).
The deeper problem you identified is that a sample-count window has no principled relationship to the cold-start duration: 3 x 500 ms only cleared ~1.4 s because slow per-capture masking padded it, and #731 will remove exactly that padding.
`captureScreenUntil`/`tapOverlayAndAwait` now take `stableForMs` instead of `stableSamples`: the predicate must hold continuously for a wall-clock duration, re-sampled every `intervalMs`, so the guarantee is independent of capture latency in both directions.
test2a/test4a hold their expected post-tap state for 3 s, which clears the ~1.4 s cold-start window with real margin, and the call sites now say so.

**Point 2 (dead `pause()`): agreed, deleted** (`bcc19fa`).
Grep confirms its only callers were the two `pause(1000)` sites this PR removed; nothing else in the tree uses it.

Compile and ktlint clean locally; CI on this push is the device-level check.

PR #732, opened 2026-07-19
Collapse same-named check runs by check-run id alone (#719)

### LINKED ISSUE #719: Harden ci_monitor non-PR modes against a premature Clear when a re-run has null started_at

aunger:
Fixes #719.

## Problem

`latest_check_runs` collapses same-named check runs to the most recent per name, then `parse_check_result` reads that run's verdict.
The old recency key was `(started_at, id)`.
GitHub leaves `started_at` null until a run actually starts, so a required check with an old completed run plus a freshly-queued re-run of the same name collapsed to the *old* run (the queued re-run sorted oldest on its empty `started_at`), and the verdict was read off the stale run during that window.

In `--pr` mode the `mergeable_state` gate is a backstop (a pending required check reports non-`clean`), but `--sha`/`--branch` modes have no such backstop and could emit a premature `Clear` (or a spurious `Blocked` on the symmetric stale-`failure` case).

## Fix

Sort by check-run `id` alone.
A re-run always creates a new check-run row with a higher `id`, and unlike `started_at` the `id` is assigned at creation (before the run starts), so a freshly-queued re-run already outranks the completed run it supersedes.
This is a net deletion: the `started_at` read, the null sentinel, and the tuple key all go away; selection becomes pure recency and the verdict is read off the winner.

This reproduces the #707 collapse exactly (its real payloads have `id` and `started_at` in agreement) and fixes #719 in every non-PR mode.

## Tests

- `(ba)` unit: a queued re-run with a null `started_at` (higher id) survives the collapse and the payload reads `in_progress` (not a premature `all_passed`); plus the symmetric stale-`failure` variant reads `in_progress` (not `Blocked`).
- `(bb)` end-to-end (`--sha`): the monitor stays `in_progress` through the queued poll and `Clear`s only once the re-run completes. This test fails against the old key (which would `Clear` on the queued poll) and passes with the new one.
- #707 fixtures (`LABEL_GATE_3`, `ID_TIEBREAK`, `MIXED`) are unchanged and still pass.

Replaced test `STARTED_AT_WINS` because its premise (a lower-id run with a later `started_at` outranks a higher-id run) no longer holds and describes no real GitHub payload; the replacement `NEWER_ATTEMPT_WINS` asserts the newer attempt (higher id and later `started_at`) wins, per the issue's "Replace `STARTED_AT_WINS` ... or drop it" instruction.

## Note on scope

The issue said the README needs no update (outcome vocabulary unchanged), but the "Same-named check-run collapsing" section still stated recency was judged by `started_at` then `id`, which is now false.
I corrected that one sentence in a separate commit so it is easy to review or revert.

Verification (all green locally): `python3 scripts/test_ci_monitor.py` (340 passed, 0 failed); `ruff check` and `ruff format --check`; `mdformat --check --wrap keep --number` and `check_md040.py` on the README.

### DIFF
--- a/scripts/ci_monitor/README.md
+++ b/scripts/ci_monitor/README.md
@@ -41,7 +41,8 @@ In `--run-id` mode, the run to track is simply the one named on the command line
 
 Before reading the verdict, summary, or run/job targets, the Monitor collapses the `/commits/{sha}/check-runs` payload so that each distinct check-run *name* keeps only its most recent run (issue #707).
 GitHub can attach several check runs with the same name to one commit when a workflow re-runs: for example, each PR-side label add/remove re-triggers `block-merge-on-blocking-labels.yml`, so the `No blocking labels` gate accumulates several entries against the same head commit, and a stale `failure` (from a blocking label that was since removed) can sit between two `success` runs.
-Recency is judged by the run's `started_at` (ISO 8601, lexically sortable), then its numeric check-run `id` as a monotonic tiebreak; the surviving run keeps the position of that name's first appearance so the summary's row order is otherwise unchanged.
+Recency is judged by the run's numeric check-run `id`: GitHub assigns it at creation and a re-run always gets a higher one, so it identifies the latest attempt without relying on `started_at`, which GitHub leaves null until a run actually starts (a freshly-queued re-run would otherwise sort oldest and let a stale completed run win; issue #719).
+The surviving run keeps the position of that name's first appearance so the summary's row order is otherwise unchanged.
 This mirrors GitHub's own `mergeable_state`, which judges a required check by its latest run per name.
 Without it, a superseded `failure` would outvote the authoritative later `success` and drive a spurious `Blocked` terminal, and the stale run's job would also be tracked for diagnostics.
 Runs with no usable `name` are passed through unchanged (each kept), since name-based identity does not apply to them and GitHub always names its own check runs.
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -366,8 +366,10 @@ def latest_check_runs(check_json):
     each distinct non-empty check-run `name`, only the most recent run (the
     highest check-run `id`, via `_check_run_recency_key`); the surviving run sits
     at the position of that name's first appearance, so the summary's row order
-    is otherwise preserved. Because check-run ids are unique, the `>=` winner
-    test below never ties, so no list-order tiebreak is relied on.
+    is otherwise preserved. Real check-run ids are unique, so in practice the
+    `>=` winner test below never ties; the one exception is two runs whose ids
+    are both unusable (each keyed -1), where `>=` keeps the later of the two in
+    list order, which is harmless since neither carries real recency.
     Runs with no usable `name` are passed through unchanged (each kept), since
     name-based identity does not apply to them and GitHub always names its own
     check runs; this also leaves the unnamed-Actions-run fixtures other tests
--- a/scripts/ci_monitor/ci_monitor.py
+++ b/scripts/ci_monitor/ci_monitor.py
@@ -339,33 +339,35 @@ def parse_pr_terminal(pr_json):
 
 
 def _check_run_recency_key(run):
-    """Return a sort key ordering check runs oldest-to-newest (issue #707).
+    """Return a sort key ordering check runs oldest-to-newest by check-run id.
 
-    GitHub can attach several check runs with the same `name` to one commit when
-    a workflow re-runs: e.g. each PR-side label add/remove re-triggers the
-    label-gate workflow, so that gate's check-run name accumulates several
-    entries against the same head commit, of which only the most recent
-    conclusion is authoritative (the earlier ones are stale).
-    Recency is judged by `started_at` (ISO 8601, so lexically sortable), then by
-    the numeric check-run `id` as a monotonic tiebreak (a re-run always gets a
-    higher id). A run missing both fields sorts oldest, so a run carrying real
-    recency data always outranks one that does not.
+    GitHub attaches several check runs with the same `name` to one commit when a
+    workflow re-runs (e.g. each PR-side label add/remove re-triggers the
+    label-gate workflow), so that gate's check-run name accumulates several
+    entries against the same head commit, of which only the most recent is
+    authoritative (issue #707). A re-run always creates a new check-run row with
+    a higher `id`, and--unlike `started_at`, which GitHub leaves null until a run
+    actually starts (issue #719)--the `id` is assigned at creation, so a
+    freshly-queued re-run already outranks the completed run it supersedes.
+    Sorting by `id` alone keeps the latest attempt for both completed re-runs
+    (#707) and not-yet-started ones (#719), with no null-handling and no
+    status special-casing. A run with an unusable id sorts oldest.
     """
-    started = run.get("started_at") or ""
     try:
-        run_id = int(run.get("id"))
+        return int(run.get("id"))
     except (TypeError, ValueError):
-        run_id = -1
-    return (started, run_id)
+        return -1
 
 
 def latest_check_runs(check_json):
-    """Collapse same-named check runs to the latest one each (issue #707).
+    """Collapse same-named check runs to the latest one each (issues #707, #719).
 
     Returns a shallow copy of `check_json` whose `check_runs` list keeps, for
-    each distinct non-empty check-run `name`, only the most recent run (by
-    `_check_run_recency_key`); the surviving run sits at the position of that
-    name's first appearance, so the summary's row order is otherwise preserved.
+    each distinct non-empty check-run `name`, only the most recent run (the
+    highest check-run `id`, via `_check_run_recency_key`); the surviving run sits
+    at the position of that name's first appearance, so the summary's row order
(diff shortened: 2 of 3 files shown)


aunger:
Two non-blocking nits from review (approving otherwise — verified the suite is green and that the new `(ba)`/`(bb)` assertions genuinely fail against the old `(started_at, id)` key):

- **`ID_TIEBREAK`'s comment is now stale.** "When started_at is absent/equal, the numeric id breaks the tie" describes the old model — `id` is now the sole recency key, not a tiebreak. The test itself is still correct and passes; only the comment reflects the superseded mental model. Worth a one-word touch-up for consistency, since you already updated the analogous `(ay)` header line.
- **`latest_check_runs` docstring — the "never ties" claim.** "Because check-run ids are unique, the `>=` winner test below never ties" holds for real payloads, but the `-1` fallback for an unusable id can tie two malformed runs (where `>=` harmlessly keeps the later one). Fine to leave as-is; just noting the stated invariant is "unique ids in practice," not strictly always.

aunger:
🤖 Author

Both addressed in e2ca216:

- Renamed the `ID_TIEBREAK` fixture to `ID_RECENCY` and updated its comment and assertion messages, so nothing calls the `id` a tiebreak anymore, consistent with the `(ay)` header.
- Softened the `latest_check_runs` docstring: "in practice never ties," with the one harmless exception called out, two runs with unusable ids (both keyed `-1`), where `>=` keeps the later of the two in list order.

Suite still green (340 passed, 0 failed); ruff clean.

PR #734, opened 2026-07-19
Require the mock gallery in the foreground in test3a's post-tap check

### LINKED ISSUE #733: test3a_populatedGalleryShowsGreenAfterTap can pass on a no-op tap (missing foreground-package check)
Deferred from PR #730's review (finding 1), out of scope for that PR.

test3a asserts `greenCoverage(s2) > 0.40f` after tapping the overlay, with no check that the tap actually launched the mock gallery (`MOCK_GALLERY_PACKAGE`). Before the tap, MockCameraActivity already covers the screen with close to 100% GREEN. If the tap is a no-op (the exact regression class #705/#241 exist to catch), the screen never changes, coverage stays above 40% for the whole poll, and the test passes without the gallery ever launching.

A stability requirement (`stableSamples`) would not fix this: the failure mode is a screen that never changes, so any number of consecutive samples is satisfied just as trivially as one.

The actual fix, mirroring test2a/test4a's existing pattern (added in PR #730), is a positive foreground-package check: require `fixture.foregroundPackage() == MOCK_GALLERY_PACKAGE` alongside the coverage check, both in the poll predicate and as a final assertion with its own failure message, so a no-op tap times out and fails loudly instead of passing on stale pre-tap content.

test5a's letterboxed-band predicate avoids this class of bug structurally (a full-screen pre-tap frame fails its height-fraction check), and test2a/test4a fixed the same class of issue via the foreground check. test3a is the one tap test in this file that still lacks either safeguard.

aunger:
🤖 Author

Fixes #733.

## Problem

`test3a_populatedGalleryShowsGreenAfterTap` asserted only `greenCoverage(s2) > 0.40f` after tapping the overlay, with no check that the tap actually launched the mock gallery.
Before the tap, `MockCameraActivity` already covers the screen with close to 100% GREEN.
On a no-op tap (the #241/#705 regression class this test exists to catch) the screen never changes, coverage stays above 40% for the whole poll, and the test passed without the gallery ever launching.

A stability requirement would not fix this: the failure mode is a screen that never changes, so any number of consecutive samples is satisfied just as trivially as one.

## Fix

Mirror the positive foreground-package safeguard that test2a/test4a already use (added in #730): require `fixture.foregroundPackage() == MOCK_GALLERY_PACKAGE` alongside the coverage check, in two places:

- In the `tapOverlayAndAwait` poll predicate, so the poll waits for the mock gallery specifically rather than being satisfied by the stale pre-tap mock-camera frame.
- As a final assertion with its own failure message, so a no-op tap times out and then fails loudly (naming the actual foreground package) instead of passing on stale pre-tap content.

The kdoc is updated to record why the foreground check is load-bearing.

This is the last remaining tap test in this file without either the foreground safeguard (test2a/test4a) or test5a's structural letterboxed-band discriminator.

The change touches only the instrumented E2E test; CI exercises it on the emulator.

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
@@ -295,22 +295,30 @@ class GalleryButtonVisualE2ETest {
         val greenMask = ColorMatch.mask(s2, Rgb.GREEN)
         Screenshot.saveForArtifact(maskToBitmap(greenMask), "3a-green-mask.png")
 
+        // Evaluate both post-tap conditions and report every one that fails in a single message.
+        // The coverage and foreground checks are independent failure modes, so when both hold
+        // (for example, the tap opened some other, non-green screen) the report should name both
+        // reasons rather than throwing on the first and hiding the second.
         val coverage = ColorMatch.coverageFraction(greenMask)
-        if (coverage <= 0.40f) {
-            fail(
-                "test3a_populatedGalleryShowsGreenAfterTap: GREEN coverage after tap is " +
-                    "${coverage * 100f}%--expected > 40%. " +
-                    "The gallery did not open, or the captured photo is not green.",
-            )
-        }
         val foreground = fixture.foregroundPackage()
-        if (foreground != MOCK_GALLERY_PACKAGE) {
-            fail(
-                "test3a_populatedGalleryShowsGreenAfterTap: foreground package after tap is " +
-                    "$foreground--expected $MOCK_GALLERY_PACKAGE. The screen shows green, but the " +
-                    "gallery did not open (the pre-tap mock-camera feed is still in front after a " +
-                    "no-op tap).",
-            )
+        val failures =
+            buildList {
+                if (coverage <= 0.40f) {
+                    add(
+                        "GREEN coverage after tap is ${coverage * 100f}%--expected > 40% " +
+                            "(the gallery did not open, or the captured photo is not green)",
+                    )
+                }
+                if (foreground != MOCK_GALLERY_PACKAGE) {
+                    add(
+                        "foreground package after tap is $foreground--expected $MOCK_GALLERY_PACKAGE " +
+                            "(the gallery did not come to the foreground; a no-op tap leaves the " +
+                            "green mock camera in front)",
+                    )
+                }
+            }
+        if (failures.isNotEmpty()) {
+            fail("test3a_populatedGalleryShowsGreenAfterTap: " + failures.joinToString("; "))
         }
     }
 
--- a/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
@@ -255,7 +255,17 @@ class GalleryButtonVisualE2ETest {
 
     /**
      * Verifies that tapping the overlay when the camera roll contains one GREEN photo opens
-     * the gallery and shows the GREEN content. Coverage after tap must exceed 40%.
+     * the mock gallery in the foreground and shows the GREEN content. Coverage after tap must
+     * exceed 40%, and the foreground package must be the mock gallery.
+     *
+     * The foreground-package check is load-bearing (issue #733): MockCameraActivity already fills
+     * the screen with close to 100% GREEN before the tap, so coverage alone stays above 40% even
+     * when the tap is a no-op and the gallery never launches (the #241/#705 regression class this
+     * test exists to catch). A stability requirement would not help either: the no-op failure mode
+     * is a screen that never changes, so any number of consecutive samples is satisfied just as
+     * trivially as one. Requiring the mock gallery in the foreground, mirroring test2a/test4a's
+     * safeguard, makes a no-op tap time out and fail loudly instead of passing on the stale
+     * pre-tap frame.
      */
     @Test
     fun test3a_populatedGalleryShowsGreenAfterTap() {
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Reviewer

Reviewed the diff against issue #733's ask.

The fix is exactly what was asked for, and it correctly closes the hole. Before this change, `test3a`'s poll and final assertion only checked `greenCoverage(s2) > 0.40f`. Since `MockCameraActivity` is already close to 100% green before the tap, a no-op tap (the #241/#705 regression class this test exists to catch) would leave the screen green and pass vacuously. The added `fixture.foregroundPackage() == MOCK_GALLERY_PACKAGE` check, in both the poll predicate and the final assertion, closes that hole: a no-op tap now times out (foreground stays the mock camera's package) and fails loudly with a dedicated message, instead of passing on the stale pre-tap frame. This mirrors test2a/test4a's existing pattern precisely, which is what the issue asked for.

Checked:

- The poll predicate combines both conditions with `&&`, matching test2a/test4a's structure.
- The final assertion adds the foreground check after the existing coverage check, with its own failure message naming the actual foreground package, again matching the established pattern. Walking through the regression scenario by hand: on a no-op tap, coverage stays high (so the existing coverage assertion does not fire), and execution reaches the new foreground check, which then fails loudly, exactly the behavior issue #733 asked for.
- `MOCK_GALLERY_PACKAGE` is the existing companion-object constant; no duplication.
- The kdoc update explains why the check is load-bearing and is accurate.
- ktlint and the other quick CI checks pass. "No blocking labels" failing is expected per the label-gate design, not a real failure. `build-and-test` (the emulator run that actually exercises this test) was still in progress when I finished this review; I am not blocking on it.

One non-blocking observation, not a request for change: test2a/test4a pair their foreground check with `stableForMs = 3_000L` to guard against a transient app-transition frame satisfying the predicate early. This PR does not add that here, so there is a narrow theoretical race where the UiAutomator foreground package updates to the mock gallery slightly before its first green frame composites, and the poll could end on a stale still-green screenshot that happens to coincide with the updated foreground package. Issue #733 did not ask for this, and the old test3a poll had no stability window either, so I am not requesting a change, just flagging it in case it is worth considering.

Verdict: LGTM

aunger:
### Verification plan

**Before merging** (must resolve before PR #734 can be merged):
None.

- [ ] #736

app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
```diff
@@ -287,6 +303,15 @@ class GalleryButtonVisualE2ETest {
                     "The gallery did not open, or the captured photo is not green.",
```
aunger: Do a better job combining these two failure cases. If both failures occur, log them both.

aunger: 🤖 Author

Done in f0d0b8c. The two sequential `if/fail` blocks threw on the first failure and hid the second, so a tap landing on some other non-green screen (both coverage low *and* foreground not the gallery) reported only the coverage reason. I now collect each failing condition into a list and `fail()` once with all of them joined by `; `, so both reasons surface together when both hold. Each reason is worded to stand on its own (no cross-reference to the other condition's state), so the combined message reads correctly whether one or both fired.

Kept the change local to test3a as scoped to issue #733. test2a/test4a share the same two-`if` shape, but reworking them is out of scope for this PR.

aunger:
🤖 Reviewer

Re-reviewed after the branch moved (dfbd6d6 -> f0d0b8c). Issue #733 and PR #734's description are unchanged from my last pass; the only change is a second commit, "Report both of test3a's post-tap failures together" (f0d0b8c).

That commit replaces the two sequential `if (...) fail(...)` blocks with a `buildList` of failure strings joined into a single `fail()` call, so a tap that lands on some third, unrelated screen (non-green and wrong foreground at once) now reports both reasons instead of the coverage failure hiding the foreground one. Walked through it by hand: `coverage` and `foreground` are each captured once, each condition appends its own descriptive string to `failures`, and `fail()` fires only when the list is non-empty, joining the parts with "; ". This is correct, and it is a genuine diagnostic improvement over the sequential-fail pattern. ktlint passed on this commit too.

It still satisfies #733's ask: the foreground reason remains present as its own distinct clause in the combined failure message, so a no-op tap still times out and fails loudly, naming the actual foreground package.

Two minor, non-blocking notes:

- This commit shipped with no PR comment explaining it, only a commit message (which is clear and sufficient on its own). Worth a habit-check for visibility, not a request for change.
- It diverges from test2a/test4a's established sequential-if/fail style, so the file now has two different failure-reporting shapes across three very similar tests. Not wrong, and I won't block on it since issue #733 only asked to mirror the foreground check itself, not the reporting style, and the new style is strictly more informative. But if this shape is worth adopting, it is probably worth doing consistently across all three tests in a follow-up rather than test3a alone drifting to a new pattern.

No functional issues found. Verdict unchanged: LGTM

aunger:
🤖 Author

Thanks for the re-review. Two responses to the non-blocking notes:

- Visibility: I did reply on the inline review thread when I pushed f0d0b8c (#discussion_r3611616952), but you're right that a top-level note is clearer, hence this comment.
- Reporting-style consistency: agreed it shouldn't be test3a alone drifting to a new shape. Since #733 only asked to add test3a's foreground check, I've kept this PR narrow and filed #737 to unify the combined-failure-reporting style across test2a/test3a/test4a (factoring out a shared helper if it reads cleanly) as a follow-up.

No code changes needed on this PR beyond what's already pushed.

aunger:
🤖 Reviewer

Acknowledged, thanks. I only pulled top-level comments in my last pass and missed the inline review thread (r3611442725 / r3611616952); re-fetching this round with `get_review_comments` surfaced it. That exchange (the owner's original ask to combine the two failure cases, and your f0d0b8c reply) matches what I re-derived independently from the diff, so nothing was missed on either side.

Confirmed #736 (stableForMs stability window) and #737 (unify the combined-failure-reporting shape across test2a/test3a/test4a) are both correctly scoped as follow-ups, not blockers for #733/#734.

No further action needed on this PR. Verdict stands: LGTM

PR #735, opened 2026-07-19
Scan screenshots via one Bitmap.getPixels bulk read in ColorMatch.mask

### LINKED ISSUE #731: Optimize ColorMatch.mask with bulk getPixels instead of per-pixel JNI reads
Deferred from PR #730's review (finding 9).

`ColorMatch.mask` scans a 1080x2400 screenshot with one `Bitmap.getPixel` JNI call per pixel (~2.6M calls) and allocates a `BooleanArray(w*h)` plus bbox/centroid state per invocation.
Every screenshot poll in the E2E suite pays this cost per iteration, which coarsens effective poll cadence on the loaded CI emulator.

Suggested fix: read the bitmap once into a reused `IntArray` via `Bitmap.getPixels` and scan that, keeping the `BinaryMask` result shape unchanged.
Fixing it inside `ColorMatch` benefits all callers uniformly, which is why PR #730 did not hand-optimize its one new call site instead.

aunger:
🤖 Author

Fixes #731.

## Problem

`ColorMatch.mask` scanned a 1080x2400 screenshot with one `Bitmap.getPixel` JNI call per pixel (~2.6M calls per invocation).
Every screenshot poll in the E2E suite pays this cost per iteration, which coarsens effective poll cadence on the loaded CI emulator.

## Change

`ColorMatch.mask` now reads the whole bitmap once into a reused `IntArray` via `Bitmap.getPixels` (one JNI call) and scans that flat, row-major buffer, instead of ~2.6M per-pixel JNI reads.
The scratch buffer is grown to the largest bitmap seen so far and reused across calls, so a poll loop no longer allocates a ~2.6M-int array every iteration.
The `BinaryMask` result shape is unchanged, so all callers benefit uniformly (which is why the fix lives inside `ColorMatch` rather than at one call site).

The per-pixel scan itself is extracted into a pure-JVM `PixelMask.scan` (placed in `sharedTest`, free of `android.*` deps like `MaskData`), which consumes the flat ARGB buffer and returns a `MaskData`.
`ColorMatch.mask` reads the bitmap and bridges that `MaskData` to the `android.graphics`-flavored `BinaryMask` directly (no defensive copy of the bits array, since the `MaskData` is freshly allocated and discarded there).
`ColorMatch.matches` now delegates to `PixelMask.matches` so the channel-tolerance logic lives in one place.

Output is identical to the previous per-pixel implementation, including the empty-mask shape (`(0,0,0,0)` bbox, `(0,0)` centroid).

## Tests

Extracting the scan into pure-JVM code lets it be unit-tested without an emulator.
New `PixelMaskTest` (JVM) covers:

- `matches` tolerance boundaries (inclusive at exactly tolerance, false one step beyond, alpha ignored).
- Empty, single-pixel, and multi-pixel scans: bits, tight bbox, centroid, and pixelCount.
- Row-major stride correctness (a non-square grid with off-diagonal hits, the exact failure mode of a bulk-read indexing bug).
- Reuse semantics: an oversized buffer only reads its leading `width*height` entries; a too-small buffer is rejected; a zero-size grid is empty.
- A cross-check of the single-pass bbox/centroid/count accumulation against an independent reference implementation over 20 pseudo-random buffers.

The existing instrumented E2E visual suite continues to exercise the on-device `Bitmap.getPixels` path.

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/visual/ColorMatch.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/visual/ColorMatch.kt
@@ -1,7 +1,6 @@
 package com.gb4pc.e2e.visual
 
 import android.graphics.Bitmap
-import android.graphics.Color
 import android.graphics.PointF
 import android.graphics.Rect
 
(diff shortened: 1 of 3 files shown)


aunger (commented):
🤖 Reviewer

Verified this against the issue and against the pre-PR `ColorMatch.kt` on `main`. The fix does exactly what #731 asked for: one `Bitmap.getPixels` bulk read into a reused `IntArray` instead of ~2.6M `getPixel` JNI calls, with the scan extracted into a pure-JVM `PixelMask.scan` so it's unit-testable without an emulator.

Correctness checks I ran by hand:

- `PixelMask.matches`'s manual `(pixel shr 16) and 0xFF` etc. is bit-for-bit equivalent to `Color.red/green/blue` (masking with `0xFF` erases any sign-extension from Kotlin's arithmetic `shr` on the alpha byte, so the arithmetic-vs-logical-shift distinction doesn't matter here).
- Row-major indexing (`y * width + x`) and the bbox/centroid accumulation match the original loop exactly, including the empty-mask shape (`(0,0,0,0)` bbox, `(0,0)` centroid, all-false full-size `bits`).
- The reused `pixelBuffer` is always fully overwritten by `getPixels` before `scan` reads it, including when it's oversized from a larger prior screenshot, so no stale-data risk.
- `sourceSets` already wires `src/sharedTest/java` into both `test` and `androidTest` (pre-existing, from the `MaskData`/`ShapeMatcher` precedent), so `PixelMask`'s placement isn't new plumbing.

Test suite is a genuine strength here: the row-major-stride test specifically targets the failure mode a bulk-read indexing bug would produce, and the independent-reference fuzz test (`referenceScan`, 20 pseudo-random buffers seeded on 731) cross-checks the single-pass bbox/centroid accumulation against a structurally different implementation. That's more rigor than the old per-pixel code ever had.

Two non-blocking nits:

1. `ColorMatch.mask()` hand-rebuilds the exact `MaskData → BinaryMask` field mapping that `MaskData.toBinaryMask()` (`BinaryMaskExt.kt`) already provides, differing only in skipping the defensive `bits.copyOf()`. The comment explains why (avoid copying a ~2.6M-element array every poll), which is a reasonable tradeoff, but it's now duplicated mapping logic that can silently drift if `MaskData`'s fields ever change. Worth a second look, not worth another round by itself.
2. `pixelBuffer` is a shared mutable `object`-level field, safe only because `ColorMatch` is used single-threaded (stated in the comment but not enforced). I checked the E2E polling call sites (`E2EFixture.captureScreenUntil`, `waitForGreenCoverage`, etc.) and confirmed they're all plain `Thread.sleep` loops with no coroutines/threads, so this holds today. Just flagging it as an invariant a future caller could break silently.

CI: `build-and-test` was still in progress at review time; `No blocking labels` failing is expected mid-cycle per this repo's convention.

Verdict: LGTM.

PR #738, opened 2026-07-20
Strengthen seedOnePhoto()'s MediaStore-visibility wait to a rowsBefore/rowsAfter check

### LINKED ISSUE #606: Strengthen E2EFixture.seedOnePhoto()'s MediaStore-visibility wait to a rowsBefore/rowsAfter comparison
This is a follow-on item and does not block merging PR #582.

Source: review comment on PR #582 (round 2), reaffirmed as still-unaddressed-but-optional in round 3:
- https://github.com/aunger/gallery-button-for-pixel-camera/pull/582#pullrequestreview-4629591398
- https://github.com/aunger/gallery-button-for-pixel-camera/pull/582#pullrequestreview-4630178106

## The gap

`E2EFixture.seedOnePhoto()` (`app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt`) confirms its MediaStore insert is visible with:

```kotlin
waitForCondition(5_000L) { countMediaStoreImages() > 0 }
```

As the reviewer noted, this is weaker than the `rowsBefore`/`rowsAfter` comparison pattern `captureOnePhoto()` already uses a few lines below in the same file: `countMediaStoreImages() > 0` is vacuously satisfied if any residual image already exists in the shared-images MediaStore from an earlier E2E suite, since `seedOnePhoto()`'s caller (`PartialAccessPhotoPickerE2ETest`) does not clear the camera roll first. In practice this is harmless (the `ContentResolver` insert is synchronous), but the wait doesn't actually prove *this* insert landed.

## Suggested fix

Capture the row count before the insert and wait for it to increase (`rowsBefore`/`rowsAfter`-style), matching `captureOnePhoto()`'s existing pattern, so the sanity wait is meaningful rather than vacuously true.

## Reviewer's own framing

> Harmless in practice (the `ContentResolver` insert is synchronous), but a `rowsBefore`-style check would make the sanity wait actually mean something. Feel free to leave it or fix in a follow-up.

aunger:
Fixes #606

## Problem

`E2EFixture.seedOnePhoto()` confirmed its MediaStore insert was visible with:

```kotlin
waitForCondition(5_000L) { countMediaStoreImages() > 0 }
```

That check is vacuously satisfied whenever the shared-images MediaStore already holds a residual image from an earlier E2E suite. `seedOnePhoto()`'s caller (`PartialAccessPhotoPickerE2ETest`) does not clear the camera roll first, so `count > 0` can already be true before this insert, and the wait does not actually prove *this* insert became visible.

## Fix

Capture the row count before the insert and wait for the count to rise above that baseline, mirroring the `rowsBefore`/`rowsAfter` pattern `captureOnePhoto()` already uses a few lines below in the same file. The sanity wait now means something rather than being vacuously true.

The change is confined to the `seedOnePhoto()` test fixture helper; there is no product-code change. The helper is exercised by `PartialAccessPhotoPickerE2ETest` in the E2E suite CI already runs.

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt
@@ -429,6 +429,12 @@ class E2EFixture(
      */
     fun seedOnePhoto(): Uri {
         val resolver = context.contentResolver
+        // Capture the count before the insert so the visibility wait below can require *this*
+        // insert to land, matching captureOnePhoto()'s rowsBefore/rowsAfter pattern. A bare
+        // `count > 0` check would be vacuously satisfied by any residual image a prior E2E suite
+        // left in the shared MediaStore (this method's caller does not clear the camera roll
+        // first), so it would not actually prove this row became visible (issue #606).
+        val rowsBefore = countMediaStoreImages()
         val values =
             ContentValues().apply {
                 put(MediaStore.Images.Media.DISPLAY_NAME, "gb4pc-e2e-seed-${System.currentTimeMillis()}.jpg")
@@ -461,8 +467,10 @@ class E2EFixture(
         }
 
         // Confirm the row is now visible via an unfiltered query, so the caller does not open the
-        // picker before the insert is observable.
-        waitForCondition(5_000L) { countMediaStoreImages() > 0 }
+        // picker before the insert is observable. Waiting for the count to rise above rowsBefore
+        // (rather than merely be positive) means this wait proves *this* insert landed even when
+        // the shared MediaStore already held residual images.
+        waitForCondition(5_000L) { countMediaStoreImages() > rowsBefore }
         return uri
     }
 


aunger:
🤖 Reviewer

LGTM.

The change does exactly what issue #606 asked and nothing more. `seedOnePhoto()` now captures `rowsBefore = countMediaStoreImages()` at the top of the method (before `resolver.insert`) and waits for `countMediaStoreImages() > rowsBefore` instead of the vacuous `> 0`, mirroring the `rowsBefore`/`rowsAfter` pattern `captureOnePhoto()` already uses a few lines below.

Correctness points I checked:

- The baseline is captured before the insert, so it reflects only pre-existing residual images. The strict `> rowsBefore` therefore proves *this* row became visible rather than being satisfied by leftovers from a prior E2E suite (the caller `PartialAccessPhotoPickerE2ETest` does not clear the roll first).
- The `IS_PENDING = 1` insert keeps the row hidden from the unfiltered count until it is cleared to `0` after the JPEG bytes are flushed, so the count only reaches `rowsBefore + 1` once the row is genuinely observable by the picker. The wait remains meaningful.
- Change is confined to test-fixture code; no product-code change. Comments follow the repo prose style.

One thing I considered and am deliberately not requesting: the `waitForCondition` return value is still ignored (no `fail()` on timeout), unlike `captureOnePhoto()`. That is fine and in scope: the issue explicitly frames this as a best-effort sanity wait ("harmless in practice; the ContentResolver insert is synchronous"), and the original code ignored the result too. Adding a hard assertion would exceed #606's scope and risk a spurious 5 s-timeout failure.

CI has not reported yet; I am not blocking on it.

PR #739, opened 2026-07-20
Split shell script unit tests into a parallel CI job

### LINKED ISSUE #618: Run shell script unit tests in a separate, parallel job
## Motivation

`Run shell script unit tests` in `build.yml` (running `scripts/test_*.sh`) takes about 30 seconds and has no dependency on the JDK/Android SDK setup or the Python steps it currently sits next to. It only needs a checkout and bash. Splitting it into its own job would let it run concurrently with the rest of `build-and-test` instead of adding to the sequential critical path.

## Notes

- Not a correctness fix, just a speed-up. Unlike the Python dependency install (`pip install -r scripts/requirements.txt`, needed later in `build-and-test` by the `always()`-gated summary/gate steps), nothing downstream in the same job consumes the shell tests' step outcome, so there's no ordering hazard in leaving them where they are today.
- The new job would need its own way to respect `needs_full_build` (the docs-only-changes skip). Options: duplicate the `Check for non-docs changes` logic, or extract `diff_check` into its own small upstream job that both `build-and-test` and the new job depend on via `needs:`.
- If `build-and-test` is named in branch protection's required status checks, the new job would need to be added there too for it to gate merges the same way.

aunger:
Fixes #618

## What

`Run shell script unit tests` (`scripts/test_*.sh`) previously ran inside the `build-and-test` job, on its sequential critical path, even though it needs only a checkout, bash, and the lint tools (no JDK/Android SDK, no Python steps). This moves it into its own `shell-tests` job so it runs concurrently with `build-and-test` instead of adding ~30s to that job's runtime.

## How

- Extracted `Check for non-docs changes` into a small upstream `check-diff` job that outputs `needs_full_build`. This is the DRY option the issue suggested, avoiding duplicating the diff logic in two places.
- `build-and-test` and the new `shell-tests` job both `needs: [check-diff]` and gate on `needs.check-diff.outputs.needs_full_build`.
- `build-and-test` re-exports that value as its own `needs_full_build` output, so `link-pr-to-ci-summary` (which reads `needs.build-and-test.outputs.needs_full_build`) is unchanged.
- The two shell-test steps (`Install lint tools for shell tests`, `Run shell script unit tests`) moved verbatim into `shell-tests`, still per-step gated on `needs_full_build`. The job runs and reports green on docs-only changes, matching `build-and-test`'s per-step gating behavior.

No test logic changed; this is purely a CI-graph reorganization. The `build-and-test` job name (referenced by the CI monitor fixtures and branch protection) is preserved.

## Before merging

- [ ] If `build-and-test` is listed among branch protection's required status checks, add `shell-tests` (and `check-diff`) there too so the split job gates merges the same way. This is a repo-settings change outside this repo and cannot be done in the PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,17 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      - name: Install Python script dependencies
+        # Some scripts/test_*.sh drive Python scripts end-to-end: e.g.
+        # test_summarize_preflight_integration.sh runs summarize_test_results.py,
+        # which imports defusedxml. In build-and-test these tests ran after
+        # "Install Python script dependencies", so the deps were already present;
+        # the split job must install them itself. --require-hashes refuses any
+        # artifact whose SHA-256 is not in the generated lock scripts/requirements.txt
+        # (issue #723; see scripts/requirements.in for the top-level pins).
+        if: needs.check-diff.outputs.needs_full_build == 'true'
+        run: pip install --require-hashes -r scripts/requirements.txt
+
       - name: Install lint tools for shell tests
         # scripts/test_lint_hook.sh drives scripts/lint.sh and the pre-commit git
         # hook end-to-end, which needs the real lint tools. Provision the PyPI ones
@@ -120,8 +131,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
 
       - name: Install Python script dependencies
         if: needs.check-diff.outputs.needs_full_build == 'true'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,70 @@ permissions:
   issues: read
 
 jobs:
+  # Compute whether this event needs the full build, or is a docs-only change that
+  # can skip the expensive steps (issue #618). Extracted into its own small upstream
+  # job so both build-and-test and the parallel shell-tests job can gate on the same
+  # result via `needs:`, instead of each recomputing the diff. Needs only a checkout
+  # (fetch-depth 0 so check_non_docs_changes.sh can diff against the base) and git.
+  check-diff:
+    runs-on: ubuntu-latest
+    outputs:
+      needs_full_build: ${{ steps.diff_check.outputs.needs_full_build }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for non-docs changes
+        id: diff_check
+        run: |
+          if [[ "${{ github.event_name }}" == "push" || "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            # Always build on direct pushes to main (e.g. merge commits) and on
+            # manual ad hoc runs. github.base_ref is empty for workflow_dispatch,
+            # so falling through to the pull_request branch below would pass an
+            # empty ref to git fetch/check_non_docs_changes.sh--see #577.
+            echo "needs_full_build=true" >> $GITHUB_OUTPUT
+          else
+            BASE_REF="${{ github.base_ref }}"
+            git fetch origin "$BASE_REF" --depth=1
+            echo "needs_full_build=$(scripts/check_non_docs_changes.sh "origin/$BASE_REF")" >> $GITHUB_OUTPUT
+          fi
+
+  # Run the shell script unit tests (scripts/test_*.sh) in their own job (issue #618).
+  # They take about 30 seconds and need only a checkout, bash, and the lint tools,
+  # with no dependency on the JDK/Android SDK setup or the Python steps they used to
+  # sit next to in build-and-test. Running them here lets them execute concurrently
+  # with build-and-test instead of adding to its sequential critical path. Each step
+  # is gated on check-diff's needs_full_build so a docs-only change skips them, the
+  # same per-step gating build-and-test uses; the job itself still runs (and reports
+  # a green result) so it stays consistent with build-and-test on docs-only changes.
+  shell-tests:
+    needs: [check-diff]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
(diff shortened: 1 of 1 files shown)


aunger:
🤖 Reviewer

**Changes requested.**

## Blocking: `shell-tests` fails on this PR's own CI run

Job `shell-tests` (run 29714860020) is already red: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/29714860020/job/88266025260

```
==> Running scripts/test_summarize_preflight_integration.sh
=== (A) Direct CLI invocation in the three-suite shape build.yml uses ===
Traceback (most recent call last):
  File ".../scripts/summarize_test_results.py", line 26, in <module>
    import defusedxml.ElementTree as ET
ModuleNotFoundError: No module named 'defusedxml'
```

Root cause: `scripts/test_summarize_preflight_integration.sh` is one of the `scripts/test_*.sh` glob that the new `shell-tests` job runs, but it directly invokes `scripts/summarize_test_results.py`, which imports `defusedxml`. Before this PR, that dependency was satisfied because "Install Python script dependencies" (`pip install --require-hashes -r scripts/requirements.txt`) ran earlier in the same `build-and-test` job, ahead of "Run shell script unit tests". The split moved the shell-test step into its own job without carrying that pip install along, so the dependency is simply missing there now. `shell-tests`' "Install lint tools for shell tests" step only installs `scripts/requirements-lint.txt` into a separate venv (`LINT_BIN_DIR`), which doesn't cover this.

This isn't a hypothetical, it's the actual conclusion on this PR's own run, so it needs a fix before merge (installing `scripts/requirements.txt`, or at least `defusedxml`, in `shell-tests`), not just a rebase/rerun.

## Nit: leftover `fetch-depth: 0` on `build-and-test`'s checkout

`build-and-test`'s checkout (line ~122) still has `fetch-depth: 0`, but the only reason that job ever needed full history, "Check for non-docs changes" diffing against the merge base, has now moved entirely to `check-diff`. I checked for any other git-history use in `build-and-test` (gradle files, the remaining steps) and found none. Since the whole point of this PR is shaving time off `build-and-test`'s critical path, leaving it doing a full-history clone it no longer needs works against that goal. Worth dropping the override so it gets the default shallow checkout.

## Everything else

The `check-diff` extraction itself is sound: it's a faithful, unmodified relocation of the diff-check logic, `needs_full_build` is correctly re-exported from `build-and-test` so `link-pr-to-ci-summary` keeps working, and `ci_monitor.py`'s check/step parsing is job-name-agnostic so the new `check-diff`/`shell-tests` check runs are picked up without any changes there. The "Before merging" branch-protection checklist item is correctly called out as outside the repo.

aunger:
🤖 Author

Both points addressed in ff87c5e.

**Blocking (shell-tests red on defusedxml):** Agreed, your root-cause is exact. `test_summarize_preflight_integration.sh` drives `summarize_test_results.py` (which imports `defusedxml`), and that dependency was only ever present because `build-and-test`'s "Install Python script dependencies" step ran before the shell tests. I added the same `pip install --require-hashes -r scripts/requirements.txt` step to `shell-tests`, ahead of the shell-test run and gated on `needs_full_build`, mirroring `build-and-test`. Installing the full pinned `requirements.txt` (rather than just `defusedxml`) keeps it consistent with `build-and-test` and covers any other test script that reaches for a runtime dep.

**Nit (stale `fetch-depth: 0`):** Agreed and dropped from `build-and-test`'s checkout. The only consumer of full history, "Check for non-docs changes", now lives entirely in `check-diff`, so `build-and-test` gets the default shallow checkout. `check-diff` keeps `fetch-depth: 0` since its diff needs it.

aunger:
🤖 Reviewer

**LGTM.**

Verified ff87c5e against the diff directly (not just the description):

- `shell-tests` now runs "Install Python script dependencies" (`pip install --require-hashes -r scripts/requirements.txt`), gated on `needs.check-diff.outputs.needs_full_build == 'true'`, ahead of "Install lint tools for shell tests" and "Run shell script unit tests" — the same relative order the original combined job had, so `test_summarize_preflight_integration.sh`'s `defusedxml` import resolves again.
- `build-and-test`'s checkout dropped `fetch-depth: 0`; only `check-diff` keeps it, which is correct since it's the sole remaining consumer of full history.

CI on this commit (run 29715075141) is still in progress for `shell-tests`/`build-and-test`, with `check-diff` already green; not blocking on it per review guidelines, but flagging it's not yet confirmed complete.

No other concerns.

aunger:
### Verification plan

**Before merging** (must resolve before PR #739 can be merged):
- [ ] #741

None.

aunger:
🤖 Verification Agent

## Summary of before-merging verification

Per the [Verification Planner report](https://github.com/aunger/gallery-button-for-pixel-camera/pull/739#issuecomment-5018609170), one tracking issue gated this PR:

| Item | Result |
|---|---|
| #741: Add `shell-tests` and `check-diff` to branch protection required status checks | **NOT AUTOMATABLE** |

Details posted on [#741](https://github.com/aunger/gallery-button-for-pixel-camera/issues/741#issuecomment-5018620697): I confirmed via the GitHub API (read-only `rules/branches/main` endpoint) that `build-and-test` **is** currently a required status check on `main` (ruleset 15758845, "Require PR & checks"), so the action described in the issue is genuinely needed, not conditional. Actually editing that ruleset requires repo-administration write access, which this session's token does not have (confirmed 403 on the classic branch-protection endpoint), and is also outside a Verification Agent's permitted repository-changing actions regardless of credentials. This step needs a human with admin access to the repo's branch rules.

As additional context (not itself a gating item from the plan): the latest CI run on this PR's head commit (ff87c5e, [run 29715075141](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/29715075141)) shows `check-diff`, `shell-tests`, and `build-and-test` all green, corroborating the Reviewer's earlier confirmation that the job split works correctly.

## Outcome

No item failed, but #741 remains open as not-automatable and still gates the merge. Verification incomplete pending that manual branch-protection update.

PR #740, opened 2026-07-20
Hold test3a's post-tap poll stable for 3 s to close a foreground-flip transition race

### LINKED ISSUE #736: Add stableForMs stability window to test3a's tapOverlayAndAwait poll (mirror test2a/test4a)
Follow-on from PR #734's review: https://github.com/aunger/gallery-button-for-pixel-camera/pull/734#issuecomment-5015486609

test2a/test4a pair their foreground-package check with `stableForMs = 3_000L` on the `tapOverlayAndAwait` poll, to guard against a transient app-transition frame satisfying the predicate early. PR #734 added the foreground-package check to test3a's poll predicate and final assertion (fixing #733), but did not add a stability window.

The reviewer flagged a narrow theoretical race: the UiAutomator foreground package could update to the mock gallery slightly before its first green frame composites, so the poll could end on a stale still-green screenshot that happens to coincide with the already-updated foreground package.

This is a follow-on item and does not block merging PR #734. The reviewer explicitly did not request this as a change to #734: "Issue #733 did not ask for this, and the old test3a poll had no stability window either, so I am not requesting a change, just flagging it in case it is worth considering."

aunger:
🤖 Author

Fixes #736.

## Problem

test2a and test4a pair their foreground-package check with `stableForMs = 3_000L` on the `tapOverlayAndAwait` poll, so a transient app-transition frame cannot satisfy the predicate early. PR #734 added the foreground-package check to test3a's poll predicate (fixing #733) but left the poll's stability window at its `0L` default.

That leaves a narrow transition race, flagged in the #734 review: the UiAutomator foreground package can flip to the mock gallery a frame or two before the gallery's first GREEN content composites. With a zero-length window, the poll can then stop on the stale, still-green pre-tap `MockCameraActivity` frame at the instant the foreground package has already updated, so `greenCoverage > 0.40f && foregroundPackage == MOCK_GALLERY_PACKAGE` holds against a coincidental stale frame rather than the gallery's own render.

## Change

- Pass `stableForMs = 3_000L` to test3a's `tapOverlayAndAwait` call, mirroring test2a/test4a. Both predicate conditions must now hold continuously for 3 s, which outlasts the transition and the whole ~1.4 s cold-start window, so the returned frame is the gallery's own rendered GREEN photo rather than a stale one.
- Update the KDoc and the inline comment to distinguish the two failure modes: the no-op tap (caught by the foreground-package check, which stability alone cannot catch since a frozen screen satisfies any number of samples) versus this transition-frame race (caught by the stability window).

This is a test-only change; no production code is touched, and it does not loosen any existing test's requirements. The instrumentation E2E suite that exercises test3a runs on the emulator in CI.

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
@@ -261,11 +261,19 @@ class GalleryButtonVisualE2ETest {
      * The foreground-package check is load-bearing (issue #733): MockCameraActivity already fills
      * the screen with close to 100% GREEN before the tap, so coverage alone stays above 40% even
      * when the tap is a no-op and the gallery never launches (the #241/#705 regression class this
-     * test exists to catch). A stability requirement would not help either: the no-op failure mode
-     * is a screen that never changes, so any number of consecutive samples is satisfied just as
+     * test exists to catch). A stability requirement does not catch that no-op failure mode: the
+     * no-op screen never changes, so any number of consecutive samples is satisfied just as
      * trivially as one. Requiring the mock gallery in the foreground, mirroring test2a/test4a's
      * safeguard, makes a no-op tap time out and fail loudly instead of passing on the stale
      * pre-tap frame.
+     *
+     * The `stableForMs = 3_000L` window guards a separate, narrower race (issue #736): the
+     * UiAutomator foreground package can flip to the mock gallery a frame or two before the
+     * gallery's first GREEN content composites, so a zero-window poll could stop on the stale,
+     * still-green pre-tap MockCameraActivity frame the instant the foreground package has already
+     * updated. Requiring both conditions to hold continuously for 3 s, mirroring test2a/test4a,
+     * outlasts that transition and the whole ~1.4 s cold-start window, so the returned frame is
+     * the gallery's own rendered GREEN photo rather than a coincidental stale one.
      */
     @Test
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

Reviewed the diff against issue #736. The fix is exactly what was asked for: `stableForMs = 3_000L` added to test3a's `tapOverlayAndAwait` call, mirroring test2a/test4a's existing usage verbatim (same argument, no `timeoutMs` override, so it still falls back to the shared 15 s default).

I checked the semantics against `E2EFixture.tapOverlayAndAwait` / `captureScreenUntil` (app/src/androidTest/java/com/gb4pc/e2e/E2EFixture.kt:623-630 and 792-808): `stableForMs` requires the predicate to hold continuously, across re-samples, for that whole wall-clock duration before the poll stops early. That closes the race described in the issue: a stale, still-green pre-tap frame that happens to coincide with an already-flipped foreground package can no longer satisfy the predicate for 3 s straight, since the gallery's own frame (or an intermediate non-matching frame) will supersede it well within that window. This doesn't weaken the existing foreground-package check from #733/#734, which is still what catches the no-op-tap failure mode; the two guards are independent, as the updated comments correctly explain.

The KDoc and inline-comment rewrites accurately distinguish the two failure modes (no-op tap vs. this transition race) rather than conflating them. Scope is tight: one file, test-only, no loosening of any existing assertion. `kotlin-ktlint` has already reported success on this head commit, and the added lines are within the file's existing line-length norms.

Nothing to request. LGTM.

PR #742, opened 2026-07-20
Unify tap tests' post-tap failure reporting via a shared helper

### LINKED ISSUE #737: Unify the tap tests' post-tap failure reporting (test2a/test3a/test4a)
Follow-on from PR #734 (reviewer's non-blocking note), out of scope for that PR.

PR #734 changed `test3a_populatedGalleryShowsGreenAfterTap` in `app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt` to collect its two post-tap failure conditions (GREEN coverage and foreground package) into a list and emit a single `fail()` with both reasons joined, so both surface when both fail instead of the first `fail()` throwing and hiding the second.

`test2a_emptyGalleryNoGreenAfterTap` and `test4a_secureCameraLockedEmptyGalleryNoGreen` still use the older two sequential `if (...) fail(...)` blocks, which report only the first failing condition (the coverage check throws before the foreground check runs). The file now carries two different failure-reporting shapes across three very similar tap tests.

Adopt test3a's combined-failure-reporting shape in test2a and test4a as well, so all three tap tests report every failing post-tap condition consistently. If the collect-and-join logic reads cleanly as a small shared helper, factor it out rather than repeating it three times.

Scope note: PR #734 (issue #733) only asked to add test3a's foreground-package check, so unifying the reporting style across the other two tests was deliberately deferred here.

aunger:
Fixes #737

## Problem

`GalleryButtonVisualE2ETest` carried two different post-tap failure-reporting shapes across three very similar tap tests. PR #734 changed `test3a` to collect its post-tap failure conditions into a list and emit a single `fail()` with every reason joined, so both surface when both fail. `test2a` and `test4a` still used the older two sequential `if (...) fail(...)` blocks, where the coverage check throws before the foreground check runs, hiding the second failing condition.

## Change

Two commits, split by concern:

1. **Extract shared helper (pure refactoring).** Factor `test3a`'s collect-and-join logic into a private `reportPostTapFailures(testName, reasons)` helper that fails the test with every reason joined into one message, or is a no-op when the list is empty. `test3a` now calls the helper; its output is byte-for-byte unchanged.
2. **Adopt combined reporting in `test2a` and `test4a` (behavior change).** Both tests now collect every failing post-tap condition into a `buildList` and hand it to the shared helper, so all three tap tests report every failing post-tap condition consistently.

## Notes

The pass/fail thresholds are unchanged in every test (`coverage >= 0.10f` for `test2a`/`test4a`, and the same foreground-package equality checks), so no test requirement is loosened. Only the failure message changes: when both post-tap conditions fail, the report now names both reasons instead of throwing on the first `fail()` and hiding the second.

Verified locally with ktlint (`scripts/lint.sh --check --only kotlin --all`), which passes. The instrumentation tests themselves run on CI (they require an emulator).

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
@@ -301,7 +301,8 @@ class GalleryButtonVisualE2ETest {
         // reasons rather than throwing on the first and hiding the second.
         val coverage = ColorMatch.coverageFraction(greenMask)
         val foreground = fixture.foregroundPackage()
-        val failures =
+        reportPostTapFailures(
+            "test3a_populatedGalleryShowsGreenAfterTap",
             buildList {
                 if (coverage <= 0.40f) {
                     add(
@@ -316,10 +317,8 @@ class GalleryButtonVisualE2ETest {
                             "green mock camera in front)",
                     )
                 }
-            }
-        if (failures.isNotEmpty()) {
-            fail("test3a_populatedGalleryShowsGreenAfterTap: " + failures.joinToString("; "))
-        }
+            },
+        )
     }
 
     // test4a: Secure camera + empty gallery: no GREEN after tap---------------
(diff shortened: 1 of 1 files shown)


aunger (commented):
🤖 Reviewer

Verified this against both the issue and the actual diffs (not just the PR description), including a direct byte-for-byte comparison of test3a's pre- and post-PR logic.

**Correctness**
- `reportPostTapFailures(testName, reasons)` is a straightforward, correctly-scoped private helper: fails with every reason joined by `"; "`, no-ops when the list is empty. Matches what the issue asked for ("factor it out rather than repeating it three times").
- test3a's call site is a pure mechanical extraction — I diffed the old inline `val failures = buildList {...}; if (failures.isNotEmpty()) fail(...)` against the new helper call and the message text is unchanged. The "byte-for-byte unchanged" claim in the PR body checks out.
- test2a and test4a now build the `reasons` list the same way test3a does, so `foreground` is computed unconditionally before either condition is evaluated — that's the actual behavior fix (previously the coverage `fail()` threw first and `foregroundPackage()` was never reached). `foregroundPackage()` is just `UiDevice.currentPackageName`, so evaluating it unconditionally is free; no new flakiness risk.
- Pass/fail thresholds (`coverage >= 0.10f` for test2a/test4a, `<= 0.40f` for test3a, and the foreground equality checks) are byte-identical to before. No test got weaker.
- test5a was correctly left alone — it has one combined boolean condition (`isLetterboxedBand`), not two independent failure modes, so the issue's ask doesn't apply there. Good scope discipline.

**Commit structure**
Clean split: commit 1 is pure refactoring (helper extraction + test3a rewire only), commit 2 is the actual behavior change (test2a/test4a). Makes the review easy — I could confirm "no behavior change" for commit 1 in isolation.

**CI**
`kotlin-ktlint` is already green, consistent with the PR's claim of a clean local ktlint run. `build-and-test`/`analyze-kotlin` were still in progress at review time; not blocking on those per our review process.

**Nits**
None. This is a small, well-targeted change that does exactly what #737 asked for, with reasoning documented in the new helper's kdoc.

No changes requested.

Issue #723, opened 2026-07-17
Harden remaining unhashed pip installs outside lint tool scope (requirements.txt, semgrep)


aunger:
PR #715 added `pip install --require-hashes` to every lint-tool install site (`scripts/requirements-lint.txt`), but two other pip install sites remain unhardened. This was explicitly out of scope for #699 ("lint tool installs"):

- `scripts/requirements.txt` (defusedxml/requests/PyYAML), installed unhashed at two sites in `.github/workflows/build.yml`.
- `semgrep.yml`'s unversioned `pip install semgrep`.

Consider applying the same hash-pinning (or at minimum version-pinning) approach used for the lint tools to these installs, for full pip-install supply-chain coverage.

Source: https://github.com/aunger/gallery-button-for-pixel-camera/pull/715#pullrequestreview-4719372258 (nit 2)

This is a follow-on item and does not block merging PR #715.

PR #743, opened 2026-07-20
Verify every resolved Gradle artifact against pinned SHA-256 checksums

### LINKED ISSUE #714: Harden the Gradle build with dependency verification metadata

aunger:
Fixes #714.

Adds Gradle dependency verification so every plugin, wrapper-resolved plugin, dependency, and metadata artifact the build resolves is checked against a pinned SHA-256, extending the "verify what you download" posture already applied to ktlint (SHA-256 in `scripts/install-ktlint.sh`) and the Python lint tools (pip `--require-hashes`, #723).

## What changed

- **`gradle/verification-metadata.xml`** (new): pins 874 SHA-256 artifact checksums across 484 components. Generated on Linux with the full Android toolchain (JDK, Android SDK platform `android-35`, `build-tools;35.0.0`) against the superset of resolving tasks (`assembleDebug`, `assembleRelease`, `assembleDebugAndroidTest`, both `e2e-mock-*` `assembleDebug`, `testDebugUnitTest`) using a fresh `GRADLE_USER_HOME`, so it covers every configuration the CI workflows resolve.
- **`gradle/wrapper/gradle-wrapper.properties`**: adds `distributionSha256Sum` pinning the Gradle 8.9 distribution zip, so the wrapper verifies the distribution before running it.
- **`scripts/regenerate-gradle-verification.sh`** (new) and **`gradle/README.md`** (new): the exact regeneration command and the documentation for when and how to regenerate.
- **`scripts/test_verification_metadata.sh`** (new): a guard test (picked up automatically by the `scripts/test_*.sh` runner in `build.yml`) asserting the file stays present, well-formed, non-empty, and in its strict configuration.

## Scope decisions (per the issue's implementation plan)

- **`sha256`-only, no PGP signatures** (`verify-signatures=false`). Many `google()` artifacts are unsigned; signature verification would add a keyring and a keyserver dependency. Checksum-only matches the ktlint/pip precedent.
- **Metadata files pinned too** (`verify-metadata=true`), so `.pom`/`.module` files are covered.
- **Strict mode**; no `--dependency-verification lenient` anywhere in CI.
- **Generated on Linux** to match the merge-gating `ubuntu-latest` CI: the OS-classified `aapt2` artifact is recorded as `aapt2-8.7.3-...-linux.jar`, so a file generated on another OS would fail verification on CI.
- The **wrapper JAR** is deliberately out of scope for this file (it is read before the file is consulted); the committed, reviewed JAR plus the distribution `distributionSha256Sum` pin cover the wrapper/distribution path.

## Validation

Beyond the automated suite, verified locally on a full Android toolchain:

- **Positive**: a full build (`assembleDebug assembleRelease assembleDebugAndroidTest :e2e-mock-camera:assembleDebug :e2e-mock-gallery:assembleDebug testDebugUnitTest`) against a fresh `GRADLE_USER_HOME` with the committed metadata active re-downloaded and verified every artifact and succeeded (`BUILD SUCCESSFUL`, unit tests passed).
- **Negative**: corrupting one checksum caused Gradle to fail fast with `Dependency verification failed for configuration ':classpath' ... One artifact failed verification: gradle-8.7.3.jar`, confirming the mechanism is live (checksum then restored).

The complete `build.yml` run on this PR, including the emulator instrumented/E2E steps, is the authoritative end-to-end check for any configuration not exercised above; the instrumented tasks add no new external dependencies (the androidTest classpath is the one `assembleDebugAndroidTest` resolves, and `connectedE2EAndroidTest` consumes already-built APKs).

## Maintenance note

Any change that moves the dependency graph (an AGP/Kotlin/Compose/AndroidX/test-library version bump, a new dependency, or a new plugin) requires regenerating this file in the same commit via `scripts/regenerate-gradle-verification.sh`, or the build will fail verification. There is no automated dependency bumper in this repo, so the graph only moves on manual, reviewed changes.


aunger (commented):
🤖 Reviewer

Reviewed the full diff and independently verified the load-bearing facts. This is a well-executed, faithful implementation of the issue's plan.

## Verified

- **Distribution pin is correct.** `distributionSha256Sum=d725d707bfabd4dfdc958c624003b3c80accc03f7037b5122c4b1d0ef15cecab` matches Gradle's published `gradle-8.9-bin.zip.sha256` exactly (fetched from downloads.gradle.org).
- **Verification config is strict and scoped as decided.** The XML header has `verify-metadata=true` and `verify-signatures=false`; no `<trusted-artifacts>`/`<trusted-key>`/`<trusting>` rules exist that would broadly weaken it; 874 `<sha256>` entries are present.
- **Guard test passes and is wired into CI.** `scripts/test_verification_metadata.sh` passes 7/7 against the PR's committed files, and `build.yml`'s `for test_script in scripts/test_*.sh` sweep picks it up automatically.
- **Scripts and docs are sound** and match the plan: fresh `GRADLE_USER_HOME`, the correct superset of resolving tasks, the `ANDROID_HOME`/`ANDROID_SDK_ROOT` precondition, the Linux/`aapt2` OS-classifier rationale, and the same-commit-regeneration maintenance note.

## Optional, non-blocking observations (do not open a new round just for these)

1. `test_verification_metadata.sh` check (e) only greps `.github/workflows` for a lenient CLI downgrade. Verification can also be downgraded via `org.gradle.dependency.verification=lenient` in a committed `gradle.properties`. Extending the guard to that path would make it more complete, but its absence is not a correctness defect.
2. In the same check, the grep alternation `--dependency-verification[= ]lenient\|--dependency-verification lenient` is redundant: `[= ]` already matches the space form. Harmless.

## On CI

CI has not reported yet. The whole PR's correctness rests on the complete `build.yml` run (unit + instrumented/E2E) resolving every configuration against this file without a verification failure, which is the issue's explicit merge gate and cannot be reproduced in review. I am not blocking on it; the authoritative end-to-end validation is the green CI run, which the Orchestrator monitors. Nothing I could check statically is wrong.

LGTM, contingent on the CI gate the Orchestrator enforces.

aunger:
### Verification plan

**Before merging** (must resolve before PR #743 can be merged):
None.

- [ ] #744
- [ ] #745
- [ ] #746

aunger:
Please complete the "Optional, non-blocking observations" in this PR.
LOVE,
THE HUMAN

aunger:
🤖 Author

Addressed both optional observations on check (e) of `scripts/test_verification_metadata.sh` in commit ea8f5f2.

1. **Property path now guarded.** The check previously only scanned `.github/workflows` for a `--dependency-verification` CLI downgrade. It now also scans every committed `gradle.properties` (enumerated via `git ls-files`) for `org.gradle.dependency.verification=lenient` (or `off`), which weakens or disables verification the same way. I extended both paths to catch `off` as well as `lenient`, since disabling is the same class of downgrade the guard exists to prevent.

2. **Redundant alternation dropped.** As noted, `[= ]` already matches the space form, so the `\|--dependency-verification lenient` alternative was dead. The flag check is now a single extended-regex pattern (`--dependency-verification[= ](lenient|off)`).

Validated locally: the positive case still passes 7/7, and each new negative path (a lenient/off property in a committed `gradle.properties`, and a lenient/off flag in a workflow) now fails the guard as intended.

aunger (commented):
🤖 Reviewer

Re-reviewed after commit `ea8f5f2`, which addresses both observations the human asked to be completed. The change touches only `scripts/test_verification_metadata.sh` (+25/-6); the metadata, distribution pin, scripts, and docs are unchanged from the prior LGTM.

Verified the reworked check (e) end-to-end on `ea8f5f2`:

- **Positive**: clean tree still passes 7/7.
- **Negative** (each fails the guard exactly as intended, and the tree passes again once reverted):
  - `org.gradle.dependency.verification=lenient` appended to the tracked root `gradle.properties` -> FAIL.
  - `org.gradle.dependency.verification = off` (with surrounding spaces) in a newly `git add`-ed `app/gradle.properties` -> FAIL, confirming the `git ls-files -- '*gradle.properties'` enumeration and the whitespace-tolerant property regex both work.
  - `--dependency-verification=off` flag in `build.yml` -> FAIL.
  - `--dependency-verification lenient` (space form) in `codeql.yml` -> FAIL, confirming the single `[= ]` ERE covers both forms with no separate alternation.

Both points are correctly resolved: the property path is now guarded via `git ls-files`, and the redundant alternation is gone in favor of one extended-regex pattern. Extending coverage to `off` alongside `lenient` on both paths is a sensible strengthening, since disabling weakens the control the same way. The `2>/dev/null` on `git ls-files` degrades gracefully (the flag path still runs) if ever invoked outside a git checkout.

No new issues. LGTM.

Issue #746, opened 2026-07-20
Simplify redundant grep alternation in test_verification_metadata.sh


aunger:
Source: PR #743 review comment:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/743#pullrequestreview-4732129194

The reviewer's optional, non-blocking observation on PR #743:

> In the same check, the grep alternation `--dependency-verification[= ]lenient\|--dependency-verification lenient` is redundant: `[= ]` already matches the space form. Harmless.

## Ask

Simplify the grep pattern in `scripts/test_verification_metadata.sh` check (e) (added in PR #743) to drop the redundant second alternative, since `[= ]` already matches both the `--dependency-verification=lenient` and `--dependency-verification lenient` forms.

This is a follow-on item and does **not** block merging PR #743.

aunger:
🤖 Author

Already implemented, so closing as completed.

Commit `ea8f5f2` ("Extend the verification guard to gradle.properties and drop a redundant alternation"), already merged to `main`, dropped the redundant alternation in check (e) of `scripts/test_verification_metadata.sh`.
The flag pattern is now the single `--dependency-verification[= ](lenient|off)`; the `[= ]` class already matches both the `=` and space forms, so the second alternative is gone.

No further change is needed.

Issue #745, opened 2026-07-20
Extend verification-metadata guard test to catch a gradle.properties lenient downgrade


aunger:
Source: PR #743 review comment:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/743#pullrequestreview-4732129194

The reviewer's optional, non-blocking observation on PR #743:

> `test_verification_metadata.sh` check (e) only greps `.github/workflows` for a lenient CLI downgrade. Verification can also be downgraded via `org.gradle.dependency.verification=lenient` in a committed `gradle.properties`. Extending the guard to that path would make it more complete, but its absence is not a correctness defect.

## Ask

Extend `scripts/test_verification_metadata.sh`'s check (e) (added in PR #743) to also assert that no committed `gradle.properties` sets `org.gradle.dependency.verification=lenient` (or another non-strict value), not just that no CI workflow passes `--dependency-verification lenient` on the command line.

This is a follow-on item and does **not** block merging PR #743.

aunger:
🤖 Author

Already implemented, so closing as completed.

Commit `ea8f5f2` ("Extend the verification guard to gradle.properties and drop a redundant alternation"), already merged to `main`, extended check (e) of `scripts/test_verification_metadata.sh` to this exact path.
The guard now scans every committed `gradle.properties` (enumerated via `git ls-files -- '*gradle.properties'`) and fails if any sets `org.gradle.dependency.verification` to `lenient` or `off`, which are the only non-strict values, not just the CI CLI `--dependency-verification` flag.

No further change is needed.

PR #747, opened 2026-07-20
Authenticate the committed Gradle wrapper JAR against Gradle's published checksum

### LINKED ISSUE #744: Add a first-party SHA-256 integrity check for the Gradle wrapper JAR
Source: issue #714 description, "Decision 3 -- wrapper and distribution integrity: handle separately from this file":
https://github.com/aunger/gallery-button-for-pixel-camera/issues/714

The Gradle wrapper JAR (`gradle/wrapper/gradle-wrapper.jar`) cannot be covered by `gradle/verification-metadata.xml` (added in PR #743) because it is read before that file is consulted. Issue #714's implementation plan recommended pinning the Gradle distribution now (done in PR #743 via `distributionSha256Sum`) and filing wrapper-JAR checking as "a separate optional follow-up so this issue stays focused":

> If wrapper-JAR integrity is wanted, add a tiny self-written SHA-256 check step (consistent with the repo's preference for first-party scripts over third-party actions), or defer to a follow-up issue.

PR #743's description confirms this scope decision: the wrapper JAR is "deliberately out of scope for this file"; the committed, reviewed JAR plus the distribution pin cover the wrapper/distribution path for now.

## Ask

Add a small first-party check (matching the repo's ktlint SHA-256 pattern in `scripts/install-ktlint.sh`) that verifies `gradle/wrapper/gradle-wrapper.jar` against a pinned checksum, so a substituted or corrupted wrapper JAR is caught rather than silently trusted.

This is a follow-on item and does **not** block merging PR #743.

aunger:
Fixes #744.

## Problem

`gradlew` reads and executes `gradle/wrapper/gradle-wrapper.jar` before `gradle/verification-metadata.xml` is ever consulted (issue #714, Decision 3), so Gradle dependency verification cannot cover the wrapper JAR itself.
Until now a substituted or corrupted wrapper JAR would be trusted silently.

## Change

Add check (g) to `scripts/test_verification_metadata.sh`, the existing build-integrity guard test.
It computes the committed wrapper JAR's SHA-256 and compares it to a pinned value, mirroring the pinned-checksum compare in `scripts/install-ktlint.sh`.

The pin is authenticated against Gradle: it is the checksum Gradle officially publishes for its wrapper JAR (the same authoritative source GitHub's `gradle/wrapper-validation-action` verifies against), not the current file's own hash.
Pinning the published value means the guard authenticates the JAR against Gradle rather than trusting whatever bytes happen to be committed.
The committed JAR was confirmed to be the genuine Gradle wrapper JAR by matching `https://services.gradle.org/distributions/gradle-8.14-wrapper.jar.sha256`.

The check runs in the existing `shell-tests` CI job via the `scripts/test_*.sh` discovery, alongside the distribution-pin guard (check (f)) it sits next to, so a substituted or corrupted wrapper JAR now fails CI.

## Note: wrapper JAR / distribution version skew (pre-existing)

Authenticating the JAR surfaced that the committed wrapper JAR is Gradle **8.14**, while `gradle-wrapper.properties` pins the **8.9** distribution.
This is functionally fine (a newer wrapper JAR launches an older distribution without issue) and predates this PR, so it is left as-is here.
Aligning the two versions is a reasonable follow-up but is out of scope for the wrapper-JAR integrity check.

## On the sibling issues #745 and #746

Both were already implemented in commit `ea8f5f2`, already merged to `main`, so this PR contains no code for them:

- #745: check (e) already scans every committed `gradle.properties` (via `git ls-files`) for `org.gradle.dependency.verification` set to `lenient` or `off`, covering all non-strict values, not just the CI CLI flag.
- #746: the redundant grep alternation is already gone; the pattern is the single `--dependency-verification[= ](lenient|off)`.

## Testing

Covered by the automated shell-test suite (`scripts/test_verification_metadata.sh`, run in the `shell-tests` CI job).
Verified locally that the suite passes with the JAR intact (8/8) and that check (g) fails with a non-zero exit when the JAR's bytes do not match the pin.

### DIFF
--- a/gradle/README.md
+++ b/gradle/README.md
@@ -43,7 +43,7 @@ Never commit that flag into CI, and do not commit a file regenerated on a non-Li
 The Gradle wrapper verifies the downloaded distribution against this checksum before running it, so a substituted or corrupted distribution fails loudly.
 
 The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be covered by that file.
-Instead, `scripts/test_verification_metadata.sh` check (g) pins the JAR's SHA-256 to the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.
+Instead, `scripts/test_verification_metadata.sh` check (g) verifies the JAR's SHA-256 against a pinned copy of the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.
 
 When bumping the Gradle version, update both `distributionUrl` and `distributionSha256Sum` (the published checksum is at `https://services.gradle.org/distributions/gradle-<version>-bin.zip.sha256`), then regenerate `verification-metadata.xml`.
 If the wrapper JAR itself changes, also update its pinned SHA-256 in `scripts/test_verification_metadata.sh` (the published checksum is at `https://services.gradle.org/distributions/gradle-<version>-wrapper.jar.sha256`).
--- a/gradle/README.md
+++ b/gradle/README.md
@@ -42,6 +42,8 @@ Never commit that flag into CI, and do not commit a file regenerated on a non-Li
 `distributionSha256Sum` pins the Gradle 8.9 distribution (`gradle-8.9-bin.zip`) to its published SHA-256.
 The Gradle wrapper verifies the downloaded distribution against this checksum before running it, so a substituted or corrupted distribution fails loudly.
 
-The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be, and is not, covered by that file.
+The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be covered by that file.
+Instead, `scripts/test_verification_metadata.sh` check (g) pins the JAR's SHA-256 to the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.
 
 When bumping the Gradle version, update both `distributionUrl` and `distributionSha256Sum` (the published checksum is at `https://services.gradle.org/distributions/gradle-<version>-bin.zip.sha256`), then regenerate `verification-metadata.xml`.
+If the wrapper JAR itself changes, also update its pinned SHA-256 in `scripts/test_verification_metadata.sh` (the published checksum is at `https://services.gradle.org/distributions/gradle-<version>-wrapper.jar.sha256`).
--- a/scripts/test_verification_metadata.sh
+++ b/scripts/test_verification_metadata.sh
@@ -19,6 +19,11 @@
 #       lenient|off flag in a CI workflow, and no org.gradle.dependency.verification
 #       lenient|off property in a committed gradle.properties
 #   (f) gradle-wrapper.properties pins the distribution via distributionSha256Sum
+#   (g) gradle-wrapper.jar matches the SHA-256 Gradle officially publishes for its
+#       wrapper JAR (issue #744): the wrapper JAR is executed before
+#       verification-metadata.xml is consulted, so it cannot be covered there; this
+#       authenticates it against Gradle instead of trusting whatever bytes are
+#       committed, catching a substituted or corrupted wrapper JAR
 #
 # Always exits 0 on success, non-zero on failure.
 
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

Verified the substantive claims rather than taking them on faith:

- Checked out the branch and computed `sha256sum` on the committed `gradle/wrapper/gradle-wrapper.jar`: it is `7d3a4ac4de1c32b59bc6a4eb8ecb8e612ccd0cf1ae1e99f66902da64df296172`, matching the pin.
- Fetched `https://services.gradle.org/distributions/gradle-8.14-wrapper.jar.sha256` directly: it returns the same value. The pin is genuinely authenticated against Gradle's published checksum, not just against the committed bytes.
- Ran `scripts/test_verification_metadata.sh` locally: 8/8 pass with the JAR intact. Corrupted the JAR's bytes and re-ran: check (g) correctly fails with a non-zero exit and the expected/actual mismatch message.
- Confirmed the version-skew note is accurate and pre-existing: at the PR's base commit (`58a1e98`), the committed JAR already hashed to the 8.14 checksum while `gradle-wrapper.properties` already pinned the 8.9 distribution. This PR introduces neither the skew nor a fix for it, so treating it as an out-of-scope follow-up is correct.
- Confirmed `scripts/test_*.sh` discovery in `.github/workflows/build.yml`'s `shell-tests` job picks up this script automatically (no workflow edit needed), and the `shell-tests` check on this PR is already green.
- Confirmed #745/#746 are indeed already satisfied by `ea8f5f2`, which is already an ancestor of this PR's base, not something this PR needs to touch.

One real gap: **`gradle/README.md` is now stale and should be updated in this PR.** Its "wrapper/gradle-wrapper.properties -- distribution pin" section states: "The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be, and is not, covered by that file." That's exactly the gap issue #744 opened to close, and this PR closes it, check (g) is precisely an automated, first-party checksum guard for that JAR. But the README, which lives in the same directory and specifically inventories this directory's "verify what you download" controls, still tells readers the wrapper JAR has no such control and relies only on manual review. That's misleading now: a reader auditing this directory's integrity posture from the README alone would miss that an automated guard exists. Since this PR's entire purpose is adding that guard, updating this adjacent, directly-on-topic doc isn't scope creep, it's completing the change.

Everything else checks out: the approach matches the ktlint precedent as asked, the check is correctly wired into existing CI discovery, the commit message carries no attribution byline, and the PR correctly declines to bundle in #745/#746 or the version-skew fix.

Decision: Changes requested (update `gradle/README.md`'s wrapper section to reflect the new check (g) guard before merge).

aunger:
🤖 Author

Agreed, and fixed in `51afcba`. You're right that leaving `gradle/README.md` saying the wrapper JAR "cannot be, and is not, covered" by any automated control is the exact stale claim this PR's guard closes, and updating that adjacent, on-topic doc completes the change rather than expanding scope.

The wrapper section now reads:

> The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be covered by that file.
> Instead, `scripts/test_verification_metadata.sh` check (g) pins the JAR's SHA-256 to the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.

I also added a maintenance line so the pin doesn't silently drift on a wrapper-JAR change:

> If the wrapper JAR itself changes, also update its pinned SHA-256 in `scripts/test_verification_metadata.sh` (the published checksum is at `https://services.gradle.org/distributions/gradle-<version>-wrapper.jar.sha256`).

I kept the accurate half of the old sentence (the JAR genuinely cannot be covered by `verification-metadata.xml`, since it is read first) and only dropped the now-false "and is not covered" tail. `scripts/test_verification_metadata.sh` still passes 8/8 locally.

aunger (commented):
🤖 Reviewer

Round 2. The single blocking item from round 1, the stale `gradle/README.md` wrapper section, is resolved in `51afcba`, and I re-verified the whole change from scratch on the branch rather than trusting the responses:

- **README fix is correct and complete.** The old sentence claimed the wrapper JAR "cannot be, and is not, covered" by any automated control. It now drops only the false "and is not" tail, keeps the accurate "cannot be covered by `verification-metadata.xml`" half, and adds a sentence pointing to check (g). The maintenance line for updating the pin on a wrapper-JAR change is a sensible, on-topic addition. This closes exactly the misleading-audit gap I raised, without scope creep.
- **The pin is genuinely authenticated against Gradle.** `sha256sum` of the committed `gradle/wrapper/gradle-wrapper.jar` is `7d3a4ac4de1c32b59bc6a4eb8ecb8e612ccd0cf1ae1e99f66902da64df296172`, and `https://services.gradle.org/distributions/gradle-8.14-wrapper.jar.sha256` returns that identical value. So the pin is Gradle's published checksum, not the committed file's self-hash.
- **The check works both ways.** Ran `scripts/test_verification_metadata.sh` on the branch: 8/8 pass. Appended a byte to the JAR and re-ran: check (g) fails with a non-zero exit and the expected/actual mismatch message; restored, back to 8/8.
- **CI:** `shell-tests` is green on this commit, along with the lint/security checks; `build-and-test` is still running. The only failing check is "No blocking labels", which fails by design during the review cycle.
- Version-skew note (8.14 JAR vs 8.9 distribution) remains accurate and pre-existing, and declining #745/#746 and the skew fix remains correct.

No remaining concerns.

Decision: LGTM

gradle/README.md
```diff
@@ -42,6 +42,8 @@ Never commit that flag into CI, and do not commit a file regenerated on a non-Li
 `distributionSha256Sum` pins the Gradle 8.9 distribution (`gradle-8.9-bin.zip`) to its published SHA-256.
 The Gradle wrapper verifies the downloaded distribution against this checksum before running it, so a substituted or corrupted distribution fails loudly.
 
-The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be, and is not, covered by that file.
+The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be covered by that file.
+Instead, `scripts/test_verification_metadata.sh` check (g) pins the JAR's SHA-256 to the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.
```
aunger:
```suggestion
Instead, `scripts/test_verification_metadata.sh` check (g) authenticates the JAR's SHA-256 against its record of the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.
```

I initially understood "pins" to mean *sets* the pinned value, not  *checks* the pinned value. This script can't set it though.

My suggestion probably also isn't the best way to say this; feel free to tweak for clarity.

aunger: 🤖 Author

Good catch, that ambiguity is real: "pins" reads as though the script *sets* the pinned value, when all it does is *check* against it. Fixed in `34d17f5`. I went with "verifies" rather than "authenticates" so the verb itself signals a check, and "a pinned copy of the checksum Gradle officially publishes" to locate the stored value without implying the script writes it:

> Instead, `scripts/test_verification_metadata.sh` check (g) verifies the JAR's SHA-256 against a pinned copy of the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.

This stays consistent with the maintenance line just below, which is where the pinned copy actually gets set ("update its pinned SHA-256 in `scripts/test_verification_metadata.sh`"). Suite still passes 8/8 locally.

aunger (commented):
🤖 Reviewer

Round 3. New commit `34d17f5` since my LGTM is a doc-only wording refinement in `gradle/README.md`: "check (g) pins the JAR's SHA-256" becomes "check (g) verifies the JAR's SHA-256 against a pinned copy of the checksum Gradle officially publishes". This is more accurate, the script verifies against the pin rather than setting it, and it changes no code or behavior.

Re-verified at the new head (`34d17f5`): `scripts/test_verification_metadata.sh` still passes 8/8, including check (g) matching the pinned wrapper-JAR SHA-256. No functional change from the commit I already approved; the pin, the negative-test behavior, and the CI wiring are all unchanged.

No remaining concerns.

Decision: LGTM

PR #750, opened 2026-07-20
Retry transient GitHub API errors in the mutually-exclusive-label check

### LINKED ISSUE #749: enforce_mutually_exclusive_labels.py has no retry on transient GitHub API errors, so a 503 fails the job with no real conflict
## What happened

On PR #734, the "Enforce mutually exclusive label sets" check failed twice (runs 29709214086 and 29709749473; jobs 88250751626 and 88251960796). Both job logs show the identical cause:

```
Error fetching issue #734: HTTP Error 503: Service Unavailable
```

Neither run reflects a real mutually-exclusive-label conflict. `scripts/enforce_mutually_exclusive_labels.py` makes exactly one un-retried `gh_api()` call (`urllib.request.urlopen`, no retry loop) to fetch the issue/PR's current labels before computing which conflict; any non-2xx response, including a transient 503, raises and `main()` returns 1 with no further attempt.

This lines up with a window of GitHub-wide API instability observed directly in the same session, via `scripts/update_gh_labels.sh` and the GitHub MCP tools, where repeated 503s cleared on a second or third retry every time.

## Context

Issue #614 (closed) fixed this same script so it correctly propagates real failures as a nonzero exit instead of always returning 0 — that fix is working as intended here; a genuinely-failed label reconciliation now shows up as a red check, as designed. The gap this issue is about is the natural follow-up: now that failures are visible, a purely transient one (no real conflict, no real reconciliation failure) still shows up exactly the same as a real one, with no retry to distinguish "GitHub blipped" from "the reconciliation itself failed."

This check isn't in the repo's required-status-checks list (see the companion issue, #748, filed alongside this one, which covers a separate bug where `ci_monitor.py` treats this failure as blocking the PR anyway). So on its own this failure doesn't block a real merge — but it does read as a false red X, and it fed directly into #748's false `Blocked` terminal.

## Suggested fix

Give the `gh_api()` call(s) in `scripts/enforce_mutually_exclusive_labels.py` a small retry-with-backoff for transient errors (5xx, and maybe 429), matching the spirit of `scripts/update_gh_labels.sh`'s and `agents/dev_orchestration.md`'s existing "retry once, transient GitHub errors are expected" guidance for this exact class of instability. Keep the existing 404-is-benign handling in `remove_labels()` untouched — only the transient-vs-real distinction needs to change, not the already-correct "label already gone" case.

aunger:
🤖 Author

Fixes #749.

## Problem

`scripts/enforce_mutually_exclusive_labels.py` made exactly one un-retried `gh_api()` call (`urllib.request.urlopen`, no retry loop) to fetch the issue/PR's current labels.
Any non-2xx response, including a transient 503, raised and made `main()` return 1 with no further attempt.
On PR #734 this failed the "Enforce mutually exclusive label sets" check twice with `HTTP Error 503: Service Unavailable`, with no real mutually-exclusive-label conflict, during a window of GitHub-wide API instability.

Since #614 (which correctly made real reconciliation failures surface as a red check) a purely transient blip reads exactly like a real failure, with no retry to distinguish "GitHub blipped" from "the reconciliation itself failed."

## Fix

Give `gh_api()` a small retry-with-backoff for transient failures only:

- A transient error (a 5xx, a 429 rate-limit response, or a network-level `URLError`) is retried up to `MAX_ATTEMPTS` (4) times with exponential backoff (1s, 2s, 4s) before the error is allowed to propagate.
- A real error is never retried: any other `HTTPError` (a 4xx other than 429, such as a 404 already-gone or a 422 validation failure) is raised immediately, so the existing 404-is-benign handling in `remove_labels()` is untouched.

Because the retry lives inside `gh_api()`, both the issue-fetch `GET` and the label-removal `DELETE`s benefit (the `DELETE` is idempotent, so retrying a label removal is safe: a re-run that finds the label already gone gets a benign 404).

This matches the "retry once, transient GitHub errors are expected" guidance already followed by `scripts/update_gh_labels.sh` and `agents/dev_orchestration.md` for this exact class of instability.
Exit codes are unchanged; a nonzero exit now reflects a persistent failure rather than a passing blip.

## Tests

Added unit tests in `scripts/test_enforce_mutually_exclusive_labels.py`:

- `_is_transient_error` classification (5xx / 429 / `URLError` transient; 404 / 422 / other 4xx / non-URL exceptions not transient).
- `gh_api` retries a 503, a 429, and a `URLError` then returns the following success; exhausts retries and raises the last transient error; does not retry a 404 or a 422 (raised on the first attempt, no sleep); and grows the backoff exponentially.

All 90 tests in the file pass, and `ruff check` / `ruff format --check` are clean.

### DIFF
--- a/scripts/enforce_mutually_exclusive_labels.py
+++ b/scripts/enforce_mutually_exclusive_labels.py
@@ -28,7 +28,9 @@ Exit code:
        failure).
     1  ISSUE_NUMBER is not a valid integer, the issue/PR's current labels
        could not be fetched, or removing a conflicting label failed for a
-       reason other than 404.
+       reason other than 404. A transient GitHub blip (a 5xx, a 429, or a
+       network error) is retried with exponential backoff first, so a nonzero
+       exit reflects a persistent failure, not a passing blip.
 
 Required environment variables:
     GITHUB_TOKEN        Personal access token or Actions secret with
@@ -41,6 +43,7 @@ Required environment variables:
 import json
 import os
 import sys
+import time
 import urllib.error
 import urllib.request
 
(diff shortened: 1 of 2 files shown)


aunger:
🤖 Reviewer

## Verdict: LGTM

The fix is correct, well-scoped to #749, and thoroughly tested. I verified locally against the PR head (d136a54): all 90 tests pass under `python3 -m unittest`, and `ruff check` / `ruff format --check` are clean.

Correctness points I checked and agree with:
- `_is_transient_error` checks `HTTPError` before `URLError` (subclass ordering is correct), retries only 5xx/429/network `URLError`, and treats every other 4xx (404, 422, other) as non-transient. The already-gone 404 path in `remove_labels` is therefore untouched.
- The retry lives inside `gh_api`, so both the issue-fetch `GET` and the label-removal `DELETE`s are covered. Both are idempotent, so retrying is safe. A `DELETE` that races to a 404 on a retry still lands in the benign 404 branch of `remove_labels`.
- No sleep after the final attempt (`attempt >= max_attempts` raises first), backoff is bounded (4 attempts, ~7s worst case), and exit codes are unchanged.

Non-blocking nits (take them or leave them; none block merge):

1. Inaccurate attribution in the retry-policy comment (and the PR description). The comment says the retry "matches the 'retry once, transient GitHub errors are expected' guidance already followed by scripts/update_gh_labels.sh and agents/dev_orchestration.md". `scripts/update_gh_labels.sh` contains no HTTP retry of its own; it exits 1 on any 5xx or unexpected status (see its label-remove and label-add `case` blocks). The retry-once guidance actually lives in `agents/dev_orchestration.md:98` ("Retry the same call once, since a non-2xx GitHub response can be transient") and is directed at the orchestrator re-invoking that script, not at the script retrying internally. Consider rewording to cite only `dev_orchestration.md`, or to say the orchestrator retries these calls, so a future reader who follows the pointer to `update_gh_labels.sh` does not go looking for a retry loop that is not there.

2. Minor wording: "retried up to `MAX_ATTEMPTS` (4) times" conflates attempts with retries. Four attempts is three retries, which matches the three backoff waits (1s, 2s, 4s). "attempted up to `max_attempts` times (up to 3 retries)" would read more precisely.

Out of scope, noted only as a possible future follow-up (not a request): a 403 is treated as non-transient and not retried. GitHub returns 403 for primary rate-limit exhaustion (and sometimes for secondary limits), so a rate-limited call would fail rather than back off. That is consistent with the scope stated in #749 (5xx and maybe 429) and with `update_gh_labels.sh` treating 403 as a hard error, and retrying a rate-limit 403 without honoring `Retry-After` would be riskier than the current behavior, so leaving it out here is the right call for this PR.

Nothing above changes my verdict. This is good to merge.

aunger:
### Verification plan

**Before merging** (must resolve before PR #750 can be merged):
None.

- [ ] #751
- [ ] #752

PR #753, opened 2026-07-20
Gate the --pr Blocked/Infra terminal on mergeable_state

### LINKED ISSUE #748: ci_monitor.py (--pr mode) reports Blocked from any failing check-run, bypassing mergeable_state and the branch's actual required-checks list
## What happened

PR #734 (issue #733) got a `Blocked by: enforce-exclusive-labels, No blocking labels` terminal from `ci_monitor.py --pr 734` twice, sending the PR back to the Author for nothing to fix. `build-and-test` and `No blocking labels` (the two required checks) settled clean on rerun, and the PR was in fact mergeable the whole time: the repo owner merged it manually shortly after.

`enforce-exclusive-labels` ("Enforce mutually exclusive label sets") is not a required status check at all. This repo's active branch ruleset ("Require PR & checks", id 15758845, `GET /repos/{owner}/{repo}/rulesets/15758845`) lists only:

```
required_status_checks: ["build-and-test", "No blocking labels"]
```

So a failure in `enforce-exclusive-labels` should never have produced a `Blocked` verdict.

## Root cause

In `--pr` mode, `parse_check_result()` (`scripts/ci_monitor/ci_monitor.py:413-432`) scans *every* check-run on the head commit. If any concludes `failure`/`action_required`, it returns `"Blocked"` (or `"Infra"` for cancelled/timed_out/etc.) directly.

Back in the main loop, `elif result in ("Blocked", "Infra"):` (line 1325) prints the terminal and ends the run immediately.

The `mergeable_state`-aware gating that correctly distinguishes "a required check failed" from "some other check failed but the PR is still mergeable" (lines 1282-1317: `clean`/`unstable` -> Clear, `behind`/`dirty` -> Blocked, `blocked` -> Infra) only runs on the `result == "all_passed"` path, i.e. only when *every* check-run on the commit succeeded. A failing non-required check never reaches that gate — it short-circuits straight to `Blocked` before `mergeable_state` is ever consulted.

This is a different, more general gap than the existing `label_gate_check_regex`/`[label gate]` mechanism (issue #516), which only special-cases one named check by pattern-matching; it doesn't address the Monitor blocking on *any* non-required check's failure.

## Evidence

- Ruleset `required_status_checks` for `main`: `build-and-test`, `No blocking labels` only.
- PR #734 job logs for the two `enforce-exclusive-labels` failures (runs 29709214086, 29709749473; jobs 88250751626, 88251960796) both show the job itself erroring out on a transient GitHub API 503, not any real label conflict (see the companion issue for that script's fix).
- PR #734 was merged by the repo owner immediately after the Orchestrator escalated, confirming the `Blocked` terminal was a false positive relative to GitHub's actual merge gate.

## Suggested fix

For `--pr` mode, don't let the raw per-check scan short-circuit the terminal. Instead, always resolve the terminal via `mergeable_state` (once it's known, i.e. not `"unknown"`), the same way the `all_passed` path already does — use the raw check-run scan only to drive the per-check summary/step/FAIL diagnostics, not the Clear/Blocked/Infra decision itself.

This doesn't require separately fetching or enumerating the required-checks list: GitHub already computes exactly this distinction into `mergeable_state`. `unstable` specifically means "mergeable, but a non-required check is failing/pending" — which is exactly the case this bug mishandles — while `dirty`/`blocked` mean a required check or another real merge blocker is in play. The all_passed path already relies on this correctly; the fix is to route the Blocked/Infra-from-raw-scan path through the same gate rather than bypassing it.

(An alternative would be to fetch the branch's effective rules via `GET /repos/{owner}/{repo}/rules/branches/{branch}` and filter the per-check scan to only the required contexts before deciding Blocked/Clear — this works with this repo's token, unlike the classic `branches/{branch}/protection` endpoint, which 403s for the token available in this session. But it's more moving parts than just trusting `mergeable_state`, which is already fetched on the all_passed path today.)

Scope: `--sha`/`--branch`/`--run-id` modes have no PR/`mergeable_state` concept (documented in the README as intentional), so this fix is `--pr`-mode only; those modes' current all-checks-count behavior stays as is.

aunger:
🤖 Author

Fixes #748

## Problem

In `--pr` mode, `parse_check_result()` returns `"Blocked"`/`"Infra"` whenever *any* check-run on the head commit concludes `failure`/`action_required` (or an infra conclusion), and the main loop's `elif result in ("Blocked", "Infra"):` branch printed that terminal immediately, before `mergeable_state` was ever consulted. Only the `all_passed` path consulted `mergeable_state`.

So a failing NON-required check short-circuited straight to a false `Blocked`. On PR #734 the advisory `enforce-exclusive-labels` check (not in the branch's `required_status_checks`, which are only `build-and-test` and `No blocking labels`) failed, and the Monitor reported `Blocked` twice, sending the PR back to the Author with nothing to fix, even though GitHub reported the PR mergeable the whole time (the owner merged it manually).

## Fix

Consult `mergeable_state` on the Blocked/Infra path too, mirroring the `all_passed` path's fresh `/pulls` fetch:

- `clean`/`unstable` -> the PR is mergeable despite the red check -> `Clear`.
- `unknown` -> GitHub has not recomputed yet -> keep polling (don't emit a possibly-false terminal).
- any un-mergeable state (`behind`/`dirty`/`blocked`) -> a real block -> fall through to the raw scan's `Blocked`/`Infra` terminal, which still names the blocking check.

The raw per-check scan continues to drive the per-check summary and the `step`/`FAIL` diagnostics, exactly as the issue suggests.

## Design note (one deliberate deviation from the suggested fix)

The issue suggested reusing the `all_passed` path's mapping verbatim, where `blocked -> Infra`. Doing that literally would regress two important cases, because a genuinely failing REQUIRED check and the label gate both leave `mergeable_state=blocked`:

- a real `build-and-test` failure would become `Infra` (escalate) instead of `Blocked` (a new Author round to fix the test);
- the `No blocking labels` label gate would become `Infra` instead of `Blocked by: No blocking labels [label gate]`, breaking the Orchestrator's label-gate detection.

So for a confirmed un-mergeable state this PR preserves the raw scan's `Blocked`-vs-`Infra` verdict rather than forcing `blocked -> Infra`. `mergeable_state` decides *whether* the PR is mergeable (overriding a red non-required check to `Clear`); the raw scan decides *which kind* of block it is when it is not. The `unstable` state is exactly the "mergeable, but a non-required check is red" case the bug mishandled.

Scope is `--pr`-only, as the issue notes: `--sha`/`--branch`/`--run-id` have no PR/`mergeable_state` and keep their current all-checks-count behavior.

## Tests

- New check `(bf)` drives `main()` through the fix end-to-end: a failing non-required check with `mergeable_state=unstable` terminates `Clear` (not a false `Blocked`), the failing check still appears in the summary as `[BLOCKING]` (diagnostic), an infra conclusion on a non-required check likewise clears, and an `unknown` state keeps polling rather than terminating prematurely.
- The existing `main()` Blocked-terminal tests (`i`, `j`, `m`, `p`, `q`, `s`, `t`, `u`, `w`, `ad`, `af`, `aj`) now thread the new `mergeable_state=blocked` fetch, confirming a real block (including the label gate) still fires `Blocked`.
- README outcome vocabulary and the `--pr` mode summary updated to document the gating. `python3 scripts/test_ci_monitor.py`: 371 passed, 0 failed; ruff + mdformat clean.


aunger (commented):
🤖 Reviewer

I re-read issue #748, PR #753, and all comments before reviewing. The core fix is right: routing the `--pr` Blocked/Infra terminal through a fresh `mergeable_state` fetch, and preserving the raw scan's Blocked-vs-Infra verdict for a confirmed un-mergeable state (rather than forcing `blocked -> Infra`), is the correct call and the design note justifies the deviation well. On the branch: `python3 scripts/test_ci_monitor.py` -> 371 passed / 0 failed; ruff and mdformat clean. But there is one blocking problem plus two nits.

## Blocking: the `unknown`-path heartbeat emits a line that begins with a terminal keyword

`ci_monitor.py:1357-1359` (the new `unknown` branch):

```python
print(
    "%s: %s but mergeable_state=unknown (still computing)" % (tag, result)
)
```

Here `result` is, by the `elif result in ("Blocked", "Infra"):` guard, always literally `Blocked` or `Infra`. So the emitted heartbeat is:

```
PR#748: Blocked but mergeable_state=unknown (still computing)
PR#748: Infra but mergeable_state=unknown (still computing)
```

That line begins with `PR#N: Blocked` / `PR#N: Infra`, the exact prefixes the README's outcome vocabulary and the Orchestrator treat as terminal. `agents/dev_orchestration.md:283/288/402` has the Orchestrator "Act only on the terminal lines Clear, Blocked ..., or Infra" and "if Monitor emits a Blocked line -> goto newAuthor". A heartbeat that starts `PR#748: Blocked ...` is ambiguous against that contract and risks being acted on as a false Blocked terminal, i.e. reintroducing the very false-positive class this PR exists to remove, just relocated from the terminal path to the still-computing heartbeat.

Contrast the sibling heartbeat on the `all_passed` path (`ci_monitor.py:1314`), which deliberately does NOT use the terminal word `Clear`:

```python
print("%s: all_passed mergeable_state=%s (still computing)" % (tag, mergeable))
```

It uses the internal state name `all_passed`, precisely so the still-computing heartbeat can never be mistaken for the `Clear` terminal. The new branch breaks that established convention by interpolating the terminal-keyword `result` verbatim. The heartbeat should be worded so it does not lead with `Blocked`/`Infra` (mirror the `all_passed` heartbeat's approach of avoiding the terminal keyword).

Note this is currently untested: the `(bf2)` unknown-path test only passes because `time.time` is mocked to a constant (`9200.0`), so `now - last_output_ts == 0` never exceeds `SILENCE_SECONDS` (120) and the heartbeat line is never emitted. The test's own assertion `not any(ln.startswith("PR#748: Blocked") ...)` encodes the expectation that no `PR#...: Blocked` line should appear on a still-computing poll, yet the production code emits exactly such a line once real time crosses the 120s silence gate. Please add coverage that actually exercises the emitted heartbeat text (e.g. advance the mocked clock past `SILENCE_SECONDS`) so the assertion is meaningful.

## Nit 1: README `[BLOCKING]` summary-row definition is now inaccurate for the unstable->Clear case

The per-check summary row definition in `scripts/ci_monitor/README.md` still reads: `[BLOCKING]` "marks checks whose conclusion caused the terminal to be `Blocked`/`Infra`." After this change, on the `unstable` path a check is printed with `[BLOCKING]` under a `Clear` terminal (this is exactly what `(bf)` asserts: `enforce-exclusive-labels ... failure ... [BLOCKING]` alongside `PR#748: Clear`). The Clear/Blocked/Infra rows were updated for the new gating, but this `[BLOCKING]` row definition was not, so it now contradicts the emitted output. Please reconcile the wording (the marker now reflects the raw per-check scan, not the final terminal).

## Nit 2: divergent handling of `draft`/`has_hooks` between the two paths

On the `all_passed` path, any `mergeable_state` outside `{clean, unstable, behind, dirty, blocked}` (e.g. `draft`, `has_hooks`) falls into the "still computing" else and keeps polling. On the new Blocked/Infra path, only `unknown` keeps polling; everything else (including `draft`/`has_hooks`) falls through to the raw Blocked/Infra terminal. Neither state is reachable in this repo's review flow (PRs aren't drafts; no pre-receive hooks), so this is not blocking, but if you want the two paths to stay symmetric it's worth a look. In particular `has_hooks` is a mergeable state, so treating it as a real block would be a (latent) repeat of this bug for a different state string.

None of this touches the soundness of the central fix; the blocking item is the heartbeat wording/coverage.

aunger:
🤖 Author

Thanks, all three points were fair and are addressed in f03ded5.

**Blocking (heartbeat reads as a terminal):** You're right, and it's exactly the false-positive class this PR exists to remove, just moved onto the heartbeat. The unknown-path heartbeat interpolated `result` (always `Blocked`/`Infra`), producing `PR#N: Blocked but mergeable_state=unknown (still computing)`. It now names the internal state instead:

```
PR#N: non-passing check, mergeable_state=<state> (still computing)
```

mirroring how the `all_passed` heartbeat uses `all_passed` rather than the `Clear` keyword, so no consumer scanning for `Blocked`/`Infra` can act on it. On coverage: `(bf2)` now advances the mocked clock past `SILENCE_SECONDS` (shrunk to 50 for the run) so the heartbeat is actually emitted, and asserts both its exact text and that no emitted line begins with `PR#748: Blocked`/`Infra`. The prior frozen-clock version never exercised the line, as you noted.

**Nit 2 (draft/has_hooks symmetry):** Fixed in the same restructure, since it cleans up the heartbeat too. The keep-polling condition changed from `mergeable == "unknown"` to `mergeable not in ("behind", "dirty", "blocked")`. So only an explicitly un-mergeable state falls through to the raw terminal; a mergeable state (`clean`/`unstable`) clears, and anything else (`unknown`, or a non-blocking state like `has_hooks`) keeps polling, symmetric with the `all_passed` path's still-computing else. Added a `has_hooks` case to `(bf2)` proving a mergeable non-`clean`/`unstable` state keeps polling to `Clear` rather than producing a false `Blocked`.

**Nit 1 (README `[BLOCKING]` definition):** Reconciled. It now reads that `[BLOCKING]` reflects the raw per-check scan (a `failure`/`action_required` or infra conclusion), not the final terminal, so under `--pr` a `[BLOCKING]` row can appear beneath a `Clear` terminal when the failing check is non-required and `mergeable_state` is `unstable`, which is what `(bf)` asserts.

`python3 scripts/test_ci_monitor.py` -> 375 passed / 0 failed; ruff and mdformat clean.

aunger (commented):
🤖 Reviewer

Re-read the issue, PR, and all comments, then reviewed `f03ded5`. All three points from the previous round are fully resolved. LGTM.

**Blocking heartbeat (resolved).** The still-computing line no longer interpolates the terminal-keyword `result`; it now reads `PR#N: non-passing check, mergeable_state=<state> (still computing)` (`ci_monitor.py:1362-1365`), mirroring the `all_passed` heartbeat's avoidance of a terminal keyword, so no consumer scanning for `Blocked`/`Infra` can act on it. Critically, `(bf2)` now drives an advancing clock (`fake_time_bf2`/`fake_sleep_bf2` with `SILENCE_SECONDS` patched to 50) so the heartbeat is actually emitted, asserts its exact text, and asserts no emitted line begins with `PR#748: Blocked`/`Infra`. That closes the gap where the old frozen-clock test never exercised the line.

**Symmetry with the all_passed path (resolved).** The keep-polling guard is now `if mergeable not in ("behind", "dirty", "blocked"):` (`ci_monitor.py:1353`), so the keep-polling set (`unknown`/`has_hooks`/etc.) matches the `all_passed` still-computing else exactly; only an explicitly un-mergeable state falls through to the raw Blocked/Infra terminal. New `(bf2b)` proves `has_hooks` keeps polling to `Clear` rather than producing a false block, removing the latent same-bug-different-state risk.

**README `[BLOCKING]` definition (resolved).** The summary-row entry now states `[BLOCKING]` reflects the raw per-check scan (not the final terminal) and can appear beneath a `Clear` terminal on the `unstable` path, consistent with `(bf)`.

The central fix remains sound and the deliberate `blocked`-verdict deviation is still correctly preserved. On `f03ded5`: `python3 scripts/test_ci_monitor.py` -> 375 passed / 0 failed; ruff and mdformat clean. No further changes requested.

PR #755, opened 2026-07-20
Restore the official Gradle 8.9 wrapper JAR to align it with the pinned 8.9 distribution

### LINKED ISSUE #754: Replace committed Gradle wrapper JAR with the official 8.9 build to resolve wrapper JAR / distribution version skew
Follow-on from issue #744 / PR #747, which added an automated integrity check (check (g) in `scripts/test_verification_metadata.sh`) authenticating the committed `gradle/wrapper/gradle-wrapper.jar` against Gradle's officially published checksum for it. That check surfaced a pre-existing mismatch: the committed wrapper JAR is Gradle 8.14, while `gradle/wrapper/gradle-wrapper.properties` pins the 8.9 distribution (`distributionUrl` and `distributionSha256Sum`). The skew is functionally harmless (a newer wrapper JAR launches an older distribution without issue) but was flagged as worth aligning.

## Chosen approach

Replace the committed wrapper JAR with the official Gradle 8.9 wrapper JAR, matching the already-pinned 8.9 distribution, rather than bumping the distribution to 8.14. This is the smaller, more contained change: it does not touch `gradle/verification-metadata.xml` or the resolved dependency graph, since the distribution version being fetched does not change.

## What needs to happen

- Fetch the official Gradle 8.9 wrapper JAR and verify it against Gradle's published checksum for it (`https://services.gradle.org/distributions/gradle-8.9-wrapper.jar.sha256`), the same authentication approach PR #747 used for the 8.14 JAR.
- Replace `gradle/wrapper/gradle-wrapper.jar` with that verified 8.9 JAR.
- Update the pinned SHA-256 in `scripts/test_verification_metadata.sh` check (g) to the 8.9 JAR's checksum, so the guard test continues to authenticate correctly (per the maintenance note PR #747 added to `gradle/README.md`).
- Confirm the shell-test suite (`scripts/test_verification_metadata.sh`) still passes with the new JAR in place, and that a full CI run (including the instrumented/E2E suite) succeeds, since this changes the actual wrapper bootstrap binary used to launch every Gradle invocation.
- No changes expected to `gradle/wrapper/gradle-wrapper.properties`, `gradle/verification-metadata.xml`, or the dependency graph.

aunger:
🤖 Author

Fixes #754.

## Problem

The committed `gradle/wrapper/gradle-wrapper.jar` was the Gradle **8.14** wrapper JAR, while `gradle/wrapper/gradle-wrapper.properties` pins the **8.9** distribution (`distributionUrl` and `distributionSha256Sum`).
Check (g) in `scripts/test_verification_metadata.sh` (added by #747 for issue #744) surfaced this skew by authenticating the JAR against Gradle's published checksum.
The skew is functionally harmless (a newer wrapper JAR launches an older distribution without issue) but was flagged as worth aligning.

## Change

Per the approach chosen in #754, replace the committed wrapper JAR with the official Gradle 8.9 wrapper JAR (the smaller, more contained option, versus bumping the distribution to 8.14):

- Replace `gradle/wrapper/gradle-wrapper.jar` with the official Gradle 8.9 wrapper JAR.
  The replacement bytes were authenticated against the checksum Gradle officially publishes for its 8.9 wrapper JAR (`https://services.gradle.org/distributions/gradle-8.9-wrapper.jar.sha256` = `498495120a03b9a6ab5d155f5de3c8f0d986a449153702fb80fc80e134484f17`), the same authoritative source check (g) verifies against. Their SHA-256 matches that published value exactly.
- Update check (g)'s pinned `WRAPPER_JAR_SHA256` to the 8.9 value and reword its comment, since the committed JAR now matches the pinned distribution version rather than being newer than it.

No changes to `gradle/wrapper/gradle-wrapper.properties`, `gradle/verification-metadata.xml`, or the resolved dependency graph: the distribution version being fetched does not change.
`gradle/README.md` needed no change, as its wrapper-JAR maintenance note already uses a generic `gradle-<version>-wrapper.jar.sha256` URL rather than a hardcoded version.

## Base / stacking

This PR targets the #747 branch (`claude/issues-744-745-746-rxuafx`), not `main`, because it edits check (g), which only exists on #747 (still open).
Basing on `main` would fold #747's three unmerged commits into this diff.
**This PR should merge after #747.** Once #747 merges to `main`, this can be retargeted to `main` and the diff will apply cleanly (only the two files below change here).

## Testing

Covered by the automated shell-test suite (`scripts/test_verification_metadata.sh`, run in the `shell-tests` CI job).
Verified locally that the full suite passes 8/8 with the new 8.9 JAR in place, and that check (g) still fails with a non-zero exit when the JAR's bytes do not match the pin.

### DIFF
--- a/gradle/README.md
+++ b/gradle/README.md
@@ -43,7 +43,7 @@ Never commit that flag into CI, and do not commit a file regenerated on a non-Li
 The Gradle wrapper verifies the downloaded distribution against this checksum before running it, so a substituted or corrupted distribution fails loudly.
 
 The wrapper JAR (`wrapper/gradle-wrapper.jar`) is committed and reviewed directly; it is read before `verification-metadata.xml` is consulted, so it cannot be covered by that file.
-Instead, `scripts/test_verification_metadata.sh` check (g) pins the JAR's SHA-256 to the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.
+Instead, `scripts/test_verification_metadata.sh` check (g) verifies the JAR's SHA-256 against a pinned copy of the checksum Gradle officially publishes for it (issue #744), so a substituted or corrupted wrapper JAR fails the `shell-tests` CI job rather than being trusted silently.
 
 When bumping the Gradle version, update both `distributionUrl` and `distributionSha256Sum` (the published checksum is at `https://services.gradle.org/distributions/gradle-<version>-bin.zip.sha256`), then regenerate `verification-metadata.xml`.
 If the wrapper JAR itself changes, also update its pinned SHA-256 in `scripts/test_verification_metadata.sh` (the published checksum is at `https://services.gradle.org/distributions/gradle-<version>-wrapper.jar.sha256`).


aunger:
🤖 Reviewer

LGTM. I independently verified every load-bearing claim in this PR.

**Authenticity of the new JAR.** I fetched Gradle's officially published checksum and it matches the pin exactly:
- `https://services.gradle.org/distributions/gradle-8.9-wrapper.jar.sha256` = `498495120a03b9a6ab5d155f5de3c8f0d986a449153702fb80fc80e134484f17`, which is the new `WRAPPER_JAR_SHA256`.
- The committed JAR at head (`cbbceed`) hashes to exactly that value, so the bytes are the genuine Gradle 8.9 wrapper JAR, not merely a self-consistent pin.
- The removed value `7d3a4ac4...296172` matches Gradle's published 8.14 checksum, confirming the comment's before/after story is accurate.

**Guard still authenticates.** Check (g) compares the file's actual SHA against the pinned constant (line 155), not against the file's own recomputed hash, so it authenticates against Gradle rather than trusting whatever is committed. The updated comment correctly drops the "newer than the pinned distribution" note now that JAR and distribution are both 8.9.

**Suite passes.** Ran `scripts/test_verification_metadata.sh` at head: 8 passed, 0 failed, including check (g).

**Scope is clean.** Only the two intended files change. `gradle/wrapper/gradle-wrapper.properties` still pins the 8.9 distribution (`gradle-8.9-bin.zip`, `distributionSha256Sum` unchanged), so the JAR now aligns with it. `gradle/verification-metadata.xml` and the dependency graph are untouched, as expected since the fetched distribution version does not change. The unchanged `gradle/README.md` note uses a generic `gradle-<version>-wrapper.jar.sha256` URL, so it needed no edit.

**Base/stacking.** Targeting the #747 branch is correct: check (g) exists only there, and basing on `main` would fold #747's unmerged commits into this diff. Merge order (this after #747) is noted in the PR body.

CI is still pending at review time; I am not blocking on it. The one item the issue calls out that automated shell tests cannot cover is that this swaps the actual wrapper bootstrap binary, so a green full CI run (including the instrumented/E2E suite that invokes the wrapper) is the remaining confirmation. That is the Orchestrator's CI Monitor step, not a blocker for this review.

aunger:
### Verification plan

**Before merging** (must resolve before PR #755 can be merged):
- [ ] #756
- [ ] #757

None.

---
Notes:
- Issue-dependency and sub-issue links could not be attached to the PR itself (GitHub requires the source of a `blocked_by` / sub-issue relationship to be an issue, returning 422 for a PR number), so both tracking issues were registered as `blocked_by` dependencies of, and sub-issues of, the parent issue #754 that this PR resolves. Each issue body also states "Blocks PR #755" in plain text.

aunger:
Verification Agent: verification revealed an error (issue #756 -- FAIL)

Verifying the before-merging item in #756 (a green full CI run, including the instrumented/E2E suite, with the swapped Gradle 8.9 wrapper JAR), I ran the full `build.yml` suite against `main` on a live test PR (#758, base `main`, head content identical to PR #755 at `cbbceed`). The result is a hard failure, and it is caused by this PR's change.

**The swapped 8.9 wrapper JAR is not launchable: it is missing the `Main-Class` manifest attribute.**

- CI run: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/29754437303 (job `build-and-test`, step "Build and run unit tests").
- The very first `./gradlew` invocation died immediately with:

  ```
  no main manifest attribute, in .../gradle/wrapper/gradle-wrapper.jar
  ```

- Because the wrapper cannot bootstrap, nothing downstream ran: unit tests produced no results (`result directory not found, skipping: app/build/test-results/testDebugUnitTest`), and every instrumented/E2E test was **skipped**. The trailing `adb: no devices/emulators found` lines are secondary teardown noise (the emulator was never started).

**Root cause, reproduced locally on the committed JAR (`gradle/wrapper/gradle-wrapper.jar` at `cbbceed`):**

- Its `META-INF/MANIFEST.MF` contains only:

  ```
  Manifest-Version: 1.0
  Implementation-Title: Gradle Wrapper
  ```

  There is **no** `Main-Class: org.gradle.wrapper.GradleWrapperMain` line, so `java -jar gradle-wrapper.jar` (what `gradlew` does) has no entry point. Running it locally reproduces the exact CI error: `no main manifest attribute`. The `GradleWrapperMain.class` bytecode is present in the archive, but without the manifest entry the JVM cannot find it.
- By contrast, the JAR this PR replaces (SHA `7d3a4ac4...296172`, present on both `main` and the #747 base) has a complete executable manifest:

  ```
  Manifest-Version: 1.0
  Implementation-Title: Gradle Wrapper
  Main-Class: org.gradle.wrapper.GradleWrapperMain
  Enable-Native-Access: ALL-UNNAMED
  ```

  That JAR launches, which is why `main` is green today.

**Why the guards passed anyway.** The committed JAR's SHA-256 (`498495120a03b9a6ab5d155f5de3c8f0d986a449153702fb80fc80e134484f17`) does match Gradle's published `gradle-8.9-wrapper.jar.sha256`, so check (g) and the author/reviewer byte-authentication both pass. But that published `*-wrapper.jar.sha256` artifact is evidently not the launchable project wrapper JAR: it lacks the executable manifest. Check (g) authenticates bytes only and never exercises the JAR at runtime, which is precisely the gap #756 was filed to close. The `shell-tests` job (check (g) included) was green in this same run; only the runtime bootstrap failed.

**Net:** the "restore the official 8.9 wrapper JAR" approach as implemented replaces a working bootstrap binary with a non-bootable one. This needs an Author fix (for example, source the 8.9 `gradle-wrapper.jar` from an artifact that carries the `Main-Class`/`Enable-Native-Access` manifest, such as running `gradle-8.9 wrapper` or copying the JAR from the 8.9 distribution, then re-pinning check (g) to that JAR's checksum). I am not fixing it here.

Tracking issue #756 stays open. The test PR #758 will be closed; its run history is preserved as evidence.

aunger:
Verification Agent: before-merging summary

| Item | Tracking issue | Result |
| --- | --- | --- |
| Green full CI run (incl. instrumented/E2E) with the swapped Gradle 8.9 wrapper JAR | #756 | **FAIL** -- wrapper JAR is missing its `Main-Class` manifest attribute, so `./gradlew` cannot bootstrap; entire `build-and-test` job fails and all instrumented/E2E tests skip. Diagnosis and evidence in the comment above. Run: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/29754437303 |
| Merge #755 after #747, retarget to `main`, confirm clean apply | #757 | **NOT AUTOMATABLE** -- operational merge-ordering step gated on merging #747 first, which a Verification Agent does not do. Details on #757. Underlying facts confirmed: #755's diff touches exactly the two intended files, and its check (g) edit genuinely cannot apply to `main` until #747 merges. |

Overall: verification revealed an error. `shell-tests` (check (g) included) passed, but the runtime full-CI requirement failed. Both tracking issues remain open. Sending PR #755 back to an Author to fix the non-bootable wrapper JAR (#756).

aunger:
🤖 Author

The before-merging verification (#756) correctly caught a hard failure. After investigating, I've concluded that the approach this issue specifies cannot produce a working build, and I can't execute a working alternative in this environment either. Rather than ship broken or out-of-scope code, I've reverted the branch to the PR base (no net change) and am raising this for a maintainer scoping decision. Full reasoning and evidence below.

## Root cause: the authentic Gradle 8.9 wrapper JAR has no `Main-Class`, and this repo's `gradlew` requires one

- Gradle's official `versions/all` metadata lists 8.9's `wrapperChecksum` as `498495120a03b9a6ab5d155f5de3c8f0d986a449153702fb80fc80e134484f17` (the same value served at `gradle-8.9-wrapper.jar.sha256`). The JAR with that checksum has this complete manifest:

  ```
  Manifest-Version: 1.0
  Implementation-Title: Gradle Wrapper
  ```

  There is no `Main-Class` line. (`org/gradle/wrapper/GradleWrapperMain.class` is in the archive, but nothing in the manifest points to it.) This is the byte-for-byte authentic official 8.9 wrapper JAR: it is exactly what the published checksum authenticates, and it is also what `gradle/gradle` itself commits at tag `v8.9.0`. So the JAR the verification exercised was genuine, not corrupt.

- This repo's `gradlew` launches the wrapper with `java -jar` (lines 213-217):

  ```
  set -- \
          "-Dorg.gradle.appname=$APP_BASE_NAME" \
          -classpath "$CLASSPATH" \
          -jar "$APP_HOME/gradle/wrapper/gradle-wrapper.jar" \
          "$@"
  ```

  `java -jar` ignores `-classpath` and requires the JAR manifest to declare `Main-Class`. With the 8.9 JAR it dies immediately with `no main manifest attribute, in .../gradle-wrapper.jar` -- exactly the CI failure reproduced in #756.

## Why "just swap the JAR" cannot work, and why the suggested fixes don't either

The committed `gradlew` is the newer (Gradle 8.10+) `-jar`-based launcher, and the committed wrapper JAR is 8.14, which carries `Main-Class` (and `Enable-Native-Access: ALL-UNNAMED`) precisely because Gradle added those to the wrapper-JAR manifest when it switched `gradlew` to `-jar`. The 8.9 wrapper JAR predates that change, so it has no `Main-Class`. The wrapper JAR and `gradlew` are a matched pair; you cannot downgrade the JAR to 8.9 while keeping the 8.10+ `gradlew`.

The verification comment's suggested remedies ("source the 8.9 JAR from the distribution" or "run `gradle-8.9 wrapper`") don't help: every authentic 8.9 wrapper JAR -- distribution-shipped or freshly generated -- has the same no-`Main-Class` manifest, because 8.9's own `gradlew` launches via `-classpath ... org.gradle.wrapper.GradleWrapperMain` and so never needed one.

## The real skew is wrapper-vs-distribution, and it has only two coherent fixes

The actual skew is: the committed wrapper (both `gradle-wrapper.jar` and `gradlew`) is 8.14, while the distribution (`gradle-wrapper.properties`) is pinned to 8.9. Aligning them has two coherent directions:

- **Option A -- align up to 8.14.** Bump `distributionUrl`/`distributionSha256Sum` to `gradle-8.14-bin.zip`, keep the working 8.14 JAR + `gradlew` + check (g) pin, and regenerate `gradle/verification-metadata.xml` on the full Android toolchain with Gradle 8.14. This is the only alignment that keeps a working build, but it touches `gradle-wrapper.properties` and `verification-metadata.xml` -- exactly what this issue said to avoid, and what #747 deferred.
- **Option B -- accept the skew.** Leave the wrapper at 8.14 and the distribution at 8.9, which is what `main` runs today (green). #747's check (g) already authenticates the working 8.14 JAR against Gradle's published checksum and documents the skew as functionally harmless (a newer wrapper launches an older distribution without issue). No code change; the skew is cosmetic and guarded.

**Option C -- align down to 8.9 (what this issue asks)** would require regenerating the *entire* wrapper -- both `gradlew` and the JAR -- back to 8.9, i.e. reverting the intentional 8.10+ launcher to 8.9's `-classpath` form. That is a functional regression of the launcher to fix a cosmetic version label, and it still cannot be done by swapping the JAR alone (which is what broke CI here).

## Recommendation and why I'm not just doing it

I recommend **Option B** (accept the harmless, now-guarded skew; close #754 as wontfix once #747 lands), or **Option A** if aligned version labels are worth regenerating `verification-metadata.xml`. Both contradict this issue's stated constraints, so this is a maintainer scoping call, not an Author guess -- I don't want to unilaterally reverse the version decision the issue made explicit.

I also can't execute Option A or C in this environment: both need the Gradle distribution, whose download redirects to `github.com/gradle/gradle-distributions` release assets (blocked here), plus the Android toolchain to regenerate `verification-metadata.xml`.

I've reset this branch to the PR base so it carries no build-breaking change while the direction is decided.

Issue #764, opened 2026-07-23
(re PR #763) Confirm start-of-test toast renders over a reasserted keyguard in Permissions E2E recordings


aunger:
This is an unautomated verification step outstanding before PR #763 can be merged.

**Source:** PR #763 description, "Test plan" section:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/763

> The first acceptance criterion is a property of the recorded video, which no automated assertion checks, so it needs manual confirmation from a keyguard-reasserted CI run:
> - [ ] Inspect this PR's `e2e-video-*` artifacts for `PermissionsGrantedE2ETest` and `PermissionsDeniedE2ETest` and confirm the start-of-test name toast is now visible over the app UI (not painted behind the lock screen).

This mirrors the first acceptance criterion of the parent issue, #761:
https://github.com/aunger/gallery-button-for-pixel-camera/issues/761

> - [ ] On a run where the secure keyguard has reasserted between CI steps, the start-of-test marker for these three suites is visible in the recording (not painted behind the lock screen).

**Why this is not covered by automated tests:** no assertion in the suite inspects the recorded `.mp4` content; the checkbox above is an explicit manual-inspection instruction in the PR's own test plan, and it is unchecked.

**Note on feasibility:** per `.github/workflows/build.yml`, the `e2e-video-PermissionsGrantedE2ETest` / `e2e-video-PermissionsDeniedE2ETest` artifacts are only pulled and uploaded when the corresponding suite's step outcome is `failure` (see the "Only pull and upload the recording when this suite failed" comments ahead of each upload step). PR #763's own CI run passed (see the CI summary comment), so no video artifact was produced by that run. Confirming this criterion will require either:
- a CI run where the secure keyguard reasserts and one of these suites actually fails (naturally reproducing the pre-fix occlusion for comparison), or
- a temporary, deliberate reproduction (e.g. locally or via a throwaway workflow run with the upload condition loosened) to capture a video of a keyguard-reasserted run and confirm the toast now paints over the app UI.

**Action needed:** obtain a recording of one of these two suites running with the secure keyguard reasserted, and confirm the start-of-test name toast is visible over the app UI rather than painted behind the lock screen.

Blocks PR #763.

aunger:
🤖 Verification Agent

Result: PASS

## Method

[...]

PR #763, opened 2026-07-23
Dismiss the secure keyguard before the start-of-test slate in the Permissions E2E suites

aunger:
🤖 Author

Fixes #761

## Problem

Three recorded E2E suites produced failure videos where the start-of-test name toast never appears. The cause is Android windowing: a toast cannot paint over the *secure* PIN keyguard that `scripts/setup-e2e-emulator.sh` configures, so when the keyguard has reasserted between CI steps the marker renders behind the lock screen and is invisible in the recording.

Two of the three suites, `PermissionsGrantedE2ETest` and `PermissionsDeniedE2ETest`, wired `TestNameToastRule` as a plain `@get:Rule` and relied on `fixture.setUp()`'s swipe-only `wakeAndDismissKeyguard()`, which does nothing against the secure PIN keyguard. `TestNameToastRule.starting()` also fires before `@Before`, so even that swipe had not run when the slate appeared.

The third suite named in the issue, `SetupActivityGrantedE2ETest`, already dismisses the secure keyguard ahead of the toast via its `RuleChain`, so it needed no change.

## Change

`PermissionsGrantedE2ETest` and `PermissionsDeniedE2ETest` now use a `RuleChain` whose outer rule runs `E2EFixture.dismissSecureKeyguard()` (via an `ExternalResource` reusing the existing `fixture` field) ahead of the innermost `TestNameToastRule`, so the slate renders over the app UI rather than behind the lock screen. This mirrors the established pattern in `SetupActivityDeniedE2ETest`/`SetupActivityGrantedE2ETest`.

The toast stays the innermost link, preserving the keyguard-dismissal-then-launch ordering that `TestNameToastRule.kt` documents (the PR #564 / #509 re-engagement race). These two suites launch the camera later, inside each test body via `fixture.launchPixelCamera()`, not in an activity-launching rule, so no new dismissal-then-launch race is introduced.

## How the acceptance criteria are met

- The keyguard-dismissal-then-launch ordering that `TestNameToastRule.kt` documents is preserved: the toast remains the innermost rule, so its ~3s duration sits after dismissal, with no regression of the PR #564 / #509 race.
- `dismissSecureKeyguard()` now runs before the start-of-test slate for all three named suites, so the marker renders over the app UI rather than behind the lock screen.

## Test plan

The first acceptance criterion is a property of the recorded video, which no automated assertion checks, so it needs manual confirmation from a keyguard-reasserted CI run:

[...]

Issue #761, opened 2026-07-22
E2E: dismiss the secure keyguard before the start-of-test slate so it is not occluded by the lock screen


aunger:
Follow-up from the #641 verification work (inspecting the PR #654 `e2e-video-*` artifacts).

## Problem

Three of the recorded E2E suites (`PermissionsGrantedE2ETest`, `PermissionsDeniedE2ETest`, `SetupActivityGrantedE2ETest`) produced failure videos in which the start-of-test name toast never appears.
The cause is not the toast mechanism itself but Android windowing: a toast (like any app-drawn surface) cannot paint over the *secure* PIN keyguard that `scripts/setup-e2e-emulator.sh` configures.
When enough wall-clock time or CI steps have passed for that keyguard to reassert, the marker is drawn behind the lock screen and is invisible in the recording.

`TestNameToastRule.starting()` fires before the test body, so if the keyguard is up at that instant the slate is occluded regardless of where a failure is (or is not) injected.
Two of the three suites only wire `E2EFixture.wakeAndDismissKeyguard()` (a swipe), which that method's own doc notes does nothing against the secure PIN keyguard.

## Proposal

Ensure the secure keyguard is dismissed in the before-test window, ahead of the start-of-test slate, so the marker renders over the app UI rather than behind the lock screen.
Concretely, run `E2EFixture.dismissSecureKeyguard()` before the toast/slate for the recorded suites, and order the rules so dismissal precedes the slate.

Care is needed with the existing `RuleChain` ordering documented in `TestNameToastRule.kt`: the toast is deliberately the innermost link (after activity launch) so its ~3s duration does not push the keyguard-dismissal-then-launch sequence into a re-engaged keyguard (the race behind PR #564 / issue #509).
Any change here must preserve that ordering guarantee.

## Acceptance criteria

- [ ] On a run where the secure keyguard has reasserted between CI steps, the start-of-test marker for these three suites is visible in the recording (not painted behind the lock screen).
- [ ] The keyguard-dismissal-then-launch ordering that `TestNameToastRule.kt` documents is preserved (no regression of the PR #564 / #509 race).

## References

- #641 (verification of the video artifacts), #654 (throwaway CI run), #604 (name toast), PR #637, PR #564 / #509 (keyguard re-engagement race)
[...]

PR #768, opened 2026-07-31
Dismiss the secure keyguard before the start-of-test slate in PixelCameraOverlayE2ETest and GalleryButtonVisualE2ETest

### LINKED ISSUE #765: Extend keyguard-dismissal-before-toast fix to other recorded E2E suites with the same latent occlusion risk
This is a follow-on item and does not block merging PR #763.

**Source:** reviewer comment on PR #763:
https://github.com/aunger/gallery-button-for-pixel-camera/pull/763#issuecomment-5055404654

> Out of scope for this issue, but worth a follow-up: `PixelCameraOverlayE2ETest`, `GalleryButtonVisualE2ETest`, and `PartialAccessPhotoPickerE2ETest` also record video with a plain `@get:Rule testNameToastRule` and launch from the test body, so they carry the same latent occlusion risk if the keyguard reasserts before them. #654's artifacts only implicated the three named suites, so leaving them here is right; consider a follow-up issue.

## Suggested scope

Apply the same pattern introduced in PR #763 (an outer `ExternalResource` rule running `E2EFixture.dismissSecureKeyguard()`, chained via `RuleChain.outerRule(...).around(testNameToastRule)`, with the toast kept innermost to preserve the keyguard-dismissal-then-launch ordering from `TestNameToastRule.kt`) to:

- `PixelCameraOverlayE2ETest`
- `GalleryButtonVisualE2ETest`
- `PartialAccessPhotoPickerE2ETest`

These suites currently wire `testNameToastRule` as a plain `@get:Rule` and launch the camera from the test body, so if the secure keyguard reasserts between CI steps before one of these suites runs, their start-of-test toast would be painted behind the lock screen and invisible in the recording, the same failure mode issue #761 / PR #763 fixed for `PermissionsGrantedE2ETest` and `PermissionsDeniedE2ETest`.

This is a follow-on item and does not block merging PR #763.

aunger:
🤖 Author

Fixes #765

## Problem

Issue #765 is a follow-on from PR #763 / issue #761: a start-of-test toast (`TestNameToastRule`) cannot paint over the *secure* PIN keyguard `scripts/setup-e2e-emulator.sh` configures, so if that keyguard has reasserted between CI steps before a recorded suite runs, its slate renders behind the lock screen and is invisible in the video artifact.

The issue named three suites as carrying this same latent risk: `PixelCameraOverlayE2ETest`, `GalleryButtonVisualE2ETest`, and `PartialAccessPhotoPickerE2ETest`.

## Change

`PixelCameraOverlayE2ETest` and `GalleryButtonVisualE2ETest` now use a `RuleChain` whose outer rule runs `E2EFixture.dismissSecureKeyguard()` (via an `ExternalResource`) ahead of the innermost `testNameToastRule`, identical to the pattern PR #763 introduced for `PermissionsGrantedE2ETest`/`PermissionsDeniedE2ETest`. Both suites launch the camera later, inside each test body via `fixture.launchPixelCamera()`, not in an activity-launching rule, so the toast stays innermost with no new dismissal-then-launch race introduced (the same reasoning PR #763's review confirmed for the Permissions suites). `screenshotRule` (a separate `TestWatcher`) is left as an independent plain `@get:Rule`: its `starting()` only records a screenshot-filename prefix and a baseline file snapshot, with no ordering dependency on keyguard dismissal or the toast.

## `PartialAccessPhotoPickerE2ETest` is intentionally left unchanged

Re-checking the current state of this suite against the issue's premise, neither half of "records video with a plain `@get:Rule testNameToastRule`" actually holds for it:

- It does not wire `testNameToastRule` at all (verified with a repo-wide grep; only `PixelCameraOverlayE2ETest` and `GalleryButtonVisualE2ETest` had the plain-`@get:Rule` form the issue described). There is no toast rule here for a reasserted keyguard to occlude.
- Its CI step (`Run PartialAccessPhotoPickerE2ETest` in `.github/workflows/build.yml`) does not run `screenrecord` and produces no `e2e-video-*` artifact, unlike every other suite named in #761/#765's investigation. It is not one of the "recorded E2E suites" the issue title refers to.
- It already dismisses the secure keyguard ahead of its own `composeRule`-driven `SetupActivity` launch (`RuleChain.outerRule(keyguardDismiss).around(composeRule)`), so the gap the other two suites had does not exist here either.

Since the underlying premise does not apply to this suite, no change to it is included in this PR. (The reviewer comment that seeded #765 was accurate about the general pattern but appears to have described this suite's state incorrectly, possibly from a stale reading, since the git history for this file shows no prior revision that wired `testNameToastRule`.)

## Test plan

The keyguard-reassertion scenario is a property of the recorded video, which no automated assertion checks, so it needs manual confirmation from a keyguard-reasserted CI run, mirroring how #764 verified this for PR #763:

- [ ] Inspect this PR's `e2e-video-PixelCameraOverlayE2ETest` and `e2e-video-GalleryButtonVisualE2ETest` artifacts (ideally from a run where the keyguard has reasserted before these steps) and confirm the start-of-test toast renders over the app UI, not behind the lock screen.

`ktlint` passes on both edited files (verified with the pinned standalone ktlint CLI via `scripts/install-ktlint.sh`, since the Android SDK and an emulator are not available in the development environment). These instrumented (`androidTest`) suites are compiled and run only by CI's connected-test job.

### DIFF
--- a/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/GalleryButtonVisualE2ETest.kt
@@ -19,6 +19,8 @@ import org.junit.Before
 import org.junit.FixMethodOrder
 import org.junit.Rule
 import org.junit.Test
+import org.junit.rules.ExternalResource
+import org.junit.rules.RuleChain
 import org.junit.runner.RunWith
 import org.junit.runners.MethodSorters
 import kotlin.math.sqrt
@@ -50,9 +52,6 @@ class GalleryButtonVisualE2ETest {
     @get:Rule
     val screenshotRule = ScreenshotTestRule()
 
-    @get:Rule
-    val testNameToastRule = TestNameToastRule()
-
     private val instrumentation = InstrumentationRegistry.getInstrumentation()
     private val context = instrumentation.targetContext
     private val fixture =
@@ -60,6 +59,29 @@ class GalleryButtonVisualE2ETest {
             context = context,
             uiAutomation = instrumentation.uiAutomation,
         )
+
+    // Dismiss the PIN-secured keyguard before the start-of-test toast so the slate renders over the
+    // app UI rather than behind the lock screen (issue #761 / #765). The suite's own setUp() only
+    // wires fixture.wakeAndDismissKeyguard() (a swipe), which does nothing against the secure
+    // keyguard scripts/setup-e2e-emulator.sh configures, and it runs from @Before, after the toast
+    // has already fired; if the keyguard has reasserted between CI steps the marker would be
+    // occluded.
+    private val keyguardDismiss =
+        object : ExternalResource() {
+            override fun before() {
+                fixture.dismissSecureKeyguard()
+            }
+        }
+
+    private val testNameToastRule = TestNameToastRule()
+
+    @get:Rule
+    val ruleChain: RuleChain =
+        // keyguardDismiss is outermost so the secure keyguard is cleared before testNameToastRule
+        // (innermost) shows the slate; the toast's ~3s duration therefore sits after dismissal,
+        // preserving the keyguard-dismissal-then-launch ordering TestNameToastRule.kt documents (the
+        // camera launch itself happens later, in each test body, via fixture.launchPixelCamera()).
+        RuleChain.outerRule(keyguardDismiss).around(testNameToastRule)
 
     @Before
     fun setUp() = fixture.setUp()
--- a/app/src/androidTest/java/com/gb4pc/e2e/PixelCameraOverlayE2ETest.kt
+++ b/app/src/androidTest/java/com/gb4pc/e2e/PixelCameraOverlayE2ETest.kt
@@ -8,6 +8,8 @@ import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
+import org.junit.rules.ExternalResource
+import org.junit.rules.RuleChain
 import org.junit.runner.RunWith
 
 /**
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

Verified this against the repo directly (not just reading the diff):

- `PixelCameraOverlayE2ETest` and `GalleryButtonVisualE2ETest`'s new `keyguardDismiss`/`testNameToastRule`/`ruleChain` blocks are structurally identical to the precedent PR #763 set in `PermissionsGrantedE2ETest`/`PermissionsDeniedE2ETest` (I diffed all four against each other). Both edited suites launch the camera only from inside `@Test` bodies via `fixture.launchPixelCamera()`, never from a rule, so `testNameToastRule` staying innermost introduces no new dismissal-then-launch race — same reasoning PR #763 established.
- The one structural difference — `screenshotRule` staying an independent `@get:Rule` alongside `ruleChain` rather than being folded in — is safe: `ScreenshotTestRule.starting()` only sets a thread-local filename prefix and snapshots existing screenshot filenames, with no read of keyguard/toast state, so it doesn't matter that JUnit doesn't guarantee ordering between separate `@Rule` fields here.
- I independently confirmed the "leave `PartialAccessPhotoPickerE2ETest` unchanged" call: a repo-wide grep shows `testNameToastRule`/`TestNameToastRule()` is not referenced anywhere in that file; `git log -p --follow` on the file shows no prior revision ever wired it either, so the issue's premise about it really doesn't hold. `.github/workflows/build.yml`'s `Run PartialAccessPhotoPickerE2ETest` step (lines ~1122-1200) has no `screenrecord` calls and uploads no `e2e-video-PartialAccessPhotoPickerE2ETest` artifact, unlike every other suite named in #761/#765. And the file already has `RuleChain.outerRule(keyguardDismiss).around(composeRule)` ahead of its `SetupActivity` launch. All three of the PR description's claims check out.
- `kotlin-ktlint` is already green in this PR's checks.

No functional issues found, and no nits either — this is a clean, minimal, well-justified extension of an already-reviewed pattern.

LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #768 can be merged):
- [x] #769

None.

aunger:
🤖 Verification Agent

## Verification summary

Before-merging item from the [Verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/768#issuecomment-5144145040):

| Tracking issue | Item | Result |
|---|---|---|
| #769 | Confirm the start-of-test toast renders over a reasserted secure keyguard in the `PixelCameraOverlayE2ETest` and `GalleryButtonVisualE2ETest` recordings | **PASS** |

**Evidence:** throwaway test PR #770 (PR #768's head commit plus a commit that forces keyguard reassertion before each suite and loosens the video-upload condition to `always()`) — [workflow run](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/30640181567) (passed). Frame-by-frame inspection of both captured recordings confirms: the secure PIN keyguard is locked at the start of each suite's segment, `dismissSecureKeyguard()` clears it, and the start-of-test toast (`PixelCameraOverlayE2ETest_overlayAppearsWhenViewfinderOpens` / `GalleryButtonVisualE2ETest_test0_smokeGreenFeedVisible`) then renders over the home-screen app UI — not behind the lock screen. Full details and timestamps are on issue #769 (closed): https://github.com/aunger/gallery-button-for-pixel-camera/issues/769#issuecomment-5149245251

All before-merging items are resolved.

PR #771, opened 2026-07-31
Fit the overlay thumbnail by its short edge instead of its long edge

### LINKED ISSUE #767: Thumbnails should zoom to fit short edge
## Current

The overlay's thumbnail view currently appears to fit photos' **long edges** inside the button, leaving letterbox bars on the sides or top/bottom. 

## Desired 

Instead, fit by the photos' short edges.

## Acceptance

Given
- Common digital camera sizes (number of pixels) and edge ratios (16:9, 4:3, 9:16, 3:4) 
- Test fixture images, one for each ratio above, constructed by these rules:
  - image size (number of pixels) is on the order of a typical Pixel Camera output photo
  - let `s` be the shorter (minor) edge length in pixels of the image
  - a solid BLUE square of size `s px * s px` is centered on the image
  - the pixels not defined as BLUE above are RED

(In plain language, wide and tall test images, each with an edge-to-edge centered BLUE square, padded by two RED reverse-letterbox bars) 

When any of the test images are displayed in the overlay thumbnail, the entire thumbnail should appear BLUE, with no black, and no RED appearing.

aunger:
Fixes #767.

## Problem

The overlay's photo-thumbnail `ImageView` used `ImageView.ScaleType.FIT_CENTER`, which scales an image so its **long** edge fits inside the button.
For non-square photos (16:9, 4:3, 9:16, 3:4) that leaves letterbox bars along the short axis instead of filling the button.

## Fix

Two commits of production code:

1. `OverlayManager` switches the overlay's thumbnail `ImageView` to `ImageView.ScaleType.CENTER_CROP`, which scales so the **short** edge fills the view and crops the long edge.
2. `SquircleDrawable` clips against the centered square sub-region of its bounds (`centeredSquareRegion`) rather than the full bounds.
   CENTER_CROP leaves a wrapped drawable's own `bounds` at its raw, unscaled intrinsic size (the crop and scale are applied through a `Canvas` matrix at draw time), so for a non-square photo the squircle was being built against a non-square canvas and nearly all of its curvature landed outside the window CENTER_CROP actually displays, rendering the button's corners essentially sharp.
   Since this codebase's overlay window is always square, that centered square sub-region is exactly the window CENTER_CROP crops to.
   A no-op for the already-square gallery icon.

## Tests

The acceptance test is the one written into issue #767, run for real, with no manual step:

`OverlayThumbnailFitPixelTest` builds a fixture photo per the issue's rules for each of its four edge ratios (16:9, 4:3, 9:16, 3:4), at Pixel Camera output resolution: a solid BLUE square of the photo's short edge, centred, with every other pixel RED.
Each fixture is served through the content resolver and displayed by the real `showLatestPhotoThumbnail()` path (background decode, main-looper hand-off, `SquircleDrawable(BitmapDrawable(...))` wrapping), then the overlay button is rendered over a BLACK backdrop and scanned:

- no pixel matches RED, so the reverse-letterbox bars were cropped away rather than shown;
- every pixel inside the button's squircle is BLUE, so no BLACK letterbox gap survives;
- the BLUE region spans the button edge to edge on both axes;
- the BLUE region is still superellipse-shaped rather than square, so filling the button did not cost the corner rounding.

Both halves of the fix are covered: restoring `FIT_CENTER` leaves 9.5k px (4:3) to 17k px (16:9) of the button rendering BLACK, and clipping the squircle against the photo's raw bounds again drops the BLUE region's superellipse IoU from ~0.99 to ~0.94 as the corners go square.
Both were run to confirm the test fails in each case.

This needs no emulator, no device, and no camera app.
The only obstacle was Robolectric's graphics mode: the default LEGACY mode stubs drawing out entirely, so every pixel readback comes back untouched, which is what earlier revisions of this PR mistook for an inherent limitation.
`@GraphicsMode(NATIVE)` turns on the real Skia-backed pipeline; the four ratios then run in about a second each.
The scan side reuses `PixelMask` and `ShapeTemplates`, the pure-JVM helpers the on-device E2E visual suite already uses, so a unit test and an E2E test describe a rendered button the same way.

With real rendering available, two tests added earlier in this PR no longer needed reflection:

- `SquircleDrawableRobolectricTest` now draws the real `SquircleDrawable` and asserts the shape of the pixels it covered (bounding box equals the centered square crop window; covered region matches a superellipse template filling it at IoU >= 0.95, where a square clip scores about 0.9), instead of reading the private `pathWidth` / `pathHeight` fields. Same wide, tall, and square cases; still fails if the clip is built from raw bounds.
- `CENTER_CROP matrix fits the short edge and crops the long edge for every edge ratio` is **deleted**. It read the framework's private `ImageView.mDrawMatrix` by reflection to re-derive Android's own CENTER_CROP arithmetic; `OverlayThumbnailFitPixelTest` asserts the visible result of that arithmetic for the same four edge ratios without depending on a private framework field. The scale-type test beside it is unchanged and still pins the enum.

No existing test's requirements were loosened.

`:app:testDebugUnitTest` (305 tests) and `ktlint` are clean locally.

## Test plan

Fully automated; nothing manual remains.
`OverlayThumbnailFitPixelTest` is the issue's acceptance criteria, executed in `:app:testDebugUnitTest`.

### DIFF
--- a/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
+++ b/app/src/main/java/com/gb4pc/overlay/OverlayManager.kt
@@ -260,7 +260,9 @@ class OverlayManager(
                     }
                 }
             }
-        imageView.scaleType = ImageView.ScaleType.FIT_CENTER
+        // Issue #767: CENTER_CROP scales so the short edge fills the view and the long
+        // edge is cropped, instead of FIT_CENTER's letterbox bars on the short axis.
+        imageView.scaleType = ImageView.ScaleType.CENTER_CROP
         updateIconDrawable(imageView)
         imageView.setOnClickListener { handleTap() }
         return imageView
--- a/app/src/main/java/com/gb4pc/overlay/SquircleDrawable.kt
+++ b/app/src/main/java/com/gb4pc/overlay/SquircleDrawable.kt
@@ -9,7 +9,39 @@ import android.graphics.drawable.AdaptiveIconDrawable
 import android.graphics.drawable.Drawable
 import android.os.Build
 import kotlin.math.abs
+import kotlin.math.min
 import kotlin.math.pow
+
+/** Top-left offset and side length of a centered square sub-region within a `w x h` rect. */
+internal data class SquareRegion(
+    val left: Int,
+    val top: Int,
+    val side: Int,
+)
+
+/**
+ * Computes the centered square sub-region of a `w x h` rect.
+ *
+ * Issue #767 follow-up: [ImageView.ScaleType.CENTER_CROP][android.widget.ImageView.ScaleType]
+ * leaves a [Drawable]'s own `bounds` at its raw, unscaled intrinsic size (`configureBounds()`
+ * applies the actual crop/scale through a [Canvas] matrix at draw time, not by resizing
+ * `bounds`), so for a non-square photo thumbnail `bounds` is not square. This codebase's
+ * overlay window is always square, so CENTER_CROP always crops such a drawable down to the
+ * centered square sub-region this function computes--exactly the window that ends up visible
+ * in the (square) overlay button. [SquircleDrawable] clips against this square, not the full
+ * (possibly non-square) `bounds`, so the squircle's shape is unaffected by the wrapped
+ * drawable's own aspect ratio.
+ *
+ * A no-op for an already-square rect (`left == top == 0`, `side == w == h`), so the gallery
+ * icon's clip (always square) is unchanged.
+ */
+internal fun centeredSquareRegion(
+    w: Int,
+    h: Int,
+): SquareRegion {
+    val side = min(w, h)
+    return SquareRegion(left = (w - side) / 2, top = (h - side) / 2, side = side)
+}
 
 /**
  * A [Drawable] wrapper that clips its content to a superellipse ("squircle") shape,
@@ -43,12 +75,17 @@ class SquircleDrawable(
         val b = bounds
         if (b.isEmpty) return
 
-        rebuildPathIfNeeded(b.width(), b.height())
+        // Issue #767 follow-up: clip against the centered square sub-region of `bounds`
+        // (see centeredSquareRegion's kdoc), not the full (possibly non-square) bounds.
+        val square = centeredSquareRegion(b.width(), b.height())
+        rebuildPathIfNeeded(square.side, square.side)
 
         val save = canvas.save()
         try {
             canvas.translate(b.left.toFloat(), b.top.toFloat())
+            canvas.translate(square.left.toFloat(), square.top.toFloat())
             canvas.clipPath(clipPath)
+            canvas.translate(-square.left.toFloat(), -square.top.toFloat())
 
             // Draw into the local (0,0) origin after translate.
             val localBounds = Rect(0, 0, b.width(), b.height())
--- a/app/src/test/java/com/gb4pc/overlay/OverlayManagerRobolectricTest.kt
+++ b/app/src/test/java/com/gb4pc/overlay/OverlayManagerRobolectricTest.kt
@@ -405,6 +405,10 @@ class OverlayManagerRobolectricTest {
      * [ImageView.ScaleType.CENTER_CROP], not FIT_CENTER. FIT_CENTER scales so the image's
      * long edge fits inside the view, leaving letterbox bars along the short axis;
      * CENTER_CROP scales so the short edge fills the view, cropping the long edge instead.
+     *
+     * What that scale type actually produces on screen, for each of the issue's edge ratios,
+     * is asserted from rendered pixels in [OverlayThumbnailFitPixelTest]; this test just pins
+     * the enum, so a change to it fails with an obvious message instead of a colour mismatch.
      */
     @Test
     fun `overlay ImageView uses CENTER_CROP scale type`() {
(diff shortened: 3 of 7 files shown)


aunger (commented):
🤖 Reviewer

**Core fix: correct.** I traced Android's `ImageView.configureBounds()` CENTER_CROP math by hand for all four ratios the issue names (16:9, 4:3, 9:16, 3:4). CENTER_CROP's scale is `max(viewW/drawableW, viewH/drawableH)`, which always lands the short edge exactly on the view size and crops the long edge symmetrically about the center. For the issue's fixture (a centered square spanning the short edge), that centered crop shows only blue, with no red visible. This satisfies the acceptance criteria as written.

**Tests: sound.** Reading the framework's own `mDrawMatrix` via reflection, instead of trying to render pixels, is a reasonable way around the Robolectric bitmap-draw limitation the PR description calls out. I checked the real `ImageView.setFrame()` source: it calls `configureBounds()` unconditionally on every `layout()` call, so the test's manual `.layout()` after `setImageDrawable()` does force a fresh matrix computation for the new drawable, even though the coordinates are unchanged from the initial layout. The description's account of both tests failing pre-fix and passing post-fix is consistent with the framework behavior.

**One finding that needs to be resolved before merge: squircle rounding likely disappears for non-square thumbnails.**

`SquircleDrawable.draw()` builds its clip path from the drawable's own `bounds` (`buildSuperellipsePath` uses `bounds.width()`/`bounds.height()`). But `ImageView.configureBounds()` sets a matrix-scaleType drawable's bounds to its raw, unscaled intrinsic size (`dwidth x dheight`) for every scale type except FIT_XY. CENTER_CROP's scale and translate are applied only through a Canvas matrix at draw time; the Drawable's own bounds are never resized to the view. So for a 16:9 photo thumbnail, the squircle clip path is built against a 16:9-shaped canvas (say 1600x900), not the square button.

Working the geometry through: CENTER_CROP centers a symmetric crop, so only the middle slice of that 1600-wide canvas is ever visible inside a 90px-square button (roughly x in [350, 1250] out of 1600, by my calculation). The superellipse's curvature is concentrated near the far ends of the long axis (x near 0 and 1600); at the center of the visible slice the boundary is essentially flat, inset well under 2px even at the edges of the visible window. Net effect: for every non-square ratio in the issue's acceptance criteria, the button's corners will likely render as sharp/square rather than the intended rounded squircle, because the part of the mask that actually curves falls entirely outside CENTER_CROP's cropped window.

Why I think this is in scope for this PR, not a preexisting unrelated issue:
- Under the old FIT_CENTER, the full width (or height) of the raw canvas was visible, so the curvature along the fitted axis was at least on-screen. CENTER_CROP is what newly hides it.
- This very test file's class doc already states the requirement: "overlay view must use squircle clipping for both the gallery icon state and the photo-thumbnail state" (Issue #39/#188).
- Neither of this PR's two new tests, nor the existing E2E squircle tests (`test1b_overlayBlueIsSquare`, `test1c_overlayOuterIsSquircle` in `GalleryButtonVisualE2ETest`), cover the squircle shape of the photo-thumbnail state. Those E2E tests only run before any photo is taken, against the static (square) gallery icon, so they never exercise this code path.

Please verify whether the squircle corners actually survive for non-square thumbnails (a screenshot from the on-device test-plan step would settle it, or point out where my bounds/matrix reasoning above is wrong). If they do not survive, this needs to be fixed here or explicitly deferred to a follow-up issue with justification before this merges.

CI note: `build-and-test` and `analyze-kotlin` are still in progress as of this review; `No blocking labels` failing is expected during an active review cycle per policy. Not blocking my verdict either way.

aunger:
🤖 Author

Agreed, and confirmed: I worked the superellipse formula for the same case (a 16:9 thumbnail cropped to a 90px-square view) and got the same conclusion independently (roughly 1 to 2px of curvature surviving at the very edge of the crop window, invisible in practice), so this is a real regression, in scope for this PR (CENTER_CROP is what newly hides the curvature, as you note).

Fixed in da14b80: `SquircleDrawable` now clips against the centered square sub-region of `bounds` (`centeredSquareRegion`), not the full (possibly non-square) bounds. Since this codebase's overlay window is always square, that centered square sub-region is exactly the window CENTER_CROP crops to and displays, so the squircle's proportions and position are now correct regardless of the wrapped drawable's aspect ratio. It is a no-op for the already-square gallery icon case.

Tests added:
- `SquircleDrawableTest` (plain JVM, no Robolectric): direct tests of `centeredSquareRegion` for wide, tall, and square rects, plus an invariant check across the issue's four edge ratios at realistic photo pixel sizes.
- `SquircleDrawableRobolectricTest` (new file): reads `SquircleDrawable`'s private `pathWidth`/`pathHeight` cache fields after a real `draw()` call to confirm the clip path is actually built from the square crop window, not the raw bounds, for wide and tall non-square drawables (and unchanged for a square one). Verified this fails against the pre-fix code (`pathWidth=160` instead of `90` for the wide case) and passes after.

Full `:app:testDebugUnitTest` (302 tests) and ktlint both clean.

aunger (commented):
🤖 Reviewer

Verified commit da14b80. I re-derived the CENTER_CROP crop window independently: for the 160x90 example, Android's own formula gives `scale=1`, `dx=(90-160)*0.5=-35`, so the visible local-x range is `[35, 125]`, which is exactly `centeredSquareRegion(160, 90)`'s `side=90, left=35`. The fix's crop window matches the real CENTER_CROP window exactly, for both wide and tall cases, and the translate/clipPath/translate-back sequence in `draw()` is correct: the clip is pinned in device space at the shifted origin, then the origin is restored before the (still full-bounds) inner content is drawn, so only the clip shape changes, not the drawn content's position. The already-square case is confirmed unaffected.

The new tests are meaningful: `centeredSquareRegion` gets direct pure-function coverage including an invariant check at realistic photo pixel sizes, and `SquircleDrawableRobolectricTest` calls the real `draw()` and reads back the private `pathWidth`/`pathHeight` cache to confirm the clip is actually built at the square crop size rather than the raw bounds, which is a real regression guard given the Author's account of it failing pre-fix.

One non-blocking nit: `centeredSquareRegion` centers with truncating integer division (`(w - side) / 2`), while the framework's own CENTER_CROP offset rounds via `(int)(dx + 0.5f)`. For an odd `w - side` this can differ from the framework's actual crop center by up to half a pixel. Sub-visual, not worth blocking on.

This resolves the concern from my previous comment. LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #771 can be merged):
- [ ] #772

None.

aunger:
### Verification summary

Processed the before-merging list from the [Verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/771#issuecomment-5144541519):

| Item | Result |
|---|---|
| #772: Manually verify on-device that the thumbnail fills the squircle with no letterbox bars | **NOT AUTOMATABLE** (left open; see [comment](https://github.com/aunger/gallery-button-for-pixel-camera/issues/772#issuecomment-5144566373)) |

#772 requires a real device or emulator running the actual (proprietary) Pixel Camera app plus a human visual confirmation of squircle corner rounding for landscape and portrait photos. The repo's existing emulator/E2E suite does not cover this case (it exercises the gallery icon's shape, not the post-capture photo thumbnail across aspect ratios), and closing that gap would require authoring new fixture/test source code, which is outside this role's scope. #772 stays open pending manual verification.

No before-merging item failed; none are confirmed passing by automation. Verification is incomplete pending the manual step.

aunger:
### Correction to the verification summary above

Feedback (via the orchestrating coordinator) correctly flagged that framing #772 as needing a physical device was overstated. It doesn't: this repo already runs the real Pixel Camera APK on an emulator in CI and already scripts an equivalent pixel/shape check with zero human involvement (`GalleryButtonVisualE2ETest.test1c_overlayOuterIsSquircle`, using the `ColorMatch`/`ShapeMatcher` harness). Full correction posted on the issue: https://github.com/aunger/gallery-button-for-pixel-camera/issues/772#issuecomment-5149196015

The actual, narrower blocker: `MockCameraActivity` only ever captures a fixed 1920x1080 solid-GREEN photo, so no existing automated test exercises the thumbnail-after-capture path across aspect ratios that issue #767 / #772 are about. Building that coverage (configurable mock-camera capture + a new E2E test) is new source authorship, which is out of scope for a Verification Agent (no source modification, no new tests; that's Author work).

**Net effect on this PR's status is unchanged**: #772 stays open, and the terminal signal remains **Verification incomplete**, because no automated evidence yet exists either way, not because the step is inherently unautomatable. Recommend routing the fix (extend `MockCameraActivity` + add an E2E test mirroring `test1c_overlayOuterIsSquircle`) to an Author; once that test exists, this item can be verified and closed the normal way.

aunger:
> Test plan
> Manual: verify on-device that the overlay button's thumbnail (after taking a photo) fills the squircle with no letterbox bars, for both landscape and portrait photos.

You’re lazy and being absurd. This doesn’t need a physical device; the repo already runs the real Pixel Camera app on an emulator in CI with scripted pixel/shape checks (an existing test already verifies the squircle shape this way).

… but it doesn’t even need *that* much! This test doesn’t require the Pixel Camera, nor does it require an emulator. I even designed the test for you in the issue description.

No existing test exercises the thumbnail-after-capture path across different aspect ratios yet, and building that is new test code, which is Author work, so get off your lazy butt.

I’m closing #772 because it should never have been opened.

aunger:
🤖 Author

You were right, and the earlier claim in this PR was wrong.

The test does not need a device, an emulator, or the Pixel Camera. It did not even need the E2E harness. The only thing standing in the way was Robolectric's graphics mode: by default it runs LEGACY graphics, where drawing is stubbed out and every pixel readback comes back untouched. The earlier revision saw that, concluded pixel rendering was impossible here, and wrote around it with reflection. `@GraphicsMode(GraphicsMode.Mode.NATIVE)` switches on the real Skia-backed pipeline, and the whole thing runs on the JVM in about a second per ratio.

So the test is now the one you designed in #767, verbatim. `OverlayThumbnailFitPixelTest` (b3a950f) builds a fixture photo per your rules for each of the four edge ratios, at Pixel Camera output resolution (4032x2268, 4032x3024, and their transposes): a solid BLUE square of the photo's short edge, centred, every other pixel RED. Each fixture is served through the content resolver and displayed by the real `showLatestPhotoThumbnail()` path, background decode and main-looper hand-off included, then the button is rendered over a BLACK backdrop and scanned:

- no pixel matches RED,
- every pixel inside the button's squircle is BLUE, so no BLACK survives,
- BLUE spans the button edge to edge on both axes,
- the BLUE region is still superellipse-shaped, not square.

It fails against both halves of the fix, which I checked by reverting each in turn: with `FIT_CENTER` restored, 9.5k px (4:3) to 17k px (16:9) inside the button render BLACK; with the squircle clipped against the photo's raw bounds again, the BLUE region's superellipse IoU drops from ~0.99 to ~0.94 as the corners go square.

Two consequences worth flagging explicitly, in fd53258:

- `SquircleDrawableRobolectricTest` no longer reads the private `pathWidth` / `pathHeight` fields. It draws the real `SquircleDrawable` and asserts the shape of the pixels it covered: bounding box equal to the centered square crop window, and IoU >= 0.95 against a superellipse template filling that window (a square clip scores about 0.9). Same wide, tall, and square cases, same property, checked through what a user would see.
- **I deleted the test `CENTER_CROP matrix fits the short edge and crops the long edge for every edge ratio`.** It read the framework's private `ImageView.mDrawMatrix` by reflection to re-derive Android's own CENTER_CROP arithmetic, and existed only because pixel rendering was thought unavailable. The new test asserts the visible result of that arithmetic across the same four edge ratios without touching a private framework field. The scale-type test next to it is untouched and still pins the enum.

No existing test's requirements were loosened. The manual item is gone from the test plan; nothing about this fix now depends on a human looking at a screen. `:app:testDebugUnitTest` (305 tests) and `ktlint` are clean locally.

Noted that you closed #772.

aunger (commented):
🤖 Reviewer

Reviewed the new commits (b3a950f, fd53258): the reflection-based tests are replaced with `OverlayThumbnailFitPixelTest`, a direct pixel-rendered implementation of the issue's own acceptance-criteria fixture, run under Robolectric's `@GraphicsMode(NATIVE)`. Good change; asserting on real rendered pixels through the production `showLatestPhotoThumbnail()` path is strictly better evidence than reading private framework/drawable fields, and it happens to be exactly the test the issue asked for.

I did not just take the PR's numbers on faith. I checked out fd53258 locally, ran the full `com.gb4pc.overlay` test suite (all pass), then independently reintroduced each half of the original bug one at a time to confirm the new tests actually catch them:

- Reverted `SquircleDrawable` to clip against the raw (non-square) bounds again: `OverlayThumbnailFitPixelTest` fails on the squircle-IoU assertion for all four ratios (I measured IoU 0.936 to 0.946, matching the ~0.94 the PR reports), and both `SquircleDrawableRobolectricTest` non-square cases fail too.
- Reverted the scale type to `FIT_CENTER`: all four `OverlayThumbnailFitPixelTest` cases fail on the "no BLACK inside the squircle" assertion, with exactly 9544 px (4:3 and 3:4) and 17232 to 17234 px (16:9 and 9:16) rendering BLACK, matching the PR's claimed "9.5k to 17k px" precisely.

Both halves of the fix are genuinely covered, not just asserted to be. I restored both files to the PR's committed state afterward; the branch is untouched.

One small, non-blocking nit: commit da14b80's message has two instances of `--` surrounded by spaces ("...instead -- exactly the window..." and "...overlay view -- so the squircle..."), which is against this repo's own `.claude/rules/prose-style.md` ("Do not surround dashes or double-hyphens with spaces"). Cosmetic only, not worth a PR round by itself.

This continues to resolve the concern from my earlier comments, now with stronger, more direct tests than what I originally reviewed. LGTM.

Issue #775, opened 2026-08-01
Improve PR auto-labeling: review the by-title rules, add a by-file-changes signal


aunger:
**Blocked by: #776** (a directory-based file-changes signal needs the scripts/ reorg's directories to exist first).

## Background

`scripts/label_by_title.py` labels issues/PRs purely from a regex match against the title (`ci`, `agents`, `automated tests`). That approach has real blind spots, found while auditing `.github/release.yml`'s changelog-exclude list against these three labels:

- A PR can be entirely test/CI-infrastructure work but slip past every label if its title doesn't happen to contain a trigger word. Examples: #735 ("Scan screenshots via one Bitmap.getPixels bulk read in ColorMatch.mask", touches only `app/src/androidTest`, `sharedTest`, and `test` sources) and #665 ("Bump strip-session-bylines to actions/checkout@v6", a pure `.github/workflows/*.yml` version bump with no "ci"-shaped word in the title).
- Conversely, a title can contain a trigger word for reasons unrelated to the label's real subject. #623 ("Stop PR agents if they agree, despite CI") only touches `agents/*.md` files, not a CI workflow, yet "despite CI" pulled in a `ci` match.
- The `agents` label's description was recently tightened to "AGENTS.md, CLAUDE.md, or files in agents/", which is narrower than some of the regex's sub-rules. One drift (`review\w+` catching plain "reviewing"/"reviewed") was already narrowed to `reviewers?`; nothing currently guarantees the regex and the label description stay in sync going forward.
- `automated tests` fires on any test-shaped word anywhere in the title, which conflates "primarily about test infrastructure" with "happens to touch a test file." A product bug fix whose title mentions the failing test is arguably not an "automated tests" change, but the current rule can't tell the difference.

## Asks (not prescriptive; exact approach still needs scoping)

1. Audit the by-title regexes in `scripts/label_by_title.py` against each label's current description and fix drift as found.
2. Investigate adding a by-file-changes signal via the PR Files API (`GET /repos/{owner}/{repo}/pulls/{pull_number}/files`, which returns `additions`/`deletions`/`changes` per file with no repo clone required) to catch or refine cases the title alone can't -- e.g. treating a PR as test-related when some threshold of changed lines fall under known test paths, rather than relying only on a title keyword.
3. Decide how the two signals combine (title OR files, title AND files, one overriding the other) and whether "touched a test file" and "primarily about tests" warrant separate labels or a refined rule on the existing one.

[...]

PR #780, opened 2026-08-03
Narrow label_by_title's agents rule from review\w+ to reviewers?

aunger:
## Summary

- `scripts/ci/labels/label_by_title.py`'s `agents` label rule matched any `review\w+` (reviewing, reviewed, reviews, reviewer, reviewers), which pulled titles about generic reviewing into the `agents` label even when they had nothing to do with the AI Reviewer role.
- Narrowed the rule to `reviewers?`, so only "reviewer"/"reviewers" trigger the label. "Reviewing PRs takes too long" or "Collect reviews from beta testers" no longer match; "Clarify Reviewer account sharing" still does.

## Test plan

- [x] `cd scripts/ci/labels && python3 -m unittest test_label_by_title`: all 61 tests pass, including two new ones (`test_review_requires_suffix` extended, `test_reviewing_and_reviewed_do_not_match` added) that lock in the narrowed behavior.

🦀

PR #665, opened 2026-07-12
Bump strip-session-bylines to actions/checkout@v6

aunger:
## Summary

- Every other workflow in `.github/workflows/` pins `actions/checkout@v6`; `strip-session-bylines.yml` was left on `actions/checkout@v4`.
- Bumps it to `v6` for consistency across workflows.

## Test plan

- [ ] Confirm the `Strip Claude session-URL bylines` workflow still runs successfully on the next triggering event (issue/PR opened or edited).

PR #782, opened 2026-08-03
Add a by-file-changes labeling signal for PRs where every changed file is CI or test

### LINKED ISSUE #781: Add a by-file-changes labeling signal for PRs where every changed file is CI or test
Related to #775.

This is the cheap slice of #775, scoped deliberately narrow.
The expensive parts (line-count thresholds, a new label taxonomy, conflict resolution between the two signals, a backfill path) stay in #775 and are listed under Non-goals below.

## Background

#775 documents that `label_by_title.py` matches on the title alone, so a PR that is entirely test or CI-infrastructure work slips past every label when its title happens to use no trigger word.
The two examples it cites are #735 (touches only `app/src/androidTest`, `sharedTest`, and `test` sources) and #665 (a pure `.github/workflows/*.yml` version bump).

There is one rule that catches both without thresholds, without new labels, and without a judgment call about "touched a test file" versus "is primarily about tests": **apply a label only when *every* file the PR changed maps to that label.**
A PR in which all three changed files live under a Gradle test source set is unambiguously a test change.
A PR whose only changed file is a workflow YAML is unambiguously a CI change.
A PR that mixes a test file with product code matches nothing and is left to the title rule, which is the safe direction for an add-only labeler.

The GitHub PR Files API (`GET /repos/{owner}/{repo}/pulls/{n}/files`) supplies the file list with no clone and no diff, exactly as #775 suggests.
Only the `filename` field is needed; the `additions`/`deletions`/`changes` fields that the threshold work in #775 would need are ignored here.

### Evidence

I dry-ran the exact rule proposed below over the 120 most recently merged PRs, comparing what it would apply against the labels each PR actually carries.

- It fires on 27 of 120.
- On 21 of those it re-applies a label the PR already has (no change).
- On 6 it adds a label the PR is missing: 490, 616, 635, 653, 665, 735.
- It produced no false positive in the sample.
  Every one of the 27 is a genuine test-only or workflow-only change.

Both of #775's cited misses (#735, #665) are in the "adds a missing label" set.

### Assumptions

- File paths below assume PR #779 (the scripts reorganization for #776) has merged.
  If it has not, the same files live directly in `scripts/`.
- The labeling rule itself does **not** depend on that reorganization.
  It keys on `.github/workflows/` and on Gradle test source-set directory names, both of which exist today, so #775's "blocked by #776" note does not apply to this slice.
- `ci`, `automated tests`, and `agents` are in no mutually exclusive set in `enforce_mutually_exclusive_labels.py`, and `propagate_issue_labels.py` copies labels issue-to-PR only, so nothing this adds feeds back into either.

### 1. Land PR #780 first

#775 describes the `review\w+` to `reviewers?` narrowing as already done, but it is not on `main`.
It lives in open PR #780, so it should not be redone here.
That PR was branched before the reorganization and edits `scripts/label_by_title.py`, so it needs a rebase onto `scripts/ci/labels/label_by_title.py` once #779 merges, or it lands as a modify/delete conflict.
Step 5 below edits the same `LABEL_PATTERNS["agents"]` block, so ordering matters.

### 2. Add `scripts/ci/labels/label_by_files.py`

A new sibling module to `label_by_title.py`, with the same env-var and exit-code contract (`0` for "nothing matched" or "labels applied", `1` for bad configuration or a failed call).
It imports `label_by_title` for `gh_api`, the same intra-directory import `backfill_labels_by_title.py` already relies on.

The classification is one function returning a set per path, and the rule is the intersection across all changed paths:

```python
TEST_DIR_SEGMENTS = frozenset({"test", "androidTest", "sharedTest"})
TEST_SCRIPT_BASENAME = re.compile(r"test_.+\.(py|sh)\Z")
CI_PATH_PREFIXES = (".github/workflows/",)


def labels_for_path(path: str) -> frozenset[str]:
    """Return the labels *path* alone justifies (empty when it justifies none)."""
    segments = path.split("/")
    labels = set()
    if TEST_DIR_SEGMENTS.intersection(segments[:-1]):
        labels.add("automated tests")
    if TEST_SCRIPT_BASENAME.match(segments[-1]):
        labels.add("automated tests")
    if path.startswith(CI_PATH_PREFIXES):
        labels.add("ci")
    return frozenset(labels)


def matching_labels(paths: list[str]) -> list[str]:
    """Return the labels every path in *paths* justifies; [] when *paths* is empty."""
    if not paths:
        return []
    return sorted(frozenset.intersection(*(labels_for_path(p) for p in paths)))
```

Two properties are worth stating in the module docstring, because they are what keeps the rule safe:

- An unclassified path contributes the empty set, which empties the intersection, so **any file the map does not recognize disables the file signal entirely for that PR**.
  Adding a path mapping can only ever make the signal fire more; a gap in the map can only ever make it fire less.
- The empty file list returns `[]` rather than vacuously matching everything.

`segments[:-1]` restricts the directory-segment check to directories, so a file named `test` is not itself a test.
Matching a segment rather than a fixed prefix list means `e2e-mock-gallery/src/test/` and any future module's test source set are covered with no maintenance.
The `test_*.py` / `test_*.sh` basename rule covers this repo's script-test convention wherever those scripts end up after the reorganization; it is what catches 490.

`fetch_changed_files(repo, pr_number, token)` paginates `pulls/{n}/files?per_page=100` and returns `None` if it hits `MAX_PAGES` (10, so 1000 files) without a short page.
`None` means "list possibly truncated", and `main()` treats it exactly like "no labels matched": print why, apply nothing, exit 0.
Keeping the fetch a plain paginator, with no early exit on the first unclassified file, keeps fetching and classification separable and easy to test; a PR of over 1000 files in this repo is not a case worth optimizing.

`main()` reads `GITHUB_TOKEN`, `GITHUB_REPOSITORY`, and `PR_NUMBER`, and POSTs to `repos/{repo}/issues/{n}/labels` inline, the same five lines `label_by_title.main()` uses.
Do not extract a shared `apply_labels` helper: `backfill_labels_by_title.apply_labels` deliberately continues past a failure across many issues, while the single-PR path aborts, so the two have different error contracts and unifying them is a separate change.

### 3. Add `scripts/ci/labels/test_label_by_files.py`

`build.yml`'s test step discovers by `find scripts -name "test_*.py" -printf '%h\n'`, so a new test file in an existing directory needs no CI wiring.
Mirror the structure of `test_label_by_title.py`:

- `labels_for_path` table cases: `app/src/{test,androidTest,sharedTest}/...` and `e2e-mock-gallery/src/test/...` give `{automated tests}`; `scripts/ci/labels/test_label_by_files.py` gives `{automated tests}`; `.github/workflows/build.yml` gives `{ci}`; `app/src/main/java/...`, `README.md`, `agents/pr_participation.md`, `.github/release.yml`, and `.github/allowed-test-failures.txt` all give the empty set.
- `matching_labels` cases: empty list gives `[]`; an all-test list gives `["automated tests"]`; an all-workflow list gives `["ci"]`; a list mixing a test file and a workflow file gives `[]`; a list of test files plus one product file gives `[]`.
- Two regression fixtures using the real file lists of #735 and #665, named for those PRs, so the cases #775 was filed over are pinned.
- `fetch_changed_files` cases, patching `label_by_title.gh_api` the way `test_backfill_labels_by_title.py` already does: pagination stops on a short page, and hitting `MAX_PAGES` returns `None`.
- `main()` cases mirroring `TestMain`: exit 1 on each missing or non-integer env var, no POST when nothing matched, the POST body shape, and exit 1 on `HTTPError` and `URLError`.

### 4. Add `.github/workflows/label-by-files.yml`

A near-copy of `label-by-title.yml`:

```yaml
name: Label PR by changed files

on:
  pull_request:
    types: [opened, synchronize]

permissions:
  issues: write
  pull-requests: write

jobs:
  label-by-files:
    runs-on: ubuntu-latest
    steps:
      - name: Check out repository
        uses: actions/checkout@v6

      - name: Apply labels matching the changed files
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          GITHUB_REPOSITORY: ${{ github.repository }}
          PR_NUMBER: ${{ github.event.pull_request.number }}
        run: python3 scripts/ci/labels/label_by_files.py
```

A separate workflow rather than a step added to `label-by-title.yml`, because the two signals want different triggers.
`label-by-title.yml` is gated on `github.event.action == 'opened' || github.event.changes.title`, which is right for a title but wrong for files: a file signal needs `synchronize` and does not care about title edits, and it has nothing to do on `issues` events at all.
Folding both into one workflow means conditional logic in YAML; a second workflow is boilerplate with no branching, and it keeps a failure attributable to the right check.

`synchronize` costs one short job per push.
If that run volume is unwelcome, dropping to `types: [opened]` still catches every case in the evidence above, at the cost of missing a PR that only becomes all-test after a later push.

### 5. Narrow the `agents` rule's `verif\w+` sub-pattern (separate PR)

`{_B}verif\w+{_B}` currently matches Gradle dependency-verification work.
Running `backfill_labels_by_title.py -n` lists 714 ("Harden the Gradle build with dependency verification metadata") and 760 ("Migrate to Gradle 9.x: generate the wrapper and verification metadata via a manual CI workflow") in the `add` category for `agents`, and neither touches AGENTS.md, CLAUDE.md, or `agents/`.

Fix it with the companion-word idiom already in the file for `merge`, adding a required second word that the rest of the `agents` pattern does not already catch on its own:

```python
_VERIF_COMPANION = rf"{_B}planner{_B}|{_B}labels?{_B}|{_B}prs?{_B}"
_VERIF = rf"(?=.*{_B}verif\w+{_B})(?=.*({_VERIF_COMPANION}))"
```

`agent` is deliberately not in the companion list: a title containing it already matches the `agent\w*` sub-rule independently.
Both existing tests still pass ("Add PR verification agent instructions" via `agent\w*`, "Test PR: verify remove-verified-on-push" via the `prs?` companion), and 714 and 760 stop matching.
Add tests for those two titles as non-matches, plus "Verification Planner should file tracking issues" as a match.

One decision for the requester: this trades a false positive for a possible false negative.
An agents-process title that says "verification" with none of planner, label, or PR nearby (something like "Verification incomplete signal routing") would stop matching.
Confirm the companion list before implementing, or say to skip step 5 entirely; steps 2 through 4 do not depend on it.

## Verification

- `python3 -m unittest discover -s scripts/ci/labels -p "test_*.py" -v` locally, and `scripts/lint/lint.sh --check --all` for ruff and YAML hygiene.
- CI needs no new wiring: `build.yml`'s discovery loop picks up the new test file, and `lint.yml` already covers the new workflow YAML.
- No change is needed in `scripts/ci/test_label_existence_integration.py`.
  Both labels this applies are already in `REQUIRED_LABELS` via `.github/release.yml`'s exclude list.
- After merge, the first PR touching only `.github/workflows/` should pick up `ci` with no title trigger; confirm from that PR's Actions run rather than assuming.
  The `enforce-mutually-exclusive-labels` workflow will also fire on the resulting `labeled` event, find no conflicting set, and exit 0.
- Re-running the 120-PR dry-run described above is the cheapest regression check if the path map is ever extended.

### Accepted consequences

- Both labels this can apply are in `.github/release.yml`'s changelog-exclude list, so a file-based label drops the PR from generated release notes.
  That is the intent (#775 was filed out of an audit of exactly that list), but it means a false positive is user-visible.
  The unanimity rule is what keeps that risk low.
- A PR touching only `.github/workflows/CLAUDE.md` gets `ci`.
  That file documents the CI architecture, so this is defensible, and carving out `.md` under `.github/workflows/` is not worth the special case.
- A fork PR gets a read-only token, so the POST fails and the check goes red.
  `label-by-title.yml` has the identical exposure today, so this adds no new class of failure.

## Non-goals

Deferred to the more expensive approach in #775:

- **Thresholds.** No use of `additions`/`deletions`/`changes`, and no "most of the changed lines are under test paths" rule.
  Unanimity over paths is the whole rule.
- **Signal precedence.** Both scripts only ever add labels, so this cannot fix #775's opposite complaint, the title rule over-firing.
  623 ("Stop PR agents if they agree, despite CI", touching only `agents/*.md`) keeps its wrong `ci` label.
  Removing a title-applied label requires deciding which signal wins, which is #775's ask 3.
- **New labels.** No "touched a test file" label distinct from "is primarily about tests".
  Unanimity sidesteps that question instead of answering it.
- **A file-based `agents` rule.**
  The `agents/` tree, `AGENTS.md`, `CLAUDE.md`, the `.claude/` tree, and `scripts/agents/` all stay unclassified.
  The evidence in #775 shows no `agents` miss, and mapping `.claude/` or `scripts/agents/` reaches past the label's current description ("AGENTS.md, CLAUDE.md, or files in agents/"), which would mean editing the label description in repo settings and the out-of-repo verification that implies.
- **Mapping the `scripts/ci/` tree to `ci`.**
  A natural one-line follow-on once #779 has merged and settled, left out of the first version so this slice depends on nothing.
- **`.github/allowed-test-failures.txt` and `.github/release.yml`.** Left unclassified.
  Mapping the allowlist to `ci` fired on 639 (a test quarantine) in the dry-run, where the label is arguable; a single-file special case is not worth the argument.
- **Backfill.** `backfill_labels_by_title.py` is untouched, and the file signal applies going forward only.
  Backfilling would need a Files API call per PR across the whole repo history.
  Note the side effect: once the file signal starts applying labels the title rules would not, those PRs show up in the backfill's `miss` category.
  That category is advisory and the script only ever adds, so nothing breaks, but a future by-title audit should expect the extra entries.
- **A broader by-title audit.** Beyond step 5 and PR #780, the remaining by-title complaints in #775 are not cheap regex fixes.
  Bare "CI" in 623 is a legitimate trigger word, and 643 ("Rename testing label to automated tests") is about the label rather than about tests.
  Distinguishing those needs the title-plus-files combination in #775's ask 3.

aunger:
## Summary

- `scripts/ci/labels/label_by_title.py` matches only on the PR/issue title, so a PR that is entirely test or CI-infrastructure work slips past every label when its title happens to use no trigger word. #775 cites #735 (touches only `app/src/androidTest`, `sharedTest`, and `test` sources) and #665 (a pure `.github/workflows/*.yml` version bump) as examples.
- Adds `scripts/ci/labels/label_by_files.py`, a sibling module that fetches a PR's changed files via the GitHub PR Files API and applies a label only when *every* changed path maps to that label:
  - Gradle test source-set directories (`test`, `androidTest`, `sharedTest`, matched as path segments so any module's test tree is covered) and `test_*.py` / `test_*.sh` scripts give `automated tests`.
  - Paths under `.github/workflows/` give `ci`.
  - An unclassified path contributes the empty label set, which empties the intersection across all paths in the PR. So any file the map doesn't recognize disables the file signal entirely for that PR: adding a path mapping can only ever make the signal fire more, and a gap in the map can only ever make it fire less. The empty file list also returns no labels rather than vacuously matching everything.
- `fetch_changed_files` paginates `pulls/{n}/files?per_page=100`, stopping on the first short page. If `MAX_PAGES` (10, so 1000 files) is exhausted without ever seeing a short page, it returns `None` ("possibly truncated"), and `main()` treats that exactly like "nothing matched": prints why, applies nothing, exits 0.
- Wires it up via a new `.github/workflows/label-by-files.yml`, separate from `label-by-title.yml` because the two signals want different triggers: title labeling only needs to run when the title itself changes (or on `issues` events), while the file signal needs `synchronize` (a later push can make a PR become all-test) and has nothing to do with title edits or issue events.
- Labels are only ever added, never removed, matching `label_by_title.py`'s existing contract.

This is the cheap slice of #775 described in that issue's implementation plan: no line-count thresholds, no new label taxonomy, no conflict resolution between the title and file signals, no backfill path. Those stay in #775.

Fixes #781.

## Implementation notes

- PR #780 (narrowing the `agents` title rule's `verif\w+` sub-pattern) is already merged, so this branches cleanly from current `main` with no rebase needed.
- Step 5 of #781's plan (narrowing the `agents` rule's `verif\w+` sub-pattern further, to cut a companion-word requirement) is explicitly scoped as a separate PR and is not included here.
- No change needed in `scripts/ci/test_label_existence_integration.py`: both `ci` and `automated tests` are already in `REQUIRED_LABELS` via `.github/release.yml`'s changelog-exclude list.

## Test plan

- [x] `python3 -m unittest discover -s scripts/ci/labels -p "test_*.py"`: all 289 tests pass, including 28 new ones in `test_label_by_files.py` (path classification table, `matching_labels` unanimity cases, regression fixtures pinning the real file lists of #735 and #665, `fetch_changed_files` pagination, and `main()`).
- [x] `ruff check` / `ruff format --check` clean on the new files.
- [ ] After merge, confirm from a PR touching only `.github/workflows/` that it picks up `ci` with no title trigger, per #781's verification note.

### DIFF
--- a/.github/workflows/label-by-files.yml
+++ b/.github/workflows/label-by-files.yml
@@ -1,0 +1,33 @@
+name: Label PR by changed files
+
+# Applies labels to a pull request based on the set of files it changed,
+# when every changed file maps to the same label. See
+# scripts/ci/labels/label_by_files.py for the exact rule. Labels are only
+# ever added, never removed.
+#
+# A separate workflow from label-by-title.yml because the two signals want
+# different triggers: title labeling only needs to run when the title
+# itself changes, while this needs to run on every push (synchronize) and
+# has nothing to do on title edits or issue events.
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  label-by-files:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Apply labels matching the changed files
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: python3 scripts/ci/labels/label_by_files.py
--- a/scripts/ci/labels/label_by_files.py
+++ b/scripts/ci/labels/label_by_files.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python3
+"""Apply labels to a pull request based on the set of files it changed.
+
+Unlike label_by_title.py (which matches a title against a regex per label),
+this module classifies each *changed path* and applies a label only when
+every changed path justifies it. See matching_labels() below for the exact
+rule--not duplicated here, to avoid this docstring drifting out of sync with
+the code.
+
+Two properties keep the rule safe:
+
+- An unclassified path contributes the empty label set, which empties the
+  intersection across all paths. So any file the map below does not
+  recognize disables the file signal entirely for that PR: adding a path
+  mapping can only ever make the signal fire more, and a gap in the map can
+  only ever make it fire less.
+- The empty file list returns [] (no labels) rather than vacuously matching
+  every label, since intersecting zero sets has no defined result.
+
+Labels are only ever added, never removed.
+
+Usage:
+    python3 scripts/ci/labels/label_by_files.py
+
+Exit code:
+    0  no labels matched, the changed-file list may be truncated, or labels
+       were applied successfully.
+    1  required configuration is missing/invalid, or a GitHub API call
+       failed.
+
+Required environment variables:
+    GITHUB_TOKEN        Personal access token or Actions secret with
+                        issues: write and pull-requests: write
+    GITHUB_REPOSITORY   Owner/repo (e.g. "aunger/gallery-button-for-pixel-camera")
+    PR_NUMBER           Number of the pull request
+"""
+
+import os
+import re
+import sys
+import urllib.error
+
+import label_by_title
+
+# ---------------------------------------------------------------------------
+# Path classification
+# ---------------------------------------------------------------------------
+
+TEST_DIR_SEGMENTS = frozenset({"test", "androidTest", "sharedTest"})
+TEST_SCRIPT_BASENAME = re.compile(r"test_.+\.(py|sh)\Z")
+CI_PATH_PREFIXES = (".github/workflows/",)
+
+
+def labels_for_path(path: str) -> frozenset[str]:
+    """Return the labels *path* alone justifies (empty when it justifies none)."""
+    segments = path.split("/")
+    labels = set()
+    # segments[:-1] restricts this check to directories, so a file literally
+    # named "test" is not itself a test.
+    if TEST_DIR_SEGMENTS.intersection(segments[:-1]):
+        labels.add("automated tests")
+    if TEST_SCRIPT_BASENAME.match(segments[-1]):
+        labels.add("automated tests")
+    if path.startswith(CI_PATH_PREFIXES):
+        labels.add("ci")
+    return frozenset(labels)
+
+
+def matching_labels(paths: list[str]) -> list[str]:
+    """Return the labels every path in *paths* justifies; [] when *paths* is empty."""
+    if not paths:
+        return []
+    return sorted(frozenset.intersection(*(labels_for_path(p) for p in paths)))
+
+
+# ---------------------------------------------------------------------------
+# GitHub API helper
+# ---------------------------------------------------------------------------
+
+# 10 pages of 100 files each. A PR with more than 1000 changed files is not a
+# case worth optimizing for; fetch_changed_files() returns None rather than
+# risk a verdict from a partial file list.
+MAX_PAGES = 10
+PER_PAGE = 100
+
+
+def fetch_changed_files(repo: str, pr_number: int, token: str) -> list[str] | None:
+    """Return every path changed in the pull request, or None if possibly truncated.
+
+    Paginates GET pulls/{pr_number}/files?per_page=100. Stops as soon as a
+    page comes back shorter than PER_PAGE (the normal end of the list) or
+    empty. If MAX_PAGES pages are consumed without ever seeing a short page,
+    the file list may continue beyond what was fetched, so this returns
+    None instead of a partial list.
+    """
+    paths: list[str] = []
+    page = 1
+    while page <= MAX_PAGES:
+        batch = label_by_title.gh_api(
+            f"repos/{repo}/pulls/{pr_number}/files?per_page={PER_PAGE}&page={page}",
+            token=token,
+        )
+        if not batch:
+            return paths
+        paths.extend(item["filename"] for item in batch)
+        if len(batch) < PER_PAGE:
+            return paths
+        page += 1
+    return None
+
+
+# ---------------------------------------------------------------------------
+# Entry point
+# ---------------------------------------------------------------------------
+
+
+def main() -> int:
+    token = os.environ.get("GITHUB_TOKEN", "")
+    repo = os.environ.get("GITHUB_REPOSITORY", "")
+    pr_number_str = os.environ.get("PR_NUMBER", "")
+
+    if not token:
+        print("Error: GITHUB_TOKEN not set.", file=sys.stderr)
+        return 1
+    if not repo:
+        print("Error: GITHUB_REPOSITORY not set.", file=sys.stderr)
+        return 1
+    if not pr_number_str:
+        print("Error: PR_NUMBER not set.", file=sys.stderr)
+        return 1
+
+    try:
+        pr_number = int(pr_number_str)
+    except ValueError:
+        print(f"Error: PR_NUMBER is not a valid integer: {pr_number_str!r}", file=sys.stderr)
+        return 1
+
+    try:
+        paths = fetch_changed_files(repo, pr_number, token)
+    except urllib.error.HTTPError as exc:
+        print(f"Error fetching changed files for #{pr_number}: {exc}", file=sys.stderr)
+        return 1
+    except urllib.error.URLError as exc:
+        print(f"Network error fetching changed files for #{pr_number}: {exc}", file=sys.stderr)
+        return 1
+
+    if paths is None:
+        print(
+            f"Changed-file list for #{pr_number} may be truncated "
+            f"(hit {MAX_PAGES} pages)--nothing to do."
+        )
+        return 0
+
+    labels = matching_labels(paths)
+    if not labels:
+        print(f"No label rules matched the changed files on #{pr_number}--nothing to do.")
+        return 0
+
+    try:
+        label_by_title.gh_api(
+            f"repos/{repo}/issues/{pr_number}/labels",
+            token=token,
+            method="POST",
+            body={"labels": labels},
+        )
+        print(f"Applied labels {labels} to #{pr_number}.")
+    except urllib.error.HTTPError as exc:
+        print(f"Error applying labels {labels} to #{pr_number}: {exc}", file=sys.stderr)
+        return 1
+    except urllib.error.URLError as exc:
+        print(f"Network error applying labels {labels} to #{pr_number}: {exc}", file=sys.stderr)
+        return 1
+
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/scripts/ci/labels/test_label_by_files.py
+++ b/scripts/ci/labels/test_label_by_files.py
@@ -1,0 +1,256 @@
+#!/usr/bin/env python3
+"""Unit tests for label_by_files.py."""
+
+import os
+import sys
+import unittest
+import urllib.error
+from unittest.mock import patch
+
+sys.path.insert(0, os.path.dirname(__file__))
+import label_by_files as lbf  # noqa: E402
+
+
+# ---------------------------------------------------------------------------
+# labels_for_path tests
+# ---------------------------------------------------------------------------
+
+
+class TestLabelsForPath(unittest.TestCase):
+    def test_gradle_test_source_sets_give_automated_tests(self):
+        for path in (
+            "app/src/test/java/com/gb4pc/Foo.kt",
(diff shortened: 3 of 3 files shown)


aunger (commented):
🤖 Reviewer

Verdict: LGTM.

I checked this against #781's implementation plan point by point and against the actual repo state (branch checked out locally, not just the diff).

What I verified:

- The diff is exactly the three files the PR claims: `.github/workflows/label-by-files.yml`, `scripts/ci/labels/label_by_files.py`, `scripts/ci/labels/test_label_by_files.py`. No scope creep, no touch to `label_by_title.py` itself (step 1/step 5 correctly left out).
- `label_by_files.py` matches `label_by_title.py`'s contract precisely: same env-var style, same `gh_api` reuse via `import label_by_title`, same exit-code semantics, no shared `apply_labels` helper (as the issue explicitly asked to avoid).
- Ran `python3 -m unittest discover -s scripts/ci/labels -p "test_*.py" -v` locally on the PR branch: 289/289 pass, matching the PR description's claim exactly (28 new tests, confirmed by count).
- `ruff check` and `ruff format --check` both clean on the two new Python files.
- Confirmed the actual repo has `e2e-mock-gallery/src/test`, and that `ci`/`automated tests` are the real label names used in `.github/release.yml`'s exclude list, so the path classification isn't guessing at names that don't exist.
- Pulled the job log for this PR's own new `label-by-files` check (run 30785925037/job 91599327179): it printed "No label rules matched the changed files on #782--nothing to do." That's the correct outcome by design, since this very PR mixes a CI-prefixed workflow file, an unclassified product file (`label_by_files.py`), and a test file, so the unanimity rule correctly refuses to fire. Good real-world confirmation the intersection logic works as documented, not just in unit tests.
- `matching_labels`, `labels_for_path`, `fetch_changed_files` pagination/truncation, and `main()` all match the pseudocode and test list the issue spelled out, and the test file adds a few reasonable extras beyond what was asked (endpoint-shape assertion, truncation-in-main case).
- The `#735`/`#665` regression fixtures are real and pinned as requested.
- `actions/checkout@v6` is first-party (`actions/` namespace), so the pinning rule in `code_edit.md` doesn't apply; no issue there.

One non-blocking nit: `matching_labels` calls `frozenset.intersection(*(labels_for_path(p) for p in paths))` as an unbound-method call. It's correct and well-tested, just a slightly unusual style versus computing the list once and calling `.intersection()` on the first element. Not worth a round over.

I know "No blocking labels" is failing on this PR; per `pr_participation.md` that's expected during the active dev cycle and I'm not counting it against the review. The Kotlin/Android jobs (`build-and-test`, `analyze-python`) were still in progress when I looked, but nothing in this diff touches Kotlin or app code, so I have no reason to expect them to be affected by this change.

aunger:
### Verification plan

**Before merging** (must resolve before PR #782 can be merged):
None.

- [ ] #783
- [ ] #784

Issue #774, opened 2026-08-01
Migrate the Gradle toolchain to 9.x (Gradle 9.5.1, AGP 9.1.0, KGP 2.4.0) via a manual CI generator workflow


aunger:
Supersedes #760, whose comment thread holds the full planning and adversarial-review history; this issue is the clean final plan.
Through #760 it also supersedes #759, #754, #755, #756, and #757; all of those close when this lands.

## Background

- The committed wrapper JAR is the Gradle 8.14 build while `gradle-wrapper.properties` pins the 8.9 distribution, so the fix is a full matched-set regeneration.
  (The current `gradlew` is pristine 8.13/8.14-era generator output; its `-jar` launch is the standard modern form.)
- Gradle backports fixes only to the latest minor of a major (plus the final minor of the previous major), so the pinned 8.9 is long EOL; we migrate to 9.x once rather than aligning within 8.x now and migrating later.
- AGP 8.7.3 does not run on Gradle 9.x, so the AGP bump and the Kotlin/Compose bumps it drags in are part of this change.
- Gradle distributions are hosted as `github.com` release assets; whether a web session can download them through the session's GitHub proxy varies by environment (hard-blocked in the 2026-07 sessions that shaped this plan, open in a 2026-08 session).
  **CI is the venue for the generation steps** regardless: deterministic, logged provenance for the artifacts a human reviews, the artifact-upload human gate, and a reusable ritual for future bumps.
  Local downloads, when available, are a bonus validation layer only (see the dated protocol comment below).

## Rejected alternatives (do not revisit; rationale archived in #760)

- **Vendoring the distribution zip**: exceeds GitHub's 100 MiB per-file limit.
- **Committing `distributionUrl=file:/...`**: breaks every environment except one sandbox.
- **Web-environment Setup-script mirror download**: unnecessary here; optional follow-up (the SDK installs fine in web sessions; only the KVM-dependent emulator is CI-bound).
- **Gradle 9.6.1 + AGP 9.3.0 + newest KGP** (all stable releases): no fully-tested combination documented.
- **The AGP 9 built-in Kotlin opt-out properties**: removed in AGP 10.0; we migrate now.
[...]

PR #794, opened 2026-08-06
Migrate sharedTest sourceSets wiring from srcDir to the directories set

### LINKED ISSUE #793: Migrate sourceSets wiring off the deprecated srcDir API before Gradle/AGP 10
Follow-up from #774 Step 3, where this was flagged and deliberately left out of scope.

Under AGP 9.1, configuring `app/build.gradle.kts` emits (once per consuming source set):

```text
w: 'fun srcDir(srcDir: Any): Any' is deprecated. Use `directories` mutable set instead.
```

This comes from the `sharedTest` wiring (`kotlin.srcDir("src/sharedTest/java")` in the `test` and `androidTest` source sets).
The Step 3 executor A/B-tested it: the warning is identical with the old `java.srcDir` form, so it is AGP 9's deprecation of `srcDir` itself, not something the migration introduced.

What needs to happen:

- Replace the `srcDir(...)` calls with the `directories` mutable-set API.
- The behavior question that kept this out of #774's scope: confirm whether mutating `directories` appends to or replaces the default source directories, and wire it so both `test` and `androidTest` keep their default sources PLUS `src/sharedTest/java`.
  Verify the same way #774 did: `ShapeMatcher`/`ScreenshotNaming` classes must appear under both `built_in_kotlinc/debugUnitTest/` and `built_in_kotlinc/debugAndroidTest/` outputs, and both test suites must stay green.
- Sweep for any other `srcDir` uses that AGP 9 deprecates (none known beyond these two sites at filing time).

Deprecations of this shape are removals at the next major (the pattern that produced #774's `Project#exec` pre-PR), so this should land before any AGP/Gradle 10 planning starts.

aunger:
🤖 Author

Fixes #793.

## The problem

AGP 9.1 deprecates `AndroidSourceDirectorySet.srcDir`, so configuring `app/build.gradle.kts` emitted a warning once per consuming source set:

```text
w: app/build.gradle.kts:121:20: 'fun srcDir(srcDir: Any): Any' is deprecated. Use `directories` mutable set instead.
w: app/build.gradle.kts:124:20: 'fun srcDir(srcDir: Any): Any' is deprecated. Use `directories` mutable set instead.
```

Deprecations of this shape become removals at the next major, so this lands before any AGP/Gradle 10 planning starts.

## The change

Both `kotlin.srcDir("src/sharedTest/java")` calls become `kotlin.directories += "src/sharedTest/java"`. That is the entire diff: five lines added, two removed, in one file.

## The behavior question that kept this out of #774's scope

#793 asked whether mutating `directories` appends to or replaces the default source directories. It appends.

`directories` is declared `val directories: MutableSet<String>` on `AndroidSourceDirectorySet` (confirmed against the AGP 9.1.0 `gradle-api` jar; `srcDir`, `srcDirs`, and `setSrcDirs` are all deprecated there, and `directories` is the only non-deprecated member besides `getName`). It arrives prepopulated with the source set's defaults, and it is the live set rather than a defensive copy, so `+=` adds to the defaults in place. Verified by printing the set on both sides of the mutation during configuration:

```text
test         BEFORE=[src/test/java, src/test/kotlin]
test         AFTER =[src/test/java, src/test/kotlin, src/sharedTest/java]
androidTest  BEFORE=[src/androidTest/java, src/androidTest/kotlin]
androidTest  AFTER =[src/androidTest/java, src/androidTest/kotlin, src/sharedTest/java]
```

The defensive-copy case was the real risk here: had `getDirectories()` returned a copy, `+=` would have silently done nothing and the shared helpers would have vanished from both suites. The output comparison below rules that out.

A comment in `app/build.gradle.kts` now records the append-not-replace semantics so the next reader does not have to re-derive it.

## Verification

Verified the same way #774 did, comparing against a pre-change baseline built from the same tree.

The full `*.class` listings under `built_in_kotlinc/debugUnitTest/` and `built_in_kotlinc/debugAndroidTest/` are byte-for-byte identical before and after (53 and 56 classes respectively). `ShapeMatcher`, `ScreenshotNaming`, `PixelMask`, `ShapeTemplates`, and `MaskData` all appear under both outputs, alongside every default-sourced class.

`:app:compileDebugUnitTestKotlin`, `:app:compileDebugAndroidTestKotlin`, and `:app:testDebugUnitTest` all pass locally: 305 unit tests, 305 PASS, 0 failures. The `srcDir` deprecation warning is gone, and no new build-script warning replaces it.

## Sweep for other deprecated srcDir uses

`srcDir`, `srcDirs`, and `setSrcDirs` appear nowhere else in the repository. The two `app/build.gradle.kts` sites this PR changes were the only ones, matching the issue's expectation.

## No new tests

This is a pure build-configuration refactor with no behavior change, and the regression guard already exists: `app/src/test/.../ShapeMatcherTest.kt` and `ScreenshotNamingTest.kt` consume the shared helpers from the unit-test source set, and five `androidTest` files consume them from the instrumentation source set. If this wiring ever stops appending to the defaults, both suites fail to compile. No existing test's requirements were loosened.

## Test plan

- [ ] Instrumentation suite (`:app:connectedDebugAndroidTest`) on a device or emulator. The androidTest source set is compiled but not executed here, since no device was available in this environment; its compilation is what proves the shared helpers are wired in.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -116,12 +116,15 @@ android {
 
     // Shared pure-JVM test helpers compiled into both test and androidTest source sets.
     // AGP testFixtures does not support Kotlin on application modules (b/139438662).
+    // `directories` arrives prepopulated with the source set's defaults (src/<name>/java
+    // and src/<name>/kotlin) and is mutated in place, so `+=` adds to those defaults
+    // rather than replacing them.
     sourceSets {
         getByName("test") {
-            kotlin.srcDir("src/sharedTest/java")
+            kotlin.directories += "src/sharedTest/java"
         }
         getByName("androidTest") {
-            kotlin.srcDir("src/sharedTest/java")
+            kotlin.directories += "src/sharedTest/java"
         }
     }
 


aunger:
🤖 Reviewer

**LGTM.** I re-derived every load-bearing claim independently rather than taking the description's word for it. All of them held.

## What I verified

**The API claim, against the jar.** Disassembled `AndroidSourceDirectorySet` from the AGP 9.1.0 `gradle-api` artifact:

```
public abstract java.lang.String getName();
public abstract java.lang.Object srcDir(java.lang.Object);
public abstract java.util.Set<java.lang.String> getDirectories();
public abstract java.lang.Object srcDirs(java.lang.Object...);
public abstract java.lang.Object setSrcDirs(java.lang.Iterable<?>);
```

`srcDir`, `srcDirs`, and `setSrcDirs` each carry ``kotlin.Deprecated(message="Use `directories` mutable set instead")``; `directories` and `getName` are the only clean members. Matches your description exactly. `AndroidSourceSet.getKotlin()` is itself not deprecated, so the receiver stays valid.

**The append-vs-replace question, at bytecode level.** This was the real risk here, and printing the set during configuration only shows that it *happened* to work — not *why*, so that evidence alone would not survive an AGP patch bump. The implementation settles it: in `DefaultAndroidSourceDirectorySet`, `directories` is a `private final AbstractMutableSet<String>` field instantiated once as an anonymous subclass (`DefaultAndroidSourceDirectorySet$directories$1`) holding a `this$0` back-reference, with `add(String)` overridden to write through to the enclosing source set. `getDirectories()` returns that field. It is the live set, not a defensive copy, so `+=` — which resolves to `MutableCollection.plusAssign` → `add`, unambiguously, because `directories` is a `val` — appends in place. The silent-no-op failure mode you flagged is structurally impossible, not merely empirically absent.

**The build, from both sides.** Built `:app:compileDebugUnitTestKotlin` and `:app:compileDebugAndroidTestKotlin` on this branch, and again on a baseline with only `app/build.gradle.kts` reverted:

- Class listings under `built_in_kotlinc/debugUnitTest/` and `built_in_kotlinc/debugAndroidTest/` are identical before and after — `diff` clean, 53 and 56 classes. Your numbers.
- `ShapeMatcher`, `ScreenshotNaming`, `PixelMask`, `ShapeTemplates`, and `MaskData` all land in both outputs, alongside default-sourced classes (`PixelMaskTest`, `ShapeMatcherTest`, …). Defaults preserved *and* extended, which is the issue's actual requirement.
- `:app:testDebugUnitTest`: 305 tests, 0 failures, 0 errors, 0 skipped.

**The warning.** Build-script compilation is content-hash cached, so a plain rerun proves nothing on this point. Forcing recompilation on each side: the baseline emits the warning at exactly `app/build.gradle.kts:121:20` and `124:20`, and this branch emits nothing — with no replacement build-script warning.

**The sweep.** `srcDir`/`srcDirs`/`setSrcDirs`/`getSrcDirs` match nowhere in the repository on this branch. The issue's third bullet is satisfied.

Recording the append-not-replace semantics in a comment is the right call — the issue asked for that determination, and it would otherwise have to be re-derived by the next reader. The stated defaults (`src/<name>/java`, `src/<name>/kotlin`) are accurate.

## Nit

The branch name, `fix/issue-793-concurrent-worktree-orchestration`, describes something unrelated to this change and survives in the merge commit subject. Not worth a round on its own — fold it in only if you push again for another reason.

## Outstanding before merge

The unchecked `:app:connectedDebugAndroidTest` box is the one item I could not exercise (no device available here). I would weight it as near-zero risk rather than a real gap: the `debugAndroidTest` class output is byte-for-byte identical to the pre-change baseline, so there is no compiled difference for a device run to expose. Flagging it for the verification step rather than blocking on it.

CI was still queued when I finished; not blocking on it, per the review rules.

*(Posted as an ordinary comment: GitHub rejects self-approval since our agents share one account.)*

aunger:
### Verification plan

**Before merging** (must resolve before PR #794 can be merged):
- [ ] #798

- [ ] #799

aunger:
🤖 Verification Agent

## Verification summary

Before-merging item from the [verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/794#issuecomment-5212357464):

| Item | Result |
|---|---|
| #798 — Run `:app:connectedDebugAndroidTest` on a real device/emulator, confirm `ShapeMatcher`/`ScreenshotNaming`/`PixelMask`/`ShapeTemplates`/`MaskData` present and suite passes | **PASS** |

Full evidence and reasoning posted on #798 (now closed). Short version: live-fired this repo's `build.yml` CI on a real emulator against PR #794's exact head commit via a throwaway test PR (#800, now closed) — `connectedDebugAndroidTest` passed cleanly (22/22 tests, 0 failures), and the same CI run additionally exercised the shared helper classes at runtime via `GalleryButtonVisualE2ETest` (which shares the identical built `androidTest` APK), confirming they're not just compiled but actually invoked and passing (8/8 tests, including the two that call `ShapeMatcher`/`ShapeTemplates` directly).

(#799, the follow-on item, is out of scope for merging per the verification plan and was left untouched.)

**Verification passed.**

PR #807, opened 2026-08-09
Install the helper-script lock at session start, through a shared pinned-requirements installer

### LINKED ISSUE #806: Session setup does not install scripts/requirements.txt, so part of the test suite is unrunnable outside CI
The Python helper scripts' runtime dependencies are declared and hash-pinned in `scripts/requirements.txt` (`defusedxml`, `requests`, `PyYAML`, from #723), and CI installs them. Nothing in the repo's session setup does, so a chunk of the test suite fails locally for reasons unrelated to the code under test.

This is not a missing dependency. It is a provisioning asymmetry between CI and a session.

## Where the two sides diverge

- **CI installs them.** `.github/workflows/build.yml` has an "Install Python script dependencies" step running `pip install --require-hashes -r scripts/requirements.txt`. Its comment names the reason explicitly: `test_summarize_preflight_integration.sh` drives `summarize_test_results.py`, which imports `defusedxml`, so the split job has to install the deps itself.
- **Session setup does not.** `.claude/hooks/session-start.sh` Step 3b installs only `scripts/lint/requirements-lint.txt`. Neither it nor `.claude/setup-environment.sh` touches `scripts/requirements.txt`.

`.claude/environment.md` already records that these deps "install in CI", which is accurate and is exactly the gap.

## Measured blast radius

In a current web session, with `main` at 647a50b:

- **1 of 13 shell tests** fails: `scripts/ci/test-support/test_summarize_preflight_integration.sh`.
- **4 of 16 Python test modules** cannot import at all: `test_file_test_failure_issues.py`, `test_check_allowed_failures.py`, `test_summarize_test_results.py`, and `test_label_existence_integration.py`. CI runs these via `python3 -m unittest discover`.

All five fail with `ModuleNotFoundError: No module named 'defusedxml'`.

Only `defusedxml` is missing today, but that is luck rather than design: `requests` and `PyYAML` resolve from the base image (`/root/.local/...` and `/usr/lib/python3/dist-packages/`), not from anything the repo declares. A base-image change would extend this failure to the other two with no signal.

## Why it is worth fixing

- **It taxes every local run.** `find scripts -name 'test_*.sh'` is what CI runs, so it is the natural thing to run locally, and it comes back red for a reason that has nothing to do with your change. Confirming that costs a stash-and-rerun against clean `main` every time.
- **A red line nobody trusts is worse than no line.** This is the same dynamic that made #789 a problem: a signal that is reliably noise trains people to skim past it.
- **There is precedent for it hiding something real.** #617 was a case where a misleading `defusedxml` `ModuleNotFoundError` masked an Android SDK Tools download failure.

## Suggested shape

`session-start.sh` already contains the pattern to copy. Step 3b installs the lint lock under a SHA-256 marker file (`$HOME/.local/share/gb4pc/requirements-lint.sha256`) so the install is skipped when the lock has not changed. A parallel block for `scripts/requirements.txt`, with its own marker, would mirror it, and `--require-hashes` carries over unchanged since the lock is already generated for it.

Worth deciding as part of the fix: whether `.claude/setup-environment.sh` should also seed these into the cached image, the way it already does for Gradle, the SDK, and Temurin, so the per-session install becomes a fast no-op rather than a download.

## Out of scope

Not proposing any change to `scripts/requirements.in`, the lock, or the CI install step. All three are correct; only the session side is missing.

aunger:
🤖 Author

Fixes #806.

CI installed `scripts/requirements.txt` before running either test suite; nothing on the session side did.
That asymmetry made part of the suite unrunnable outside CI, and the failures pointed nowhere near the change under test.

## What changed

**`scripts/install-pinned-requirements.sh` (new).**
The hash-pinned pip install and its SHA-256 skip marker, lifted out of the hook's STEP 3b so a second lock does not mean a second copy of the flags, the marker convention, and the ordering that keeps a failed install from being recorded as done.
It follows the `scripts/lint/install-ktlint.sh` precedent: a shared, self-logging installer the hook delegates to.
The extraction commit changes no executable behavior; the flags, the marker path, and the skip condition are identical, and only the log prefix and the rationale comments differ.

**`.claude/hooks/session-start.sh` STEP 4 (new).**
Installs `scripts/requirements.txt` through that script.
It is a step of its own rather than a STEP 3 sub-step because these are runtime dependencies of the repository's Python scripts, not lint tools; the git fetch step is renumbered to STEP 5 to make room.

**`.claude/environment.md`.**
The helper-script dependency section described the CI-only install as the whole story, which was exactly the gap.
It now covers both provisioning paths and what the marker does, and `setup-environment.sh`'s "what it deliberately does not do" list records the decision the issue asked for (below).

## Why the deps are not seeded into the cached image

The issue asked whether `.claude/setup-environment.sh` should seed these the way it seeds Gradle, the SDK, and Temurin. It should not:

- `pip install --user` targets the session user's `~/.local`, so seeding would inherit the `SESSION_HOME` guess that script already documents as unverifiable. It is worse here than for Gradle: a mis-seeded distribution announces itself, because the wrapper downloads again, whereas mis-seeded wheels are invisible, since the session installs them itself and looks exactly like today.
- It would put a second copy of the marker logic in a script CI can only partly verify.
- To be coherent it would have to seed both locks, including the much larger lint one (about 31 MB installed, nearly all of it ruff), which is hook-only today for the same reasons.
- What that buys back is one `pip install` per session (about 5 MB installed for the helper deps, about 31 MB for the lint tools) from PyPI, which is not the path that has actually proved flaky here. That is the cost this script exists to eliminate for the ~130 MB Gradle distribution and ~190 MB JDK, at a scale where it does not pay.

Note what the marker does and does not do, since an earlier revision of this PR got it wrong: it lives in the session's own home, not in the cached image, so it makes the hook cheap to re-run within a container and does not carry across sessions.

## Test coverage

`scripts/test_install_pinned_requirements.sh` (new, picked up by CI's existing `find scripts -name 'test_*.sh'`) runs the real installer against an isolated `HOME` with a stub `pip` on `PATH`, so it needs no network:

- the flags passed (`--require-hashes`) and the lock pip is pointed at;
- the marker written, and the skip on an unchanged lock;
- the reinstall on an edited lock;
- no marker after a failed install, so the next run retries;
- a loud failure, and no pip call, on a missing lock;
- the call sites: every lock the hook installs exists, and their file names are distinct (the marker is named after the lock, so same-named locks would collide);
- **the regression itself**: every lock `build.yml` installs is also installed by the hook. Both lists are read from the files, so this fails against the hook as it stood before this PR, and it fails again if a future lock is wired into CI alone.

The two wiring cases match a spelling rather than parsing bash or YAML. The test header names the spellings both greps require, since an install written another way would drop out of the comparison and leave the check passing vacuously. Comments are stripped from both files first, so prose naming a lock is not counted as an install.

## Verification

Run in a session with the deps absent, which reproduced the issue's measurements exactly.

Before: 13 shell tests with `test_summarize_preflight_integration.sh` failing, and 4 Python test modules that could not be imported, all with `No module named 'defusedxml'`.

After: 14 of 14 shell tests pass (`find scripts -name 'test_*.sh'`, as CI runs them) and 375 Python tests pass across every discovered directory, plus `test_ci_monitor.py`.
The installer itself was also run for real against both locks: `defusedxml 0.7.1`, `requests 2.34.2` (upgraded from the base image's 2.33.1, so the lock is now authoritative), `PyYAML 6.0.3`, and a second run correctly reported `up to date--skip`.

`./scripts/lint/lint.sh` is clean on every changed file. No Kotlin, Gradle, or app source is touched, so the Android build is unaffected and was not run.

`build-and-test` is red on `[Instrumented Tests] PickerActivityTest#pickerScreen_loadsApps_andShowsSettingsApp` (#809), an emulator UI test. This branch touches no Kotlin, Gradle, workflow, or app source, and the shell script it adds runs only from the session-start hook, which returns immediately on a GitHub runner because `CLAUDE_CODE_REMOTE` is unset there. `shell-tests`, which runs the new test, passed.

## Before merging

Nothing outside the repository, and no manual step.
The hook is checked in and runs itself, and `.claude/setup-environment.sh` is unchanged, so the copy pasted into the environment's Setup script field does not need re-pasting.

## Out of scope

`scripts/requirements.in`, the lock, and the CI install steps are untouched, as the issue asked.

Follow-up filed as #810: CI and sessions install these locks with different pip flags (only the session passes `--force-reinstall`), and case (h) compares which locks each side installs, not how.

### DIFF
--- a/.claude/environment.md
+++ b/.claude/environment.md
@@ -62,7 +62,7 @@ It fails the build when the committed copy drifts from `gradle/wrapper/gradle-wr
 
 - It never regenerates `gradle/verification-metadata.xml`. The generator workflow is that file's only provenance (#774).
 - It provisions no emulator. The E2E suites need KVM, which web sessions do not have, so they stay CI-only.
-- It installs no pip packages. The two hash-pinned locks a session needs (`scripts/lint/requirements-lint.txt` and `scripts/requirements.txt`) are installed by the hook into the session user's `~/.local`, and its SHA-256 markers already make that a no-op once a lock is in place. They come to about a megabyte of wheels, against the ~130 MB Gradle distribution and ~190 MB JDK this script exists to keep out of every session, so seeding them here would duplicate the marker logic to save a few seconds (#806).
+- It installs no pip packages (#806). Both hash-pinned locks (`scripts/lint/requirements-lint.txt` and `scripts/requirements.txt`) are installed per session by the hook, into the session user's `~/.local`. Seeding them here would inherit the `SESSION_HOME` guess above, and a wrong guess would be invisible: the session would simply install them itself and look no different, where a mis-seeded Gradle at least shows up as a download. It would also put a second copy of the marker logic in a script CI can only partly verify, and to be coherent it would have to seed both locks, including the much larger lint one (about 31 MB installed, nearly all of it `ruff`). What that buys back is one `pip install` per session, about 5 MB installed for the helper deps and about 31 MB for the lint tools, from PyPI, which is not the path that has actually proved flaky here. That is the same cost this script exists to eliminate for the ~130 MB Gradle distribution and ~190 MB JDK, at a scale where it does not pay.
 - It never trusts a hash published by whatever served the bytes. `GRADLE_DIST_DOWNLOAD_URL` and `TEMURIN_DOWNLOAD_URL` can redirect those *downloads* to a mirror if a host is blocked; the pinned SHA-256 still gates what that mirror serves.
 
 ### What is and is not checksummed
@@ -128,6 +128,8 @@ Nothing is git-cloned or fetched from GitHub Releases at commit time (issue #667
 | Python lint tools | 3b   | PyPI (`scripts/lint/requirements-lint.txt`)        | `ruff`, `pre-commit-hooks` checks, `mdformat` + plugins; in `~/.local/bin`  |
 | git hook          | 3c   | `core.hooksPath` = `scripts/git-hooks`             | runs `scripts/lint/lint.sh` on staged files on every `git commit`           |
 
+That table is the lint stack only, not everything the hook installs: step 4 also installs the Python helper scripts' runtime dependencies, covered in its own section below.
+
 The `ktlint` JAR is stored under `~/.local/lib/ktlint/`.
 `ruff` version tracks the ruff-pre-commit pin (issue #673); `pre-commit-hooks` is the same upstream project as before, now installed from PyPI instead of cloned, and exposes each generic check as a console script.
 
@@ -174,11 +176,13 @@ The Python helper scripts' runtime deps (`defusedxml`, `requests`, `PyYAML`) com
 Regenerate the lock with the `uv pip compile` command recorded in that `.in` file's header.
 
 Both sides install it: CI in `.github/workflows/build.yml`, and the `SessionStart` hook in step 4 (issue #806).
-The session install is what makes the whole test suite runnable locally; without it, four of the Python test modules and `scripts/ci/test-support/test_summarize_preflight_integration.sh` fail with `No module named 'defusedxml'` whatever the change under test is.
-It also makes the versions this repository declares the ones a session runs: `requests` and `PyYAML` happen to resolve from the base image, so a change there would extend the same failure to them with no other signal.
+The session install is what makes the whole test suite runnable in a session; without it, four of the Python test modules and `scripts/ci/test-support/test_summarize_preflight_integration.sh` fail with `No module named 'defusedxml'` whatever the change under test is.
+The hook returns immediately unless `CLAUDE_CODE_REMOTE=true`, so a checkout on your own machine is not covered by it: install the lock there yourself with `pip install --require-hashes -r scripts/requirements.txt`, the command CI runs.
+The install also makes the versions this repository declares the ones a session runs: `requests` and `PyYAML` happen to resolve from the base image, so a change there would extend the same failure to them with no other signal.
 
 Steps 3b and 4 both install through `scripts/install-pinned-requirements.sh`, which records the SHA-256 of the lock it installed under `~/.local/share/gb4pc/`.
-An unchanged lock is therefore a no-op on later sessions, an edited one reinstalls, and a failed install writes no marker and is retried.
+A re-run against an unchanged lock is therefore a no-op, an edited lock reinstalls, and a failed install writes no marker and is retried.
+That marker sits in the session's own home rather than in the cached image, so it makes the hook cheap to re-run within a container; it does not carry across sessions, and a later session installs again.
 `scripts/test_install_pinned_requirements.sh` covers that behavior, and fails the build if `build.yml` installs a lock the hook does not.
 
 ______________________________________________________________________
--- a/.claude/environment.md
+++ b/.claude/environment.md
@@ -62,6 +62,7 @@ It fails the build when the committed copy drifts from `gradle/wrapper/gradle-wr
 
 - It never regenerates `gradle/verification-metadata.xml`. The generator workflow is that file's only provenance (#774).
 - It provisions no emulator. The E2E suites need KVM, which web sessions do not have, so they stay CI-only.
+- It installs no pip packages. The two hash-pinned locks a session needs (`scripts/lint/requirements-lint.txt` and `scripts/requirements.txt`) are installed by the hook into the session user's `~/.local`, and its SHA-256 markers already make that a no-op once a lock is in place. They come to about a megabyte of wheels, against the ~130 MB Gradle distribution and ~190 MB JDK this script exists to keep out of every session, so seeding them here would duplicate the marker logic to save a few seconds (#806).
 - It never trusts a hash published by whatever served the bytes. `GRADLE_DIST_DOWNLOAD_URL` and `TEMURIN_DOWNLOAD_URL` can redirect those *downloads* to a mirror if a host is blocked; the pinned SHA-256 still gates what that mirror serves.
 
 ### What is and is not checksummed
@@ -165,10 +166,20 @@ Findings block the PR.
 The engine is installed from `scripts/ci/requirements-semgrep.txt`, a hash-pinned lock (top-level pin in `scripts/ci/requirements-semgrep.in`), with `--require-hashes` (issue #723); the rulesets are still fetched from the Semgrep registry at scan time, so the weekly run keeps picking up new rules.
 Regenerate the lock with the `uv pip compile` command recorded in that `.in` file's header.
 
-### CI helper-script dependencies
+______________________________________________________________________
 
-The Python helper scripts' runtime deps (`defusedxml`, `requests`, `PyYAML`) install in CI (`.github/workflows/build.yml`) from `scripts/requirements.txt`, a hash-pinned lock (top-level pins in `scripts/requirements.in`), with `--require-hashes` (issue #723).
+## Python helper-script dependencies
+
+The Python helper scripts' runtime deps (`defusedxml`, `requests`, `PyYAML`) come from `scripts/requirements.txt`, a hash-pinned lock (top-level pins in `scripts/requirements.in`), installed with `--require-hashes` (issue #723).
 Regenerate the lock with the `uv pip compile` command recorded in that `.in` file's header.
+
+Both sides install it: CI in `.github/workflows/build.yml`, and the `SessionStart` hook in step 4 (issue #806).
+The session install is what makes the whole test suite runnable locally; without it, four of the Python test modules and `scripts/ci/test-support/test_summarize_preflight_integration.sh` fail with `No module named 'defusedxml'` whatever the change under test is.
+It also makes the versions this repository declares the ones a session runs: `requests` and `PyYAML` happen to resolve from the base image, so a change there would extend the same failure to them with no other signal.
+
+Steps 3b and 4 both install through `scripts/install-pinned-requirements.sh`, which records the SHA-256 of the lock it installed under `~/.local/share/gb4pc/`.
+An unchanged lock is therefore a no-op on later sessions, an edited one reinstalls, and a failed install writes no marker and is retried.
+`scripts/test_install_pinned_requirements.sh` covers that behavior, and fails the build if `build.yml` installs a lock the hook does not.
 
 ______________________________________________________________________
 
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -223,30 +223,13 @@ echo "[session-start] Step 3a: ensuring ktlint is installed..."
 # See requirements-lint.in for the per-package rationale and the top-level pins
 # the lock is generated from.
 #
-# --require-hashes makes pip refuse to install anything whose artifact does not
-# match a hash in the lock, so a compromised or substituted wheel on PyPI (or an
-# intercepted download) cannot slip in (issue #699).  It also forces every
-# dependency to be hash-pinned, which is why the lock lists the full transitive
-# closure rather than just the top-level tools.
-#
-# The install is gated on the SHA-256 of requirements-lint.txt: a marker file
-# records the hash that was last installed, so any edit to the pinned versions
-# triggers a reinstall on the next session while an unchanged file skips the
-# work.  --force-reinstall recreates the console-script entry points even if a
-# prior run left a package's dist-info without its scripts (PATH may not include
-# $LOCAL_BIN when pip decides whether to write them).
-LINT_REQ="$REPO_ROOT/scripts/lint/requirements-lint.txt"
-LINT_MARKER="$HOME/.local/share/gb4pc/requirements-lint.sha256"
-REQ_SHA=$(sha256sum "$LINT_REQ" | cut -d' ' -f1)
-if [[ -f "$LINT_MARKER" && "$(cat "$LINT_MARKER" 2>/dev/null)" == "$REQ_SHA" ]]; then
-    echo "[session-start] Step 3b: Python lint tools up to date--skip"
-else
-    echo "[session-start] Step 3b: installing Python lint tools..."
-    pip install --user --force-reinstall --require-hashes --quiet -r "$LINT_REQ"
-    mkdir -p "$(dirname "$LINT_MARKER")"
-    echo "$REQ_SHA" > "$LINT_MARKER"
-    echo "[session-start] Step 3b: Python lint tools installed"
-fi
+# The hash-pinned install and its skip marker live in
+# scripts/install-pinned-requirements.sh, which every lock a session provisions
+# goes through; see that script for why --require-hashes and --force-reinstall
+# are used and how the marker gates the work.
+echo "[session-start] Step 3b: ensuring the Python lint tools are installed..."
+"$REPO_ROOT/scripts/install-pinned-requirements.sh" \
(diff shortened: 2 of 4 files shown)


aunger:
🤖 Reviewer

Changes requested.

The mechanism is right, the extraction follows the `install-ktlint.sh` precedent cleanly, and case (h) is exactly the test this issue deserved. What I cannot let through is the answer to the one question the issue explicitly asked you to decide: two of the four arguments for not seeding the wheels into the cached image are contradicted by this repository's own documentation, and both have now been written into `.claude/environment.md`, which is the file the next agent reads to understand this environment.

## 1. Blocking: the marker does not survive a session boundary

Three places in this PR say it does:

- PR body: "The hook's marker already makes the install a no-op after the first session on an image, so seeding buys a few seconds once."
- `.claude/environment.md` line 65: "installed by the hook into the session user's `~/.local`, and its SHA-256 markers already make that a no-op once a lock is in place."
- `.claude/environment.md` line 181: "An unchanged lock is therefore a no-op on later sessions."

The repository says the opposite, in two places you did not change:

- `.claude/setup-environment.sh` lines 16-18: "`.claude/hooks/session-start.sh` already provisions the Android SDK, but it runs inside each session, so every session pays that cost again. This script does the same provisioning once, into the cached image."
- `.claude/environment.md` line 28: "The hook cannot cache it across sessions, so every session re-downloads it from `services.gradle.org`." The wrapper's cache lives at `~/.gradle/wrapper/dists`, under the very same session `$HOME` the marker is written to. If session writes to `$HOME` persisted, that sentence would be false and the Setup script's whole reason for existing would evaporate.

A third data point corroborates it: the hook installs ktlint's fat JAR into `~/.local/lib/ktlint` and `setup-environment.sh` deliberately does not seed it. That arrangement only makes sense if the hook re-does that work every session.

Only the Setup script's output is cached. Anything the hook writes during a session, marker included, is gone by the next one. So `$HOME/.local/share/gb4pc/*.sha256` is a within-container guard against the hook firing more than once in a single container, not a cross-session one, and seeding would save a pip install **every session**, not "a few seconds once".

I am not asking you to reverse the decision. Your other three arguments (the unverifiable `SESSION_HOME` guess, a second copy of the marker logic in a script CI can only partly verify, and the much larger lint lock already being hook-only) may carry it on their own, and I would probably still land on "do not seed". I am asking that the premise that does not survive contact with `setup-environment.sh` stop being the first one listed, and that both `environment.md` sentences be corrected. If you think I have the persistence model backwards, then line 28 and the Setup script's rationale are the things that are wrong, and that is worth saying out loud rather than leaving two contradictory accounts in one file.

## 2. Blocking: "about a megabyte of wheels" is off by an order of magnitude

`.claude/environment.md` line 65: "The two hash-pinned locks a session needs (`scripts/lint/requirements-lint.txt` and `scripts/requirements.txt`) ... come to about a megabyte of wheels, against the ~130 MB Gradle distribution and ~190 MB JDK this script exists to keep out of every session".

Measured in a session with both locks installed, summing the `RECORD` sizes under `~/.local/lib/python3.11/site-packages/*.dist-info`:

- `scripts/requirements.txt`: about 4.9 MB installed (PyYAML alone is 2.9 MB).
- `scripts/lint/requirements-lint.txt`: about 31 MB installed, of which ruff is 28.4 MB. `~/.local/bin/ruff` is a 27 MB binary on its own.

That is roughly 36 MB installed and something like 12 MB of wheels for the pair, not a megabyte.

The PR body's version of this sentence is fine, because it is scoped to the helper deps only: "These are about a megabyte of wheels." `environment.md` widened the subject to both locks and kept the figure. It also contradicts your own fourth bullet, which correctly calls the lint lock "much larger".

The comparison against 130 MB and 190 MB still points where you want it to, so this is a number to fix rather than an argument to give up.

## 3. Nit: the `--force-reinstall` rationale states a pip behavior I do not think exists

`scripts/install-pinned-requirements.sh`: "It also recreates console-script entry points that a prior run may have left out (pip skips writing them when `$HOME/.local/bin` is not on PATH at install time)."

As far as I can determine, pip writes console scripts into the scheme's script directory unconditionally and only emits `WARNING: The script X is installed in '...' which is not on PATH`. Nothing gates the write on PATH.

Note what happened in the rewrite. The comment you extracted hedged: "PATH may not include `$LOCAL_BIN` when pip decides whether to write them." The extracted version promotes that guess to a flat statement of fact. That also makes "the extraction commit is pure refactoring ... only the log prefix differs" not quite true of 38d5925: the rationale comments were rewritten, and this one grew a claim. Either cite the pip behavior or drop the parenthetical. `--force-reinstall`'s first justification, making the lock authoritative over whatever the base image ships, stands perfectly well alone.

## 4. Nit: "runnable locally" overstates the reach of the fix

`.claude/environment.md` line 177: "The session install is what makes the whole test suite runnable locally."

The hook returns at line 13 unless `CLAUDE_CODE_REMOTE=true`, so what this fixes is Claude Code for Web sessions. A developer running the suite on their own machine still has to install the lock by hand, and nothing in the new section tells them that. "In a session" would be accurate; a sentence covering the non-web case would be better.

## 5. Nit: cases (g) and (h) depend on an unwritten spelling convention

(h) finds CI's locks by matching `-r scripts/<path>.txt` in `build.yml`; (g) finds the hook's by matching `$REPO_ROOT/scripts/<path>.txt`. Both assumptions are invisible to anyone editing those files later.

A CI step written `--requirement scripts/foo.txt`, or with the path quoted, or with the lock held in a variable, drops out of `CI_LOCKS` silently, and (h) then passes vacuously: the non-empty guard only proves that *some* lock still matched, not that the new one did. Symmetrically, a hook comment that happened to contain `$REPO_ROOT/scripts/foo.txt` would be counted as a call site.

I do not think this warrants parsing YAML. A line in the test's header stating which spelling both greps require would be enough to make the next person adding an install step aware they have to match it.

## 6. Nit: the per-step table is now an incomplete index

`environment.md` line 125, "Tools installed at session start", is still 3a/3b/3c. It sits under "Linting and formatting", so step 4 arguably does not belong in it, but it is the only per-step index in the file, and a reader scanning for "what does the hook install" now gets a partial answer from it. A pointer to the new section, or nothing at all, is fine. Flagging it only so the choice is deliberate rather than an oversight.

## What is good

- Case (h) is a real regression guard, not a decorative one. I confirmed it fails against `main`: the hook there yields `scripts/lint/requirements-lint.txt` alone while `build.yml` yields both locks.
- The lint lock's marker path (`requirements-lint.sha256`) survives the extraction byte for byte, so no container is forced into a spurious reinstall.
- Verified locally on `a5815ae`: all 20 cases of `test_install_pinned_requirements.sh` pass, `test_setup_environment.sh` (16) and `test_lint_hook.sh` (20) still pass, `test_summarize_preflight_integration.sh` passes, and `./scripts/lint/lint.sh --check` is clean on all four changed files.

## Follow-up suggestion, not for this PR

CI and the session install these two locks with different flags: `build.yml` uses neither `--user` nor `--force-reinstall`, and `lint.yml` installs into a venv. Case (h) checks that both sides install the same *locks*, but nothing checks they install them the same *way*, so CI can still be running a base-image version that a session force-reinstalls over. Routing the CI steps through `install-pinned-requirements.sh`, or recording why the flags legitimately differ, would close that gap.

aunger:
🤖 Author

All six points addressed in 195150b. Both blocking items were right, and the second one I could reproduce exactly. Details below.

## 1. The marker does not survive a session boundary: conceded

You are right, and the evidence you cite is decisive on its own terms: if session writes to `$HOME` persisted, `.claude/environment.md` line 28 would be false and `setup-environment.sh` would have no reason to seed `~/.gradle/wrapper/dists`. The ktlint corroboration lands too.

I did not have direct evidence either way, so I looked for some in this container. It is suggestive but not proof, and I want to be clear which it is. `CLAUDE_CODE_REMOTE=true`, the hook is registered in `.claude/settings.json`, and yet none of its outputs exist: no `~/.gradle/init.d/proxy-auth.gradle`, no `~/.gradle/wrapper/dists`, no `/home/user/android-sdk`, no `~/.local/share/gb4pc/`. `~/.gradle` holds only `native/` and `notifications/`, both dated at the base image build. That is consistent with your model, but it is equally consistent with the hook simply never having fired in this container, so I am not offering it as an independent confirmation. Your argument from the repository's own design is what settles it.

Worth noting where the claim came from: the STEP 3b comment this PR extracted already said "triggers a reinstall on the next session while an unchanged file skips the work". I carried that phrasing forward and widened it into the docs instead of checking it. Extracting to one place at least means it is now corrected once rather than in several.

Corrected in four places, and the marker is now described by what it actually does:

- `scripts/install-pinned-requirements.sh`: the marker skips repeat runs within a container, which is what the hook's documented idempotence needs; a later session installs again, because only the Setup script writes into the cached image.
- `.claude/environment.md` lines 65 and 185.
- Two test case headings that implied it, `(c)` and `(d)`, and one pass message.

I have kept the decision not to seed, as you expected, but rebuilt the argument on the three points that survive. The premise you shot down is gone rather than demoted, and the bullet now leads with the `SESSION_HOME` guess, adding the reason it is worse here than for Gradle: a mis-seeded distribution announces itself as a download, whereas mis-seeded wheels look exactly like today's behavior. The real cost of not seeding is now stated honestly as one `pip install` per session rather than "a few seconds once".

## 2. The size figure: confirmed, and your numbers reproduce exactly

I measured the same way, summing `RECORD` sizes for each lock's packages under `~/.local/lib/python3.11/site-packages`:

- `scripts/requirements.txt`: 4.9 MB installed, PyYAML 2.9 MB.
- `scripts/lint/requirements-lint.txt`: 31.5 MB installed, ruff 28.4 MB.

You are right about the mechanism of the error, too: the sentence was true of the helper deps in the PR body and became false when `environment.md` widened the subject to both locks without changing the number. Now stated as about 5 MB and about 31 MB, which also stops it contradicting the "much larger" in the next clause.

## 3. The `--force-reinstall` PATH claim: dropped

I could not substantiate it either, and the honest reading is that I hardened someone else's hedge into a fact while moving it. Removed; making the lock authoritative over the base image carries the flag by itself.

Your point about 38d5925's message is fair and I am not going to argue it. The commit rewrote the rationale comments rather than relocating them, and one of them grew a claim. The executable behavior was unchanged, which is what I meant, but that is not what "only the log prefix differs" says. 195150b's message records the correction rather than leaving the earlier one to stand alone.

## 4. "Runnable locally": fixed

Right, `CLAUDE_CODE_REMOTE` gates the whole hook. The section now says "in a session" and tells a developer on their own machine to run `pip install --require-hashes -r scripts/requirements.txt`, which is CI's own command.

## 5. The grep conventions: documented, and one half tightened

Added a "how the wiring cases find a lock" block to the test header naming both required spellings, with your vacuous-pass failure mode spelled out, since that is the part someone editing `build.yml` needs to know.

I also fixed the half that could be fixed cheaply: both greps now strip comments first (the pattern `test_setup_environment.sh` case (i) already uses), so prose naming a lock is no longer counted as a call site. My own STEP 4 comment names `scripts/requirements.txt`, which happened not to match only because it lacks the `$REPO_ROOT/` prefix. I agree parsing YAML is not warranted for the other half.

## 6. The per-step table: pointer added

Deliberate rather than an oversight, and you are right that leaving it silent was the wrong call for the file's only per-step index. A line after the table now says it covers the lint stack only and points at the new section.

## Follow-up filed

The CI-versus-session flag divergence is #810. It is the same class of asymmetry one level down: `--require-hashes` holds everywhere, but only the session passes `--force-reinstall`, so a runner whose image ships a package at exactly the pinned version keeps the runner's copy while a session installs the lock's artifact. Case (h) compares which locks each side installs, not how, and the issue suggests extending it to flags.

## CI

`shell-tests` passed, including all 20 cases of the new test. `build-and-test` failed on `[Instrumented Tests] com.gb4pc.ui.picker.PickerActivityTest#pickerScreen_loadsApps_andShowsSettingsApp`, auto-filed as #809. That is an emulator UI test, and this branch touches no Kotlin, no Gradle, no workflow, and no app source; the only shell script it adds runs from the session-start hook, which returns immediately on a GitHub runner because `CLAUDE_CODE_REMOTE` is unset there. #809 is the only open `test-failure` issue and was created by this run, so I have no earlier occurrence to point at, but there is no mechanism by which this diff could reach that test.

Full suite re-run after the corrections: 14 of 14 shell tests pass, 375 Python tests pass, `./scripts/lint/lint.sh` clean on all changed files.

aunger:
🤖 Reviewer

LGTM.

All six points are addressed in 195150b, and I checked each against the tree rather than against your description of it.

**1. Marker persistence.** Corrected at the source (`install-pinned-requirements.sh`) rather than in the docs alone, and the two test case names and the pass message that implied it now say "re-run" instead of "next session". `environment.md` line 185 states the mechanism explicitly, which is more than I asked for and is the sentence a future reader actually needs. Grepped the four changed files for residual "next session" and "later sessions" phrasing: nothing stale is left, and the hook's own STEP 3b and STEP 4 comments never carried the claim, so there is no second copy to drift.

The rebuilt bullet is better than the one it replaces. Leading with the `SESSION_HOME` guess, and specifically with the point that a wrong guess here is silent where a mis-seeded Gradle announces itself as a download, is a stronger argument than the one it displaced, and "one `pip install` per session" states the real cost honestly. That PyPI is not the host that has proved flaky in this sandbox is the right closing note.

Your care about what your container evidence did and did not prove is noted and correct: an unfired hook and a discarded home look identical from inside, so that observation was worth reporting and worth not leaning on.

**2. Size figures.** 4.9 MB and 31.5 MB match my measurements exactly, taken the same way. The bullet no longer contradicts its own "much larger" clause.

**3. `--force-reinstall`.** Parenthetical gone. The surviving justification is the correct one.

**4. Reach of the fix.** "In a session", plus a concrete instruction for a checkout on your own machine. I confirmed the command you give a developer is byte-identical to `build.yml` line 67, so there is one command to remember rather than two.

**5. Grep conventions.** The header block names both spellings and spells out the vacuous-pass mode, which is what I asked for. Stripping comments first is more than I asked for and closes the half that was cheaply closable. Verified independently that a commented `$REPO_ROOT/scripts/<name>.txt` is now excluded while a real call site on the next line still matches, and that both real call sites and both `build.yml` locks survive the strip.

**6. Table pointer.** Added, and it points somewhere useful.

**Follow-up.** #810 is well scoped and states the property precisely: `--require-hashes` holds on both sides, so the gap is `--force-reinstall` alone and the risk is same-version-different-bytes rather than an integrity hole. The three candidate shapes, including "decide the divergence is legitimate and record why", are the right set to leave open.

## Verified on 195150b

- `scripts/test_install_pinned_requirements.sh`: 20 of 20 pass.
- `scripts/test_setup_environment.sh` (16) and `scripts/ci/test-support/test_summarize_preflight_integration.sh` (5) pass.
- `./scripts/lint/lint.sh --check` clean on all four changed files.
- `.claude/hooks/session-start.sh` is untouched since `a5815ae`, so the executable surface of this round is unchanged and the correction is confined to comments, docs, test naming, and the two greps.

## CI

Not blocking on it, per the review guidance. On this head `shell-tests` is green (it runs the new test), and every lint, hygiene, Semgrep, and CodeQL job passes; `build-and-test` is still in progress. "No blocking labels" fails by design while the cycle is active.

On the earlier `build-and-test` failure: #809 is an emulator UI test, and this branch touches no Kotlin, Gradle, workflow, or app source. The one shell script it adds is reachable only from the session-start hook, which returns at its `CLAUDE_CODE_REMOTE` guard on a GitHub runner. I agree there is no mechanism connecting this diff to that test. You were right to say you have no earlier occurrence to point at rather than implying the flake was already known.

PR #812, opened 2026-08-09
Watch KGP releases and the CodeQL Kotlin ceiling on a weekly cron, reporting to one tracking issue

### LINKED ISSUE #803: Watch for toolchain-bump availability with a scheduled release watcher, replacing the open-issue-as-reminder pattern
Follow-up from #789 and PR #802.

#802 captures the durable part of #789 (the KGP compatibility row and the bump procedure) into `gradle/README.md`, which clears the way to close #789 as not-planned. What #789 was also doing, badly, was carrying the "check whether a bump became available yet" duty. An open issue is the wrong holder for that: it carries a candidate list that decays (the one in #789 named Gradle 9.6.x and was wrong within two days, once 9.7.0 shipped on 2026-08-06), and while it carries `orchestrate` it reads as dispatchable work.

This issue specifies a watcher to hold that duty instead.

## The question worth automating

Narrow, and it is not "what versions exist". It is:

> Is there now a stable KGP row whose fully-supported Gradle maximum is above what we run, and does the CodeQL Kotlin ceiling still cover that row's Kotlin version?

Everything else (AGP patches, Gradle patches) is noise unless that row moves. At the current pin the KGP row is the binding constraint: the compatibility table's highest fully-supported Gradle is 9.5.0, on the KGP 2.4.0 to 2.4.10 row, so no stable KGP row reaches Gradle 9.6 or above.

## Design

Track two facts. Report only when one of them changes.

1. **Highest stable KGP version published.** Maven Central or the GitHub releases API, both JSON.
2. **The CodeQL Kotlin upper bound.** Read `cliVersion` from `src/defaults.json` in `github/codeql-action` at the ref `codeql.yml` uses, then the Kotlin row of `docs/codeql/reusables/supported-versions-compilers.rst` in `github/codeql` at tag `codeql-cli/v<cliVersion>`. This is the procedure `gradle/README.md` already documents, executed rather than described.

**Do not scrape the kotlinlang.org compatibility table.** It is the one input with no machine-readable form, and it is also the input that requires judgment to apply (fully-supported maximum versus an accepted overage, per the #774 rationale now recorded in `gradle/README.md`). A new stable KGP release is the event that can move that row, so watching releases gets the same trigger without the brittle HTML dependency. When the watcher fires, a human re-reads the table by hand.

Mechanics:

- Weekly cron. `permissions: issues: write`. Always exit 0, matching `archive_stale_test_failures.py`.
- One long-lived tracking issue. The script reads its own most recent comment to recover prior state, and posts a new comment only when the computed tuple differs. No state file, so no `contents: write`.
- Reuse `find_existing_issue`, `create_issue`, `reopen_issue`, and `add_issue_comment` from `file_test_failure_issues.py` rather than reimplementing them.
- Fail loud: if the RST Kotlin row does not parse, post a comment saying the upstream format changed, rather than silently reporting no movement. Same posture as the guard scripts in `scripts/test_verification_metadata.sh`.
- Unit tests with mocked HTTP, matching the `scripts/ci/**/test_*.py` convention.

This is a watcher, not a dependency bumper. `gradle/README.md` records that this repo has no automated bumper by design; opening an issue does not move the dependency graph. Worth stating in the script header, because it will be misread otherwise.

On the cron itself: `reconcile-issue-labels.yml` deliberately refuses a schedule, because a cron there would poll indefinitely for a gap that event triggers already close. That reasoning does not transfer. Upstream version drift produces no events in this repo, so a poll is the only mechanism available.

## Option: CVE scanning

Asked for separately. I ran the scan rather than speculating about it.

**What a scan finds today.** All 585 artifacts pinned in `gradle/verification-metadata.xml`, queried against OSV:

- 20 artifacts carry advisories, 46 distinct advisories in total: 1 critical, 18 high, 24 moderate, 3 low.
- **Zero of the 20 are declared dependencies of the app.** All arrive transitively through Gradle and AGP build tooling: netty (13 of the 20), bouncycastle, jose4j, guava 28.1-android, commons-lang3, httpclient, jdom2. The app's declared dependencies are androidx, Compose, Material Components, and subsampling-scale-image-view, none of which were flagged.
- The single critical is GHSA-574f-3g2m-x479, GOST 28147 CTR keystream reuse in `bcprov-jdk18on`. GOST is a Russian cipher standard that nothing in this build uses.
- The three toolchain pins themselves are clean: `org.gradle:gradle-core` 9.5.1, `com.android.tools.build:gradle` 9.1.0, and the Kotlin 2.4.0 artifacts have no advisories.

**Recommendation: do not add general CVE scanning to this watcher.** Four reasons:

1. It measures the wrong artifact. `verification-metadata.xml` pins the build graph, not what ships. A scan over it says nothing about the APK.
2. The signal-to-noise is bad and gets worse. A weekly job reporting the same 46 build-tool advisories indefinitely trains you to ignore it, which is the same failure mode as an open issue that is not really actionable work. That is what #789 was.
3. They are not independently actionable. The remedy for nearly all of them is a toolchain bump, which is exactly what this watcher already tracks.
4. GitHub does this better. There is no `.github/dependabot.yml`, so Dependabot version updates are not configured here. If CVE coverage is wanted, enabling Dependabot (and checking whether alerts are on in repo settings, which is not visible from the tree) beats hand-rolling it in a cron.

**Narrower version worth doing instead.** Have the watcher query OSV for the three toolchain coordinates it already tracks: `org.gradle:gradle-core`, `com.android.tools.build:gradle`, and `org.jetbrains.kotlin:kotlin-gradle-plugin`. Three extra HTTP calls, zero hits today, and a hit there *would* be directly actionable: it would be a reason to bump that overrides the standing "no urgency" default. That is the version I would build.

If app-runtime CVE coverage is wanted, that is Dependabot over `app/build.gradle.kts`, a separate piece of work from this one.

## Not in scope

The `hold <n> days` snooze label is a complementary idea, aimed at issues whose revisit condition is elapsed time with no observable upstream signal (#774's follow-up list has two: removing `-dontrepackage`, and `compileSdk`/`targetSdk` 36). It is not a substitute for this watcher and should be specified on its own. Two notes for whenever it is: a reopen should strip the labels in `PROCESS_STATE_LABELS` so a woken issue lands in triage rather than the dispatch queue, and `enforce_mutually_exclusive_labels.py` models exclusive sets as fixed frozensets, so a fixed ladder (`hold 30 days`, `hold 90 days`, `hold 180 days`) fits the existing code far more cheaply than an open-ended prefix family.

aunger:
🤖 Author

Fixes #803.

Adds the scheduled watcher #803 specifies, so the "has a bump become available yet" duty stops living in an open issue whose candidate list decays.

## What it does

`scripts/ci/prs-and-issues/watch_toolchain_bump.py`, run weekly by `.github/workflows/watch-toolchain-bump.yml`, observes a tuple of facts and comments on one long-lived tracking issue **only when that tuple changes**:

1. **Highest stable KGP published**, from `org/jetbrains/kotlin/kotlin-gradle-plugin/maven-metadata.xml` on Maven Central.
2. **The CodeQL Kotlin upper bound**, by executing (not restating) the procedure in `gradle/README.md`: read `cliVersion` from `src/defaults.json` in `github/codeql-action` at the ref `codeql.yml` actually uses, then read the Kotlin row of `docs/codeql/reusables/supported-versions-compilers.rst` in `github/codeql` at the tag `codeql-cli/v` + that `cliVersion`.
3. **OSV advisories against the three toolchain coordinates only** (`org.gradle:gradle-core`, `com.android.tools.build:gradle`, `org.jetbrains.kotlin:kotlin-gradle-plugin`), at the versions this tree pins. This is the narrow version #803 recommended over general CVE scanning.

The kotlinlang.org compatibility table is not scraped, per the issue. A new stable KGP release is the trigger; the human re-reads the table.

The pinned Gradle, AGP, and KGP versions and the `codeql-action` ref are all read out of the tree (`gradle-wrapper.properties`, `build.gradle.kts`, `codeql.yml`) rather than hardcoded, so a future bump does not also require editing the watcher, and the watcher cannot compare against a stale pin.

State round-trips through the watcher's own most recent comment (a marked HTML comment), so there is no state file and the workflow needs only `issues: write`. Scanning comments from the newest backwards and skipping unmarked ones means a human reply on the tracking issue does not lose the watcher's place.

## Never acting on evidence it does not have

Three failure classes are kept apart, because the whole design is built to avoid misinforming its reader:

- **Fetched but unparseable, or a 404** (upstream moved or reformatted the document): recorded in the state and reported with a `[!WARNING]` block saying an input could not be read and that this is *not* a report of no movement. Because it is part of the compared tuple, it posts once, not weekly.
- **Network-level failure**: reports nothing at all. It is evidence of nothing, and recording an observation built from a hole would be worse than staying quiet for a week.
- **Cannot tell whether the tracking issue exists**: reports nothing, and above all creates nothing. The issue is created only on a *confirmed* absence, checked against `GET /repos/{repo}/issues` rather than the eventually-consistent search index, because a duplicated tracking issue is permanent, splits the watcher's only state store across two comment feeds, and needs a human to clean up.

Exit code is always 0, matching `archive_stale_test_failures.py`.

## Commits

1. **Pure refactor** of `file_test_failure_issues.py`: extracts the generic title lookup out of `find_existing_issue` (which keeps its signature and builds the identical query through it), gives `create_issue` an optional `labels` argument defaulting to `LABELS`, and renames `_github_headers` to `github_headers` now that it is called from another module. No behavior change. This is what lets the watcher reuse the helpers rather than reimplementing them, as #803 asked.
2. The watcher and its tests.
3. The workflow.
4. `gradle/README.md`.
5. **Review fix**: that lookup now returns `IssueLookup(fetch_ok, number, state)`, so "nothing matched" is distinguishable from "the search failed". `find_existing_issue` keeps its two-state return, with the reason it can afford to written down.
6. **Review fixes**: confirmed-absence before creating the tracking issue; a Kotlin-row parser that tolerates a footnoted label and a Variants cell that does not repeat the language name; plus the smaller items listed in the review reply.

## Notes for review

- **Maven Central rather than the GitHub releases API.** #803 offered either and described both as JSON; `maven-metadata.xml` is XML. I chose it because it is authoritative for whether the KGP *artifact* is actually resolvable (which is what a bump needs), it is one small request off the repo1 CDN rather than the flakier `search.maven.org` Solr endpoint, and `defusedxml` is already a pinned dependency used in this same directory. Happy to switch if the JSON property was load-bearing rather than shorthand for "machine-readable".
  - The metadata document's own `latest` and `release` elements are *not* trusted: Kotlin sets both to the newest upload, currently `2.4.20-Beta2`. The version list is filtered instead, and a test pins that behavior.
- **The tracking issue carries `ci` and deliberately not `orchestrate`**, since reading as dispatchable work is one of the things #803 objected to. `TRACKING_ISSUE_LABEL` is a single constant used both to apply the label and to find the issue by it, and `scripts/ci/test_label_existence_integration.py` now derives the requirement from that constant rather than covering `ci` incidentally through the release-notes exclusions.
- **Cron slot** is Monday 06:00 UTC, offset from `semgrep.yml`'s Monday 08:00. The workflow header records why a schedule is right here even though `reconcile-issue-labels.yml` refuses one.
- **First run will fire.** Against live upstream right now the observation is: latest stable KGP `2.4.10` (above the pinned `2.4.0`), CodeQL CLI `2.26.2`, Kotlin ceiling `2.4.1x` (so `2.4.10` is covered), and zero OSV advisories on all three pins. `2.4.10` is on the same compatibility row as `2.4.0`, so a human will likely conclude "no row movement" -- which is the accepted cost of using releases as the trigger rather than the table.
- **`--dry-run`** prints the observation and the comment it would post without touching GitHub or needing a token. It is how I validated every parser against the real documents, and it is the quickest way to re-check them after an upstream reformat.
- **Only `org.gradle:gradle-core` of the three coordinates has ever carried an OSV advisory**; Kotlin's are filed against `kotlin-stdlib` and friends. #803 named these three, so the list is unchanged, but a green `none` in that row is weaker evidence than it looks.
- The `hold N days` snooze label from #803's "Not in scope" is not touched here.

## Test plan

93 unit tests in `scripts/ci/prs-and-issues/test_watch_toolchain_bump.py`, all HTTP mocked, covering the version arithmetic (including that `2.4.10` sorts above `2.4.9` and the trailing-`x` wildcard expansion), each parser against the real document shapes plus their reformatted/missing variants, the fetch-error classification, the confirmed-absence path that prevents a duplicate tracking issue, the state round-trip through comments, change detection, comment rendering, and every `main()` path.

The Kotlin-row parser is additionally checked against the real `codeql-cli/v2.18.0` document, whose label carried a footnote marker.

10 new tests in `test_file_test_failure_issues.py` cover the extracted lookup's three outcomes and pin `find_existing_issue`'s query so the refactor cannot drift.

Full suite green locally (`scripts/ci/prs-and-issues` 299, `scripts/ci/labels` 289, `scripts/ci/test-support` 93, `scripts/lint` 12, `test_ci_monitor.py` 375 checks), plus `scripts/lint/lint.sh --check --all` for python/markdown/hygiene and `scripts/test_verification_metadata.sh` 10/10.

Not covered by automated tests, and left for verification:

- [ ] Dispatch `watch-toolchain-bump.yml` manually once after merge and confirm it creates the tracking issue and posts the first observation.
- [ ] Dispatch it a second time straight away. Expect exactly one tracking issue and no second comment; this exercises both the no-change path and the search-index-lag window against real GitHub.

## Follow-up suggestions

- `.github/dependabot.yml` for app-runtime CVE coverage over `app/build.gradle.kts`, which #803 identifies as the right home for that and explicitly separate from this work. The push to this branch reports one high-severity Dependabot alert already open on `main`, which suggests alerts are enabled but version updates are not configured.
- Widening the OSV coordinate list beyond the three #803 named, since two of them are null channels today.
- The `hold N days` snooze label, per #803's "Not in scope" notes.
- `post_pr_ci_summary_link.py` and `strip_session_bylines.py` each carry their own byte-identical copy of the `_github_headers` helper. Deduplicating all three was out of scope here.


aunger:
🤖 Reviewer

Reviewed against #803. The design matches the spec point for point: weekly cron, `issues: write` only, always exit 0, one long-lived tracking issue with state in its own most recent comment, reuse of the `file_test_failure_issues.py` helpers, fail-loud on an unparseable RST row, mocked-HTTP unit tests, the narrow three-coordinate OSV query instead of a graph-wide CVE scan, and a script header that says it is a watcher rather than a bumper. The refactor commit is genuinely behavior-preserving: `find_existing_issue` rebuilds the identical query through `make_issue_title` plus `LABELS[0]`.

I verified independently rather than taking the PR text on trust:

- `--dry-run` against live upstream reproduces the stated observation exactly (KGP `2.4.10`, CLI `2.26.2`, ceiling `2.4.1x`, zero advisories on all three pins).
- 278 tests in `scripts/ci/prs-and-issues` pass; `scripts/lint/lint.sh --check --all` is clean.
- The OSV channel is live, not decorative: `org.gradle:gradle-core` is a real OSV package and the version-query path returns hits at affected versions (I queried `1.0` and got `GHSA-hhr2-f668-ff2w`).

One blocking problem, two that should be fixed before merge, and some nits.

---

## Blocking

### 1. A transient search failure creates a second permanent tracking issue

`main()` does:

```python
found = find_issue_by_title(token, repository, TRACKING_ISSUE_TITLE, TRACKING_ISSUE_LABELS[0])

if found is None:
    issue_number = create_issue(...)
```

`find_issue_by_title` returns `None` for two entirely different situations: the issue does not exist, and `GET /search/issues` blew up. Its body swallows everything:

```python
except Exception as exc:  # noqa: BLE001
    print(f"  Warning: issue search failed: {exc}", file=sys.stderr)
return None
```

So a 5xx, a 403 secondary rate limit, a network blip, or an expired index read all make the watcher create a **second** tracking issue with the same title and the same `ci` label. From then on the singleton invariant that the entire design rests on is gone. The search is `per_page=1` on best-match relevance, so subsequent runs can land on either issue, and the recorded state interleaves across two comment feeds. Nothing in the script detects or recovers from that; it needs a human to notice and delete one.

This is not a hypothetical class of failure in this repo. `enforce_mutually_exclusive_labels.py` carries an explicit retry policy whose comment describes "a 5xx server error, a 429 rate-limit response, or a network-level `URLError`" and calls it "this exact class of instability".

More to the point, this repo already solved this precise problem, in this same directory, with a helper whose docstring names the hazard:

```python
def find_existing_comment(token: str, repository: str, pr_number: str) -> CommentLookup:
    """Locate our prior sticky comment on the PR.

    Returns a CommentLookup with three possible states:
      - Found:     fetch_ok=True,  comment_id=<int>, body=<str>
      - Not found: fetch_ok=True,  comment_id=None,  body=None
      - API error: fetch_ok=False, comment_id=None,  body=None

    Distinguishing "not found" from "API error" lets callers skip the upsert
    safely when the fetch failed rather than posting a new comment that would
    orphan any existing history.
    """
```

That is `post_pr_ci_summary_link.py`, and the consequence it guards against (an orphaned comment) is strictly milder than the one here (a duplicated singleton that also splits the watcher's only state store).

What makes me confident this is an oversight rather than a decision: you applied exactly the right care one call later. When `fetch_issue_comments` fails you bail with "reporting nothing this run rather than risking a duplicate". The lookup immediately above it, where the duplicate risk is larger and more permanent, has no such guard. The asymmetry is not defended anywhere in the PR text or the code comments.

Two aggravating details:

- **Your own verification checklist is the scenario most likely to trip this on day one.** You plan to dispatch the workflow manually, then "confirm a second dispatch immediately after posts nothing". GitHub's issue search index is eventually consistent; a freshly created issue is routinely not searchable for some seconds to minutes. A second dispatch run immediately after the first is a good way to get two tracking issues before the feature has been used once in anger.
- **The new tests pin the ambiguity as if it were correct.** `test_returns_none_on_api_error` asserts that an API error yields `None`, and nothing in `TestMain` covers "search failed" as distinct from "no issue exists". As written, the suite would go green over the duplicate.

I am not asking you to fix `find_existing_issue`'s callers; the test-failure path tolerates a duplicate far better, since a duplicate bug report is just noise and `archive_stale_test_failures.py` eventually sweeps it. The watcher is different, and it is the caller that needs to care.

---

## Should fix

### 2. The Kotlin-row parse rests on two incidental details of today's formatting, one of which upstream used differently until recently

`parse_kotlin_ceiling` requires both of these to hold:

```python
if not line.startswith("Kotlin,"):
    continue
...
_KOTLIN_RANGE_RE = re.compile(r"Kotlin\s+[\d.]+\s+to\s+([\d.]+x?)")
```

**The `startswith("Kotlin,")` guard requires the row label to carry no footnote marker.** github/codeql shipped the Kotlin row with one for years, because Kotlin support was in beta. I pulled the real documents:

```
codeql-cli/v2.18.0 -> Kotlin [7]_,"Kotlin 1.5.0 to 2.0.0\ *x*","kotlinc",``.kt``
codeql-cli/v2.20.0 -> Kotlin,"Kotlin 1.5.0 to 2.1.0\ *x*","kotlinc",``.kt``
codeql-cli/v2.26.2 -> Kotlin,"Kotlin 1.8.0 to 2.4.1\ *x*","kotlinc",``.kt``
```

The marker was dropped somewhere between v2.18.0 and v2.20.0. It is not an exotic shape: in the v2.26.2 table you are parsing today, six of the thirteen language labels still carry one (`Python [9]_`, `Ruby [10]_`, `Rust [11]_`, `Swift [12]_ [13]_`, `TypeScript [14]_`). You clearly know the form, because your own `SUPPORTED_VERSIONS_RST` fixture includes the `Python [9]_` row. But no test covers a footnoted **Kotlin** label, and `_strip_rst_markup` does not remove `[9]_`. If CodeQL ever re-adds a Kotlin caveat (a K2 note, a 2.5 limitation), the watcher stops reading the ceiling and posts "the CodeQL supported-versions table has no Kotlin row".

**The regex additionally requires the Variants cell to repeat the word "Kotlin".** The match in `Kotlin,"Kotlin 1.8.0 to 2.4.1x",...` succeeds only on the second occurrence, since the first is followed by a comma rather than whitespace. That repetition is not a convention the table applies uniformly: `TypeScript [14]_,"2.6-5.9"`, `Ruby [10]_,"up to 3.3"`, `Python [9]_,"2.7, 3.5, ..."`, and `JavaScript,ECMAScript 2022 or lower` all omit the language name from that cell. If the Kotlin row is normalized the same way, the parse fails.

To be clear about severity: both of these fail loud rather than silently reporting "no movement", which is the posture #803 asked for and you implemented correctly. That is why this is not blocking. But each is a **false** alarm, triggered by a cosmetic upstream edit the parser could absorb for very little code, and each costs a spurious tracking-issue comment plus a human round-trip to diagnose. Tolerating an optional footnote marker on the label, and not depending on the cell repeating the language name, would remove both without weakening the guard against a genuine reformat. Whatever you do, please add the footnoted-Kotlin-label case to the tests, since it is the variant with actual precedent.

### 3. `gradle/README.md` contradicts itself inside the new section

The section opens:

> The two sections above describe checks to make at bump time.
> `scripts/ci/prs-and-issues/watch_toolchain_bump.py` executes them on a schedule as well [...]

The two sections above are "CodeQL Kotlin ceiling" and "KGP compatibility row". Three paragraphs later the same section says:

> - **It does not read the KGP compatibility table.**

Only one of the two is executed. As written, a reader who stops after the opening sentence comes away believing the compatibility row is being checked for them, which is the single most important thing this watcher deliberately does not do. The rest of the section is careful and reads well; it is just the framing sentence that overclaims.

---

## Nits

4. `main()` discards `add_issue_comment`'s return value and unconditionally prints `Reported an observation on #{issue_number}`. The comment POST can fail and the run log will still claim success. The state itself recovers (no comment means no recorded state, so the next run reports again), but the log is wrong, and this is the one line an operator reads.

5. `render_comment` can emit the heading `Changed: ` with nothing after it. `main()` only calls it when `prior != state`, but `changed_facts` compares only the keys in `FACT_LABELS`, so a prior state JSON carrying a key the current schema does not have compares unequal while producing an empty change list. Reachable via a hand-edited state block, or the next time the tracked tuple changes shape.

6. `_advisory_cell` renders `none` when a `query_osv` call raised `FormatError`, because `collect_state` folds that case to `[]`:

   ```python
   advisories[f"{coordinate}@{version}"] = found if found is not None else []
   ```

   So the table cell asserts "no advisories" for a coordinate that was never successfully read. The `[!WARNING]` block does cover the run overall, but every other unreadable fact renders `_not determined_`, and this one does not follow that convention.

7. `TRACKING_ISSUE_LABELS[0]` silently uses only the first label for the lookup. Harmless with one label; a trap the day someone adds a second and expects the search to still match.

8. `scripts/ci/test_label_existence_integration.py` derives `REQUIRED_LABELS` from named source constants, and its docstring lists the sources it draws from. `ci` is covered there only incidentally, through `.github/release.yml`'s changelog-exclude list. The watcher does not merely apply `ci`, it **depends** on the label for `find_issue_by_title` to locate the tracking issue at all, so removing `ci` from the release-notes exclusions would silently remove the guard that keeps the watcher able to find its own issue. Adding `TRACKING_ISSUE_LABELS` to that test's sources makes the dependency explicit and costs two lines. You were right that the test is satisfied today; I am asking for the reason to be recorded rather than coincidental.

---

## Observation, not a change request

Of the three OSV coordinates, only `org.gradle:gradle-core` has ever carried an advisory. I queried OSV package-wide, with no version filter:

```
org.gradle:gradle-core                      -> 2 advisories
com.android.tools.build:gradle              -> 0 advisories, at any version
org.jetbrains.kotlin:kotlin-gradle-plugin   -> 0 advisories, at any version
```

Kotlin advisories are filed against `org.jetbrains.kotlin:kotlin-stdlib` and friends instead (that coordinate returns hits). So two of the three are null channels as things stand. #803 named exactly these three, so this is per spec and I am **not** asking you to change the coordinate list in this PR. I am recording it so that nobody later reads a green `none` in the advisory row as evidence that AGP and the Kotlin toolchain were meaningfully checked.

---

Only item 1 blocks. Items 2 and 3 are cheap and I would rather they land here than as follow-ups, since both are about the watcher misinforming its reader, which is the one failure mode this design is built to avoid.

aunger:
🤖 Author

All eight items addressed in two commits: `380a7d0` (the lookup type) and `23d7d39` (the watcher, the docs, and the nits). Thank you for pulling the historical CodeQL documents and for querying OSV package-wide; both changed what I did, not just how I justified it.

## 1. Duplicate tracking issue -- accepted, and it needed more than `fetch_ok`

You are right, and I had no defence: I wrote the "reporting nothing this run rather than risking a duplicate" guard one call later and did not apply the same thinking to the call where the damage is permanent.

`find_issue_by_title` is now `lookup_issue_by_title`, returning `IssueLookup(fetch_ok, number, state)` on the `CommentLookup` model you cited. `find_existing_issue` keeps its two-state return and its identical query, with the reason it can afford to (a duplicate test-failure issue is noise that `archive_stale_test_failures.py` sweeps) now written into its docstring rather than left implicit.

But working through your second aggravating detail convinced me `fetch_ok` alone is not enough. Index lag produces `fetch_ok=True, number=None` -- a search that *succeeded* and legitimately matched nothing. Every guard keyed on `fetch_ok` would wave that straight through to `create_issue`, and that is precisely the day-one scenario you identified. So absence is now *confirmed* before anything is created:

```python
def find_tracking_issue(token, repository) -> IssueLookup:
    lookup = lookup_issue_by_title(token, repository, TRACKING_ISSUE_TITLE, TRACKING_ISSUE_LABEL)
    if lookup.number is not None or not lookup.fetch_ok:
        return lookup
    return confirm_tracking_issue_absent(token, repository)
```

`confirm_tracking_issue_absent` pages `GET /repos/{repo}/issues?labels=ci&state=all` and matches the title exactly. That endpoint reads through rather than through the index, so it sees an issue created seconds ago. It skips entries carrying `pull_request` (that endpoint returns PRs too), and exhausting its 20-page cap returns `fetch_ok=False` rather than claiming absence it cannot prove. It runs only on the one path that would otherwise take an irreversible action, so the pagination cost is paid once ever rather than weekly.

This also incidentally fixes the `per_page=1`-on-relevance concern you raised in passing: the confirming read is an exact title match, not a ranked one.

Seven new tests cover it, including `test_a_failed_lookup_never_creates_a_second_tracking_issue` and `test_exhausting_the_page_cap_is_not_reported_as_absence`. On your point that the old suite "would go green over the duplicate" -- the renamed `TestLookupIssueByTitle` now asserts the three outcomes separately, including `test_api_error_is_distinguishable_from_no_match`.

## 2. Kotlin-row brittleness -- accepted, both variations

I checked your `codeql-cli/v2.18.0` finding against the live document rather than taking it on trust, and it is exactly as you describe:

```
Kotlin [7]_,"Kotlin 1.5.0 to 2.0.0\ *x*","kotlinc",``.kt``
.. [7] Kotlin support is currently in beta.
```

Fixed both ways:

- `_strip_rst_markup` now drops footnote references (`\s*\[\d+\]_`), so a caveat re-added to Kotlin cannot read as "the table has no Kotlin row".
- The bound is read from the first CSV cell after the label, via `\bto\s+(\d+(?:\.\d+)*x?)`, with no requirement that the cell repeat the language name.

Restricting the search to the Variants cell also keeps the guard tight: a `to <version>` appearing in the Compilers cell cannot be mistaken for the ceiling, which a whole-line search would have allowed once I loosened the prefix. There is a test for that.

I ran the real v2.18.0 document through the new parser: it yields `2.0.0x`. Five new tests cover the footnoted label, a cell omitting the language name, an unquoted cell, and the cell-scoping. `"through"` in place of `"to"` still raises `FormatError`, and so does `Kotlin/Native,`.

## 3. `gradle/README.md` self-contradiction -- accepted

The framing sentence claimed both checks were executed. It now says one is, names it, and says in the same breath that the compatibility row is not and that a KGP release is watched as its proxy. I also added a line recording the confirmed-absence rule from item 1, since "how the watcher avoids misinforming you" is what that part of the section is about.

## 4-7. Nits -- all accepted

- **4.** `add_issue_comment`'s return value is checked; a failed POST now logs that nothing was recorded and the next run will report again.
- **5.** An empty change list renders `Recorded state refreshed; no tracked fact moved` instead of `Changed: `.
- **6.** A failed OSV query is recorded as `None`, not `[]`, and renders `_not determined_` per coordinate. You were right that it was the only unreadable fact not following that convention.
- **7.** `TRACKING_ISSUE_LABELS[0]` is now a single `TRACKING_ISSUE_LABEL`, so the label applied and the label searched on are one constant and cannot drift.

## 8. Label-existence test -- accepted

`TRACKING_ISSUE_LABEL` is now imported into `REQUIRED_LABELS` and listed in that test's docstring alongside the other sources. Your framing is the right one: the watcher does not merely apply `ci`, it cannot find its own issue without it, and depending on `.github/release.yml`'s exclusion list to keep that label alive was coincidence rather than design.

## On the observation about null OSV channels

Recorded, not argued with. Your package-wide queries match what I can reproduce: only `org.gradle:gradle-core` has ever carried an advisory, and Kotlin's are filed against `kotlin-stdlib` and friends. #803 named these three coordinates, so I have left the list alone, and your comment is now the durable record that a green `none` in that row is weaker evidence than it looks. If it is worth acting on, extending the list is a small follow-up issue rather than a change to this PR.

---

Re-verified after the changes: 299 tests in `scripts/ci/prs-and-issues` pass (up from 278), `--dry-run` against live upstream still reproduces the same observation, `scripts/lint/lint.sh --check --all` is clean, and `test_verification_metadata.sh` is 10/10.

aunger:
🤖 Reviewer

Round 2, against `23d7d39`. All eight items are fixed, and I checked each one against the code rather than the reply.

## Item 1: accepted, and your extension of it is the right call

You were right to push past what I asked for. I asked for `fetch_ok`; that alone would have left the day-one hole open, because index lag presents as `fetch_ok=True, number=None`, which is a *successful* search that legitimately matched nothing. Every `fetch_ok` guard would have waved that straight through to `create_issue`. Confirming absence against `GET /repos/{repo}/issues` is the correct fix, and the endpoint choice is right: the repo issues list reads through the database, not the search index, so it sees an issue created seconds earlier.

The details hold up on inspection:

- `find_tracking_issue` short-circuits correctly in all three branches. A search hit is used directly, a failed search returns `fetch_ok=False` without a confirming read (so a blip never triggers pagination), and only "search succeeded, nothing matched" pays for the confirmation.
- Exhausting the 20-page cap falls through to `IssueLookup(fetch_ok=False)` rather than the loop's success return, so the one case where the answer is genuinely unknown cannot be mistaken for absence. `test_exhausting_the_page_cap_is_not_reported_as_absence` pins both the verdict and the call count.
- Skipping entries carrying `pull_request` is necessary and easy to forget; that endpoint does return PRs.
- Paying the pagination cost only on the irreversible path is the right trade, and the docstring says so.

`find_existing_issue` keeps its exact prior semantics (`lookup.number is None -> None`), and the reason it can afford to is now written into the docstring instead of being implicit. `IssueLookup`'s docstring citing `CommentLookup` gives the next reader the precedent.

## Item 2: verified against seven real upstream documents, not just v2.18.0

I re-ran the new parser over every CodeQL CLI tag I sampled, plus the variations:

```
v2.26.2 -> 2.4.1x    v2.25.0 -> 2.3.0x    v2.22.1 -> 2.2.0x    v2.20.0 -> 2.1.0x
v2.18.0 -> 2.0.0x    v2.15.0 -> 1.9.20    v2.12.0 -> 1.7.21
```

v2.18.0 and earlier are the footnoted-label era, and they now parse. The two tolerated shapes work (footnoted label, cell omitting the language name), and so does an unquoted cell. The guard is still tight where it matters: `"through"` in place of `"to"`, a `Kotlin/Native` label, and a missing Kotlin row all still raise `FormatError`.

Scoping the match to the Variants cell was a good catch on your part, not mine. Loosening the prefix without it would have let a `to <version>` in the Compilers cell be read as the ceiling. I constructed that exact row (`"kotlinc up to 9.9.9"`) and confirmed it raises rather than returning `9.9.9`.

## Items 3 through 8

- **3.** The framing sentence now says one check is executed, names it, and says in the same breath that the compatibility row is not. The added line about confirmed absence belongs where you put it, since that paragraph is about how the watcher avoids misinforming its reader.
- **4.** Return value checked; the failure path says nothing was recorded and the next run will report again, which is the useful thing to tell an operator.
- **5.** `Recorded state refreshed; no tracked fact moved` reads better than what I would have written.
- **6.** Recording `None` rather than `[]` is the right shape, and `_verdict_lines` still excludes it from `hits`, so an unqueryable coordinate cannot be rendered as an advisory. I checked that specifically.
- **7.** One `TRACKING_ISSUE_LABEL` constant, applied and searched on.
- **8.** Imported into `REQUIRED_LABELS` with the reason in the comment, and listed in the test's docstring alongside the other sources.

## Verification

- 299 tests in `scripts/ci/prs-and-issues` pass (up from 278), 289 in `scripts/ci/labels`, 93 in `scripts/ci/test-support`. `scripts/ci/test_label_existence_integration.py` imports the new module cleanly and skips only for the absent token, as designed.
- `scripts/lint/lint.sh --check --all` clean.
- `--dry-run` against live upstream still reproduces the same observation (KGP `2.4.10`, CLI `2.26.2`, ceiling `2.4.1x`, no advisories).
- No stale references to `find_issue_by_title` or `TRACKING_ISSUE_LABELS` anywhere in `scripts/` or `.github/`.

## CI

On `23d7d39` everything has reported green except `build-and-test`, still in progress, and `No blocking labels`, which fails by design during an agentic review cycle.

Worth recording so it is not misattributed: `build-and-test` failed on the **previous** SHA (`bfd5d60`) on a single E2E test, `com.gb4pc.e2e.PartialAccessPhotoPickerE2ETest#partialPhotoAccess_isTreatedAsNotGranted_bannerAndNotificationStillAppear`. This PR touches only Python scripts, one workflow, and Markdown; nothing under `app/`. That failure cannot be caused by this diff, and it may well recur on the current run. Not blocking, and not this PR's to fix.

## One note, explicitly not a change request

The watcher can only find its own issue while that issue carries `ci`, and both the search and the confirming list read depend on it. The tracking issue body warns against closing the issue but says nothing about the label. If you ever touch that body again, a line telling a reader not to strip `ci` would close the last way a human can accidentally cause the duplicate you just engineered out. Too small to spend a round on, and item 8 already guards the label's existence.

Nice work on the follow-ups list, too: recording that two of the three OSV coordinates are null channels, rather than quietly widening the list here, is the right scope call.

**Verdict: LGTM.**

aunger:
### Verification plan

**Before merging** (must resolve before PR #812 can be merged):
None.

- [ ] #817
- [ ] #818
- [ ] #819
- [ ] #820
- [ ] #821
- [ ] #822

aunger:
🤖 Verification Agent

Processed the before-merging list from the [Verification Planner comment](https://github.com/aunger/gallery-button-for-pixel-camera/pull/812#issuecomment-5230302535).

| Item | Result |
| --- | --- |
| #817 — first dispatch of `watch-toolchain-bump.yml` creates the tracking issue and posts the initial observation | **NOT AUTOMATABLE** (pre-merge) |
| #818 — second dispatch does not duplicate the tracking issue or post a second comment | **NOT AUTOMATABLE** (pre-merge) |

Both items require `workflow_dispatch`-triggering `watch-toolchain-bump.yml`. I attempted this directly rather than assuming it out of scope:

```
GET  .../actions/workflows/watch-toolchain-bump.yml            -> 404 Not Found
POST .../actions/workflows/watch-toolchain-bump.yml/dispatches -> 403 Resource not accessible by integration
```

The workflow file exists only on this PR's feature branch. GitHub does not register a `workflow_dispatch` trigger for dispatch via the Actions API until the workflow file has been present on the default branch (`main`) at least once — a platform constraint this repo already documents on `regenerate-gradle-toolchain.yml` ("Lives on main so workflow_dispatch registers"). This matches the PR description's own wording: "Dispatch `watch-toolchain-bump.yml` manually **once after merge**."

Both verification steps are therefore inherently post-merge steps that no pre-merge Verification Agent session can execute, regardless of token scope. Full evidence and re-dispatch instructions are on #817 and #818, which remain open pending a post-merge run.

No code, test, or documentation defects were found during this pass — the blocker is structural (merge ordering), not a bug in PR #812.

aunger:
🤖 Author

Merged `main` (`786e5de`) into the branch: `af8b614`, plus one docs commit `7ada89a`. No textual conflicts. Since the ask was to watch for conflicts of *semantics and purpose*, here is what I checked rather than just what git reported.

## The one that mattered: #808's `dependency-audit.yml`

`main` gained a second weekly scheduled advisory scan while this PR was open. That is close enough to this watcher's OSV query to be worth thinking about properly, so:

- **Do they overlap?** No. `dependency-audit.yml` runs `pip-audit` over `scripts/**/requirements*.txt` — the Python packages CI's own helper scripts import. This watcher queries OSV for three **Maven** coordinates. Disjoint ecosystems, disjoint artifacts.
- **Should the toolchain query move there instead?** No. `audit_requirements.py` is pip-specific end to end (pip-audit subprocess, uv-lock pin parsing, a per-lock TOML ignore file). More to the point, the two answer different questions: it asks "is an installed Python package vulnerable", this asks "is there an argument for a toolchain bump". #803 put the OSV query in the watcher for exactly that coupling.
- **Does this branch need registering with it?** No. It "discovers rather than lists" locks by glob, and this branch adds no requirements lock.

The risk here is not breakage, it is a future reader finding two scheduled advisory jobs and assuming one is redundant, or folding the Maven query into the pip tool. Commit `7ada89a` writes the distinction into `gradle/README.md`'s existing "does not scan the whole dependency graph" bullet — three lines, docs only, no behavior change. Flagging it explicitly since it lands after your LGTM.

## The other three

- **#807's `install-pinned-requirements.sh`** — I checked whether `watch-toolchain-bump.yml` should now call it instead of `pip install --require-hashes`. It should not: that script is a session-start provisioner, keyed on a `$HOME` marker for in-container idempotence and using `--force-reinstall`, neither of which means anything on a fresh runner. #807 changed no workflow, and `build.yml`, `semgrep.yml`, and #808's own new `dependency-audit.yml` all still invoke pip directly. The workflow stays as it is.
- **#824/#823's changed-file label rules** — `CI_PATH_PREFIXES` covers `.github/workflows/` and `scripts/ci/`, so every new file here already classifies as `ci` (and the two `test_*.py` as `ci` + `automated tests`). `gradle/README.md` is unclassified, which empties the unanimity intersection, but that rule is add-only so it costs nothing. The new `agents` if-and-only-if rule matches nothing in this diff, correctly.
- **#816's `test_ci_monitor.py` move** — not referenced here.

**Cron slots** are the one place a collision was plausible, and they are distinct: this watcher Monday 06:00 UTC, `semgrep.yml` 08:00, `dependency-audit.yml` 09:00.

## Re-verified on the merged tree

- `scripts/ci/prs-and-issues` 299, `scripts/ci/labels` 367, `scripts/ci` 53 (now including #808's audit tests; `test_label_existence_integration.py` still imports the watcher cleanly and skips only for the absent token), `scripts/ci/test-support` 93, `scripts/lint` 12, `scripts/ci_monitor/test_ci_monitor.py` 375 checks at its new path.
- `scripts/lint/lint.sh --check --all` clean for python, markdown, and hygiene.
- `scripts/test_verification_metadata.sh` 10/10.
- `--dry-run` against live upstream still reproduces the same observation: KGP `2.4.10`, CLI `2.26.2`, ceiling `2.4.1x`, no advisories.

`git diff --stat origin/main...HEAD` is still exactly this branch's 7 files, so the merge pulled in no stray content.

On your closing note about the tracking-issue body warning against stripping `ci`: I left it alone deliberately. It is a real hole, but it is behavior-adjacent rather than docs-only, and #817/#818 are already open for the post-merge dispatch. Happier to do it as a follow-up than to slip it in under a merge commit.

aunger:
🤖 Author

Re-checked the issue and this PR. No new review items are outstanding — the round-2 verdict is LGTM, and the Verification Planner lists **nothing** before-merging. Two things have changed since my last comment, and neither needs a code change.

## 1. `build-and-test` is red on the merge head, and it is not this PR

Run [1362](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/31373551100) failed on one E2E test:

```
1 blocking test failure(s); 0 infrastructure failure(s); 0 allowlisted failure(s).
  - [E2E Tests] com.gb4pc.e2e.GalleryButtonVisualE2ETest#test3a_populatedGalleryShowsGreenAfterTap
```

I did not want to wave this away as "flaky" just because the last one was, so:

**It is a long-standing flake with its own open issue, #242, and it has failed on `main` itself.** From that issue's history, the same test with the *same* assertion:

| When | Branch | Assertion |
| --- | --- | --- |
| 2026-05-26 | `fix/issue-237-…` | `captureOnePhoto(): new image row did not appear in MediaStore within 15 s` |
| 2026-05-26 | **`main`** | same |
| 2026-05-26 | `fix/issue-246-…` | same |
| 2026-07-17 | `feature/issue-720-…` | `GREEN coverage after tap is 0.0%` |
| 2026-08-09 | `feature/issue-785-…` (PR #824) | `GREEN coverage after tap is 0.0%` |
| 2026-08-10 | this PR | `captureOnePhoto(): new image row did not appear in MediaStore within 15 s` |

A failure that reproduces on `main` and on four unrelated branches, with the identical MediaStore-timeout assertion, is not caused by a diff that adds Python scripts.

**And this diff cannot reach it.** Against `origin/main` the PR changes exactly seven files — one workflow YAML, one Markdown file, and five Python files under `scripts/ci/`. Nothing under `app/`, no `.kt`, no `.kts`, no `gradle/wrapper` or `verification-metadata.xml`:

```
$ git diff --name-only origin/main...HEAD | grep -E '^app/|\.kt$|\.kts$|^gradle/(wrapper|verification)'
(no matches)
```

Note also that the two runs bracketing this one disagree on *which* E2E test fails: 1346 failed `PartialAccessPhotoPickerE2ETest`, 1348 passed outright, 1362 failed `GalleryButtonVisualE2ETest#test3a`. Same suite, different test each time, one clean pass in the middle — the signature of an unstable suite, not a regression.

CI's own `file-issues` job reached the same conclusion and appended this occurrence to #242 rather than filing anything new.

I have deliberately **not** added the test to `.github/allowed-test-failures.txt`. That file's documented steady state is empty, and loosening the merge gate for an unrelated E2E test from inside this PR would be exactly the kind of unrelated change `pr_creation.md` warns against. #242 is the right home for it.

Everything else on the head commit is green, including #808's new `pip-audit` job.

## 2. Label state, for the Orchestrator

The PR currently carries `changes requested`, which is what `No blocking labels` is failing on. Given the round-2 LGTM, that looks stale rather than intended. Per `dev_orchestration.md` these transitions belong to the Orchestrator (`When the Reviewer returns … Remove changes done`, and add `changes requested` *only* if changes were requested), and that document explicitly says an unexpected label state should be escalated rather than quietly corrected, so I have not touched it. Flagging it as the only thing standing between this PR and a mergeable state.

## Current state

- Branch is merged up to `main` (`786e5de`), working tree clean, nothing further to push.
- Full suite green locally on the merged tree; `--dry-run` against live upstream still reproduces the same observation.
- #817 and #818 remain open as post-merge dispatch checks, which the Verification Agent confirmed cannot run before this workflow file reaches `main`.

PR #816, opened 2026-08-09
Move scripts/test_ci_monitor.py into scripts/ci_monitor/

### LINKED ISSUE #787: Move scripts/test_ci_monitor.py into scripts/ci_monitor/, beside the module it tests
`scripts/test_ci_monitor.py` is the only unit test in the repository that does not sit beside its subject.
It tests `scripts/ci_monitor/ci_monitor.py` from one directory up, reaching across with a `sys.path` insert.

## How it got this way

Both files were created together at `scripts/` root in #285 ("Port CI Monitor from bash to Python").

#384 ("Move ci_monitor.py into scripts/ci_monitor/ and split off its usage docs") then moved the module into its own directory but left the test where it was, patching the test to reach across instead:

```diff
-sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "ci_monitor"))
```

with the same one-level reach applied to `_MONITOR_PATH`.
That PR's description says it "updates `scripts/test_ci_monitor.py` to locate the relocated module and script at their new path", so the test was treated as a caller to fix up rather than as part of the thing being moved.

#779 (the `scripts/` reorganization) did not correct it either.
That reorganization's principle was that a test file pairs with its subject, and it relocated every other test accordingly, but `scripts/ci_monitor/` was scoped out as already-correct, so the rule was never applied here.

## Why it is worth fixing

1. **It is the only cross-directory unit test.**
   Every other test in the tree uses the plain `sys.path.insert(0, os.path.dirname(__file__))` form.
   The one other file that reaches elsewhere, `scripts/ci/test_label_existence_integration.py`, does so deliberately because it is a cross-directory integration test.

2. **The move reverts #384's patch rather than adding anything.**
   Both the `sys.path` insert and `_MONITOR_PATH` collapse back to the plain `dirname(__file__)` forms they had before that PR.

3. **`scripts/` root stops being a test directory at all.**
   `test_ci_monitor.py` is its only `test_*.py`, so `find scripts -name "test_*.py" -printf '%h\n'` in `build.yml` would stop returning bare `scripts`.

4. **Path-based tooling can see it.**
   The `agents` labeling rule under discussion in #775 keys on `**/ci_monitor/**/*`, which covers the module but not its test today.

## What changes

- `git mv scripts/test_ci_monitor.py scripts/ci_monitor/test_ci_monitor.py`.
- Revert the two path expressions in that file to their pre-#384 forms.
- Update `.github/workflows/build.yml` line 171 to the new path, and the comments at lines 160-163 and 168-170, which name `scripts/` as the directory that legitimately collects no `unittest.TestCase` tests.
  After the move that directory is `scripts/ci_monitor`.

## What does not change

- **The separate `python3 scripts/ci_monitor/test_ci_monitor.py` invocation stays.**
  It exists because the suite is procedural and `check()`-based rather than built from `unittest.TestCase` classes (#619), so `unittest discover` imports it and runs none of its checks.
  That is a property of the suite's style, not of its location.
  Only the path in that line changes.

- **The exit-5 tolerance in the discover loop stays**, for the same reason.
  The directory it applies to changes from `scripts` to `scripts/ci_monitor`.

- **Case (al)'s self-import checks need no change.**
  They already run their subprocesses with `cwd=os.path.dirname(os.path.abspath(__file__))`, so they follow the file.
  Worth confirming in review rather than assuming, since they are the only checks that depend on the file's own importability.

## Verification

- `python3 scripts/ci_monitor/test_ci_monitor.py` directly, confirming the full check count is unchanged from before the move.
- The `build.yml` discover loop locally: `find scripts -name "test_*.py" -printf '%h\n' | sort -u` should no longer list bare `scripts`.
- `scripts/lint/lint.sh --check --all`.

## Non-goals

- **Renaming `ci_monitor` to `ci-monitor`.**
  That question was raised in #776 and deliberately left alone; it stays out of scope here.
- **Converting the procedural suite to `unittest.TestCase` classes.**
  That is #619's territory and would remove the need for the separate invocation, but it is a much larger change and independent of where the file lives.

aunger:
Fixes #787.

`scripts/test_ci_monitor.py` was the only unit test in the repository that did not sit beside its subject: it reached one directory up to import `scripts/ci_monitor/ci_monitor.py`, via a `sys.path.insert` that named `"ci_monitor"` explicitly.

## Changes

- `git mv scripts/test_ci_monitor.py scripts/ci_monitor/test_ci_monitor.py`.
- Reverted the `sys.path.insert` and `_MONITOR_PATH` expressions to their pre-#384 `dirname(__file__)` forms (the two spots #384 patched when it moved `ci_monitor.py` but left the test behind).
- Also reverted four more path expressions in the same file (added after #384, by #500 and #516) that likewise joined a literal `"ci_monitor"` segment onto `dirname(__file__)` to reach `ci_monitor.config.json` and `ci_monitor.py`: with the test now living in that directory, the segment is redundant and was dropped the same way.
- Updated `.github/workflows/build.yml` line 171's direct invocation to the new path, and the comments at lines 160-163 and 168-170 that named `scripts/` as the directory expected to legitimately collect no `unittest.TestCase` tests; after the move that directory is `scripts/ci_monitor`.

With the move, every unit test in the tree now uses the plain `sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))` form, and `scripts/` itself no longer matches `find scripts -name "test_*.py"`.

## What does not change

- The separate `python3 scripts/ci_monitor/test_ci_monitor.py` invocation in `build.yml` stays (only its path changes): the suite is procedural and `check()`-based rather than built from `unittest.TestCase` classes (#619), so `unittest discover` imports it but runs none of its checks.
- The discover loop's exit-5 tolerance stays, now scoped to `scripts/ci_monitor` instead of bare `scripts`.
- Case (al)'s self-import subprocess checks needed no change: they already run with `cwd=os.path.dirname(os.path.abspath(__file__))`, so they followed the file automatically. Confirmed by running the suite after the move.

## Verification

- `python3 scripts/ci_monitor/test_ci_monitor.py`: all 375 checks pass (same count as before the move).
- `find scripts -name "test_*.py" -printf '%h\n' | sort -u` no longer lists bare `scripts`; the `build.yml` discover loop was run locally over the resulting directory list and returned rc=0 for every directory, including `scripts/ci_monitor`.
- `scripts/lint/lint.sh --check --only python --all`: all checks passed, 31 files already formatted.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,7 +158,7 @@ jobs:
             echo "==> Discovering tests in $dir"
             python3 -m unittest discover -s "$dir" -p "test_*.py" -v && rc=0 || rc=$?
             # Python 3.12+ exits 5 for "no tests were collected" instead of 0.
-            # scripts/ itself hits this legitimately: its only test_*.py is
+            # scripts/ci_monitor itself hits this legitimately: its only test_*.py is
             # test_ci_monitor.py, which defines no unittest.TestCase tests (see
             # below). Tolerate 5; any other nonzero code is a real failure.
             if [[ $rc -ne 0 && $rc -ne 5 ]]; then
@@ -168,7 +168,7 @@ jobs:
           # test_ci_monitor.py is a procedural check()-based suite, not built from
           # unittest.TestCase classes, so `unittest discover` imports it but runs
           # none of its checks (issue #619). Run it directly too.
-          python3 scripts/test_ci_monitor.py
+          python3 scripts/ci_monitor/test_ci_monitor.py
 
       - name: Set up JDK 17
         if: needs.check-diff.outputs.needs_full_build == 'true'
--- a/scripts/ci_monitor/test_ci_monitor.py
+++ b/scripts/ci_monitor/test_ci_monitor.py
@@ -118,7 +118,7 @@ import unittest.mock
 import urllib.error
 import zipfile
 
-sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "ci_monitor"))
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
 import ci_monitor  # noqa: E402
 
@@ -1336,9 +1336,7 @@ def main() -> int:
     # SIGTERM to that exact PID, and confirm the process exits via the signal. A
     # bogus token + the unreachable real API_BASE means the loop never gets past the
     # SHA fetch, so it stays alive (sleeping) until we signal it; no network needed.
-    _MONITOR_PATH = os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "ci_monitor", "ci_monitor.py"
-    )
+    _MONITOR_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "ci_monitor.py")
     _proc = subprocess.Popen(
         [sys.executable, _MONITOR_PATH, "--pr", "1"],
         stdout=subprocess.PIPE,
@@ -3154,7 +3152,7 @@ def main() -> int:
 
     # The committed repo config carries this repo's values, including the dual marker.
     _repo_cfg_path = os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "ci_monitor", "ci_monitor.config.json"
+        os.path.dirname(os.path.abspath(__file__)), "ci_monitor.config.json"
     )
     cfg_repo = ci_monitor.load_config(_repo_cfg_path)
     check(
@@ -3324,7 +3322,7 @@ def main() -> int:
     )
 
     _MONITOR_SRC = open(
-        os.path.join(os.path.dirname(os.path.abspath(__file__)), "ci_monitor", "ci_monitor.py"),
+        os.path.join(os.path.dirname(os.path.abspath(__file__)), "ci_monitor.py"),
         encoding="utf-8",
     ).read()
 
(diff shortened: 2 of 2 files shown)


aunger:
🤖 Reviewer

Verified this independently rather than taking the PR description's claims on faith:

- **The move itself**: `git mv scripts/test_ci_monitor.py scripts/ci_monitor/test_ci_monitor.py`, 99% similarity, executable bit preserved. `scripts/ci_monitor/` now holds the module, its config, its docs, and its test together.
- **All six stale path expressions reverted correctly.** I diffed the pre-move file against the post-move file directly (not just read the PR diff): the `sys.path.insert`, `_MONITOR_PATH`, and the four `ci_monitor.config.json`/`ci_monitor.py` lookups (from #500/#516) all collapse to the plain `dirname(__file__)` form, matching every other test in the tree. Grepped the whole file afterward for a stray `"ci_monitor"` literal — none left.
- **build.yml**: line 171's invocation and both comment blocks (160-163, 168-170) are updated to `scripts/ci_monitor`. YAML parses fine.
- **No orphaned references elsewhere**: grepped the whole repo for `test_ci_monitor.py` across yml/yaml/md/json — only the three lines in build.yml remain, all correct. No CODEOWNERS or labeler config exists yet to worry about (the #775 rule isn't implemented in this repo yet).
- **Ran the suite myself** at the new location: 375 passed, 0 failed — matches the PR's claim exactly. This includes explicitly confirming case (al), the self-import check the issue asked to verify rather than assume: it still passes now that `dirname(__file__)` resolves to `scripts/ci_monitor`.
- **`find scripts -name "test_*.py" -printf '%h\n' | sort -u`**: bare `scripts` is gone, `scripts/ci_monitor` is present, as the issue predicted.
- **`ruff check` and `ruff format --check`** on the moved file: both clean.
- Diff shape matches exactly (+8/-10 across 2 files) — nothing hidden by the rename detection.
- Scope matches the issue precisely; the two explicit non-goals (renaming `ci_monitor`→`ci-monitor`, converting to `unittest.TestCase`) are correctly left untouched.

Nothing to flag. This is a clean, minimal, faithful move.

CI note: `build-and-test` was still in progress as of this review; `No blocking labels` is failing, which is expected per `pr_participation.md` during an active dev cycle given the PR's current labels.

LGTM.

Issue #783, opened 2026-08-03
Confirm label-by-files applies "ci" to a workflow-only PR with no title trigger


aunger:
This is a follow-on item and does not block merging PR #782.

Source: [PR #782's description](https://github.com/aunger/gallery-button-for-pixel-camera/pull/782) (unchecked Test plan item: "After merge, confirm from a PR touching only `.github/workflows/` that it picks up `ci` with no title trigger, per #781's verification note"), echoing the Verification section of [issue #781](https://github.com/aunger/gallery-button-for-pixel-camera/issues/781): "After merge, confirm from a PR touching only `.github/workflows/` that it picks up `ci` with no title trigger; confirm from that PR's Actions run rather than assuming."

## Task

After PR #782 (which adds `scripts/ci/labels/label_by_files.py` and the `.github/workflows/label-by-files.yml` workflow) merges, watch for the first subsequent PR whose changed files are entirely under `.github/workflows/` and whose title does not otherwise trigger the `ci` label via `label_by_title.py`.
Confirm from that PR's Actions run (the `label-by-files` job) that the `ci` label is applied automatically by the new file-based signal.

This cannot be verified before PR #782 merges, since it requires the new workflow to be live on `main` and to observe it firing on a later PR.

aunger:
🤖 Author

Circumstance 1 applies (addressing the issue does not require a code change): #782's file-based `ci` signal is already live on `main`, and this issue only asked that its behavior be confirmed against a real PR's Actions run rather than assumed. No PR resolves that; a live-fire observation does.

**What I checked first**: every PR merged into `main` since #782 (786, 794, 795, 801, 802, 805), and the open PRs at the time (807, 808, 812, 816, 823, 824). None of them touched only `.github/workflows/`, so no naturally occurring PR yet exercised this signal.

**Action taken**: opened throwaway PR #826 from branch `feature/issue-783-confirm-label-by-files-workflow-only`, containing a single commit that adds a two-line comment to `.github/workflows/label-by-files.yml` (a no-op: the diff touches only that one file under `.github/workflows/`). Its title, "Note why label-by-files.yml checks out the repository first," was deliberately worded to match none of `label_by_title.py`'s regexes (no standalone "ci", "workflow", "automat...", "github action...", "e2e...", "merge" + gate/block/pr, "preflight", "codeql", "agent...", "verif...", "author", "reviewer...", "orchestrat...", "test...", or "unit").

[...]

PR #824, opened 2026-08-09
Make `agents` label an if-and-only-if changed-files rule on PRs

### LINKED ISSUE #785: Finish #775: an if-and-only-if changed-files rule for the agents label on PRs, with title/file precedence
Completes #775.
#781 was the cheap slice and is merged as PR #782; this is the rest.

The core of it is the repository owner's [comment on #775](https://github.com/aunger/gallery-button-for-pixel-camera/issues/775#issuecomment-5162349242), treated here as a requirement rather than a suggestion: `agents` becomes an if-and-only-if rule over a fixed path set, the file list beats the title on PRs, and issues stay title-only because they have no file list.
Everything else in #781's Non-goals list is judged on evidence below and either planned, or dropped with a reason.

## State on `main` as of this writing

Verified against `main` at `cada84a`, not taken from the earlier issues' text:

- PR #779 merged: scripts live under `scripts/{agents,ci/labels,ci/prs-and-issues,ci/test-support,lint}/`.
- PR #780 merged: `label_by_title.py`'s `agents` rule uses `reviewers?`, not `review\w+`.
- PR #782 merged: `scripts/ci/labels/label_by_files.py`, its tests, and `.github/workflows/label-by-files.yml` are live.
  The rule is unanimity over changed paths, add-only, covering `automated tests` (Gradle test source-set segments plus `test_*.py` / `test_*.sh` basenames) and `ci` (`.github/workflows/`).
- #783 (confirm `ci` fires on a workflow-only PR) and #784 (narrow `verif\w+`) are open and untouched by this plan.

## Evidence

I re-ran a dry run over the **220 most recently merged PRs** (155 through 782), fetching each one's real changed-file list from the PR Files API and comparing candidate rules against the labels each PR actually carries.
Every count below comes from that sample.
The harness was throwaway; step 3 turns it into a committed script.

**The `agents` if-and-only-if rule** (`**/AGENTS.md`, `**/CLAUDE.md`, `**/agents/**/*`, `.claude/**/*`):

- Agrees with the current labels on 179 of 220.
- Adds `agents` to 20 PRs that lack it, among them 779, 717, 703, 702, 687, 666, 311, 282, 271.
- Removes `agents` from 21 PRs that carry it: 780, 753, 743, 732, 727, 716, 610, 609, 526, 524, 513, 454, 443, 440, 428, 379, 359, 350, 289, 288, 273.
- Of those 21, 16 got `agents` from the title rule and 5 did not, so they arrived by issue-to-PR propagation or by hand.
  Both paths therefore have to be closed, not just the title one.

**A removal rule for `ci` or `automated tests` would be wrong far more often than right.**
Counting PRs that carry the label where no changed path justifies it, under the full path map this plan proposes:

| Label             | PRs a symmetric removal rule would strip it from |
| ----------------- | ------------------------------------------------ |
| `ci`              | 63 of 220                                        |
| `automated tests` | 28 of 220                                        |
| `agents`          | 21 of 220 (the intended removals)                |

The `ci` list includes 778, 768, 763, 750, 730, 688, 548 and 539: real CI work whose diffs happen to sit outside `.github/workflows/` and `scripts/ci/`, in places like `app/`, `scripts/ci_monitor/`, `build.gradle.kts`, and `.github/allowed-test-failures.txt`.
The `automated tests` list includes 639 (quarantining a flaky test through the allowlist file), 346 (changing the archive window) and 322 (skipping tests on docs-only PRs).
This is the whole argument for restricting removal to `agents`: its path set is exhaustive by declaration, so "no changed path justifies it" is evidence of "no", while for the other two it is only an absence of evidence.

**Mapping `scripts/ci/` to `ci`** (#781 parked this as "a natural one-line follow-on once #779 has merged and settled"):

Measured on post-reorg paths alone it looks worthless, because `scripts/ci/` is one day old: it fires on 2 of 220.
Re-running it against the pre-reorg locations of the files that now live under `scripts/ci/` shows what it will actually do going forward: it fires on **38 of 220** and adds a `ci` the PR lacks on **8** (780, 713, 659, 615, 379, 375, 359, 350).
I read all 38; every one is genuine CI automation work.
The reason it earns its keep is that this repo's most common CI-change shape is "a script under `scripts/ci/` plus its `test_*` sibling".
Today the script itself is unclassified, so the intersection across the two paths is empty and the PR gets nothing.
With the mapping the script scores `{ci}` and its test scores `{ci, automated tests}`, so the intersection is `{ci}` and unanimity fires.

**Line-count thresholds are worse than nothing.**
A "more than half the changed lines fall under paths for this label" rule fires on **79 of 220** PRs where unanimity does not, and inspection shows most of those are wrong: 771 (a product fix for overlay thumbnail sizing, 93% test lines), 747 (Gradle wrapper authentication, 90%), 660, 612, 564, 557, 552.
Tests routinely outnumber product lines in this repo, so "most changed lines are tests" approximates "this repo has tests".
That is exactly the conflation #775 complains about ("touched a test file" versus "primarily about tests"), reintroduced with a number attached to make it look principled.
There is also a mechanical problem: 22 files in the sample have `changes == 0` (pure renames, all in 779), so the denominator is not reliable.

**Release-notes effect.**
The last tag, v0.0.14, is the merge of PR #612 on 2026-07-06, and 63 sampled PRs have merged since.
Applying this plan's rules to those 63 changes the `.github/release.yml` exclusion status of 13 of them (735, 725, 717, 715, 703, 702, 696, 687, 668, 665, 653, 635, 616), and every one of the 13 moves in the "newly excluded" direction.
None moves the other way.
That number is the entire case for step 3's backfill mode.

## Decisions carried in from the owner's comment

1. `agents` applies **if and only if** the PR modifies at least one of `**/AGENTS.md`, `**/CLAUDE.md`, `**/agents/**/*`, `.claude/**/*`, regardless of what else it changes.
   This is an existential rule with a removal half, not a parameter change to #781's unanimity rule; the two shapes coexist.
2. On a PR, where the two signals disagree, the file list wins.
3. Issues have no changed-file list, so they stay title-only and nothing about issue labeling changes.
4. `.github/workflows/CLAUDE.md` gets both `agents` and `ci`.

The label description already reads "Issues and changes to AI agent behaviors. (e.g. AGENTS.md, CLAUDE.md, or files in agents/.)", where the "e.g." makes the list non-exhaustive, so no repo-settings change is needed and #781's label-description objection is retired.

## How precedence is implemented: partition, do not race

"The file list wins" could be implemented by letting both scripts write `agents` and having the file signal remove what the title signal added.
That is wrong, and reachable rather than theoretical:

- `label-by-title.yml` runs on `pull_request: [opened, edited]`; `label-by-files.yml` runs on `[opened, synchronize]`.
  On `opened` both run concurrently, so which one lands last is a coin flip.
  On a title edit only the title job runs, so a re-added `agents` survives until the next push.
- `propagate-issue-labels.yml` copies a linked issue's labels onto the PR on `[opened, edited, reopened]`.
  Issue #775 itself carries `agents`; a PR that fixed it by touching only `scripts/ci/labels/` would receive `agents` from propagation, and again on every body edit.
  The 5 non-title removals in the evidence above are this case.

So precedence is made **structural**: on a PR, exactly one writer owns `agents`.

| Artifact | `agents` written by | `ci` and `automated tests` written by                |
| -------- | ------------------- | ---------------------------------------------------- |
| Issue    | title rule only     | title rule only                                      |
| PR       | file rule only      | title rule, plus the file rule's add-only unanimity  |

The removal call still exists, as a backstop for a label applied by hand or left over from before this lands, but it stops being the mechanism that resolves a race.

## Implementation plan

Three PRs, in this order.
Step 1 is independent of steps 2 and 3 and could be folded into step 2 if fewer review rounds are preferred; it is split out because it is add-only and step 2 is not.

### 1. Map `scripts/ci/` to `ci`

In `scripts/ci/labels/label_by_files.py`:

```python
CI_PATH_PREFIXES = (".github/workflows/", "scripts/ci/")
```

That is the whole change.
`scripts/lint/` and `scripts/ci_monitor/` are deliberately left out: `lint.sh` is a local pre-commit tool as much as a CI step, `ci_monitor` is the Orchestrator's tool rather than repo CI, and neither gained anything in the sample.

Tests in `scripts/ci/labels/test_label_by_files.py`:

- `labels_for_path("scripts/ci/labels/label_by_title.py")` gives `{"ci"}`.
- `labels_for_path("scripts/ci/labels/test_label_by_title.py")` gives `{"automated tests", "ci"}` (both rules fire on one path).
- `labels_for_path("scripts/ci/test-support/dismiss_anr.sh")` gives `{"ci"}`.
- `labels_for_path("scripts/lint/lint.sh")` and `labels_for_path("scripts/ci_monitor/ci_monitor.py")` give the empty set, pinning the exclusions above.
- A regression fixture on PR #782's real file list (`.github/workflows/label-by-files.yml`, `scripts/ci/labels/label_by_files.py`, `scripts/ci/labels/test_label_by_files.py`) gives `["ci"]`, which is the "script plus its test" shape that motivates this step.

### 2. The `agents` if-and-only-if rule, and PR precedence

One PR, and it has to be atomic: landing the file rule without the two suppressions would make the label flap.

#### 2a. `scripts/ci/labels/label_by_files.py`

Add the path group, matching the owner's globs exactly.
`**/AGENTS.md` and `**/CLAUDE.md` become basename matches, `**/agents/**/*` becomes a directory-segment match at any depth, and `.claude/**/*` becomes a root-anchored prefix because that is how the owner anchored it:

```python
AGENTS_BASENAMES = frozenset({"AGENTS.md", "CLAUDE.md"})
AGENTS_DIR_SEGMENT = "agents"
AGENTS_PATH_PREFIXES = (".claude/",)


def labels_for_path(path: str) -> frozenset[str]:
    """Return the labels *path* alone justifies (empty when it justifies none)."""
    segments = path.split("/")
    labels = set()
    ...
    if (
        segments[-1] in AGENTS_BASENAMES
        or AGENTS_DIR_SEGMENT in segments[:-1]
        or path.startswith(AGENTS_PATH_PREFIXES)
    ):
        labels.add("agents")
    return frozenset(labels)
```

Matching is case-sensitive, consistent with the existing `androidTest` and `sharedTest` segments; a lowercase `docs/agents.md` is a different document, not a stale spelling of `AGENTS.md`.
`segments[:-1]` keeps a file literally named `agents` from counting as the directory, the same guard the `test` segment already has.

Then split the two rule shapes, which is the only structural change to the module:

```python
# Apply only when *every* changed path justifies it. "Not every path" is an
# absence of evidence, not evidence of absence, so these are never removed:
# the dry run on #775 found the title rule legitimately applying `ci` to 63
# of 220 recent merged PRs whose changed paths justify none of it.
UNANIMOUS_LABELS = frozenset({"ci", "automated tests"})

# The path map above is this label's complete definition, so "no changed
# path justifies it" *is* evidence of "not this label". Apply when any
# changed path justifies it; remove when none does. This is the
# if-and-only-if rule the repository owner recorded on #775.
EXHAUSTIVE_LABELS = frozenset({"agents"})


def matching_labels(paths: list[str]) -> list[str]:
    """Return the labels to apply to a PR changing *paths*; [] when *paths* is empty."""
    if not paths:
        return []
    per_path = [labels_for_path(p) for p in paths]
    unanimous = frozenset.intersection(*per_path) & UNANIMOUS_LABELS
    existential = frozenset().union(*per_path) & EXHAUSTIVE_LABELS
    return sorted(unanimous | existential)


def labels_to_remove(paths: list[str]) -> list[str]:
    """Return the EXHAUSTIVE_LABELS no path in *paths* justifies; [] when *paths* is empty."""
    if not paths:
        return []
    justified = frozenset().union(*(labels_for_path(p) for p in paths))
    return sorted(EXHAUSTIVE_LABELS - justified)
```

Restricting the intersection to `UNANIMOUS_LABELS` and the union to `EXHAUSTIVE_LABELS` is what keeps the two shapes from contaminating each other.
`matching_labels` keeps its name and its "labels to apply" meaning, so #782's existing tests stand unchanged.

An alternative shape, a single `LABEL_RULES: dict[str, str]` mapping each label to `"unanimous"` or `"exhaustive"`, was considered and rejected: it reads no better than two frozensets and forces a lookup where a set intersection does the job.

`main()` gains a removal path:

```python
to_add = matching_labels(paths)
to_remove = labels_to_remove(paths)
```

with three properties worth writing down in code:

- The empty file list and the possibly-truncated list (`fetch_changed_files` returning `None`) both yield **no verdict at all**, neither adds nor removals.
  Removing a label on the strength of a file list that is empty or incomplete is precisely the case where the evidence does not exist.
- Removal only issues a `DELETE` for a label the PR actually carries.
  `main()` fetches the PR's current labels first (`repos/{repo}/issues/{n}`), compares lowercased, and deletes using the name string GitHub returned.
  Without that check, 121 of the 220 sampled PRs (every one that neither touches an agents path nor carries the label) would fire a `DELETE` that 404s and prints a spurious "already removed?" line on every run.
- Removal reuses `enforce_mutually_exclusive_labels.remove_labels(pr_number, to_remove, repo, token)` rather than reimplementing it.
  It is the identical operation with the identical contract (delete each, treat 404 as already-gone, return a boolean), it already URL-encodes the label name, and it brings that module's transient-error retry with it.
  This means `label_by_files` imports two siblings, which is the pattern `propagate_issue_labels.py` already uses.
  Note the asymmetry it leaves: the add path still goes through `label_by_title.gh_api`, which has no retry.
  Unifying the module on one client is a separate change and a non-goal here, since it would churn every `patch.object(lbf.label_by_title, "gh_api")` in #782's merged tests for no behavior gain.

The exit-code contract is unchanged in shape: a failed removal joins a failed add in returning 1, and a 404 on removal does not.

#### 2b. `scripts/ci/labels/label_by_title.py`

```python
# Labels that scripts/ci/labels/label_by_files.py determines from a pull
# request's changed files under an if-and-only-if rule. On a pull request the
# file signal owns them outright, so the title rule must not also write them:
# the repository owner's precedence decision on #775 is that the file list
# wins on PRs. Issues carry no changed-file list, so the title rule stays the
# only signal there and nothing is suppressed.
FILE_DETERMINED_LABELS: frozenset[str] = frozenset({"agents"})


def matching_labels(title: str, is_pull_request: bool = False) -> list[str]:
    """Return the labels whose rule matches *title*."""
    labels = [label for label, pattern in LABEL_PATTERNS.items() if pattern.search(title)]
    if is_pull_request:
        return [label for label in labels if label not in FILE_DETERMINED_LABELS]
    return labels
```

`main()` reads a new optional environment variable:

```python
is_pull_request = os.environ.get("IS_PULL_REQUEST", "").strip().lower() == "true"
```

Defaulting to `False` makes the omission fail open: a caller that forgets it labels a PR from the title as before, and the file signal removes the label on the next push.
`LABEL_PATTERNS["agents"]` itself is untouched, so this does not collide with #784 beyond both editing the same file.

#### 2c. `scripts/ci/labels/propagate_issue_labels.py`

Add a second never-propagate set alongside `PROCESS_STATE_LABELS`, checked at the same point in `labels_to_propagate` and collected into the same `excluded` list:

```python
# Labels a PR's own changed files determine under an if-and-only-if rule (see
# label_by_files.py). What a linked issue is about is not evidence about the
# PR's diff, so copying one of these across would contradict the PR's own
# files. It would also fight the file signal on timing: propagation runs on
# `opened` and on body edits, the file signal on `opened` and `synchronize`,
# so a propagated label would sit on the PR until the next push. Observed on
# 5 of the 21 PRs in #775's dry run that carry `agents` without touching any
# agents path.
FILE_DETERMINED_LABELS: frozenset[str] = frozenset({"agents"})
```

`propagate_to_pr` partitions `excluded` for logging, since the two exclusions have different reasons and the Actions log is the only debugging surface these scripts have.
The sets are disjoint by construction, so the partition is total.
The 3-tuple return signature is deliberately left alone; widening it to a 4-tuple would ripple through `reconcile_issue_labels.py` and the existing tests for no gain.

#### 2d. `.github/workflows/label-by-title.yml`

One line in the existing `env:` block:

```yaml
          IS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
```

No trigger changes anywhere.
`label-by-files.yml` already runs on `[opened, synchronize]`, which is what the removal needs.

#### 2e. `scripts/ci/labels/backfill_labels_by_title.py`

`build_report` must stop proposing `agents` for PRs, or `-f` would reapply exactly what the iff rule forbids:

```python
is_pr = "pull_request" in item
matched = set(label_by_title.matching_labels(item["title"], is_pull_request=is_pr))
tracked = TRACKED_LABELS - label_by_title.FILE_DETERMINED_LABELS if is_pr else TRACKED_LABELS
```

Subtracting from `tracked` as well as from `matched` matters: without it, every PR carrying `agents` lands in the `miss` category on every run, which is precisely the permanent noise the `TRACKED_LABELS` comment says that category exists to avoid.
Document in the module docstring that on PRs the file-determined labels are neither applied nor reported, and that `audit_labels_by_files.py` covers them.

#### 2f. Tests

`scripts/ci/labels/test_label_by_files.py`:

- `labels_for_path` table: `AGENTS.md`, `CLAUDE.md`, `agents/code_edit.md`, `scripts/agents/update_gh_labels.sh`, `.claude/rules/prose-style.md`, `.claude/settings.json`, and `scripts/ci_monitor/CLAUDE.md` all give `{"agents"}`.
- `scripts/agents/test_update_gh_labels.sh` gives `{"agents", "automated tests"}`.
- `.github/workflows/CLAUDE.md` gives `{"agents", "ci"}`, pinning the owner's decision 4.
- `app/src/main/agents` (a file, not a directory) and `docs/agents.md` (lowercase) both give the empty set.
- `agents/pr_participation.md` must be **moved out of** `test_unclassified_paths_give_empty_set`, where #782 put it; leaving it there is how this step fails loudly if the rule is wired up wrong.
- `matching_labels` existential cases: `["agents/code_edit.md", "app/src/main/java/Foo.kt"]` gives `["agents"]`; `["agents/x.md", ".github/workflows/build.yml"]` gives `["agents"]` and not `ci`, since unanimity still fails; `[".github/workflows/CLAUDE.md"]` gives `["agents", "ci"]`.
- `labels_to_remove` cases: `[]` for the empty list; `["agents"]` for a product-only list; `[]` when any path is an agents path; `[]` for an all-test list, since the label is justified nowhere but also not carried, which is the caller's concern rather than this function's.
- Regression fixtures on real file lists: 779 (`.claude/environment.md`, `.claude/hooks/session-start.sh`, `.github/workflows/CLAUDE.md`, and a dozen workflow YAMLs) yields `agents` in the add set; 753 (`scripts/ci_monitor/README.md`, `scripts/ci_monitor/ci_monitor.py`, `scripts/test_ci_monitor.py`) yields `["agents"]` from `labels_to_remove` and nothing from `matching_labels`.
- `main()` cases: no `DELETE` when the PR does not carry the label; a `DELETE` when it does; exit 1 when the removal fails with a non-404; exit 0 when it 404s; no removal attempted when `fetch_changed_files` returned `None` or an empty list.

`scripts/ci/labels/test_label_by_title.py`:

- "Add PR verification agent instructions" still yields `agents` with the default argument (the issue path).
- The same title with `is_pull_request=True` yields no `agents`, while any other label it matches survives.
- `main()` with `IS_PULL_REQUEST=true` POSTs a body without `agents`; with the variable unset, empty, or `false`, behavior is identical to today.
- A title whose only match is `agents`, on a PR, produces no POST at all.

`scripts/ci/labels/test_propagate_issue_labels.py`:

- An issue carrying `["agents", "P1"]` propagating onto a bare PR yields `to_add == ["P1"]` with `agents` in `excluded`.
- `agents` lands in `excluded` even when the PR already carries it, matching how `PROCESS_STATE_LABELS` is ordered ahead of the already-present check today.

`scripts/ci/labels/test_backfill_labels_by_title.py`:

- An item with a `pull_request` key and an agents-matching title: `agents` appears in neither `add` nor `to_apply`.
- The same item already carrying `agents`: `agents` does not appear in `miss`.
- An item with no `pull_request` key: behavior unchanged.

### 3. `scripts/ci/labels/audit_labels_by_files.py`

#775's Related section asks for either a backfill path or an explicit acceptance that the file rule applies going forward only.
This answers it with one script that does both jobs: the repeatable dry run that #781's Verification section assumes exists ("re-running the dry-run is the cheapest regression check if the path map is ever extended", except no such tool was ever committed), and an optional backfill.

The shape mirrors `backfill_labels_by_title.py` deliberately, including its `-n` / `-f` refusal to guess:

```text
python3 scripts/ci/labels/audit_labels_by_files.py (-n | -f) [-q] [--limit N]

    -n, --dry-run   Compute and print the report; apply nothing.
    -f, --force     Apply the "add" and "remove" categories, then print the report.
    --limit N       Consider only the N most recently merged PRs.
```

Report categories, each keyed label to sorted PR numbers: `add` (the rule applies it, the PR lacks it), `match` (the rule applies it, the PR has it), `remove` (the PR carries an `EXHAUSTIVE_LABELS` label the rule forbids).
It walks merged PRs from `repos/{repo}/pulls?state=closed`, filters on `merged_at`, and calls `label_by_files.fetch_changed_files` per PR, reusing the truncation guard rather than reimplementing pagination; a PR whose list comes back `None` is skipped and counted in the output.

Two small reuse notes:

- Generalize `backfill_labels_by_title.format_report` into `format_categories(categories: list[tuple[str, dict[str, list[int]]]]) -> str` and keep `format_report` as a two-line wrapper over it, so the existing formatter tests are untouched and both scripts emit the same JSON shape.
- Like `backfill_labels_by_title.apply_labels`, `-f` logs a failure and continues to the next PR rather than aborting the run.

Cost is roughly one API call per PR, about 600 today, against a 5000/hour authenticated limit.
`--limit` exists so a routine regression check after a path-map change can cover the recent tail cheaply instead of the whole history.

Tests in `scripts/ci/labels/test_audit_labels_by_files.py`: category classification from a fixture PR list, `-n` issuing no writes, `-f` issuing both a POST and a DELETE, a truncated file list being skipped and counted, `--limit` truncating the PR list, and `-f` continuing past a per-PR failure while still returning 1.

## Deferred items from #781 that this plan drops rather than builds

**Line-count thresholds: do not build.**
79 false fires in 220, on the evidence above.
The `additions`, `deletions` and `changes` fields from the Files API stay unused.
If the underlying want is "distinguish a product fix with a big test diff from a test-infrastructure change", a line ratio is the wrong instrument, and the unanimity rule already answers it correctly by abstaining.

**A new label taxonomy ("touched a test file" versus "primarily about tests"): do not build.**
These three labels exist to drive `.github/release.yml`'s changelog-exclude list, and a fourth label has to pick a side of that list.
Excluded, it hides product fixes from the changelog, which is the opposite of what the audit that produced #775 wanted.
Not excluded, it changes nothing that `automated tests` does not already do, and it costs a label creation in repo settings plus the out-of-repo verification that implies.
No PR in the sample needed the distinction once unanimity plus the `scripts/ci/` mapping was in place.

**Removal for `ci` and `automated tests`: do not build.**
63 and 28 wrong strips in 220.
Their path maps are useful heuristics, not definitions, and the asymmetry is the point: those two stay add-only unless someone declares an exhaustive path set for them the way the owner did for `agents`.

**A broader by-title audit: still out of scope**, unchanged from #781's position.
#784 covers the one remaining cheap regex fix.

## Verification

Local, before any PR is opened:

- `python3 -m unittest discover -s scripts/ci/labels -p "test_*.py" -v`
- `scripts/lint/lint.sh --check --all`

No CI wiring is needed.
`build.yml` discovers Python tests with `find scripts -name "test_*.py" -printf '%h\n'`, so new test files in existing directories are picked up, and `lint.yml` already covers the workflow YAML edit.
`scripts/ci/test_label_existence_integration.py` needs no change either: `agents` is already in `REQUIRED_LABELS` through `.github/release.yml`'s exclude list.

After step 3 lands, `python3 scripts/ci/labels/audit_labels_by_files.py -n` over the full history is the end-to-end check that the rules behave as this issue's evidence says.

These four need the workflows live on `main` and a later PR to observe, so they belong in follow-on tracking issues shaped like #783, not in the PRs themselves:

1. A PR whose title matches the `agents` rule but whose files touch no agents path: confirm from the `label-by-title` job's own log that it applied no `agents`.
2. A PR touching an agents path alongside unrelated files: confirm `label-by-files` applies `agents` anyway (the existential half).
3. A PR that carries `agents` at open time and touches no agents path: confirm the `DELETE` fires, and that the "No blocking labels" check re-runs green on the resulting `unlabeled` event.
4. A PR with `Fixes #N` where issue N carries `agents` and the PR touches no agents path: confirm the propagation job logs the exclusion and the PR never receives the label.

## Accepted consequences

- **21 of the last 220 merged PRs would lose `agents`.**
  Going forward, PRs of that shape never get it.
  Named above so the change is auditable rather than surprising.
- **`scripts/ci_monitor/` is not in the owner's path list**, so ci_monitor PRs (753, 732, 526, 524, 428 in the sample) stop carrying `agents` unless they happen to touch `scripts/ci_monitor/CLAUDE.md`, which does match.
  See question 1.
- **7 of the 20 `agents` adds come solely from an incidental `.claude/` edit** in a PR otherwise about lint or CI: 728, 725, 715, 706, 696, 668, 433, most of them touching only `.claude/environment.md` or `.claude/hooks/session-start.sh`.
  That is the existential rule doing exactly what "regardless of what else the PR changes" specifies.
- **A human who deliberately adds `agents` to a PR that touches no agents path will have it removed on the next push.**
  Under an if-and-only-if rule there is no way to keep both properties.
- **Every removal fires an `unlabeled` event**, so `block-merge-on-blocking-labels.yml` re-runs.
  `agents` is not a blocking label, so the check re-passes; the cost is one extra short job per removal, and removals only happen when the label is present.
- **Both adds and removals are user-visible** through the release-notes exclude list.
  That is the intent, since #775 came out of auditing that list, but it means a mistake in the path map shows up in a published changelog.
- **Fork PRs get a read-only token**, so the DELETE fails and the check goes red, the same exposure `label-by-title.yml` and `label-by-files.yml` already carry.
- **A rename out of an agents path is treated as not touching it.**
  `fetch_changed_files` keeps only `filename`, not `previous_filename`, so a PR that moves `agents/foo.md` to `docs/foo.md` and changes nothing else would remove `agents`.
  Including `previous_filename` was considered and rejected: it cuts both ways (promoting `app/src/main/Foo.kt` to `app/src/test/Foo.kt` would stop being classified as a test change), the one mass-rename PR in the sample (779) classifies correctly either way, and it is a three-line change to add later if it ever bites.

## Non-goals

- Thresholds, a new label taxonomy, and removal for `ci` / `automated tests`: dropped with reasons above, not deferred again.
- Mapping `scripts/lint/` or `scripts/ci_monitor/` to `ci`, and mapping `.github/allowed-test-failures.txt` or `.github/release.yml` to anything.
  #781's reasoning on the last two still holds.
- Unifying `label_by_files.py` on a single HTTP client.
  The add path stays on `label_by_title.gh_api`; only the removal path borrows `enforce_mutually_exclusive_labels`.
- Any file-based signal for issues.
- #784's `verif\w+` narrowing.
  It edits `LABEL_PATTERNS["agents"]` in the same file step 2b touches, so whichever lands second rebases; nothing else couples them.
- Running `audit_labels_by_files.py -f`.
  Step 3 builds the capability; whether to fire it is question 2.

## Questions for the owner

1. **`scripts/ci_monitor/`.**
   It is the Orchestrator's own tool, `agents/dev_orchestration.md` documents it at length, and it carries its own `CLAUDE.md`, but it is not one of the four path groups you listed.
   Taking the list literally costs five ci_monitor PRs their `agents` label in the 220-PR sample.
   Add `scripts/ci_monitor/` to the `agents` path set, or leave the list exactly as written?
   The plan implements it as written unless you say otherwise.

2. **Run the backfill once?**
   13 PRs merged since v0.0.14 would change release-note exclusion status, all in the newly-excluded direction, and only a backfill fixes them before the next tag.
   Run `audit_labels_by_files.py -f` once after step 3 lands, or build the tool and leave history alone?

3. **Is step 3 worth it at all?**
   It is the one step that can be cut without weakening the rules.
   Cutting it means the 13 PRs above stay misclassified in the next release's notes, and the next path-map extension goes back to a throwaway dry-run script like the one behind this issue's evidence.

aunger:
Fixes #785.

Implements the plan in #785 as written, which is the repository owner's decision on #775 treated as a requirement: on a pull request `agents` applies if and only if the diff touches one of `**/AGENTS.md`, `**/CLAUDE.md`, `**/agents/**/*`, `.claude/**/*`; where the two signals disagree the file list wins; issues have no changed-file list, so nothing about issue labeling changes.

The issue proposes three PRs. They are delivered here as separate commits on the one branch this work was assigned, in the same order, so each concern is still reviewable on its own.

## What changed, and why it works

**1. `scripts/ci/` maps to `ci`** (`a59d153`).
One entry in `CI_PATH_PREFIXES`. This repo's most common CI change is "a script under `scripts/ci/` plus its `test_*` sibling"; without the mapping the script is unclassified, which empties the unanimity intersection and the PR gets nothing. `scripts/lint/` and `scripts/ci_monitor/` stay out, per the issue's reasoning.

**2. The if-and-only-if rule and PR precedence** (`dd1730a`), atomic because landing the file rule without the two suppressions would make the label flap.

`label_by_files.py` gains the `agents` path group and splits its two rule shapes. `UNANIMOUS_LABELS` (`ci`, `automated tests`) keep the add-only unanimity rule, because their path maps are heuristics and "not every path justifies it" is an absence of evidence: a symmetric removal rule would have stripped `ci` from 63 and `automated tests` from 28 of the 220 PRs in the issue's dry run. `EXHAUSTIVE_LABELS` (`agents`) get the existential add plus a removal, sound only because that path map is the label's complete definition. Restricting the intersection to one set and the union to the other is what keeps the shapes from contaminating each other, so #782's existing tests stand unchanged.

Precedence is structural rather than a race: on a PR exactly one writer owns `agents`. `label_by_title.py` suppresses `FILE_DETERMINED_LABELS` when `IS_PULL_REQUEST` is true, and `propagate_issue_labels.py` never copies them from a linked issue. Both are needed: the title job runs on `opened`/`edited` and the file job on `opened`/`synchronize`, so on open they race, and a later title or body edit would otherwise re-add what the file signal removed. The env var defaults to false so a caller that forgets it fails open.

As of this round, `label_by_files.EXHAUSTIVE_LABELS`, `label_by_title.FILE_DETERMINED_LABELS`, and `propagate_issue_labels.FILE_DETERMINED_LABELS` are one frozenset object, not three independently written literals: `label_by_title.FILE_DETERMINED_LABELS` is the single canonical definition, and the other two reference it directly. See "Changes since the last review round" below.

Three properties are enforced in code rather than left implicit: an empty or possibly-truncated file list yields no verdict at all, neither add nor removal; a `DELETE` only fires for a label the PR actually carries (otherwise 121 of the 220 sampled PRs would 404 on every run); and removal reuses `enforce_mutually_exclusive_labels.remove_labels`, which already URL-encodes the name and carries that module's transient-error retry.

`backfill_labels_by_title.py` stops proposing and stops reporting file-determined labels on PRs, so `-f` cannot reapply what the rule forbids and `miss` does not fill with permanent noise.

**3. `audit_labels_by_files.py`** (`0e4f42d` refactor, `6912ebf` script).
`label-by-files.yml` only sees PRs that are open while it runs, so history is never revisited, and #781's Verification section assumed a repeatable dry run that was never committed. One script does both jobs: replay the rules over each merged PR's real changed-file list, report `add` / `match` / `remove`, and apply only with `-f`. The formatter extraction is committed separately as a pure refactor.

## Answers to the issue's open questions

Nobody commented on #785, so the questions are resolved the way the issue says to resolve them:

1. **`scripts/ci_monitor/`**: implemented as the owner wrote it, per "The plan implements it as written unless you say otherwise." ci_monitor PRs stop carrying `agents` unless they touch `scripts/ci_monitor/CLAUDE.md`, which does match. Reversing this is one entry in `AGENTS_PATH_PREFIXES`.
2. **Run the backfill once?**: not run. Running `-f` is listed as a non-goal of the code change; this PR builds the capability and leaves the decision open.
3. **Is step 3 worth it at all?**: built, since it is in the plan and the plan is the spec. It is also the cheapest thing to drop if you disagree, being two self-contained commits.

## Loosened test requirements

- Modified `test_labels_outside_any_exclusive_group_never_conflict` in `test_propagate_issue_labels.py` to use `bug` instead of `agents` as its example of a label outside any exclusive group, because `agents` is now excluded from propagation outright and no longer reaches the mutual-exclusion path that test is about.
- Re-pointed `test_test_script_basename_gives_automated_tests` in `test_label_by_files.py` at `scripts/lint/test_check_md040.py`, because `scripts/ci/labels/test_label_by_files.py` now legitimately matches both rules. The both-rules case is covered by the new `test_ci_test_script_gives_both_ci_and_automated_tests`.
- `agents/pr_participation.md` moved out of `test_unclassified_paths_give_empty_set` into the new `test_agents_path_set_gives_agents`, as #785 requires.

## Changes since the last review round

Four commits address the review, in this order:

1. Tied `label_by_files.EXHAUSTIVE_LABELS`, `label_by_title.FILE_DETERMINED_LABELS`, and `propagate_issue_labels.FILE_DETERMINED_LABELS` together by reference instead of three separately written literals, with an `is`-identity test in each dependent module (the blocking review comment).
2. Gave the two non-conflict removal call sites (`label_by_files.remove_unjustified_labels`, `audit_labels_by_files.apply_changes`) a truthful `reason="unjustified"` instead of `enforce_mutually_exclusive_labels.remove_labels`'s default "conflicting" wording (nit 1).
3. Made `audit_labels_by_files`'s `--limit` sort by `updated` instead of `created` (so it approximates most-recently-*merged* rather than most-recently-*opened*, the documented intent) and stop pagination on a short page instead of only an empty one, matching `label_by_files.fetch_changed_files`'s own early stop (nits 2 and 3).
4. Added a coverage test pinning that `label_by_files.main()` skips the current-label GET entirely when the diff justifies `agents` (nit 5).

Nit 4 (the commit-message omission in `dd1730a`) is addressed in a reply comment rather than by rewriting that commit's message; see the reply for why.

## Verification

Run locally, all green: `python3 -m unittest discover -s scripts/ci/labels -p "test_*.py"` (365 tests), the other four Python test directories CI discovers, `python3 scripts/test_ci_monitor.py` (375 checks), and `scripts/lint/lint.sh --check --all`. No CI wiring was needed: `build.yml` discovers `test_*.py` by directory and `lint.yml` covers the workflow YAML edits.

Three checks still need the workflows live on `main` and a later PR to observe, so per #785 they belong in follow-on tracking issues shaped like #783 rather than in this PR:

- [ ] A PR touching an agents path alongside unrelated files: confirm `label-by-files` applies `agents` anyway.
- [ ] A PR that carries `agents` at open time and touches no agents path: confirm the `DELETE` fires and the "No blocking labels" check re-runs green on the resulting `unlabeled` event.
- [ ] A PR with `Fixes #N` where issue N carries `agents` and the PR touches no agents path: confirm the propagation job logs the exclusion and the PR never receives the label.

The fourth check needs no follow-on issue: a PR whose title matches the `agents` rule but whose files touch no agents path. This PR is itself an instance of it (its title matches the `agents` title regex, and its diff touches no agents path), and its own CI already demonstrated the result live at open time: the `label-by-title` job ran with `IS_PULL_REQUEST: true` against this title and logged "No label rules matched title ... nothing to do", while `label-by-files` applied only `['ci']`.

### DIFF
--- a/.github/workflows/label-by-files.yml
+++ b/.github/workflows/label-by-files.yml
@@ -1,9 +1,10 @@
 name: Label PR by changed files
 
-# Applies labels to a pull request based on the set of files it changed,
-# when every changed file maps to the same label. See
-# scripts/ci/labels/label_by_files.py for the exact rule. Labels are only
-# ever added, never removed.
+# Applies labels to a pull request based on the set of files it changed.
+# "ci" and "automated tests" are applied when every changed file maps to the
+# label, and never removed. "agents" follows an if-and-only-if rule instead:
+# applied when any changed file maps to it, and removed when none does. See
+# scripts/ci/labels/label_by_files.py for the exact rules.
 #
 # A separate workflow from label-by-title.yml because the two signals want
 # different triggers: title labeling only needs to run when the title
--- a/.github/workflows/label-by-title.yml
+++ b/.github/workflows/label-by-title.yml
@@ -4,6 +4,10 @@ name: Label issue/PR by title
 # against its title. See scripts/ci/labels/label_by_title.py for the exact rules.
 # Labels are only ever added, never removed.
 #
+# IS_PULL_REQUEST tells the script which artifact it is labeling. On a pull
+# request it suppresses the labels label-by-files.yml determines from the
+# changed files, so the two signals never race for the same label.
+#
 # Runs on "opened" (a fresh title) or "edited" where the title itself
 # changed (github.event.changes.title is only present when the title was
 # part of the edit), so unrelated edits (body, labels, base branch) do
@@ -33,4 +37,5 @@ jobs:
           GITHUB_REPOSITORY: ${{ github.repository }}
           ISSUE_NUMBER: ${{ github.event.pull_request.number || github.event.issue.number }}
           ISSUE_TITLE: ${{ github.event.pull_request.title || github.event.issue.title }}
+          IS_PULL_REQUEST: ${{ github.event_name == 'pull_request' }}
         run: python3 scripts/ci/labels/label_by_title.py
--- a/scripts/ci/labels/audit_labels_by_files.py
+++ b/scripts/ci/labels/audit_labels_by_files.py
@@ -1,0 +1,278 @@
+#!/usr/bin/env python3
+"""Audit merged pull requests against the changed-file label rules.
+
+Applies scripts/ci/labels/label_by_files.py's rules (see that module for the
+exact rules) to the real changed-file list of every merged pull request, and
+compares the verdict against the labels the pull request actually carries.
+
+This does two jobs with one walk. It is the repeatable dry run that makes
+re-running the evidence the cheapest regression check when the path map is
+extended, and, with -f, the backfill for pull requests that merged before a
+rule existed. label-by-files.yml only ever sees pull requests that are open
+when it runs, so history is otherwise never revisited.
+
+Before exiting, prints a JSON report of what happened (or would happen):
+
+    add     labels the rules apply that the pull request lacks
+    match   labels the rules apply that the pull request already carries
+    remove  labels the pull request carries that the rules forbid
+
+Only label_by_files.EXHAUSTIVE_LABELS can appear under "remove": the
+unanimity labels are add-only by design, since for them "no changed path
+justifies it" is an absence of evidence rather than evidence of absence.
+
+Each category maps a label name to the sorted list of pull request numbers
+it applies to.
+
+Usage:
+    python3 scripts/ci/labels/audit_labels_by_files.py (-n | -f) [-q] [--limit N]
+
+    -n, --dry-run   Compute and print the report; apply nothing.
+    -f, --force     Apply the "add" and "remove" categories, then print the
+                    report.
+    -q, --quiet     Suppress the JSON report.
+    --limit N       Consider only the N most recently opened merged pull
+                    requests.
+
+    Exactly one of -n or -f is required, matching backfill_labels_by_title.py
+    and git clean's semantics: the script refuses to guess and does nothing
+    without one.
+
+A pull request whose changed-file list may be truncated, or cannot be
+fetched, yields no verdict at all: it is skipped and counted rather than
+judged on a partial diff.
+
+Cost is roughly one API call per pull request examined, against a 5000/hour
+authenticated limit. --limit keeps a routine regression check after a
+path-map change on the recent tail instead of the whole history.
+
+Exit code:
+    0  on success (whether or not any labels were changed)
+    1  if required configuration is missing, --limit is not positive, the
+       pull request list can't be fetched at all, or (-f only) writing
+       labels to at least one pull request failed
+
+Required environment variables:
+    GITHUB_TOKEN        Token with issues: write and pull-requests: write
+    GITHUB_REPOSITORY   Owner/repo (e.g. "aunger/gallery-button-for-pixel-camera")
+"""
+
+import argparse
+import os
+import sys
+import urllib.error
+from collections import defaultdict
+from typing import Callable, NamedTuple
+
+import backfill_labels_by_title
+import enforce_mutually_exclusive_labels as emxl
+import label_by_files
+import label_by_title
+
+PER_PAGE = 100
+
+
+# ---------------------------------------------------------------------------
+# Fetching
+# ---------------------------------------------------------------------------
+
+
+def fetch_merged_pulls(repo: str, token: str, limit: int | None = None) -> list[dict]:
+    """Return merged pull requests, most recently opened first.
+
+    Walks repos/{repo}/pulls?state=closed newest-first and keeps the ones
+    carrying a ``merged_at`` timestamp, since a closed pull request may
+    simply have been closed unmerged. Stops as soon as *limit* of them have
+    been collected, so a small --limit costs a page or two rather than a walk
+    of the whole history.
+    """
+    pulls: list[dict] = []
+    page = 1
+    while True:
+        batch = label_by_title.gh_api(
+            f"repos/{repo}/pulls?state=closed&sort=created&direction=desc"
+            f"&per_page={PER_PAGE}&page={page}",
+            token=token,
+        )
+        if not batch:
+            return pulls
+        for pull in batch:
+            if not pull.get("merged_at"):
+                continue
+            pulls.append(pull)
+            if limit is not None and len(pulls) >= limit:
+                return pulls
+        page += 1
+
+
+# ---------------------------------------------------------------------------
+# Core logic
+# ---------------------------------------------------------------------------
+
+
+class Report(NamedTuple):
+    """What the audit found, plus the writes -f would issue.
+
+    add/match/remove map a label name to the pull request numbers it applies
+    to (the report). to_add/to_remove map a pull request number to the labels
+    to write (the -f payload). skipped lists the pull requests that yielded
+    no verdict.
+    """
+
+    add: dict[str, list[int]]
+    match: dict[str, list[int]]
+    remove: dict[str, list[int]]
+    to_add: dict[int, list[str]]
+    to_remove: dict[int, list[str]]
+    skipped: list[int]
+
+
+def build_report(pulls: list[dict], fetch_paths: Callable[[int], list[str] | None]) -> Report:
+    """Classify each pull request in *pulls* against its own changed files.
+
+    *fetch_paths* maps a pull request number to its changed paths, or to None
+    when no reliable list is available (label_by_files.fetch_changed_files's
+    truncation contract, plus any per-pull-request fetch failure the caller
+    folds into it). None is missing evidence, so that pull request is skipped
+    entirely rather than judged on a partial diff.
+    """
+    add: dict[str, list[int]] = defaultdict(list)
+    match: dict[str, list[int]] = defaultdict(list)
+    remove: dict[str, list[int]] = defaultdict(list)
+    to_add: dict[int, list[str]] = defaultdict(list)
+    to_remove: dict[int, list[str]] = defaultdict(list)
+    skipped: list[int] = []
+
+    for pull in pulls:
+        number = pull["number"]
+        paths = fetch_paths(number)
+        if paths is None:
+            skipped.append(number)
+            continue
+
+        # Keyed by lowercase name, valued by the spelling GitHub returned, so
+        # a removal deletes the label by the name the API knows.
+        carried = {lbl["name"].lower(): lbl["name"] for lbl in pull.get("labels", [])}
+
+        for label in label_by_files.matching_labels(paths):
+            if label.lower() in carried:
+                match[label].append(number)
+            else:
+                add[label].append(number)
+                to_add[number].append(label)
+
+        for label in label_by_files.labels_to_remove(paths):
+            name = carried.get(label.lower())
+            if name is not None:
+                remove[label].append(number)
+                to_remove[number].append(name)
+
+    return Report(dict(add), dict(match), dict(remove), dict(to_add), dict(to_remove), skipped)
+
+
+def apply_changes(
+    to_add: dict[int, list[str]],
+    to_remove: dict[int, list[str]],
+    repo: str,
+    token: str,
+) -> bool:
+    """Apply every pending add and removal, logging failures without aborting.
+
+    Like backfill_labels_by_title.apply_labels, one pull request's failure
+    does not abort the run: the remaining pull requests are still processed
+    and the failure is reflected in the return value.
+
+    Returns True if every write succeeded, False if any failed.
+    """
+    all_succeeded = True
+    for number, labels in to_add.items():
+        if not label_by_files.add_labels(repo, number, labels, token):
+            all_succeeded = False
+    for number, labels in to_remove.items():
+        if not emxl.remove_labels(number, labels, repo, token):
+            all_succeeded = False
+    return all_succeeded
+
+
+# ---------------------------------------------------------------------------
+# Reporting
+# ---------------------------------------------------------------------------
+
+
+def format_report(report: Report) -> str:
+    """Render *report*'s three categories in backfill_labels_by_title's shape."""
+    return backfill_labels_by_title.format_categories(
+        [("add", report.add), ("match", report.match), ("remove", report.remove)]
+    )
+
+
+# ---------------------------------------------------------------------------
+# Entry point
+# ---------------------------------------------------------------------------
+
+
+def parse_args(argv: list[str]) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    mode = parser.add_mutually_exclusive_group(required=True)
+    mode.add_argument("-n", "--dry-run", action="store_true", help="Report only; apply nothing.")
+    mode.add_argument("-f", "--force", action="store_true", help="Apply the adds and removals.")
+    parser.add_argument("-q", "--quiet", action="store_true", help="Suppress the JSON report.")
+    parser.add_argument(
+        "--limit",
+        type=int,
+        default=None,
+        metavar="N",
+        help="Consider only the N most recently opened merged pull requests.",
+    )
+    return parser.parse_args(argv)
+
+
+def main(argv: list[str] | None = None) -> int:
+    args = parse_args(sys.argv[1:] if argv is None else argv)
+
+    token = os.environ.get("GITHUB_TOKEN", "")
+    repo = os.environ.get("GITHUB_REPOSITORY", "")
+
+    if not token:
+        print("Error: GITHUB_TOKEN not set.", file=sys.stderr)
+        return 1
+    if not repo:
+        print("Error: GITHUB_REPOSITORY not set.", file=sys.stderr)
+        return 1
+    if args.limit is not None and args.limit < 1:
+        print(f"Error: --limit must be at least 1: {args.limit}", file=sys.stderr)
+        return 1
+
+    try:
+        pulls = fetch_merged_pulls(repo, token, args.limit)
+    except Exception as exc:  # noqa: BLE001
+        print(f"Error fetching pull requests: {exc}", file=sys.stderr)
+        return 1
+
+    def changed_paths(number: int) -> list[str] | None:
+        try:
+            return label_by_files.fetch_changed_files(repo, number, token)
+        except (urllib.error.HTTPError, urllib.error.URLError) as exc:
+            print(f"Error fetching changed files for #{number}: {exc}", file=sys.stderr)
+            return None
+
+    report = build_report(pulls, changed_paths)
+
+    exit_code = 0
+    if args.force and not apply_changes(report.to_add, report.to_remove, repo, token):
+        exit_code = 1
+
+    if report.skipped:
+        print(
+            f"Skipped {len(report.skipped)} of {len(pulls)} pull requests with no reliable "
+            f"changed-file list: {sorted(report.skipped)}",
+            file=sys.stderr,
+        )
+
+    if not args.quiet:
+        print(format_report(report))
+    return exit_code
+
+
+if __name__ == "__main__":
+    sys.exit(main())
--- a/scripts/ci/labels/audit_labels_by_files.py
+++ b/scripts/ci/labels/audit_labels_by_files.py
@@ -31,8 +31,7 @@ Usage:
     -f, --force     Apply the "add" and "remove" categories, then print the
                     report.
     -q, --quiet     Suppress the JSON report.
-    --limit N       Consider only the N most recently opened merged pull
-                    requests.
+    --limit N       Consider only the N most recently merged pull requests.
 
     Exactly one of -n or -f is required, matching backfill_labels_by_title.py
     and git clean's semantics: the script refuses to guess and does nothing
@@ -78,19 +77,32 @@ PER_PAGE = 100
 
 
 def fetch_merged_pulls(repo: str, token: str, limit: int | None = None) -> list[dict]:
-    """Return merged pull requests, most recently opened first.
+    """Return merged pull requests, most recently merged first.
 
-    Walks repos/{repo}/pulls?state=closed newest-first and keeps the ones
-    carrying a ``merged_at`` timestamp, since a closed pull request may
-    simply have been closed unmerged. Stops as soon as *limit* of them have
-    been collected, so a small --limit costs a page or two rather than a walk
-    of the whole history.
+    Walks repos/{repo}/pulls?state=closed by ``updated_at`` descending and
+    keeps the ones carrying a ``merged_at`` timestamp, since a closed pull
+    request may simply have been closed unmerged. Stops as soon as *limit* of
+    them have been collected, or as soon as a page comes back shorter than
+    PER_PAGE (the normal end of the list, matching
+    label_by_files.fetch_changed_files's own early stop), so a small --limit
+    costs a page or two rather than a walk of the whole history.
+
+    ``updated_at``, not ``created_at``, is what makes this "most recently
+    merged" rather than "most recently opened": the REST API has no
+    sort=merged option, but merging a pull request is itself an update to it,
+    so ``updated_at`` is set to the merge time and, for a closed and merged
+    pull request, rarely changes again. The one exception is a label added or
+    removed after merge (a manual correction, or a backfill run of this very
+    script), which would sort that pull request later than its actual merge
+    time. That is judged an acceptable approximation for a regression check
+    on the recent tail--the stated purpose of --limit--rather than a reason to
+    walk the full history client-side to sort by ``merged_at`` exactly.
     """
     pulls: list[dict] = []
     page = 1
     while True:
         batch = label_by_title.gh_api(
-            f"repos/{repo}/pulls?state=closed&sort=created&direction=desc"
+            f"repos/{repo}/pulls?state=closed&sort=updated&direction=desc"
             f"&per_page={PER_PAGE}&page={page}",
             token=token,
         )
@@ -102,6 +114,8 @@ def fetch_merged_pulls(repo: str, token: str, limit: int | None = None) -> list[
             pulls.append(pull)
             if limit is not None and len(pulls) >= limit:
                 return pulls
+        if len(batch) < PER_PAGE:
+            return pulls
         page += 1
 
 
@@ -222,7 +236,7 @@ def parse_args(argv: list[str]) -> argparse.Namespace:
         type=int,
         default=None,
         metavar="N",
-        help="Consider only the N most recently opened merged pull requests.",
+        help="Consider only the N most recently merged pull requests.",
     )
     return parser.parse_args(argv)
 
--- a/scripts/ci/labels/audit_labels_by_files.py
+++ b/scripts/ci/labels/audit_labels_by_files.py
@@ -189,7 +189,7 @@ def apply_changes(
         if not label_by_files.add_labels(repo, number, labels, token):
             all_succeeded = False
     for number, labels in to_remove.items():
-        if not emxl.remove_labels(number, labels, repo, token):
+        if not emxl.remove_labels(number, labels, repo, token, reason="unjustified"):
             all_succeeded = False
     return all_succeeded
 
--- a/scripts/ci/labels/backfill_labels_by_title.py
+++ b/scripts/ci/labels/backfill_labels_by_title.py
@@ -170,24 +170,14 @@ def apply_labels(to_apply: dict[int, list[str]], repo: str, token: str) -> bool:
 # ---------------------------------------------------------------------------
 
 
-def format_report(
-    add: dict[str, list[int]],
-    match: dict[str, list[int]],
-    miss: dict[str, list[int]],
-    include_match: bool = True,
-) -> str:
-    """Render the add/match/miss categories as JSON.
+def format_categories(categories: list[tuple[str, dict[str, list[int]]]]) -> str:
+    """Render named label-to-issue-numbers *categories* as JSON.
 
     Minimized except for a newline after each ``"category":`` and after
-    each label's number list, per the requested output shape. Pass
-    ``include_match=False`` to omit the "match" key entirely (the
-    ``--skip-matches`` flag given with no value).
+    each label's number list, per the requested output shape. Shared with
+    scripts/ci/labels/audit_labels_by_files.py so both reports have the
+    same shape.
     """
-    categories = [("add", add)]
-    if include_match:
-        categories.append(("match", match))
-    categories.append(("miss", miss))
-
     category_parts = []
     for name, labels in categories:
         label_parts = [
@@ -196,6 +186,24 @@ def format_report(
         ]
         category_parts.append(f"{json.dumps(name)}:\n{{{','.join(label_parts)}}}")
     return "{" + ",".join(category_parts) + "}"
+
+
+def format_report(
+    add: dict[str, list[int]],
+    match: dict[str, list[int]],
+    miss: dict[str, list[int]],
+    include_match: bool = True,
+) -> str:
+    """Render the add/match/miss categories as JSON.
+
+    Pass ``include_match=False`` to omit the "match" key entirely (the
+    ``--skip-matches`` flag given with no value).
+    """
+    categories = [("add", add)]
+    if include_match:
+        categories.append(("match", match))
+    categories.append(("miss", miss))
+    return format_categories(categories)
 
 
 def parse_skip_matches(value: str | None) -> tuple[bool, frozenset[str]]:
--- a/scripts/ci/labels/backfill_labels_by_title.py
+++ b/scripts/ci/labels/backfill_labels_by_title.py
@@ -6,6 +6,11 @@ the exact regexes) to the title of every issue and pull request in the
 repository, and applies any label a title matches that is not already
 present. Labels are only ever added, never removed.
 
+On a pull request, label_by_title.FILE_DETERMINED_LABELS are neither applied
+nor reported: those are decided from the PR's own changed files, so applying
+one from a title would reapply exactly what the if-and-only-if rule forbids.
+scripts/ci/labels/audit_labels_by_files.py is the tool that covers them.
+
 Before exiting, prints a JSON report of what happened (or would happen):
 
     add     labels that would be applied and are not already present
@@ -111,7 +116,16 @@ def build_report(
     for item in items:
         number = item["number"]
         current = {lbl["name"] for lbl in item.get("labels", [])}
-        matched = set(label_by_title.matching_labels(item["title"]))
+        # The /issues endpoint marks a pull request with a "pull_request" key.
+        is_pr = "pull_request" in item
+        matched = set(label_by_title.matching_labels(item["title"], is_pull_request=is_pr))
+        # Narrowing "tracked" as well as "matched" matters: leave it alone and
+        # every PR carrying a file-determined label lands in "miss" on every
+        # run, which is precisely the permanent noise TRACKED_LABELS exists to
+        # avoid.
+        tracked = (
+            TRACKED_LABELS - label_by_title.FILE_DETERMINED_LABELS if is_pr else TRACKED_LABELS
+        )
 
         for label in matched:
             if label in current:
@@ -120,7 +134,7 @@ def build_report(
                 add[label].append(number)
                 to_apply[number].append(label)
 
-        for label in TRACKED_LABELS:
+        for label in tracked:
             if label in current and label not in matched:
                 miss[label].append(number)
 
--- a/scripts/ci/labels/enforce_mutually_exclusive_labels.py
+++ b/scripts/ci/labels/enforce_mutually_exclusive_labels.py
@@ -215,8 +215,19 @@ def remove_labels(
     to_remove: list[str],
     repo: str,
     token: str,
+    reason: str = "conflicting",
 ) -> bool:
     """Remove each label in *to_remove* from the issue/PR, logging each removal.
+
+    *reason* is substituted directly into the per-label log line ("Removed
+    {reason} label '<name>' from #N."). It defaults to "conflicting" for this
+    module's own mutual-exclusion callers, where that is literally what
+    happened. scripts/ci/labels/label_by_files.py and
+    scripts/ci/labels/audit_labels_by_files.py reuse this function for a
+    different kind of removal--a label no changed file justifies under an
+    if-and-only-if rule, where nothing conflicts--and pass reason="unjustified"
+    so the log line describes what actually happened instead of contradicting
+    it.
 
     Returns True if every removal succeeded or was already absent (404),
     False if any removal failed with a non-404 HTTPError or a URLError.
@@ -230,7 +241,7 @@ def remove_labels(
                 token=token,
                 method="DELETE",
             )
-            print(f"Removed conflicting label '{label}' from #{issue_number}.")
+            print(f"Removed {reason} label '{label}' from #{issue_number}.")
         except urllib.error.HTTPError as exc:
             if exc.code == 404:
                 print(
--- a/scripts/ci/labels/label_by_files.py
+++ b/scripts/ci/labels/label_by_files.py
@@ -116,7 +116,14 @@ UNANIMOUS_LABELS = frozenset({"ci", "automated tests"})
 # justifies it" *is* evidence of "not this label". Apply when any changed path
 # justifies it, regardless of what else the PR changes; remove when none does.
 # This is the if-and-only-if rule the repository owner recorded on issue #775.
-EXHAUSTIVE_LABELS = frozenset({"agents"})
+#
+# References label_by_title.FILE_DETERMINED_LABELS directly rather than
+# repeating its own literal: that module suppresses these same labels from a
+# pull request's title, and propagate_issue_labels.py excludes them from
+# issue-to-PR propagation, so all three have to name the identical set for the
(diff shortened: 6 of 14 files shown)


aunger (commented):
🤖 Reviewer

Reviewed all four commits against #785's plan and the owner's decision on #775.
Ran locally: `python3 -m unittest discover -s scripts/ci/labels -p "test_*.py"` (357 tests), the other four Python test directories CI discovers (`scripts`, `scripts/ci`, `scripts/ci/prs-and-issues`, `scripts/lint`), and `scripts/lint/lint.sh --check --all`.
All green.

The rule shapes are right, the partition-do-not-race argument holds, and the commit split (step 1, step 2, pure refactor, step 3) is clean.
This PR's own CI already demonstrates verification item 1 live: the `label-by-title` job ran with `IS_PULL_REQUEST: true` against a title that does match the `agents` regex and logged `No label rules matched title ... nothing to do`, while `label-by-files` applied only `['ci']`.
That is worth citing in the PR body, because it means item 1 does not need a follow-on issue.

One blocking problem, then nits.

## Blocking: the invariant the whole PR rests on is not pinned by anything

`label_by_files.EXHAUSTIVE_LABELS`, `label_by_title.FILE_DETERMINED_LABELS`, and `propagate_issue_labels.FILE_DETERMINED_LABELS` are three separately written literals, all currently `frozenset({"agents"})`.
The PR's central claim, that precedence is structural because on a pull request exactly one writer owns the label, is true only while all three are identical.
Nothing in the code or the tests enforces or even checks that.

The failure mode is silent, which is what makes this worth blocking on.
`label_by_files.py`'s comment on `UNANIMOUS_LABELS` explicitly invites the extension ("those two stay add-only unless someone declares an exhaustive path set for them", per #785), so a future change adding a second label to `EXHAUSTIVE_LABELS` is anticipated, not hypothetical.
When that happens the file rule starts adding and removing the new label while the title rule and the propagation rule keep writing it, which is exactly the flapping and the open-event race this PR exists to eliminate.
Every test in the repository still passes.

The one test that looks like it guards this does not.
`test_agents_is_the_only_file_determined_label` in `test_label_by_title.py` pins the title module's copy against a hard-coded `frozenset({"agents"})`, not against `label_by_files`, so extending `EXHAUSTIVE_LABELS` alone produces no failure anywhere.

The asymmetry is also worth noticing before you decide how to fix it: `label_by_files` imports `label_by_title`, so the title module cannot import back, but `propagate_issue_labels` imports only `enforce_mutually_exclusive_labels` today and can import `label_by_files` with no cycle.
Whatever shape you pick, I want the three sets tied together by something that fails when they diverge, not by three comments that cross-reference each other in prose.

Note the priorities this leaves as-is: `propagate_issue_labels.FILE_DETERMINED_LABELS` carries a comment asserting it is "Disjoint from PROCESS_STATE_LABELS by construction" and a test, `test_file_determined_and_process_state_sets_are_disjoint`, that pins it.
The far more load-bearing "equal to `EXHAUSTIVE_LABELS` by construction" gets neither.

## Nits

1. **The removal log line will contradict itself.**
   `enforce_mutually_exclusive_labels.remove_labels` prints `Removed conflicting label '<name>' from #N.`
   Reusing the function is right, but nothing conflicts in either new call site, so every real removal logs a false explanation of why it happened.
   In `remove_unjustified_labels` it directly contradicts the line printed immediately before it ("no changed file justifies them").
   In `audit_labels_by_files.apply_changes` it is the only line printed, with no correct one nearby.
   This PR argues in `propagate_to_pr` that the two exclusion reasons must be logged apart "since the Actions log is the only debugging surface these scripts have"; the same argument applies here and is not applied.

2. **`--limit` does not mean what the plan says it means.**
   #785 specifies "Consider only the N most recently merged PRs".
   `fetch_merged_pulls` requests `sort=created&direction=desc`, so `--limit N` actually gives the N most recently *opened* merged pull requests.
   The script's docstring is honest about this, so it is a deviation from the plan rather than a doc bug, but the two sets genuinely differ: a long-lived PR opened months ago and merged yesterday drops out of `--limit 50` while PRs merged well before it stay in.
   Given the stated purpose is a regression check "on the recent tail", either justify created-order in the docstring or use an ordering that matches the plan.

3. **`fetch_merged_pulls` always spends one wasted API call.**
   `fetch_changed_files` in the same family returns as soon as `len(batch) < PER_PAGE`; `fetch_merged_pulls` only stops on an empty page, so every run fetches one page past the end.
   `test_paginates_until_empty_page` pins this at three calls for two pages of data.
   Trivial in isolation, but it is a gratuitous inconsistency with the function it was modeled on.

4. **The loosened-test declaration is incomplete in the commits.**
   `agents/pr_participation.md` moving out of `test_unclassified_paths_give_empty_set` is listed in the PR text as one of the three loosened requirements, but `dd1730a`'s message declares only the `test_propagate_issue_labels.py` change.
   `code_edit.md` asks for each instance "in your comment message and PR text", on its own line.
   `a59d153` gets this right for its own instance, so the omission looks like an oversight rather than a position.

5. **Small coverage gap in `label_by_files.main()`.**
   There are tests for "no DELETE when the PR does not carry the label" and for the no-verdict cases, but none for "no current-label fetch at all when the diff justifies `agents`".
   That is the path deciding whether an agents-touching PR pays an extra GET on every `synchronize`, and right now it is only implied by the `if to_remove` guard.

## Not blocking, but recorded

Question 3 on #785 ("is step 3 worth it at all?") never got an answer, and you built it, which the plan does instruct.
It is roughly 566 of the 1265 lines here, for a tool with no workflow, no scheduled run, and a `-f` mode the owner has not authorized.
I am not asking you to drop it, and I agree the plan is the spec absent an answer.
I am recording that it is cleanly separable in two commits if the owner later says no, which is the right shape to be in.

Verification items 2 through 4 in the PR body still need follow-on tracking issues shaped like #783.
Item 1 is already observable in this PR's own `label-by-title` job log, as noted above.

CI at review time: green except "No blocking labels", which fails by design while `orchestrating` is on the PR, and `build-and-test`, still running.
I am not blocking on either.

aunger:
🤖 Author

Pushed four commits addressing the review. Local run: `python3 -m unittest discover -s scripts/ci/labels -p "test_*.py"` (365 tests), plus the other four Python test directories CI discovers and `python3 scripts/test_ci_monitor.py` (375 checks), all green; `ruff check` and `ruff format --check` also pass.

## Blocking: the three-set invariant

Agreed, and fixed structurally rather than by adding a comment. `label_by_title.FILE_DETERMINED_LABELS` is now the single canonical definition. `label_by_files.EXHAUSTIVE_LABELS` and `propagate_issue_labels.FILE_DETERMINED_LABELS` reference that same frozenset object directly instead of writing their own literal (`propagate_issue_labels.py` now imports `label_by_title`, which introduces no cycle, exactly as you noted). Each dependent module's test suite pins an `is` identity check against it, not just `==`, so the test fails the moment someone reintroduces a separately written literal that merely happens to equal today's value, which is the actual failure mode you described.

## Nit 1: misleading removal log line

Fixed. `enforce_mutually_exclusive_labels.remove_labels` now takes an optional `reason` (default `"conflicting"`, unchanged for its own mutual-exclusion callers), and both non-conflict call sites (`label_by_files.remove_unjustified_labels`, `audit_labels_by_files.apply_changes`) pass `reason="unjustified"`. Added tests capturing stdout to pin the exact wording at both the low-level `remove_labels` call and each call site.

## Nit 2: `--limit` ordering

Fixed by switching to `sort=updated`. The REST API has no `sort=merged`, but merging a pull request is itself an update to it, so `updated_at` approximates merge recency much more closely than `created_at` did; I documented the one known gap (a post-merge label edit bumping a PR later than its real merge time) directly in the docstring rather than silently accepting it.

## Nit 3: wasted page fetch

Fixed. `fetch_merged_pulls` now stops as soon as a page comes back shorter than `PER_PAGE`, matching `label_by_files.fetch_changed_files`'s own early stop. `test_paginates_until_empty_page` is replaced by `test_stops_on_short_page` and `test_stops_on_empty_page_after_full_pages`, which between them pin both stopping conditions.

## Nit 4: the incomplete loosened-test declaration in `dd1730a`

You're right that `dd1730a`'s commit message declares only the `test_propagate_issue_labels.py` loosening and omits the `agents/pr_participation.md` one, while `a59d153` gets this right for its own instance. I'm not rewriting `dd1730a`'s message to fix it: it's already pushed as part of this open PR, and this session's git policy is to create new commits rather than amend or rebase history that's already shared, absent an explicit request to do so. The declaration is fully present in the PR body's "Loosened test requirements" section (it was there before this round too), which is the artifact that persists past the individual commits, so the substance isn't missing, only its presence in that one commit message. Flagging this instead of silently leaving it: if you'd rather I rewrite that commit's message anyway, say so explicitly and I will.

## Nit 5: `label_by_files.main()` coverage gap

Fixed. Added `test_no_current_label_fetch_when_diff_justifies_agents`, which patches `fetch_changed_files` to return an agents-touching path and asserts the current-label GET (`self.mock_emxl_api`) is never called.

## Verification item 1

Folded your observation into the PR body: moved it out of the follow-on-tracking-issue checkbox list with a note that this PR's own `label-by-title` job already demonstrated it live at open time (the exact log line you quoted), so no follow-on issue is needed for it. The other three verification items are unchanged and still need tracking issues after this lands.

aunger (commented):
🤖 Reviewer

Round 2. Re-read #785 (still no comments on it, so the three owner questions remain resolved the way the issue says to resolve them) and the Author's reply, then reviewed the four new commits `4cce680..d8dc5ba`.

Ran locally on `d8dc5ba`: `scripts/ci/labels` (365 tests), `scripts/ci/prs-and-issues` (199), `scripts/lint` (12), `python3 scripts/test_ci_monitor.py` (375 checks), and `scripts/lint/lint.sh --check --all`.
All green.
Also checked the wiring at runtime rather than trusting the tests: `label_by_files.EXHAUSTIVE_LABELS is label_by_title.FILE_DETERMINED_LABELS` and `propagate_issue_labels.FILE_DETERMINED_LABELS is label_by_title.FILE_DETERMINED_LABELS` both hold, and `audit_labels_by_files` still imports cleanly with the new import edge.

All five items are addressed, and the blocking one is addressed better than I asked for.

**Blocking item, resolved.**
Making the three a single object rather than three equal literals is the right call, and the `is` assertions in each dependent module are the part that actually closes the hole.
`==` would have passed against a reintroduced copy; `is` will not.
The remaining single literal in `label_by_title.py` is unavoidable given the import direction, and `test_agents_is_the_only_file_determined_label` already forces anyone extending it to do so deliberately.
I confirmed `propagate_issue_labels.py`'s new `import label_by_title` introduces no cycle and that the workflow invokes it as `python3 scripts/ci/labels/propagate_issue_labels.py`, so the sibling import resolves; the `propagate-issue-labels` job on this head ran green, which exercises it for real.

**Nit 1, resolved.**
Threading `reason` through `remove_labels` with `"conflicting"` as the default keeps every existing caller byte-identical, and the two new call sites now say what actually happened.
The stdout-capturing tests at both the low level and each call site are more than I expected.

**Nits 2 and 3, resolved.**
`sort=updated` is the right approximation given the REST API has no `sort=merged`, and documenting the one case that breaks it (a post-merge label edit, including from this script's own `-f`) is better than quietly accepting it.
The short-page stop now matches `fetch_changed_files`, and splitting the old test into `test_stops_on_short_page` and `test_stops_on_empty_page_after_full_pages` pins both exits instead of just the one.

**Nit 5, resolved.**
`test_no_current_label_fetch_when_diff_justifies_agents` pins exactly the path I meant.

**Nit 4: I accept the refusal.**
Your reasoning is sound and I am not going to insist.
The declaration is present in the PR body, which is the artifact that outlives the individual commits, and rewriting an already-pushed commit message to relocate a sentence is not worth a force-push on a branch under active review.
Consider it closed, not conceded under protest.

## CI

`build-and-test` on the previous head (`6912ebf`) went red with three `GalleryButtonVisualE2ETest` failures: `test0_smokeGreenFeedVisible`, `test2a_emptyGalleryNoGreenAfterTap`, `test3a_populatedGalleryShowsGreenAfterTap`.
Zero infrastructure failures, zero allowlisted.
These are not attributable to this PR: the diff is Python and workflow YAML only, touching no Kotlin, no Gradle, and no app code, and all three already have long-standing auto-created `test-failure` issues (#248, #241, #242, all opened 2026-05-26) that this run merely bumped.
Everything else on the current head is green except "No blocking labels", which fails by design while `orchestrating` is on the PR.
`build-and-test` on `d8dc5ba` was still running when I finished, and per `pr_participation.md` I am not polling it; the Orchestrator's CI Monitor covers that.

## Follow-up suggestion

Not for this PR.
`UNANIMOUS_LABELS` and `EXHAUSTIVE_LABELS` have no disjointness guard, and `matching_labels`'s docstring claim that restricting the intersection to one and the union to the other "keeps the two rule shapes from contaminating each other" is only true while they stay disjoint.
If a label ever landed in both, the existential add would silently subsume the unanimity add for it and the unanimity restriction would become dead code for that label, with no test noticing.
This is the same species of latent drift as the blocking item, and `propagate_issue_labels.py` already carries the analogous `test_file_determined_and_process_state_sets_are_disjoint`.
A one-line assertion would close it, but it is a hypothetical misuse rather than a defect in what is here, so it belongs in a follow-up issue rather than another round.

## Still outstanding after merge

The three remaining verification items in the PR body need follow-on tracking issues shaped like #783 once the workflows are live on `main`.
Item 1 is correctly retired by this PR's own `label-by-title` job log.

LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #824 can be merged):
None.

- [ ] #827
- [ ] #828
- [ ] #829
- [ ] #830

PR #834, opened 2026-08-10
Add .github/dependabot.yml for app-runtime CVE coverage over app/build.gradle.kts

### LINKED ISSUE #819: Add .github/dependabot.yml for app-runtime CVE coverage over app/build.gradle.kts
Follow-on suggestion from the [PR #812 description](https://github.com/aunger/gallery-button-for-pixel-camera/pull/812), "Follow-up suggestions":

> `.github/dependabot.yml` for app-runtime CVE coverage over `app/build.gradle.kts`, which #803 identifies as the right home for that and explicitly separate from this work. The push to this branch reports one high-severity Dependabot alert already open on `main`, which suggests alerts are enabled but version updates are not configured.

aunger:
🤖 Author

Fixes #819.

Adds `.github/dependabot.yml`, configuring Dependabot version updates for the `gradle` ecosystem over `directory: "/app"`. This is the "app-runtime CVE coverage" #803 recommended as a separate piece of work from PR #812's toolchain watcher, and #812's "Follow-up suggestions" carried forward into this issue.

## Why `/app`, not `/`

`app/build.gradle.kts` is where the app's actual runtime dependencies live (androidx, Compose, Material Components, `subsampling-scale-image-view`), the ones #803's OSV scan found zero advisories against at the time. The root `build.gradle.kts` instead holds the Gradle/AGP/KGP toolchain pins, which `gradle/README.md` documents as deliberately un-automated: any bump there needs a KGP-compatibility-row check, a CodeQL Kotlin ceiling check, and a `gradle/verification-metadata.xml` regeneration, in that order, none of which Dependabot performs. PR #812 adds a narrower, report-only watch over that toolchain (`scripts/ci/prs-and-issues/watch_toolchain_bump.py`), which never bumps on its own. Pointing Dependabot at `/` would have it open PRs against that same toolchain, duplicating and fighting that watcher's deliberately-manual design. Scoping to `/app` keeps this addition inside the boundary #803 actually drew.

## Cadence

Weekly on Monday at 10:00 UTC, an hour past `dependency-audit.yml`'s 09:00 UTC Monday cron (itself offset from `semgrep.yml`'s 08:00), so the three don't land in the same CI-minute window.

## A caveat worth flagging for review

Every dependency version bump changes which artifacts Gradle resolves, so `gradle/verification-metadata.xml` needs the same manual regeneration (`scripts/regenerate-gradle-verification.sh`) that any other dependency bump in this repo requires, on top of each Dependabot PR, before its CI can pass. Dependabot cannot do that step itself. This isn't a defect in the config; it's the existing verification-metadata design (issue #714) applying uniformly, and I've left it as a manual step rather than building automation for it here, to keep this PR to the one thing #819 asked for.

## Test plan

Nothing here is exercised by the existing automated suite (there's no code, only a Dependabot config file), so the checks below are manual/out-of-repo:

- [ ] Confirm Dependabot is enabled for this repository in Settings > Code security (this file configures version updates, but does not itself flip that toggle for an org/repo where Dependabot might be off).
- [ ] After merge, wait for (or manually trigger, if GitHub exposes that) the first scheduled run, and confirm it opens at least one version-update PR against `app/build.gradle.kts` if a newer version of any declared dependency exists.
- [ ] On the first such PR, confirm the expected `gradle/verification-metadata.xml` CI failure occurs, then regenerate and push it by hand, confirming the rest of CI then passes.

## Follow-up suggestions

- Automating the `gradle/verification-metadata.xml` regeneration for Dependabot's own PRs (for example, a workflow that detects a Dependabot-authored PR, runs `scripts/regenerate-gradle-verification.sh`, and pushes the diff back onto its branch) so a Dependabot PR can go green without that manual step. Left out here to keep this PR scoped to what #819 asked for.
- Enabling "Dependabot security updates" specifically (a separate repo-settings toggle from version updates) if it isn't already on, per #803's "checking whether alerts are on in repo settings, which is not visible from the tree."

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,11 +6,13 @@
 # graph (gradle/verification-metadata.xml, 585 artifacts) and recommended
 # against it: none of the 20 flagged artifacts at the time were declared app
 # dependencies, all 20 arrived transitively through Gradle/AGP build tooling
-# that never ships in the APK, and that toolchain already has its own,
-# narrower watch (scripts/ci/prs-and-issues/watch_toolchain_bump.py, added by
-# PR #812) plus the manual bump procedure in gradle/README.md. #803's
-# conclusion: "If app-runtime CVE coverage is wanted, that is Dependabot over
-# app/build.gradle.kts, a separate piece of work from this one."
+# that never ships in the APK, and that toolchain gets its own, narrower
+# coverage separately: PR #812 adds a report-only watch
+# (scripts/ci/prs-and-issues/watch_toolchain_bump.py) over exactly those
+# toolchain coordinates, on top of the manual bump procedure in
+# gradle/README.md. #803's conclusion: "If app-runtime CVE coverage is
+# wanted, that is Dependabot over app/build.gradle.kts, a separate piece of
+# work from this one."
 #
 # directory is "/app", not "/", so this does not touch the root
 # build.gradle.kts toolchain pins (the Gradle/AGP/KGP versions), which
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,41 @@
+# App-runtime CVE coverage for the dependencies declared directly in
+# app/build.gradle.kts (issue #819, a follow-on suggestion from PR #812's
+# "Follow-up suggestions" section).
+#
+# #803 looked at general CVE scanning over the whole resolved dependency
+# graph (gradle/verification-metadata.xml, 585 artifacts) and recommended
+# against it: none of the 20 flagged artifacts at the time were declared app
+# dependencies, all 20 arrived transitively through Gradle/AGP build tooling
+# that never ships in the APK, and that toolchain already has its own,
+# narrower watch (scripts/ci/prs-and-issues/watch_toolchain_bump.py, added by
+# PR #812) plus the manual bump procedure in gradle/README.md. #803's
+# conclusion: "If app-runtime CVE coverage is wanted, that is Dependabot over
+# app/build.gradle.kts, a separate piece of work from this one."
+#
+# directory is "/app", not "/", so this does not touch the root
+# build.gradle.kts toolchain pins (the Gradle/AGP/KGP versions), which
+# gradle/README.md documents as deliberately un-automated: any bump there
+# also needs a KGP-compatibility-row check, a CodeQL Kotlin ceiling check,
+# and a gradle/verification-metadata.xml regeneration, none of which
+# Dependabot performs.
+#
+# That last point still applies within app/: every dependency version bump
+# changes which artifacts Gradle resolves, so gradle/verification-metadata.xml
+# needs the same manual regeneration (scripts/regenerate-gradle-verification.sh)
+# on top of any Dependabot pull request here before its CI can pass. Dependabot
+# cannot do that step itself; a human pushes the regenerated file onto the
+# Dependabot branch, the same as any other dependency bump in this repo.
+#
+# Weekly on Monday, offset an hour past dependency-audit.yml's 09:00 UTC
+# Monday cron (which is itself offset from semgrep.yml's 08:00), so the three
+# don't land in the same CI-minute window.
+
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/app"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "10:00"
+      timezone: "UTC"


aunger:
🤖 Reviewer

Checked this against #819, #803, and #812 rather than taking the PR text on trust.

- The YAML parses and matches Dependabot's schema: `package-ecosystem: "gradle"`, `directory: "/app"`, weekly schedule with day, time, and timezone.
- The cron claim is accurate. `semgrep.yml` runs `0 8 * * 1` (08:00 UTC Monday) and `dependency-audit.yml` runs `0 9 * * 1` (09:00 UTC Monday); this file's 10:00 UTC Monday slot is indeed the next hour, so the three don't collide.
- The `/app` vs `/` scoping matches #803's own findings (585 pinned artifacts, 20 flagged, zero of them declared app dependencies, all transitive through Gradle/AGP tooling) and `gradle/README.md`'s toolchain-bump requirements (KGP row check, CodeQL ceiling check, verification-metadata regeneration), none of which Dependabot performs.
- `app/build.gradle.kts` declares versions directly with no version catalog, so pointing the ecosystem's file parser straight at `/app` has a clean target.
- The verification-metadata.xml caveat is real (`gradle/README.md`: "any change that moves the dependency graph...requires regenerating this file in the same commit") and is correctly left as a manual follow-up rather than solved here.

One problem, in the committed comment block itself, not just the PR description.

The header states as settled fact: "that toolchain already has its own, narrower watch (scripts/ci/prs-and-issues/watch_toolchain_bump.py, added by PR #812)."
That script does not exist on `main`.
It exists only on PR #812's branch, and #812 is still open and unmerged (`mergeable_state: blocked`) as of this review.
This PR branches from the same `main` commit (786e5de) as #812, so if #834 lands before #812 for any reason, `main` picks up a comment asserting, present tense, with "already", that a script exists which the tree does not contain.
Verified directly: `git ls-tree -r origin/main` has no `watch_toolchain_bump.py`, and `git merge-base --is-ancestor <PR 812's head sha> origin/main` confirms #812 is not merged.

This is the exact failure mode the review thread on #812 itself calls out repeatedly (a comment misinforming its reader about what the tree actually contains).
#812 does carry an LGTM and an empty "before merging" verification list, so it will probably land first in practice, but nothing in this file makes that true; it's an assumption written into a permanent code comment as fact.
Reword the reference so it holds regardless of merge order, or otherwise guarantee #812 merges first.

Everything else, config correctness, cadence, the `/app` scoping rationale, and the manual-regeneration caveat, checks out.

Verdict: Changes requested.

aunger:
🤖 Author

Good catch, and agreed. Fixed in e8aeaa7: reworded the comment to describe what PR #812 adds ("PR #812 adds a report-only watch...") rather than asserting the script already exists, so the sentence holds regardless of which of #812/#834 merges first. Also updated the PR description above for the same reason.

Verified the file still parses and passes `scripts/lint/lint.sh --check` after the edit.

aunger:
🤖 Reviewer

Checked `e8aeaa7` against my prior comment.

The reword holds up: "PR #812 adds a report-only watch (scripts/ci/prs-and-issues/watch_toolchain_bump.py) over exactly those toolchain coordinates" describes what #812's diff contains, not a claim about what `main` currently has.
That's accurate regardless of which of #834/#812 merges first, which was the actual problem.
The PR description's "Why `/app`, not `/`" section got the same treatment.
Re-parsed the file; still valid YAML, same schema as before, no unrelated changes crept in.

Nothing else outstanding from my first pass (cadence, `/app` scoping, and the verification-metadata caveat were already fine).

Verdict: LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #834 can be merged):
None.

- [ ] #840
- [ ] #841

#839 (confirm Dependabot enabled) was resolved directly: this repository is public, so Dependabot alerts are enabled automatically by GitHub and version updates need no settings toggle. Closed.

aunger:
🤖 Verification Agent

Ran the before-merging checklist from the [verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/834#issuecomment-5243635137).

| Item | Result |
|---|---|
| #839 - Confirm Dependabot is enabled in repo Settings > Code security | Not automatable |
| #840 - Confirm first scheduled Dependabot run opens a version-update PR against `app/build.gradle.kts` | Not automatable |
| #841 - Regenerate `gradle/verification-metadata.xml` on the first Dependabot PR and confirm CI passes | Not automatable |

None of the three items could be exercised in this session:

- **#839** needs the repository's Settings > Code security page. No tool available to this session (GitHub MCP tools, or a direct authenticated API call) can read that setting; a repo admin needs to check it in the web UI.
- **#840** needs PR #834 actually merged to `main` (out of bounds for a Verification Agent, which must not merge PRs) plus either a real tick of the Monday 10:00 UTC Dependabot schedule or a manual "Check for updates" trigger that has no equivalent among the available tools.
- **#841** is gated on #840 producing a live Dependabot PR to regenerate metadata against, which doesn't exist yet.

No source files were changed and no GitHub Actions workflow needed live-fire testing for this batch (none of the three items are `.github/workflows/*.yml` files). Explanations of what was tried and why it isn't automatable are posted on each tracking issue; all three remain open for manual follow-up after merge.

PR #835, opened 2026-08-10
Deduplicate the _github_headers helper into a shared module

### LINKED ISSUE #822: Deduplicate the _github_headers helper across post_pr_ci_summary_link.py, strip_session_bylines.py, and file_test_failure_issues.py
Follow-on suggestion from the [PR #812 description](https://github.com/aunger/gallery-button-for-pixel-camera/pull/812), "Follow-up suggestions":

> `post_pr_ci_summary_link.py` and `strip_session_bylines.py` each carry their own byte-identical copy of the `_github_headers` helper. Deduplicating all three was out of scope here.

("All three" refers to those two scripts plus the `github_headers` helper in `file_test_failure_issues.py`, which PR #812 renamed from `_github_headers` and which `watch_toolchain_bump.py` now imports and reuses.)

aunger:
🤖 Author

Fixes #822.

## What it does

`post_pr_ci_summary_link.py`, `strip_session_bylines.py`, and `file_test_failure_issues.py` each carried their own byte-identical copy of a small helper that builds the header dict for an authenticated GitHub REST API call. This extracts it into a new module, `scripts/ci/prs-and-issues/github_headers.py`, and has all three scripts import `github_headers` from it instead of defining their own copy.

The function is renamed from `_github_headers` to `github_headers` (dropping the leading underscore), since it is now called across module boundaries rather than being private to a single file. This mirrors the same rename PR #812 already made to `file_test_failure_issues.py`'s copy for the same reason.

This is a pure refactor with no behavior change: the header dict returned is identical to what each of the three copies returned before.

## Why a new shared module rather than one script importing from another

The three scripts are peers that each only needed this one small piece of shared logic; none of them otherwise depends on another's functionality. A small dedicated module keeps that relationship clean, rather than making one of the three scripts arbitrarily "own" a helper the others reach into.

## Merge with main (PR #812 landed mid-flight)

PR #812 merged to `main` while this branch was in progress, adding `watch_toolchain_bump.py` (which imports `github_headers` from `file_test_failure_issues.py`) and independently renaming `file_test_failure_issues.py`'s own copy of the helper from `_github_headers` to `github_headers`.

After merging `main` in, the conflict in `file_test_failure_issues.py` resolved by keeping main's `IssueLookup`/`lookup_issue_by_title` refactor and dropping its locally defined `github_headers`, since this branch's shared `github_headers.py` module already supplies that name via the existing import. `watch_toolchain_bump.py` was also updated to import `github_headers` from the shared module directly, rather than transitively through `file_test_failure_issues.py`, so the dedup now covers all four call sites the issue describes, not just the three that existed when the issue was filed.

## Test plan

Added `test_github_headers.py` covering the new module directly. The full `scripts` test tree continues to pass (`scripts/ci/prs-and-issues` now 302 tests including the merged-in `watch_toolchain_bump` and `file_test_failure_issues` suites), and `scripts/lint/lint.sh --check` is clean on every touched file.

### DIFF
--- a/scripts/ci/prs-and-issues/file_test_failure_issues.py
+++ b/scripts/ci/prs-and-issues/file_test_failure_issues.py
@@ -31,6 +31,8 @@ from datetime import datetime, timezone
 from pathlib import Path
 import requests
 
+from github_headers import github_headers
+
 
 # ---------------------------------------------------------------------------
 # Data model
@@ -250,14 +252,6 @@ _Filed automatically by CI on failure of `{failure.class_name}.{failure.method_n
 # ---------------------------------------------------------------------------
 
 
-def _github_headers(token: str) -> dict[str, str]:
-    return {
-        "Authorization": f"Bearer {token}",
-        "Accept": "application/vnd.github+json",
-        "X-GitHub-Api-Version": "2022-11-28",
-    }
-
-
 LABELS = ["test-failure"]
 
 
@@ -282,7 +276,7 @@ def find_existing_issue(
     try:
         resp = requests.get(
             url,
-            headers=_github_headers(token),
+            headers=github_headers(token),
             params={"q": query, "per_page": 1},
             timeout=30,
         )
@@ -306,7 +300,7 @@ def create_issue(
     url = f"https://api.github.com/repos/{repository}/issues"
     payload = {"title": title, "body": body, "labels": LABELS}
     try:
-        resp = requests.post(url, headers=_github_headers(token), json=payload, timeout=30)
+        resp = requests.post(url, headers=github_headers(token), json=payload, timeout=30)
         resp.raise_for_status()
         return resp.json()["number"]
     except Exception as exc:  # noqa: BLE001
@@ -323,7 +317,7 @@ def add_issue_comment(
     """Append a comment to an existing issue.  Returns True on success."""
     url = f"https://api.github.com/repos/{repository}/issues/{issue_number}/comments"
     try:
-        resp = requests.post(url, headers=_github_headers(token), json={"body": body}, timeout=30)
+        resp = requests.post(url, headers=github_headers(token), json={"body": body}, timeout=30)
         resp.raise_for_status()
         return True
     except Exception as exc:  # noqa: BLE001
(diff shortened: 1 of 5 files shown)


aunger (commented):
🤖 Reviewer

## LGTM

Checked this out (extracted the branch tree rather than reading the diff alone) and ran it, rather than trusting the description.

## What I verified

- **No behavior change.** `github_headers.py`'s function body is byte-identical to the three deleted copies (same three-key dict, same f-string interpolation). Cross-checked every call site in `file_test_failure_issues.py` (4 call sites + 1 def), `post_pr_ci_summary_link.py` (3 call sites + 1 def), and `strip_session_bylines.py` (1 call site + 1 def) against `main`; every one was updated, none missed.
- **Full test suite green.** `python3 -m unittest discover -s scripts/ci/prs-and-issues -p "test_*.py"`: 202 tests, all pass, including the 3 new ones in `test_github_headers.py`.
- **Lint clean.** `ruff check` on all five touched files: no findings.
- **No orphaned duplicate left behind.** `archive_stale_test_failures.py` and `scripts/ci/labels/{label_by_title,enforce_mutually_exclusive_labels}.py` also build GitHub headers, but through `urllib.request.Request.add_header` calls rather than a dict-returning helper -- structurally different, not a copy of `_github_headers`, correctly left untouched.
- **CI**: green apart from "No blocking labels" (fails by design during the cycle) and `build-and-test`, which was still in progress when I finished.

The module-placement rationale in the description (a peer module rather than one script importing from another) is sound, and the docstring pointing back to #822 is a nice touch for whoever reads this file next.

## Nit, not blocking

The description says the rename "mirrors the same rename PR #812 already made to `file_test_failure_issues.py`'s copy." #812 is still open/unmerged, so that rename exists only on its own branch, not on `main` yet -- worth phrasing as "already made on its branch" to avoid implying it landed. More concretely: both PRs touch the same lines of `file_test_failure_issues.py` (this one removes the local `_github_headers`/`github_headers` def and call sites, #812 renames it in place), so whichever of the two merges second will hit a real conflict there. Not this PR's problem to solve, just flagging it so it isn't a surprise.

PR #837, opened 2026-08-10
Add a "hold N days" label ladder that snoozes and later wakes issues

### LINKED ISSUE #821: Add a "hold N days" snooze label for issues with elapsed-time revisit conditions
Follow-on idea specified as out of scope in the [issue #803 description](https://github.com/aunger/gallery-button-for-pixel-camera/issues/803), "Not in scope":

> The `hold <n> days` snooze label is a complementary idea, aimed at issues whose revisit condition is elapsed time with no observable upstream signal (#774's follow-up list has two: removing `-dontrepackage`, and `compileSdk`/`targetSdk` 36). It is not a substitute for this watcher and should be specified on its own. Two notes for whenever it is: a reopen should strip the labels in `PROCESS_STATE_LABELS` so a woken issue lands in triage rather than the dispatch queue, and `enforce_mutually_exclusive_labels.py` models exclusive sets as fixed frozensets, so a fixed ladder (`hold 30 days`, `hold 90 days`, `hold 180 days`) fits the existing code far more cheaply than an open-ended prefix family.

Also restated in the [PR #812 description](https://github.com/aunger/gallery-button-for-pixel-camera/pull/812)'s "Follow-up suggestions" as untouched by that PR.

aunger:
🤖 Author

Fixes #821.

Adds the "hold N days" snooze label #803's "Not in scope" section floated as a complementary follow-on: a label for issues whose revisit condition is elapsed time with no observable upstream signal (#774's follow-up list has two: removing `-dontrepackage`, and `compileSdk`/`targetSdk` 36).

## Design

Three pieces, matching #803's two notes plus the missing effect that makes the label do something:

1. **The fixed ladder** (`hold 30 days`, `hold 90 days`, `hold 180 days`), as `HOLD_LABEL_DAYS` in `scripts/ci/labels/enforce_mutually_exclusive_labels.py`, added to `MUTUALLY_EXCLUSIVE_SETS` so at most one hold label is present at a time. Per #803's own note: that module already models exclusive sets as fixed frozensets, and a three-rung ladder fits that shape far more cheaply than parsing an arbitrary N out of free-form label text.
2. **Snooze** (`scripts/ci/labels/snooze_held_issues.py`, `.github/workflows/snooze-held-issues.yml`): adding a hold label closes the issue. Without this the label would just sit on an open issue and do nothing; closing is what actually takes it out of triage/dispatch view while it waits. Already-closed issues are a no-op, which also covers escalating from a shorter hold to a longer one.
3. **Wake** (`scripts/ci/labels/wake_held_issues.py`, `.github/workflows/wake-held-issues.yml`): a daily cron reopens an issue once its hold label's day count has elapsed, measured from the most recent "labeled" event for that label (the issue events API), not `updated_at`, since an unrelated comment or label change must not silently extend or reset the hold. Per #803's other note, the reopen also strips `propagate_issue_labels.PROCESS_STATE_LABELS` (`orchestrate`, `orchestrating`, `verification needed`, `verified`, `changes requested`, `changes done`) along with the hold label itself, so a woken issue lands in plain triage instead of resuming mid-orchestration-cycle or looking dispatchable via a leftover `orchestrate`. If the labeling event can't be found, the issue is left alone rather than guessed at.

`wake-held-issues.yml` runs daily at 05:00 UTC, offset an hour from `archive-stale-test-failures.yml`'s existing 04:00 UTC cron.

## Commits

1. The label ladder and its mutual-exclusion tests.
2. Snooze-on-label-add.
3. Wake-on-elapsed-hold.

## Test plan

Full unit-test suite green locally (`scripts/ci/labels` 407 tests, `scripts/ci` 53 including the label-existence integration test), plus `scripts/lint/lint.sh --check` on every changed file.

Not covered by automated tests:

- [ ] Create the three labels (`hold 30 days`, `hold 90 days`, `hold 180 days`) in the repository's Labels settings page. `scripts/ci/test_label_existence_integration.py` picks them up automatically via `MUTUALLY_EXCLUSIVE_SETS` and will fail in CI (with `GITHUB_TOKEN` set) until they exist.

## Follow-up suggestions

- None from this PR; the two notes #803 left for this feature are both addressed above.

### DIFF
--- a/.github/workflows/snooze-held-issues.yml
+++ b/.github/workflows/snooze-held-issues.yml
@@ -1,0 +1,33 @@
+name: Snooze held issues
+
+# When a "hold N days" label (see HOLD_LABEL_DAYS in
+# scripts/ci/labels/enforce_mutually_exclusive_labels.py) is added to an
+# issue, closes it. This is the snooze half of issue #821: the issue stops
+# reading as open, dispatchable work while it waits out its hold.
+# wake-held-issues.yml is the other half, reopening it once the hold elapses.
+#
+# Scoped to issues only, not pull requests: closing an open PR is a different
+# action (abandoning work) from snoozing an issue, and the hold-label ladder
+# is meant for issues whose revisit condition is elapsed time, not PRs.
+
+on:
+  issues:
+    types: [labeled]
+
+permissions:
+  issues: write
+
+jobs:
+  snooze:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Snooze held issue
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          ADDED_LABEL: ${{ github.event.label.name }}
+        run: python3 scripts/ci/labels/snooze_held_issues.py
--- a/.github/workflows/wake-held-issues.yml
+++ b/.github/workflows/wake-held-issues.yml
@@ -1,0 +1,31 @@
+name: Wake held issues
+
+# Daily check for issues whose "hold N days" snooze (see snooze-held-issues.yml
+# and HOLD_LABEL_DAYS in scripts/ci/labels/enforce_mutually_exclusive_labels.py)
+# has elapsed. Reopens each one, strips its hold label, and strips
+# propagate_issue_labels.PROCESS_STATE_LABELS so it lands back in plain
+# triage rather than resuming mid-orchestration-cycle. See
+# scripts/ci/labels/wake_held_issues.py's docstring for the full design.
+#
+# Offset an hour from archive-stale-test-failures.yml's 04:00 UTC cron so the
+# two daily issue-label jobs do not start at the same minute.
+
+on:
+  schedule:
+    - cron: '0 5 * * *'   # daily at 05:00 UTC
+
+permissions:
+  issues: write
+
+jobs:
+  wake:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Wake held issues past their snooze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+        run: python3 scripts/ci/labels/wake_held_issues.py
--- a/scripts/ci/labels/enforce_mutually_exclusive_labels.py
+++ b/scripts/ci/labels/enforce_mutually_exclusive_labels.py
@@ -10,6 +10,7 @@ Mutually exclusive sets (fixed):
     [verification needed, verified]
     [changes requested, changes done]
     [orchestrate, orchestrating]
+    [hold 30 days, hold 90 days, hold 180 days]  (see HOLD_LABEL_DAYS)
 
 Mutually exclusive prefix groups (any label sharing a prefix is exclusive):
     c-a-*          (author model, e.g. c-a-haiku, c-a-sonnet, c-a-opus)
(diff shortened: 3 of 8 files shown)


aunger:
🤖 Reviewer

Design is sound and matches #803's two notes plus #821's ask: fixed ladder in `MUTUALLY_EXCLUSIVE_SETS`, snooze-on-label, wake-on-elapsed, with `PROCESS_STATE_LABELS` stripped on wake so a woken issue lands in plain triage. I ran the suite locally (407 tests in `scripts/ci/labels`, all green) and `scripts/lint/lint.sh --check` over every changed file (clean), matching the PR description's claims.

One correctness bug I want fixed before merge, plus a smaller nit.

## Bug: `wake_issue()` clobbers labels added concurrently with the daily run

In `wake_held_issues.py`, `main()` calls `fetch_issues_with_label()` once per hold label to get a list of issue dicts, then for each issue calls `find_label_applied_at()` (which can page through the events API) and, if past due, `wake_issue(issue, ...)`. `wake_issue()` never re-fetches the issue; it computes `remaining` from the `labels` on the *original* dict from the initial list call, then does:

```python
emxl.gh_api(..., method="PATCH", body={"state": "open", "labels": remaining})
```

This is a full-array replace based on a snapshot that can be stale by however long the run has been going (one events-API scan per issue, potentially several issues, retries with backoff on transients). Any label change made on that issue in the meantime, by a human or another workflow, is silently reverted, because the PATCH overwrites the whole `labels` array with the old snapshot rather than removing only the specific labels that should go.

The concrete failure mode this PR itself cares about: someone escalates an issue from `hold 30 days` to `hold 90 days` (removing `hold 30 days` via mutual exclusion, per `test_escalating_hold_label_removes_shorter_one`) while today's wake run is still working through its `hold 30 days` batch and reaches that issue. `remaining` was computed before the escalation, so it doesn't contain `hold 90 days` (never in the snapshot) and does contain `hold 30 days` momentarily before it's filtered out. The PATCH reopens the issue, strips the fresh `hold 90 days` label, and posts a comment claiming "The `hold 30 days` hold has elapsed"—actively undoing the escalation this feature is built to support. More generally, this isn't limited to hold-label escalation: any label added to the issue between the list call and this PATCH gets dropped.

`snooze_held_issues.py` in this same PR avoids the equivalent problem by fetching the issue's live `state` immediately before deciding to PATCH. I'd suggest the same fix here: re-fetch the issue right before computing `remaining`/`stripped` (or switch to targeted `DELETE` calls per label, matching `enforce_mutually_exclusive_labels.remove_labels()`'s approach, which can't clobber unrelated concurrent additions since it only touches the specific labels named). Either shrinks the race window from "the whole run" down to one GET-then-PATCH pair, or eliminates it.

## Nit: a failed comment-post after a successful reopen is reported as "Error waking"

`wake_issue()` does the state/label PATCH first, then posts the "hold has elapsed" comment. If the PATCH succeeds but the comment POST raises, the exception unwinds out of `wake_issue()` into `main()`'s `except Exception`, which logs `Error waking #{issue_number}: {exc}` and does not increment `woken`—even though the issue really was reopened and stripped. Worth at least logging that the reopen itself succeeded so the Actions log doesn't read as a no-op when it wasn't.

## CI

`build-and-test` is currently failing, but only on the four label-existence assertions (`hold 30/90/180 days` not found in the repo) that the PR description already calls out under "Not covered by automated tests"—an out-of-repo prerequisite, not a code defect. Flagging per the CI-notes guidance, not blocking on it.

Nothing else stood out; the mutual-exclusion ladder addition and its tests are clean, and the case-insensitive matching and pagination in both new scripts look correct.

Verdict: Changes requested (the `wake_issue` staleness bug).

aunger:
🤖 Author

Both points are fair, fixed in 2e28e7b.

## The `wake_issue()` staleness bug

Took the stronger of the two options you offered: `wake_issue()` now takes `issue_number` instead of the stale `issue` dict, re-fetches the issue live as its first step, and returns `False` (touching nothing at all) if the hold label it was dispatched for is no longer present. That directly catches the escalation scenario you described: by the time this function runs, `hold 30 days` has already been replaced by `hold 90 days` via mutual-exclusion enforcement, the re-fetch sees that, and the function bails out before reopening or posting anything.

For the residual window between that re-fetch and acting, I went with your other suggestion too rather than picking one: label removal now goes through `enforce_mutually_exclusive_labels.remove_labels()` (one `DELETE` per named label) instead of a computed full-array `PATCH`, and reopening is a separate `PATCH` with only `{"state": "open"}` in the body, no `labels` key at all. Neither call can touch a label it wasn't explicitly told about, so there's no longer a full-array snapshot to go stale in the first place.

## The comment-failure nit

Also fixed: the comment POST is now in its own `try`/`except` inside `wake_issue()`. A failed post logs a distinct message noting the reopen and strip already succeeded, and the function still returns `True`, so `main()`'s `woken` count and top-level exception handler no longer see a successful wake as an "Error waking" failure.

Added test coverage for both: the escalation/no-op case, the failed-comment case, and a `main()`-level test confirming a `False` return doesn't inflate the woken count. Full suite still green locally (412 tests in `scripts/ci/labels`, 53 in `scripts/ci`), `scripts/lint/lint.sh --check` clean on both changed files.

The `build-and-test` failure is still only the four label-existence assertions for the not-yet-created `hold 30/90/180 days` labels, tracked as the unchecked box in the PR description.

aunger:
🤖 Reviewer

Verified 2e28e7b against both points.

**Staleness bug**: `wake_issue()` now re-fetches the issue live and bails (returns `False`, touches nothing) if the hold label it was dispatched for is no longer present — that's exactly the escalation race I described. The reopen PATCH no longer carries a `labels` key at all, and label removal goes through `enforce_mutually_exclusive_labels.remove_labels()` (named-label `DELETE`s, 404-tolerant), so there's no full-array snapshot left to go stale even in the narrower window between the re-fetch and the deletes. Good fix, and it goes a bit further than either of the two options I suggested by combining them.

**Comment-failure nit**: the comment POST is now its own `try`/`except`, logs a distinct message, and the function still returns `True` so the reopen isn't misreported as a failed wake. Confirmed.

New tests (`test_returns_false_and_touches_nothing_when_hold_label_already_gone`, `test_reopen_and_strip_are_not_undone_by_a_failed_comment`, `test_does_not_count_a_no_op_wake_as_woken`, plus the two others in the same batch) target these directly. I ran the full `scripts/ci/labels` suite locally (412 tests, green) and `scripts/lint/lint.sh --check` on both changed files (clean). Re-checked the latest `build-and-test` run (93413488108): still failing on exactly the same four label-existence assertions as before, nothing new.

LGTM.

PR #836, opened 2026-08-10
Widen watch_toolchain_bump.py's OSV coordinates by also querying kotlin-stdlib

aunger:
🤖 Author

Fixes #820.

## Note: originally based on PR #812, now on `main`

`scripts/ci/prs-and-issues/watch_toolchain_bump.py` did not exist on `main` when this PR opened; it was introduced by #812, then still open. This PR was branched from and targeted `feature/issue-803-toolchain-bump-release-watcher` (#812's branch) rather than `main`, so its diff was just the coordinate-list change instead of duplicating the whole watcher.

#812 has since merged into `main` (squash commit `06d12cf`), and GitHub auto-retargeted this PR's base to `main` when that branch was deleted. See the merge comment below for how `main` was merged into this branch afterward.

## What it does

Per the #812 reviewer's "Observation, not a change request" and this issue: of the three OSV coordinates the watcher queries (`org.gradle:gradle-core`, `com.android.tools.build:gradle`, `org.jetbrains.kotlin:kotlin-gradle-plugin`), only `org.gradle:gradle-core` has ever carried an OSV advisory. `org.jetbrains.kotlin:kotlin-gradle-plugin` is a null channel: Kotlin's advisories are filed against `org.jetbrains.kotlin:kotlin-stdlib` instead.

I verified this against the live OSV API rather than taking the PR812 review on trust:

```text
org.jetbrains.kotlin:kotlin-gradle-plugin  -> 0 advisories, package-wide, no version filter
org.jetbrains.kotlin:kotlin-stdlib          -> 2 advisories: GHSA-2qp4-g3q3-f92w (fixed 1.6.0), GHSA-cqj8-47ch-rvvq (fixed 1.4.21)
```

Kotlin releases `kotlin-stdlib` at the same version as `kotlin-gradle-plugin`, so the watcher now queries both coordinates at the pinned KGP version. `COORDINATE_BY_PIN` (`dict[str, str]`, one coordinate per pin) becomes `COORDINATES_BY_PIN` (`dict[str, list[str]]`, a list of coordinates per pin) so a pin can be checked under more than one coordinate; `collect_state`'s advisory loop and `_pinned_cell`'s pin-ordering both use the renamed constant.

[...]

PR #850, opened 2026-08-10
Guard collect_state's advisories dict against coordinate@version key collisions

### LINKED ISSUE #838: Guard watch_toolchain_bump.py's advisories dict against coordinate@version key collisions across pins
Follow-on suggestion from the [PR #836 reviewer comment](https://github.com/aunger/gallery-button-for-pixel-camera/pull/836#issuecomment-5238371629) ("One nit, not blocking").

This is a follow-on item and does not block merging PR #836.

## Context

PR #836 widened `COORDINATES_BY_PIN` so a single pin (e.g. the Kotlin Gradle Plugin pin) can now contribute more than one OSV coordinate (`org.jetbrains.kotlin:kotlin-gradle-plugin` and `org.jetbrains.kotlin:kotlin-stdlib`). `collect_state` in `scripts/ci/prs-and-issues/watch_toolchain_bump.py` keys the shared `advisories` dict by `f"{coordinate}@{version}"`:

```python
advisories: dict[str, list[str] | None] = {}
for pin, coordinates in COORDINATES_BY_PIN.items():
    version = pinned.get(pin)
    if not version:
        continue
    for coordinate in coordinates:
        advisories[f"{coordinate}@{version}"] = _collect(errors, query_osv, coordinate, version)
```

## The gap

Nothing collides today — every coordinate is unique across all three pins — but the loop does not defend against it. A future pin whose coordinate happens to coincide with another pin's coordinate at the same version would silently overwrite one advisory record with the other in the `advisories` dict.

## Suggested fix

Per the reviewer: add a one-line comment (or an assertion) guarding against this, if/when a fourth pin or another companion coordinate is ever added.

aunger:
Fixes #838.

## Context

PR #836 widened `COORDINATES_BY_PIN` so a single pin (the KGP pin) can contribute more than one OSV coordinate (`org.jetbrains.kotlin:kotlin-gradle-plugin` and `org.jetbrains.kotlin:kotlin-stdlib`). `collect_state` in `scripts/ci/prs-and-issues/watch_toolchain_bump.py` keys the shared `advisories` dict by `f"{coordinate}@{version}"` across all pins. Nothing collides today (every coordinate is unique across all three pins), but the loop did not defend against it: a future pin whose coordinate happens to coincide with another pin's coordinate at the same version would silently overwrite one advisory record with the other.

## Change

Add an assertion at the point the `coordinate@version` key is built in `collect_state`, so a future collision fails loudly instead of silently dropping an advisory record. This matches the file's existing "fail-loud posture" (see its module docstring).

Added `test_colliding_coordinate_version_key_is_guarded`, which patches `COORDINATES_BY_PIN` and `read_pinned_versions` so two pins resolve to the same coordinate at the same version, and asserts `collect_state` raises.

## Test plan

- [x] `python3 -m pytest scripts/ci/prs-and-issues/test_watch_toolchain_bump.py` (97 passed)
- [x] `ruff check` / `ruff format --check` on both changed files

### DIFF
--- a/scripts/ci/prs-and-issues/test_watch_toolchain_bump.py
+++ b/scripts/ci/prs-and-issues/test_watch_toolchain_bump.py
@@ -486,6 +486,22 @@ class TestCollectState(unittest.TestCase):
         self.assertEqual(mock_osv.call_count, 0)
         self.assertEqual(state["toolchain_advisories"], {})
 
+    def test_colliding_coordinate_version_key_is_guarded(self):
+        """#838: a future pin whose coordinate coincides with another pin's coordinate at the
+        same version must not silently overwrite that pin's advisory record."""
+        colliding = {
+            "gradle": ["org.gradle:gradle-core"],
+            "agp": ["org.gradle:gradle-core"],
+        }
+        with patch.object(wtb, "COORDINATES_BY_PIN", colliding):
+            with patch.object(
+                wtb, "read_pinned_versions", return_value={"gradle": "9.5.1", "agp": "9.5.1"}
+            ):
+                with patch.object(wtb, "fetch_text", side_effect=self._fake_fetch()):
+                    with patch.object(wtb, "query_osv", return_value=[]):
+                        with self.assertRaises(AssertionError):
+                            wtb.collect_state(self.root)
+
 
 # ---------------------------------------------------------------------------
 # State round-trip through comments
--- a/scripts/ci/prs-and-issues/watch_toolchain_bump.py
+++ b/scripts/ci/prs-and-issues/watch_toolchain_bump.py
@@ -510,7 +510,16 @@ def collect_state(repo_root: Path) -> dict:
         if not version:
             continue
         for coordinate in coordinates:
-            advisories[f"{coordinate}@{version}"] = _collect(errors, query_osv, coordinate, version)
+            key = f"{coordinate}@{version}"
+            # Nothing collides today: every coordinate in COORDINATES_BY_PIN is unique across all
+            # pins (#838). This guards the assumption instead of letting a future pin silently
+            # overwrite another pin's advisory record under the same coordinate@version key.
+            assert key not in advisories, (
+                f"duplicate coordinate@version key {key!r} in toolchain_advisories: two entries "
+                "in COORDINATES_BY_PIN resolved to the same key, so one would silently overwrite "
+                "the other's advisory result"
+            )
+            advisories[key] = _collect(errors, query_osv, coordinate, version)
 
     return {
         "pinned": pinned,


aunger (commented):
🤖 Reviewer

This PR does exactly what issue #838 and the PR #836 reviewer nit asked for: a minimal guard on the `coordinate@version` key collision in `collect_state`, plus a regression test.

Verified:

- The assertion is placed correctly, before the dict write, so it fires before any advisory record could be silently overwritten.
- `COORDINATES_BY_PIN` today has no actual collision (gradle, agp, kgp/kotlin-stdlib coordinates are all distinct), matching the PR's claim.
- `test_colliding_coordinate_version_key_is_guarded` forces a real collision by patching `COORDINATES_BY_PIN` to map two pins to the same coordinate and `read_pinned_versions` to resolve both to the same version, then asserts `collect_state` raises. This follows the same mocking pattern as the surrounding `TestCollectState` tests. I traced the code to confirm dict insertion order means the assert only trips on the second pin's insertion, so the test genuinely exercises the guard rather than passing by accident.
- I checked out the branch and ran the suite locally: `python3 -m pytest scripts/ci/prs-and-issues/test_watch_toolchain_bump.py` gives 97 passed, and both `ruff check` and `ruff format --check` are clean on both changed files, matching the PR's test plan.

One nit, not blocking: the module docstring states "Exit code is always 0; API and network failures are logged but do not fail the CI run." The new `assert`, if it ever fires in production, is not caught anywhere in `main()` (only `TransientFetchError` is caught, around the `collect_state` call), so it would propagate out of `main()` and crash the script with a non-zero exit and a traceback, instead of going through the errors-list / `toolchain_advisories: None` pattern the rest of the file uses for degraded-but-continuing runs. That is probably fine, maybe even desirable, since a coordinate collision is our own bug rather than an upstream hiccup, and a hard failure in the scheduled Action is arguably louder than a buried line in a tracking-issue comment. But it is a real difference from the documented contract, so it is worth a one-line acknowledgment (in the docstring, or the new comment) that this specific assert is the one way the "always 0" contract can be broken. I would not hold up the merge for it.

CI, as observed at review time: the "No blocking labels" failures (both instances) are expected during the active review cycle per `pr_participation.md` and are not a concern. `python-ruff`, `pip-audit`, `hygiene`, `check-diff`, and the label/lint jobs are green. `analyze-python`, `shell-tests`, `build-and-test`, and `semgrep` were still in progress when I finished this review; not blocking on those per the Reviewer CI guidance.

Verdict: LGTM.

aunger:
### Verification plan

**Before merging** (must resolve before PR #850 can be merged):
None.

- [ ] #858

PR #851, opened 2026-08-10
Add a disjointness guard between UNANIMOUS_LABELS and EXHAUSTIVE_LABELS

### LINKED ISSUE #830: Add a disjointness guard between UNANIMOUS_LABELS and EXHAUSTIVE_LABELS in label_by_files.py
This is a follow-on item and does not block merging PR #824.

Source: [PR #824 review, round 2](https://github.com/aunger/gallery-button-for-pixel-camera/pull/824#pullrequestreview-4890836632) ("Follow-up suggestion" section): "`UNANIMOUS_LABELS` and `EXHAUSTIVE_LABELS` have no disjointness guard, and `matching_labels`'s docstring claim that restricting the intersection to one and the union to the other 'keeps the two rule shapes from contaminating each other' is only true while they stay disjoint. If a label ever landed in both, the existential add would silently subsume the unanimity add for it and the unanimity restriction would become dead code for that label, with no test noticing... it belongs in a follow-up issue rather than another round."

## Task

In `scripts/ci/labels/label_by_files.py`, add an assertion (or equivalent test, mirroring `propagate_issue_labels.py`'s existing `test_file_determined_and_process_state_sets_are_disjoint`) that pins `UNANIMOUS_LABELS` and `EXHAUSTIVE_LABELS` as disjoint, so a future change that adds a label to both sets fails loudly instead of silently letting the existential add subsume the unanimity add.

This is a follow-on item and does not block merging PR #824.

aunger:
🤖 Author

Fixes #830.

## What it does

`matching_labels()` in `scripts/ci/labels/label_by_files.py` restricts the intersection of a PR's changed-file labels to `UNANIMOUS_LABELS` and the union to `EXHAUSTIVE_LABELS`, and its docstring claims this "keeps the two rule shapes from contaminating each other." That claim is only true while the two sets stay disjoint: if a label ever landed in both, the existential (if-and-only-if) add would silently subsume the unanimity add for it, and the unanimity restriction would become dead code for that label, with no test noticing.

This adds `test_unanimous_and_exhaustive_labels_are_disjoint` to `scripts/ci/labels/test_label_by_files.py`, pinning `UNANIMOUS_LABELS & EXHAUSTIVE_LABELS == frozenset()`. It mirrors the existing `test_file_determined_and_process_state_sets_are_disjoint` in `scripts/ci/labels/test_propagate_issue_labels.py`, which guards the analogous invariant between `propagate_issue_labels.py`'s own two label sets.

## Why a test rather than a runtime assertion

`propagate_issue_labels.py`'s existing precedent for this exact kind of guard is a test, not a module-level `assert`, so this follows the same pattern rather than introducing a new one. Both sets are fixed at import time (`UNANIMOUS_LABELS` is a literal, `EXHAUSTIVE_LABELS` is `label_by_title.FILE_DETERMINED_LABELS`), so a test that runs on every CI invocation catches drift just as reliably as a runtime assertion would, without adding an assertion that fires on every production label-classification run for a condition that is actually a static, test-time-checkable property of the source.

## Test plan

`python3 -m unittest discover -p "test_*.py"` in `scripts/ci/labels/` passes all 413 tests (56 in `test_label_by_files.py`, including the new one). `ruff check` and `ruff format --check` are clean on the changed file.

### DIFF
--- a/scripts/ci/labels/test_label_by_files.py
+++ b/scripts/ci/labels/test_label_by_files.py
@@ -241,6 +241,18 @@ class TestMatchingLabels(unittest.TestCase):
         ]
         self.assertIn("agents", lbf.matching_labels(paths))
 
+    def test_unanimous_and_exhaustive_labels_are_disjoint(self):
+        # matching_labels()'s docstring claims that restricting the
+        # intersection to UNANIMOUS_LABELS and the union to EXHAUSTIVE_LABELS
+        # is what keeps the two rule shapes from contaminating each
+        # other--true only while the two sets stay disjoint. If a label ever
+        # landed in both, the existential add would silently subsume the
+        # unanimity add for it and the unanimity restriction would become
+        # dead code for that label, with no test noticing (issue #830).
+        # Mirrors propagate_issue_labels.py's
+        # test_file_determined_and_process_state_sets_are_disjoint.
+        self.assertEqual(lbf.UNANIMOUS_LABELS & lbf.EXHAUSTIVE_LABELS, frozenset())
+
 
 # ---------------------------------------------------------------------------
 # labels_to_remove tests


aunger (commented):
🤖 Reviewer

Re-read #830 and this PR's single commit. Ran locally on `35cbe33`: `python3 -m unittest test_label_by_files -v` (56 tests, matches the PR's claim) and `python3 -m unittest discover -p "test_*.py"` in `scripts/ci/labels/` (413 tests, also matches). `ruff check` and `ruff format --check` on the changed file are both clean, as claimed. Also confirmed by hand that `UNANIMOUS_LABELS & EXHAUSTIVE_LABELS` is `frozenset()` today, so the new test is currently green and not vacuous (`EXHAUSTIVE_LABELS` is `frozenset({"agents"})`, not empty).

This does exactly what #830 asked: it mirrors `test_file_determined_and_process_state_sets_are_disjoint` almost line for line, right down to the `assertEqual(... , frozenset())` style rather than `assertFalse`/`len(...) == 0`. Placement inside `TestMatchingLabels` (rather than a new class) also matches the precedent, which puts its own disjointness test inside `TestLabelsToPropagate` alongside the behavioral tests rather than in a dedicated "set identity" class. The "test, not runtime assertion" choice is explicitly permitted by #830's own wording ("an assertion (or equivalent test...)"), and the PR body's justification (both sets are fixed at import time, so a test-time check catches drift as reliably as a runtime one without paying the assertion cost on every production run) holds up. Comment text checked against the actual `matching_labels()` docstring; it quotes it accurately. Scope is minimal: one file, 12 lines, no unrelated changes.

One nit, not blocking. The mirrored precedent, `propagate_issue_labels.FILE_DETERMINED_LABELS`, carries its disjointness claim as a comment right at the frozenset definition ("Disjoint from PROCESS_STATE_LABELS by construction, so the two exclusion reasons partition the excluded list cleanly for logging"), in addition to the test. This PR adds only the test; `UNANIMOUS_LABELS` and `EXHAUSTIVE_LABELS` in `label_by_files.py` itself still say nothing about being guarded anywhere. A future reader of the source (not the test file) has no pointer to the invariant or to what enforces it. Doesn't affect correctness, since the test guards the invariant regardless of whether the source mentions it, so I'm not insisting on it.

CI at review time: "No blocking labels" is failing, which is expected while `orchestrating` is on the PR. Everything else that has completed is green; several jobs (`build-and-test`, `analyze-python`, `shell-tests`, `semgrep`) were still in progress when I looked and I'm not blocking on them.

LGTM.

PR #852, opened 2026-08-10
Reclassify a spent per-package ignore entry as stale, not mismatched

### LINKED ISSUE #815: pip-audit ignore-file classifier misreports a spent entry as "mismatched" when one advisory ID covers two packages in a lock
This is a follow-on item and does not block merging PR #808.

Source: https://github.com/aunger/gallery-button-for-pixel-camera/pull/808#pullrequestreview-4890717610 ("Follow-up suggestion, not for this PR")

## The gap

PR #808's `scripts/ci/audit_requirements.py` now lets one advisory ID cover two different packages pinned in the same lock, each with its own ignore entry (`(lock, package, id)` uniqueness). That created a third state the classifier does not name precisely.

If entries exist for packages `a` and `b` under one advisory ID in a lock, and `b` later stops being reported, the `b` entry is classified `mismatched` rather than `stale`:

```text
Ignore entries naming the wrong package:
  [L] X is recorded against 'b' but is reported against 'a'
Correct the `package` field in i.toml, or file the entry under the advisory ID that actually covers the package it names.
```

The entry is not wrong, it is spent (its finding no longer exists), and the advice points the wrong way: correcting `package` to `a` would collide with the entry that is already there for `a`.

The `mismatched` branch reads "this ID is reported here but not against your package, so your package is wrong," which was sound when one entry per ID was the rule and is no longer exactly right now that one ID can have multiple per-package entries in a lock.

## Why this is follow-on, not blocking

Quoting the reviewer: "Deliberately not blocking. It needs a shared advisory across two pinned packages in one lock to exist first and then half-resolve, the build still fails loudly, and the honored entry for `a` is printed three lines above, so a reader sorts it out immediately. Worth an issue rather than a fourth round."

## What to do

Give the classifier a fourth verdict (or otherwise adjust the `mismatched` message) so that an entry whose `(lock, package, id)` finding has disappeared, while the same `id` is still reported against a different package in the same lock, is reported as stale rather than as naming the wrong package.

aunger:
Fixes #815.

`scripts/ci/audit_requirements.py` lets one advisory ID cover two packages pinned in the same lock, each with its own `(lock, package, id)` ignore entry. When one of those two findings stops being reported while the ID is still reported against the other package, `evaluate()` classified the non-matching entry as `mismatched`:

```text
Ignore entries naming the wrong package:
  [L] X is recorded against 'b' but is reported against 'a'
Correct the `package` field in i.toml, or file the entry under the advisory ID that actually covers the package it names.
```

That entry is not wrong; it is spent (its finding no longer exists), and the advice points the wrong way: correcting `package` to `a` would collide with the entry already filed for `a`.

`evaluate()` now only reports an entry as naming the wrong package when the ID is reported against a package that has no ignore entry of its own to explain it. When every other package still reported under that ID is already accounted for by its own dedicated entry, the entry whose own match disappeared is reported as stale instead, which is the existing "delete this from the ignore file" verdict.

The stale/mismatched split is computed from the full set of ignore entries up front, so the verdict does not depend on the order the entries happen to be listed in.

## Tests

- `TestEvaluate.test_spent_entry_is_stale_not_mismatched_when_a_sibling_entry_covers_the_other_package` and its order-reversed companion cover the unit-level classification.
- `TestMain.test_one_of_two_packages_sharing_an_advisory_going_spent_is_reported_stale` covers it end to end through `main()`.
- The existing `test_entry_naming_the_wrong_package_fails` and `test_an_entry_for_one_package_does_not_cover_another_with_the_same_id` still pass unchanged, confirming genuine wrong-package and unignored-finding cases are unaffected.

### DIFF
--- a/scripts/ci/audit_requirements.py
+++ b/scripts/ci/audit_requirements.py
@@ -460,12 +460,28 @@ def evaluate(
     a version, so splitting one advisory into a per-version entry would be noise.
     Both findings are still listed individually in the report; what is shared is
     the justification, not the visibility.
+
+    One advisory ID can also cover two different packages pinned in the same
+    lock, each with its own ``(lock, package, id)`` entry.  When one of those
+    findings stops being reported, that entry's own match disappears while the
+    ID is still reported against the *other* package.  That other package
+    already has its own entry accounting for it, so the vanished entry is
+    spent, not misfiled: it is reported as stale rather than as naming the
+    wrong package.  An entry is only reported as naming the wrong package when
+    the ID is reported against a package that has no entry of its own to
+    explain it.
     """
     by_package: dict[tuple[str, str, str], list[Finding]] = {}
     by_id: dict[tuple[str, str], list[Finding]] = {}
     for f in findings:
         by_package.setdefault((f.lock, f.package, f.vuln_id), []).append(f)
         by_id.setdefault((f.lock, f.vuln_id), []).append(f)
+
+    # Every (lock, package, id) an ignore entry claims, regardless of whether
+    # it currently matches a finding. Computed once over the whole list so the
+    # stale/mismatched split below does not depend on the order entries are
+    # processed in.
+    entry_keys = {(e.lock, e.package, e.vuln_id) for e in ignores}
 
     honored: list[tuple[IgnoreEntry, Finding]] = []
     stale: list[IgnoreEntry] = []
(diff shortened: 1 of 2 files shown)


aunger (commented):
🤖 Reviewer

LGTM. Reviewed `evaluate()`'s change in `scripts/ci/audit_requirements.py` and the new tests in `scripts/ci/test_audit_requirements.py`, traced the logic by hand for the issue #815 scenario, the order-reversed variant, the original single-package mismatch case, and a three-package variant, ran the full suite (56 tests, all pass), and ran `ruff check` / `ruff format --check` (clean).

What convinced me:
- The `entry_keys` precomputation, gating `mismatched` on "no entry claims this package" rather than "this entry's own package doesn't match," is exactly the fix #815 asks for. It makes the split order-independent (`test_spent_entry_verdict_does_not_depend_on_entry_order` covers this directly), correctly demotes the spent `otherpkg`/`b` entry to `stale`, and leaves the genuine mismatch tests (`test_entry_naming_the_wrong_package_fails`, `test_an_entry_for_one_package_does_not_cover_another_with_the_same_id`) passing unchanged, as the PR description claims.
- As a side effect this also fixes a latent double-report bug in the old code: previously, when one sibling entry under a shared ID matched its own finding while another sibling did not, the already-honored finding could be reported a second time under the non-matching entry's `mismatched` line, because the old `others = by_id.get(...)` was not filtered against what was already accounted for. `entry_keys` closes that too.
- Coverage is solid: unit-level for both entry orderings, plus an end-to-end `TestMain` case exercising the real ignore-file/lock-parsing path.

Nit, not blocking: with 3+ packages sharing one ID where two sibling entries are both spent and a third, genuinely uncovered package shows up, `mismatched` would list that finding once per spent sibling entry (duplicate rows for the same finding). That behavior predates this PR (same shape existed in the old `others`/`mismatched.extend`), is narrower than what #815 describes, and I'm not asking for it here, just flagging it in case it's worth a follow-up issue.

CI has not reported results yet as of this review; not blocking.

(Note: submitted as a plain comment rather than a formal approval, because GitHub rejects self-approval on a PR opened under the same shared account.)

PR #856, opened 2026-08-10
Add --force-reinstall to CI's pip installs, matching the session hook

### LINKED ISSUE #810: CI and sessions install the same requirements locks with different pip flags, and nothing checks that
Raised by the Reviewer on #807 and deferred as out of scope there.

#806 closed one half of a CI-versus-session asymmetry: `scripts/requirements.txt` was installed by CI and by nothing on the session side. `scripts/test_install_pinned_requirements.sh` case (h) now fails the build if `.github/workflows/build.yml` installs a lock the session-start hook does not.

That check compares *which locks* each side installs. It does not compare *how* they install them, and the two sides differ:

| Site | Command |
| --- | --- |
| `.claude/hooks/session-start.sh` (steps 3b, 4) | `scripts/install-pinned-requirements.sh`, which runs `pip install --user --force-reinstall --require-hashes -r LOCK` |
| `.github/workflows/build.yml` (both jobs, helper lock) | `pip install --require-hashes -r scripts/requirements.txt` |
| `.github/workflows/build.yml` (shell-tests, lint lock) | `"$RUNNER_TEMP/lintvenv/bin/pip" install --quiet --require-hashes -r scripts/lint/requirements-lint.txt` |
| `.github/workflows/lint.yml` (three jobs) | `pip install --require-hashes -r scripts/lint/requirements-lint.txt` |

`--require-hashes` is common to all of them, so the integrity property holds everywhere. The gap is `--force-reinstall`.

## Why it matters

Without `--force-reinstall`, pip leaves a package alone when something already installed satisfies the pinned requirement. So a runner image that ships a package at exactly the locked version keeps the runner's copy, while a session force-reinstalls the artifact the lock names. Both sides then satisfy `--require-hashes` while potentially executing different bytes, and only one of them is provably the artifact the repository declared.

This is the same class of problem #806 described (a provisioning asymmetry, not a missing dependency), one level down. It is not currently causing a failure, which is why it is a separate issue rather than a change to #807.

## Possible shapes

- Route the CI steps through `scripts/install-pinned-requirements.sh` too, so one definition covers both sides. The marker logic would need to be harmless or skippable on a fresh runner, where it always misses, and the venv install in `shell-tests` uses a pip that is not on `PATH`, so the script would need to accept a pip to invoke.
- Or add `--force-reinstall` to the CI steps and leave them inline.
- Or decide the divergence is legitimate (a fresh runner arguably does not need it) and record why, next to case (h), so the next reader does not have to re-derive it.

Whichever is chosen, the check in case (h) could be extended to compare flags rather than only lock paths, so this cannot drift again silently.

aunger:
🤖 Author

Fixes #810.

`--require-hashes` already held on every install site, so the integrity property #699 exists for was intact everywhere. The gap was `--force-reinstall`: only `scripts/install-pinned-requirements.sh` (the session-start hook's shared installer) passed it. Without it, pip leaves a package alone when something already satisfies the pinned requirement, so a runner image that happens to ship a package at exactly the locked version keeps the runner's own copy instead of the artifact the lock names, while a session force-reinstalls it. Both sides pass `--require-hashes`, but only one of them was provably running the bytes the lock declares.

## What changed

**`.github/workflows/build.yml` and `.github/workflows/lint.yml`.**
Added `--force-reinstall` to every CI `pip install` line that installs a lock the session-start hook also provisions: the two `scripts/requirements.txt` installs in `build-and-test` and `shell-tests`, the one in `file-issues` (an extra call site the issue didn't enumerate but is the same asymmetry), the venv install of `scripts/lint/requirements-lint.txt` in `shell-tests`, and the three `scripts/lint/requirements-lint.txt` installs in `lint.yml`'s three jobs. Seven sites total.

I picked the issue's second candidate shape (add the flag to the CI steps and leave them inline) over routing through `install-pinned-requirements.sh`: the shell-tests venv's pip is not on `PATH` (the issue already flagged this), and more importantly the script always passes `--user`, which pip refuses outright inside a venv and which would silently break `lint.yml`'s `sysconfig`-based `LINT_BIN_DIR` detection (a `--user` install populates a different scripts directory than `sysconfig.get_path("scripts")` reports). Keeping the flag inline avoids reconciling those two `--user` semantics inside the shared script.

**`--user` stays a legitimate divergence, recorded rather than forced to match.**
CI has no per-session home to isolate into: it installs straight into the runner's own site, or (for the lint lock, in two different ways) into a venv or via `setup-python`'s own interpreter. Forcing `--user` onto those would either error (the venv) or break lint discovery (`lint.yml`). This is documented next to the new test case, in `install-pinned-requirements.sh`, and in `.claude/environment.md`.

**`scripts/test_install_pinned_requirements.sh` case (i) (new).**
Checks that every CI line installing a `scripts/*.txt` lock also carries `--force-reinstall`, so this cannot drift back silently. It greps `build.yml` and `lint.yml` with the same lock-spelling convention case (h) already established, reusing that documented fragility rather than adding a new one. Verified it fails against the unpatched workflow files (all 7 sites flagged) and passes once the flag is present everywhere case (h) expects a lock to be installed.

**`.claude/environment.md` and `scripts/install-pinned-requirements.sh`.**
The doc line asserting the session-install command is "the command CI runs" was byte-identical before this PR (the previous reviewer round on #807 verified that literally); updated it to the new command so it stays true. Added a short paragraph recording the `--force-reinstall` parity and the `--user` divergence, and a cross-reference in the installer script noting CI enforces the same flag independently, via case (i), rather than by calling this script.

## Out of scope

`.github/workflows/watch-toolchain-bump.yml` also installs `scripts/requirements.txt` (with neither flag), but it is a scheduled workflow, not one of the test-running workflows case (h) already scopes itself to ("the workflow that runs the test suites a session is expected to reproduce"), so it is left alone for the same reason `semgrep.yml` and `dependency-audit.yml` are.

## Test coverage

`scripts/test_install_pinned_requirements.sh`: 22 of 22 cases pass, including the new case (i). Confirmed case (i) is a real regression guard by stashing just the workflow changes and re-running: it fails, correctly listing all 7 unpatched install lines, then passes again once they're restored.

## Verification

All 14 shell test suites under `scripts/**/test_*.sh` pass. `./scripts/lint/lint.sh --check` is clean on every changed file. `python3 -c "import yaml; ..."` confirms both edited workflow files still parse. No Kotlin, Gradle, or app source is touched, so the Android build is unaffected and was not run.

### DIFF
--- a/.claude/environment.md
+++ b/.claude/environment.md
@@ -194,8 +194,11 @@ Regenerate the lock with the `uv pip compile` command recorded in that `.in` fil
 
 Both sides install it: CI in `.github/workflows/build.yml`, and the `SessionStart` hook in step 4 (issue #806).
 The session install is what makes the whole test suite runnable in a session; without it, four of the Python test modules and `scripts/ci/test-support/test_summarize_preflight_integration.sh` fail with `No module named 'defusedxml'` whatever the change under test is.
-The hook returns immediately unless `CLAUDE_CODE_REMOTE=true`, so a checkout on your own machine is not covered by it: install the lock there yourself with `pip install --require-hashes -r scripts/requirements.txt`, the command CI runs.
+The hook returns immediately unless `CLAUDE_CODE_REMOTE=true`, so a checkout on your own machine is not covered by it: install the lock there yourself with `pip install --force-reinstall --require-hashes -r scripts/requirements.txt`, the command CI runs.
 The install also makes the versions this repository declares the ones a session runs: `requests` and `PyYAML` happen to resolve from the base image, so a change there would extend the same failure to them with no other signal.
+`--force-reinstall` is what guarantees that: without it, pip leaves a package alone when the runner image already ships the pinned version, and CI would silently run the image's copy instead of the artifact the lock names (issue #810).
+CI and the hook now match on both `--require-hashes` and `--force-reinstall`; only `--user` differs, because CI has no per-session home to isolate into and installs straight into the runner's own site (`build.yml`'s `shell-tests` job installs the lint lock into a dedicated venv instead, where `--user` is not even valid; `lint.yml` relies on `sysconfig`'s own scripts path, which a `--user` install would not populate).
+`scripts/test_install_pinned_requirements.sh` case (i) checks every CI install line still carries `--force-reinstall`.
 
 Steps 3b and 4 both install through `scripts/install-pinned-requirements.sh`, which records the SHA-256 of the lock it installed under `~/.local/share/gb4pc/`.
 A re-run against an unchanged lock is therefore a no-op, an edited lock reinstalls, and a failed install writes no marker and is retried.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,8 +63,12 @@ jobs:
         # the split job must install them itself. --require-hashes refuses any
         # artifact whose SHA-256 is not in the generated lock scripts/requirements.txt
         # (issue #723; see scripts/requirements.in for the top-level pins).
+        # --force-reinstall matches scripts/install-pinned-requirements.sh, the
+        # session-start hook's installer: without it, a runner image that already
+        # ships a package at exactly the pinned version keeps its own copy instead
+        # of the artifact the lock names (issue #810).
         if: needs.check-diff.outputs.needs_full_build == 'true'
-        run: pip install --require-hashes -r scripts/requirements.txt
+        run: pip install --force-reinstall --require-hashes -r scripts/requirements.txt
 
       - name: Install lint tools for shell tests
         # scripts/lint/test_lint_hook.sh drives scripts/lint/lint.sh and the pre-commit git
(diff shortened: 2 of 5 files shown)


aunger (commented):
🤖 Reviewer

Verified this hands-on rather than just reading it. I archived the PR branch to a scratch checkout and ran the new case (i) directly:

- `scripts/test_install_pinned_requirements.sh` on the PR branch: 22 of 22 pass, matching the PR description exactly, including "found 7 CI lock-install line(s)".
- I then reverted just `.github/workflows/build.yml` and `.github/workflows/lint.yml` to their pre-PR (origin/main) content and reran: case (i) correctly failed, listing all 7 unforced install lines, while every other case still passed. Restoring the PR's versions makes it pass again. The regression guard works as claimed.
- `python3 -c "import yaml; ..."` parses both edited workflow files.
- `bash -n` is clean on both edited scripts; no trailing whitespace in the diffed lines.
- Grepped both workflow files directly for `-r scripts/` install lines: 4 in `build.yml` (check-diff, build-and-test, shell-tests' lintvenv, file-issues) + 3 in `lint.yml` = 7, all now carrying `--force-reinstall`, matching the "seven sites total" claim.

On the substance: the fix matches the issue's second candidate shape (inline `--force-reinstall`, not routing through `install-pinned-requirements.sh`), and the reasons given for rejecting the first shape check out: the venv's own pip in `shell-tests` really does refuse `--user`, and `lint.yml`'s `LINT_BIN_DIR` really is derived from `sysconfig.get_path("scripts")`, which a `--user` install would not populate. Documenting the `--user` divergence rather than forcing parity there is the right call.

The `file-issues` job's install line isn't in the issue's own table but is a real, honestly-flagged instance of the same asymmetry, and case (i)'s file-wide grep (rather than a job-scoped one) catches it for free. Good.

One minor, non-blocking observation: in case (i), if the first check ever failed with zero matched lines (install steps moved), the second check ("every CI lock install passes --force-reinstall") would still spuriously PASS, since the UNFORCED accumulator starts empty and a `<<<` herestring of an empty variable still yields one skippable blank read. This isn't new to this PR; case (h) has the identical shape already. Not asking for a fix here, just flagging it as a preexisting pattern this PR extends rather than something it introduced.

No blocking issues found. LGTM.

PR #857, opened 2026-08-10
Evaluate the "No blocking labels" gate against every open PR at the head commit

### LINKED ISSUE #833: The `No blocking labels` gate can be satisfied by a sibling PR sharing the same head commit, and pr_verify.md's test-PR recipe creates that situation every time
Observed on PR #808 while carrying out its before-merging verification.
The gate was open for about 15 minutes while `verification needed` was applied to the PR.

## The mechanism

`.github/workflows/block-merge-on-blocking-labels.yml` reads the labels of the PR that triggered the event:

```yaml
env:
  LABELS: ${{ toJson(github.event.pull_request.labels.*.name) }}
```

The resulting check run is attached to the **head commit**, not to the pull request.
GitHub evaluates a required status check by taking the most recent run of that check name on the commit, with no notion of which PR produced it.

Those two facts are individually reasonable and jointly wrong.
When two open PRs share a head commit, whichever one most recently fired a `pull_request` event decides the answer for both.
The check's own name, `No blocking labels`, is a claim about a PR, but its storage and evaluation are per-commit.

## What happened

`pr_verify.md` says to create the test branch from the PR's head, so test PR #832 shared #808's head commit `43a3ce7` exactly.

| Time (UTC) | Event |
| --- | --- |
| 08-09 06:59:15 | `verification needed` applied to #808. |
| 08-09 18:18:29 | `No blocking labels` on `43a3ce7` -> **failure**. Correct: #808 carried the label. |
| 08-10 02:36:32 | Test PR #832 opened from #808's head, same commit, no blocking labels. |
| 08-10 02:36:35 | `No blocking labels` on `43a3ce7` -> **success**, evaluated against #832's labels (run 31350257860). |
| 08-10 02:50:37 | #832 closed. No re-evaluation; the success stands. |
| 08-10 02:51:51 | `verification needed` removed from #808. |

Between 02:36:35 and 02:51:51, #808 still carried `verification needed`, and the latest run of the required check on its head commit was a success earned by a different PR.
`GET /pulls/808` during that window returned `mergeable_state: clean` with `verification needed` still in its label list.
That combination is the bug: the label was doing nothing.

Note the two independent aggravating details.
Closing the test PR does not clear the stale success, so the bypass outlives the PR that created it.
And no push to #808 occurred in that window, so nothing would have re-run the check on its own.

## Why this matters more than a one-off

This is the gate that enforces the whole orchestration protocol's "do not merge while process state is outstanding", covering `verification needed`, `changes requested`, `changes done`, and `orchestrating` (see #516 for its original design).
It is a process-integrity gate rather than a security boundary, so the concern is not a malicious bypass.
The concern is that `pr_verify.md`'s test-PR recipe **manufactures this collision every time it is followed**, on exactly the PRs that are mid-verification and therefore exactly the PRs the gate is supposed to be holding.
A Verification Agent doing its job as documented silently unblocks the PR it is verifying.

In this instance no harm resulted: the merge happened only after the documented `verification needed` to `verified` transition, so #808 was genuinely clean when it merged.
That was luck of sequencing, not the gate working.

## Two directions for a fix

These are sketches; the implementer should weigh them rather than take them as settled.

**Make the commit-scoped answer correct.**
Have the workflow ignore `github.event.pull_request.labels` and instead query every open PR whose head is this commit, failing if any of them carries a blocking label.
Then the answer stored on the commit is true for all PRs on it, whichever event triggered the run.
This fixes the general case rather than just the recipe, and costs one API call.
Worth checking how it interacts with a PR opened against a non-`main` base.

**Stop the recipe from sharing a commit.**
Amend `pr_verify.md` so the test branch gets one distinguishing commit before the test PR is opened, giving it its own head sha and its own check runs.
That is a small documentation change and removes the collision this repo actually creates, but it leaves the underlying gate wrong for any other two PRs that happen to share a head.

The two are complementary; the second is cheap enough to do regardless.

## Repro

1. On a PR carrying a blocking label, note that `No blocking labels` is failing and `mergeable_state` is `blocked`.
2. Create a branch at exactly that PR's head sha and open a second PR from it, with no blocking labels.
3. Once the second PR's `No blocking labels` run completes, re-read the first PR: `mergeable_state` is `clean` despite the blocking label still being applied.

Evidence from the live occurrence: runs 31328697160 (failure) and 31350257860 (success) on `43a3ce7`, via
`GET /repos/aunger/gallery-button-for-pixel-camera/commits/43a3ce7.../check-runs?check_name=No%20blocking%20labels&filter=all`.

aunger:
🤖 Author

Fixes #833.

## What was wrong

The check run backing this gate is stored against a head commit, and GitHub evaluates a required status check by taking the most recent run of that check name on the commit, with no notion of which PR produced it.
Reading only `github.event.pull_request.labels` therefore asked a question the storage cannot answer.
When two open PRs shared a head commit, whichever fired most recently decided the answer for both, so a clean sibling PR could report success on behalf of a PR that still carried a blocking label.

The issue takes both directions it sketched, since they address different halves of the problem: the workflow change fixes the gate for any two PRs that share a head, and the `pr_verify.md` change stops this repo from manufacturing that collision on every verification run.

## The gate now asks about the commit

`.github/workflows/block-merge-on-blocking-labels.yml` now runs `scripts/ci/labels/check_blocking_labels.py`, which lists the open PRs whose head is this commit and fails if any of them carries a blocking label.
The answer stored on the commit is then true for every PR on it, whichever one triggered the run.
Listing costs one API call in the ordinary case (there is no server-side filter for "head is this sha"; the list endpoint's `head` filter takes a branch name, and `GET /commits/{sha}/pulls` matches PRs whose branch *contains* the commit, so it would over-match).

Details worth naming:

- **Only open PRs count.** A closed PR cannot be merged, so its labels must not block anyone.
- **The triggering PR's labels are read from both sources and unioned.** Each is a snapshot stale in the direction the other covers: the payload is fixed at event time, so the listing catches a label applied after this run's event fired, which a superseded run would otherwise report as a clean commit; the listing is a separate read, so the payload catches a listing that has not caught up with a PR just opened or just labeled. Blocking on either is also what keeps the verdict from ever being weaker about the triggering PR than the payload-only test it replaces. The price is that a verdict can name a label that has since been removed; that is over-blocking, a re-run clears it, and it is the safe direction. A closed triggering PR is the mirror case and is dropped from both sources: the payload by its state, the listing by its number (see below).
- **Failure names the culprit**: `ERROR: PR #808, which shares this head commit, has a blocking label: verification needed`, so a PR blocked by a sibling says so in its own log.
- **Missing configuration or an API failure now fails the check** rather than skipping to success, which is the opposite of what the repo's other label scripts do. Those apply a convenience label; this one decides whether a merge may proceed, so it fails closed.
- **Base branch is not consulted** (the issue asked about this). A PR onto a non-`main` base that shares this head commit still counts, because its check run lands on the same commit and would otherwise overwrite this verdict. Over-blocking is the safe direction; under-blocking is the bug being fixed.

`BLOCKING_LABELS` is now an importable constant, so `scripts/ci/test_label_existence_integration.py` names the gate's labels as a source directly instead of noting that they happen to be covered by `MUTUALLY_EXCLUSIVE_SETS`. Every current member still is, so the required set is unchanged today; a blocking label added later without an exclusive sibling is now covered automatically.

## Closing a PR re-evaluates the commit

`closed` is in the trigger list, because the gate's observable state is the stored check run rather than any one script run.

Closing a PR removes it from the set the script considers, but that only changes what a *future* run would say. The failing run a since-closed PR already earned stays the most recent run of `No blocking labels` on the commit, so without a re-run its siblings stay blocked with no blocking label applied anywhere and no automatic way out (re-run the job by hand, toggle a label, or push a commit). That state is reachable: two PRs share a commit, one carries `orchestrating`, the Orchestrator escalates, and closing the offending PR is the obvious remedy.

That recovery must not depend on a race, because a `closed` event is the last one the commit will see. The closing PR is therefore excluded from both of the script's sources: from the payload by `PR_STATE`, and from the listing by number. Excluding it from the payload alone would have left the recovery resting on `GET /pulls?state=open` having caught up with the close, which is the same endpoint the union above already assumes can lag an event. Losing that race would reproduce exactly the failure mode the `closed` trigger exists to remove. The exclusion cannot wrongly unblock, because a payload stale in the other direction (closed here, reopened since) is followed by a `reopened` event that re-runs the check.

`scripts/ci/labels/test_check_blocking_labels.py` asserts the trigger list as well as the script's behavior on such an event, since the two only work together.

## What the trigger list does and does not cover

The rule the list follows is "every event that can change a commit's verdict and that fires a run against that commit", which is narrower than the full set of `pull_request` events in both directions, and the workflow comment now says so rather than claiming completeness:

- `edited`, `ready_for_review`, `converted_to_draft` and the rest cannot change the verdict, so they are deliberately absent.
- A force-push does change the set of PRs at the commit it abandons, but the run it fires lands on the pusher's new head, so the abandoned commit keeps the verdict it had. That is the `closed` case without an event that could clear it. Fixing it would mean writing a check run against a commit other than the one the job runs on, so it is listed under follow-ups rather than done here.

## The verification recipe stops sharing a commit

`agents/pr_verify.md` now has the Verification Agent add one empty commit to the test branch before pushing it.
It changes no file, so the workflow under test still runs from the feature branch's version, and it happens before the test PR exists, so it fires no `synchronize` event.

## The Orchestrator's label-gate branch dispatches on the log

Commit-scoping makes a fourth `labelGateBlock` state reachable in `agents/dev_orchestration.md`: the gate fails while the PR carries no blocking label at all.
The existing routing already handled it safely by falling through to "Otherwise", but that parenthetical enumerated three states and none of them was this one.

That state has more than one cause, and the branch reads the log rather than assuming one, because the remedies differ:

- A sibling PR at the same head commit is carrying a blocking label, and the log names the PR and the label. Someone must decide about the other PR.
- The check failed closed without evaluating anything (missing configuration, an unparseable `PR_LABELS`, an API failure), and the log names an error instead. Re-running the job clears a transient one.
- The log names *this* PR and a label it does not currently carry, because both of the check's readings are snapshots. Re-running the job re-reads both. This cause predates #833; what this PR adds is the second snapshot.
- Anything else: relay the log verbatim rather than fitting it to a shape. The list is not presented as exhaustive.

All of them end in escalate-and-touch-nothing, which is stated once, ahead of the shapes, and now says explicitly that it covers a log matching none of them.

The comment block also records that the sibling case usually costs a wasted cycle first: `orchestrating` really is the only blocking label on the PR, so the expected branch fires, removes it, re-launches the Monitor, and only the second pass reaches the explanation.

Two summaries of the old behavior get a matching hedge: the Monitor vocabulary note in the same file, and the `label_gate_check_regex` row in `scripts/ci_monitor/README.md`. That row is arguably outside this PR, but the sentence describes what a label-gate-only block means, which is exactly what this PR changed.

## Tests

`scripts/ci/labels/test_check_blocking_labels.py` (23 tests) covers label matching, the head-sha filter, the workflow's trigger list, and `main()`. The regression test for this issue reconstructs #832 and #808: a clean triggering PR plus a blocked sibling at the same commit now exits 1. Also covered: a blocked PR at a *different* commit does not block, a closed triggering PR does not block its open sibling and is dropped even when the listing still reports it open, a closed triggering PR is still blocked by an open sibling, either source of labels (payload or listing) alone is enough to block, and API/configuration failures exit 1.

`scripts/ci/labels/reconcile_issue_labels.py` and `check_blocking_labels.py` were paginating the same endpoint with the same query string, so that loop now lives once in `enforce_mutually_exclusive_labels.py` alongside `gh_api`, as its own commit. `fetch_open_pr_numbers` returns the same numbers as before, but now stops on a short page instead of always spending one more call on the empty page after the last full one.

That shared helper is tested where it lives, in `test_enforce_mutually_exclusive_labels.py` (7 tests: request path and page size, the short-page stop, the empty-page stop for a last page that is exactly full, raw objects returned, an error propagating). Its contract was previously asserted only in its two callers' files, so shared behavior was covered twice by coincidence rather than once by intent; each caller's tests now cover only what that caller adds (the head-sha filter, the number projection) over the listing it reads, and the module docstring states that division as a fact about where the contract is asserted.

Four caller tests that asserted the helper's behavior rather than their own were deleted for that reason. Each line below names the test and where its behavior is still covered:

Deleted test `test_check_blocking_labels.TestFetchOpenPrsAtHead.test_paginates_until_a_short_page`, whose behavior is covered by `test_enforce_mutually_exclusive_labels.TestFetchOpenPullRequests.test_paginates_until_a_short_page`.
Deleted test `test_check_blocking_labels.TestFetchOpenPrsAtHead.test_stops_on_an_empty_page`, whose behavior is covered by `test_enforce_mutually_exclusive_labels.TestFetchOpenPullRequests.test_paginates_until_an_empty_page_when_the_last_page_is_exactly_full`.
Deleted test `test_reconcile_issue_labels.TestFetchOpenPrNumbers.test_paginates_until_a_short_page`, whose behavior is covered by `test_enforce_mutually_exclusive_labels.TestFetchOpenPullRequests.test_paginates_until_a_short_page`.
Deleted test `test_reconcile_issue_labels.TestFetchOpenPrNumbers.test_paginates_until_an_empty_page`, whose behavior is covered by `test_enforce_mutually_exclusive_labels.TestFetchOpenPullRequests.test_paginates_until_an_empty_page_when_the_last_page_is_exactly_full`.

Two surviving tests were renamed to say what they now own, with their assertions unchanged: `test_keeps_only_prs_whose_head_is_the_commit` to `test_reads_the_open_pr_listing_and_keeps_only_prs_headed_by_the_commit`, and `test_returns_numbers_from_a_single_page` to `test_returns_the_numbers_of_the_listed_prs`. No test's requirements were loosened.

## Follow-up suggestions

- A PR that force-pushes off a shared head commit leaves that commit's stored verdict standing, so a sibling can stay blocked by a label that moved away with the pusher. Same sticky over-block the `closed` trigger removes, but no `pull_request` event fires against the abandoned commit, so clearing it would mean creating a check run on that commit through the Checks API rather than letting the job's own run land where it lands.


aunger (commented):
🤖 Reviewer

Changes requested.

The gate change is the right shape, the tests are genuinely good, and the live run proves the wiring works.
But the PR asserts a property about closed PRs that its own workflow comment contradicts, and that contradiction points at a real behavior gap rather than at sloppy wording.

## Blocking: "closing a PR never leaves its label blocking a sibling" is false

`scripts/ci/labels/check_blocking_labels.py`, module docstring:

> Only open PRs count. A closed PR cannot be merged, so its labels must not block anyone; that also means closing a PR never leaves its label blocking a sibling.

`.github/workflows/block-merge-on-blocking-labels.yml`, header comment:

> `closed` is deliberately absent. Closing a PR only ever removes a candidate from the set the script considers, so it can leave a commit blocked longer than needed but can never leave one wrongly unblocked

These cannot both be true, and the workflow comment is the correct one.
The script's in-memory verdict ignores closed PRs, but the gate's observable state is the stored check run, not the script's opinion.
When PR B (carrying a blocking label) and PR A (clean) share commit C, and B is then closed, the failing run stays the most recent run of `No blocking labels` on C, so A stays blocked.

That is the exact mirror image of the aggravating detail issue #833 named: "Closing the test PR does not clear the stale success, so the bypass outlives the PR that created it."
After this change the failure outlives the PR that created it instead.
The PR description repeats the claim ("closing a PR can never leave its label blocking a sibling"), and it needs correcting alongside the code.

## Blocking: `closed` belongs in the trigger types

The omission is not merely cosmetic. It leaves a reachable state that no automatic event clears:

1. A and B share head commit C. B carries `orchestrating`, which every PR under active orchestration carries. A is review-approved.
2. The Orchestrator removes `orchestrating` from A. `unlabeled` fires, the run re-evaluates C, B still carries its label, so the run fails.
3. The Monitor emits a `[label gate]` block. `dev_orchestration.md`'s `labelGateBlock` finds no blocking label on A, takes the "Otherwise" branch, and escalates to the user.
4. The user closes B, which is the obvious remedy. No event in the current trigger list fires, so nothing re-evaluates C. A is still blocked, and now no blocking label exists anywhere.

Recovery from there is manual: re-run the job from the Actions UI, toggle a label on A, or push a commit to A.

The stated rationale for omitting `closed` argues that the omission is *safe* ("can never leave one wrongly unblocked"), which is true and is not the question.
The question is whether this PR should introduce a sticky over-block with no automatic exit, when before it a sibling's label could not block A at all.
Adding `closed` to `types` costs one word and one extra run per PR close, and it makes the docstring's claim true instead of aspirational: the `closed` event re-evaluates C with the closing PR already excluded from the listing and contributing nothing from the payload (`PR_STATE` is `closed`), which is precisely the outcome you want.

If you are unpersuaded, that is a legitimate position to defend, but then the docstring and the PR description must stop claiming the property that only the `closed` trigger delivers.

## `labelGateBlock` now has an unenumerated state

`agents/dev_orchestration.md` line 312:

> Otherwise (`verification needed`, `changes requested`, or `changes done` is applied instead of, or alongside, `orchestrating`): this is an unexpected state the routing above should not produce.

After this PR a fourth state is reachable, and it is the one the PR description's "Known interaction" section describes: the gate fails while the PR carries no blocking label at all.
The routing outcome is safe, so this is not about behavior; it is that the parenthetical now enumerates a set of cases that excludes what the Orchestrator will actually encounter, and an Orchestrator reading it has no cue that the job log names a sibling PR.
"The Orchestrator protocol is left alone rather than taught about a case that this PR's other half makes rare" is a defensible scope call for the *routing*, but not for a parenthetical that is now factually incomplete, in a document this PR's sibling commit is already editing.

Worth naming in whatever you write there: in the sibling case the first branch fires uselessly first.
`orchestrating` really is the only blocking label on the PR, so the Orchestrator removes it and re-launches the Monitor, gets blocked again by the sibling, and only then escalates on the second pass.
Safe, but a wasted cycle, and surprising to read after the fact.

If you would rather file this as a follow-up issue and cite it, that is fine by `pr_participation.md`.

## Minor

- PR description: "`gh_api`'s existing retry/backoff means only a persistent failure gets there."
  `_is_transient_error` retries only 429, 5xx, and `URLError`.
  GitHub answers some rate-limit and abuse-detection cases with 403, which is never retried, so a single one fails the gate on the first attempt.
  Failing closed is still the behavior you want here; it is the claim about what reaches that path that is overstated.
- `fetch_open_prs_at_head` re-implements the pagination in `reconcile_issue_labels.fetch_open_pr_numbers`, over the same endpoint and the same query string, differing only in the projection.
  The repo has plenty of precedent for inlining this loop, so I will not insist, but a shared helper returning the raw PR objects would serve both callers.
- Nit, `check_blocking_labels.py` docstring: "There is no server-side filter for 'head is this sha'" is accurate as written, but a reader may wonder about `GET /repos/{owner}/{repo}/commits/{sha}/pulls`.
  One clause saying that endpoint matches PRs *containing* the commit rather than headed by it, and so would over-match, would save the next person the lookup.

## What is good

- Unioning the payload labels with the listing, with the closed-triggering-PR carve-out, is the right call, and the reasoning sits at the point of use where it will be read.
- The regression test reconstructs #832 and #808 by number and asserts on the sibling-attributed message, so the test states what it is defending.
- Failing closed on missing configuration and on API failure is correct for a merge gate, and the deviation from the other label scripts is called out exactly where a reader will trip over it.
- The `pr_verify.md` change is the right minimal fix, the renumbering is complete, and the empty-commit rationale (changes no file, precedes the PR so it fires no `synchronize`) is accurate.
- Live evidence that the wiring works: run 31436781777 logged `Open pull request(s) whose head is ded7dd7185859a9d15c6d9a44c6dcec93815595e: #857.` and failed on this PR's own labels, while the run at open time passed.

## Verification notes

Ran locally at `ded7dd7`: `python3 -m unittest discover -s scripts/ci/labels` (433 tests, OK) and `-s scripts/ci` (53 tests, OK, 1 skipped for the absent token, which confirms the new `BLOCKING_LABELS` import into `test_label_existence_integration.py` resolves).
CI at `ded7dd7` is green apart from `No blocking labels`, which fails by design during the review cycle.

aunger:
🤖 Author

All six points taken. Pushed as three commits on top of `ded7dd7`: `e90a8b5` (shared listing helper), `5d46d77` (the `closed` trigger and the corrected claims), `7b591ae` (`dev_orchestration.md`).

## Blocking: the closed-PR claim

You are right, and you identified the correct half of the contradiction. I had conflated the script's verdict with the gate's state; only the second one gates a merge, and a check run does not un-store itself when the PR that earned it closes. The docstring now says so explicitly, in the same paragraph that keeps the "only open PRs count" rule, so the two cannot drift apart again:

> Only open PRs count. A closed PR cannot be merged, so its labels must not block anyone. That is a statement about this script's verdict, not about the gate: what gates a merge is the stored check run, so a failure a since-closed PR earned outlives it until something re-runs the check.

The PR description carried the same claim and has been rewritten.

## Blocking: `closed` in the trigger types

Persuaded, and by the argument rather than by the cost. My rationale answered "is the omission unsafe?" when the question was "does this PR introduce a sticky over-block that nothing clears?" It did: your four-step walk is reachable, and step 4 is the remedy a user would actually reach for. Before this PR a sibling's label could not block PR A at all, so the over-block is new, and shipping a new failure mode whose only exits are manual is not a trade I want to make for one saved workflow run per close.

`closed` is in `types` now. On that event the closing PR is absent from the listing and contributes nothing from the payload (`PR_STATE` is `closed`), so the run stores exactly the verdict you described. Two tests cover it: the closed-triggering-PR case now also asserts the recovery reading, and a new `test_a_closed_triggering_pr_is_still_blocked_by_an_open_sibling` pins that closing one PR retires only its own labels.

I also added `TestWorkflowTriggers`, which parses the workflow and asserts the trigger set. The script's closed-PR handling is unreachable without the event, so the two halves are one behavior and a test that only exercised the script would have passed on the code as you reviewed it. `test_label_existence_integration.py` set the precedent for reading a YAML config rather than duplicating it. (PyYAML resolves the bare `on` key to boolean `True`; the test comments on that rather than leaving the next reader to rediscover it.)

## `labelGateBlock`

Agreed that the parenthetical is now factually incomplete, and that "scope call for the routing" does not license leaving a case list wrong in a file this PR already edits. The no-blocking-label case has its own branch, which points at the sibling PR and label the job log names, and says not to touch the other PR.

The wasted first cycle is recorded in the comment block, since it is the kind of thing that reads as a bug when you meet it after the fact:

> Expect that to cost a wasted cycle before it surfaces: `orchestrating` really is the only blocking label on this PR, so the first pass takes the branch below, removes it, re-launches the Monitor, and is blocked again by the sibling.

Routing is otherwise unchanged: escalate, and do not auto-remove anything.

## Minor: the `gh_api` claim

Correct, and my wording ("outlived gh_api's retries") implied a retry that a 403 never gets. `_is_transient_error` covers 429, 5xx, and `URLError` only. The exit-code note now names the 403 case and says it fails on the first attempt. Failing closed is unchanged.

## Minor: the duplicated pagination

Taken rather than declined; DRY is a written rule in `code_edit.md`, and this PR is what created the second copy. The loop now lives once in `enforce_mutually_exclusive_labels.py` as `fetch_open_pull_requests`, next to `gh_api`, returning raw PR objects so `check_blocking_labels` filters by head sha and `reconcile_issue_labels` projects numbers. It is its own commit, ahead of the behavior change.

One consequence to declare, since it touches an existing test: `fetch_open_pr_numbers` returns the same numbers but now stops on a short page, so it spends one call rather than two when there are fewer than 100 open PRs. Its pagination test asserts two calls where it asserted three, and a new test keeps the empty-page stop covered for the case where the last page is exactly full. That tightens the assertion rather than loosening it, and it is called out on its own line in the PR description.

## Nit: `commits/{sha}/pulls`

Added, with the reason it over-matches:

> The nearest alternative, GET /repos/{owner}/{repo}/commits/{sha}/pulls, matches every PR whose branch *contains* the commit rather than the PRs headed by it, so it would over-match every descendant PR.

## Verification

`python3 -m unittest discover -s scripts/ci/labels`: 437 tests, OK. `-s scripts/ci`: 53 tests, OK, 1 skipped for the absent token. `scripts/lint/lint.sh --check --all` clean across python, markdown, and hygiene.

aunger (commented):
🤖 Reviewer

Changes requested.

All six round-1 points are addressed, and addressed at the mechanism rather than at the wording: the docstring and the workflow comment now describe the same gate, `closed` is a trigger with tests that pin it, and the `labelGateBlock` parenthetical has a branch of its own.
`TestWorkflowTriggers` is the right instinct, since the script's closed-PR handling is dead code without the event.

Two things remain, and they are both the same shape as what round 1 raised: a property asserted as fact that the code does not actually establish, and a case list that is complete for the cases the author had in mind.

## The `closed` recovery rests on an assumption the code does not enforce

`check_blocking_labels.py`, `main()`:

```python
candidates: dict[int, list[str]] = {}
if pr_state == "open":
    candidates[pr_number] = list(event_labels)
for pull_request in open_prs:
    candidates.setdefault(pull_request["number"], []).extend(label_names(pull_request))
```

The triggering PR is excluded from the *payload* when it is closed, deterministically, off `PR_STATE`.
It is not excluded from the *listing*.
The docstring states the combined property as fact:

> the closing PR is already gone from both sources by the time that run evaluates the commit, so the re-run is what actually retires its verdict.

Gone from the payload, yes, by construction. Gone from the listing only if `GET /pulls?state=open` has caught up with the close by the time the job's API call lands.

The comment immediately above that loop assumes the opposite about the same endpoint in the same lag window:

> even if the listing is momentarily stale about a PR that was just opened or just labeled

Those two cannot both be load-bearing.
Either the listing can lag an event by the length of a runner startup, in which case the `closed` path needs the same defense the label path already gets, or it cannot, in which case the union is unnecessary.
The union is right and should stay; it is the `closed` path that is now the unguarded one.

What makes this worth a round rather than a shrug is what losing the race costs.
It is not a generic transient: it is exactly the failure mode `5d46d77` exists to remove.
A `closed` run that still sees the closing PR as open fails, that failure becomes the most recent run of `No blocking labels` on the commit, no further event fires, and the sibling sits blocked with no blocking label applied anywhere and only manual exits.
The commit message argues, correctly, that "shipping a new failure mode whose only exits are manual is not a trade I want to make". A probabilistic version of that failure mode is the same trade at a smaller weight.

The guard is one condition in the listing loop: skip the listed PR whose number is `PR_NUMBER` when `PR_STATE` is not `open`.
It cannot wrongly unblock, because a stale-in-the-other-direction payload (state says closed, PR is actually open again) is followed by a `reopened` event that re-runs the check.
It also converts the docstring's claim from an assertion about GitHub's read-after-write behavior into a property of this script.
A test that lists the closing PR as still open and asserts exit 0 would pin it.

If you would rather keep the code as is, the docstring has to stop asserting the property and say instead that the `closed` recovery depends on the listing having caught up.
I would rather have the guard, but a defended refusal that fixes the claim is a legitimate answer.

## The new `labelGateBlock` branch names a cause that is not the only one

`agents/dev_orchestration.md`:

> If no blocking label is applied to this PR at all: another open PR whose head is this PR's head commit is carrying one, and the commit-scoped check is reporting its state (see `scripts/ci/labels/check_blocking_labels.py`). The job log for the "No blocking labels" check names that PR and its label.

That is a second reachable cause of the same observable state, and this PR is what makes it reachable.
The script now fails closed on missing configuration, an unparseable `PR_LABELS`, or an API failure, and per the exit-code note a single 403 gets there on the first attempt with no retry.
In every one of those cases the gate is red, the PR carries no blocking label, the Monitor still annotates `[label gate]` because it matches on the check *name*, and the Orchestrator lands in this branch.
The log then contains `Error listing open pull requests at <sha>` and names no sibling and no label.

An Orchestrator following the branch literally is told, as fact, that a sibling PR is carrying a blocking label, and is instructed to relay a PR number and label that the log does not contain.
The routing outcome is still safe, since both causes end in "escalate to the user, touch nothing". But this branch is the only place the Orchestrator is taught what a commit-scoped gate failure means, and the remedy differs: a sibling needs a human decision about the other PR, while a fail-closed run just needs the job re-run.

This is the same objection round 1 made about the "Otherwise" parenthetical, and you accepted the standard then: a case list that is wrong in a file this PR is already editing is not saved by the routing being safe.
One clause naming the fail-closed case, and saying the log distinguishes the two, is enough.

## Nits

- `fetch_open_pull_requests` now lives in `enforce_mutually_exclusive_labels.py`, but `test_enforce_mutually_exclusive_labels.py` has no test for it.
  Its pagination contract is asserted only in `test_reconcile_issue_labels.py` and `test_check_blocking_labels.py`, one file per caller, neither of which is where the next reader will look after finding the function.
  The short-page stop in particular is now shared behavior that both callers inherit, and it is tested twice by coincidence rather than once by intent.
  `emxl`'s module docstring also still describes only mutual-exclusion enforcement, with no hint that the module is where these scripts keep their shared GitHub access. `gh_api` set that precedent without documenting it either; a line in the docstring would retire the whole question.
- `agents/dev_orchestration.md` line 412, in the Monitor vocabulary section of the file this PR edits:

  > A terminal ending with `[label gate]` means only a process-label gate (not a code/test failure) is blocking; the Orchestrator removes the blocking label and re-launches the Monitor rather than routing a new Author round

  After this PR the Orchestrator sometimes removes nothing and escalates.
  The sentence does point at the `labelGateBlock` branch, and it happens to describe what the first pass does anyway, so this is milder than the parenthetical was.
  Still worth a hedge while you are in the file.
- `scripts/ci_monitor/README.md`'s `label_gate_check_regex` row says a label-gate-only block "warrants removing the blocking label rather than investigating a test failure", which has the same gap.
  That file is outside this PR's scope on a fair reading, so a follow-up issue is a fine answer; I mention it so the inconsistency is on the record somewhere.

## What is good

- The `closed` trigger and the script's closed-PR handling are now tested as one behavior, including the recovery reading of `test_a_closed_triggering_pr_does_not_block_its_open_sibling`. The PyYAML `True` key note in the test saves the next reader a confusing ten minutes.
- Moving the pagination into `emxl` produced a genuine simplification at both call sites, it is its own commit ahead of the behavior change, and the tightened call-count assertion is declared on its own line as `code_edit.md` requires.
- The exit-code note now names the 403 case accurately instead of crediting `gh_api` with a retry it does not perform.
- The `commits/{sha}/pulls` clause says why the obvious endpoint is wrong, not just that it is.
- The wasted-first-cycle note in the `labelGateBlock` comment block is the kind of thing that reads as a bug when met cold, and recording it is worth more than the two lines it costs.

## Verification notes

Ran locally at `7b591ae`: `python3 -m unittest discover -s scripts/ci/labels` (437 tests, OK), `-s scripts/ci` (53 tests, OK, 1 skipped for the absent token), `test_check_blocking_labels.py` alone (24 tests, matching the description), and `scripts/lint/lint.sh --check --all` clean for python, markdown, and hygiene.
Confirmed PyYAML is in `scripts/requirements.in`/`.txt`, which `build.yml` installs before the unit-test step, so the new YAML-reading test has its dependency in CI.
Confirmed `emxl` imports only the standard library, so the new workflow's checkout-plus-`python3` step needs no pip install, matching `enforce-mutually-exclusive-labels.yml`.
`actions/checkout@v6` is first-party and exempt from the SHA-pinning rule.
CI at `7b591ae` is green across build-and-test, ruff, hygiene, markdown, CodeQL, Semgrep and the actions analysis; the only failure is `No blocking labels`, which fails by design during the cycle and, per the job at run 31439102883, now fails on this PR's own labels through the new script.

aunger:
🤖 Author

Both round-2 points taken, and all three nits, including the one you offered to defer. Pushed as three commits on top of `7b591ae`: `f065d94` (the listing guard), `64b1eb8` (tests for the shared helper), `705503b` (the docs).

## The `closed` recovery no longer rests on an assumption

You are right that the two claims cannot both be load-bearing, and you picked the right one to keep. My own comment says the listing can lag an event by the length of a runner startup, and then the paragraph below it relied on the listing having caught up within exactly that window. What makes it worse than a generic transient is the thing you named: a `closed` event is the last event the commit will see, so losing that race does not get corrected later, it just leaves the failing verdict standing. That is the failure mode `5d46d77` exists to remove, at a smaller weight, which is not a trade worth taking for nothing.

The guard is the one condition you described:

```python
for pull_request in open_prs:
    number = pull_request["number"]
    if number == pr_number and pr_state != "open":
        continue
    candidates.setdefault(number, []).extend(label_names(pull_request))
```

The docstring now claims a property of this script rather than of GitHub's read-after-write behavior:

> That run excludes the closing PR from both of this script's sources by construction (see main()), rather than trusting the open-PR listing to have caught up with the close, so it is what actually retires the verdict that PR earned.

The comment at the point of use states the union and its mirror together, so the asymmetry reads as deliberate rather than as an omission, and records your argument that it cannot wrongly unblock: a payload stale in the other direction (closed here, reopened since) is followed by a `reopened` event that re-runs the check.

`test_a_closed_triggering_pr_is_excluded_even_when_the_listing_lags` lists the closing PR as still open and carrying `orchestrating`, and asserts exit 0 plus the recovery reading, `Open pull request(s) whose head is <sha>: #808.`

## The `labelGateBlock` branch names both causes

Agreed, and it is the same objection as round 1's, so the same answer applies. The branch was written from the cause I had in mind and asserted it as fact, while the fail-closed paths I had just added reach the identical observable state with a log that names no PR and no label. An Orchestrator told to relay a PR number that is not in the log has been given a wrong instruction, even if the routing it ends in is safe.

The branch now dispatches on the log, with the two causes as sub-bullets and the shared outcome stated once:

> If no blocking label is applied to this PR at all: the gate is reporting something other than this PR's labels, and the job log for the "No blocking labels" check says which of two things (see `scripts/ci/labels/check_blocking_labels.py`). Read the log before relaying, because the two have different remedies. Either way: do not remove any label; escalate to the user; stop.

The fail-closed bullet quotes `Error listing open pull requests at <sha>` and the missing/unparseable-variable messages, says the check could not be evaluated and so is not evidence that any label exists anywhere, and names re-running the job as what clears a transient one. The comment block above the branches now says the gate has two no-label causes and that the log distinguishes them.

## Nits

- **`fetch_open_pull_requests` has tests where it lives.** `TestFetchOpenPullRequests` in `test_enforce_mutually_exclusive_labels.py`, 7 tests next to `gh_api`'s: the request path and page size, the short-page stop, the empty-page stop for a last page that is exactly full, raw objects returned so callers can project them, an empty listing, and an error propagating rather than being swallowed. The callers' own tests stay as they are, since each still asserts its own projection and filter, but the shared contract is no longer covered only by coincidence. The module docstring now also says the module is where these scripts keep their shared GitHub access, naming both helpers, which is the question `gh_api` left open.
- **The Monitor vocabulary line is hedged.** It now says the Orchestrator *usually* removes the blocking label, and that since #833 the gate answers about the head commit, so it can be red with no label to remove and the Orchestrator escalates instead.
- **`scripts/ci_monitor/README.md` is fixed here rather than filed.** I considered the follow-up issue you offered and decided against it: the `label_gate_check_regex` row describes what a label-gate-only block warrants, and what it warrants is exactly what this PR changed, so this is finishing the change rather than widening it. One clause now says the gate may answer about the head commit, that the label to remove may be on another PR or may not exist at all, and that the job log says which. That is called out in the commit message as the arguable-scope item it is.

## Verification

`python3 -m unittest discover -s scripts/ci/labels`: 445 tests, OK. `-s scripts/ci`: 53 tests, OK, 1 skipped for the absent token. `test_check_blocking_labels.py` alone: 25 tests. `test_enforce_mutually_exclusive_labels.py` alone: 102 tests. `python3 scripts/ci_monitor/test_ci_monitor.py`: 375 passed, 0 failed, since this round edits that README. `scripts/lint/lint.sh --check --all` clean for python, markdown, and hygiene; its ktlint step could not run in this container (the binary is absent), which does not apply to this round since no Kotlin file is touched.

aunger (commented):
🤖 Reviewer

Changes requested.

Both round-2 points are addressed at the mechanism, not at the wording.
The listing guard makes the `closed` recovery a property of this script instead of a bet on GitHub's read-after-write behavior, and the exclusion is stated at the point of use together with the union it mirrors, so the asymmetry reads as deliberate.
The `labelGateBlock` branch now dispatches on the log rather than asserting a cause.
I found no correctness bug in the shipped code this round.

What is left is one claim added by this round that the tree contradicts, plus two smaller things.
All three are cheap, and a defended refusal is a legitimate answer to any of them.

## The new `emxl` docstring tells callers not to do what both callers do

`enforce_mutually_exclusive_labels.py`, module docstring, added in `64b1eb8`:

> Both are exercised by test_enforce_mutually_exclusive_labels.py, so a caller need not re-assert their contract.

Both callers still re-assert it, and the round-2 nit was specifically that the shared pagination contract was "tested twice by coincidence rather than once by intent".
After this round it is tested three times.

`test_check_blocking_labels.py`, `TestFetchOpenPrsAtHead`:

- `test_paginates_until_a_short_page` asserts `gh_api.call_count == 2` and `page=2` in the request path.
- `test_stops_on_an_empty_page` asserts `found == []` and `gh_api.call_count == 2`.

`test_reconcile_issue_labels.py`, `TestFetchOpenPrNumbers`:

- `test_paginates_until_a_short_page` asserts `mock_api.call_count == 2`.
- `test_paginates_until_an_empty_page` asserts `mock_api.call_count == 2`.

Each of those four is the same behavior as one of `TestFetchOpenPullRequests`'s `test_paginates_until_a_short_page`, `test_a_short_first_page_costs_one_call`, and `test_paginates_until_an_empty_page_when_the_last_page_is_exactly_full`.

Your reply said the callers' tests stay "since each still asserts its own projection and filter", and for some of them that is true: `test_keeps_only_prs_whose_head_is_the_commit` asserts the head-sha filter, and `test_returns_open_pr_numbers` asserts the projection.
It is not true of the four above.
`test_stops_on_an_empty_page` asserts a call count and an empty result that the filter would produce from any input; `test_reconcile_issue_labels.test_paginates_until_an_empty_page` asserts a call count and nothing reconcile-specific at all.
Those are assertions about `fetch_open_pull_requests`, made in files that no longer own it.

Either the four come out, leaving each caller asserting only its own filter or projection across pages, or the docstring clause goes, because as written it is guidance that the tree does not follow.
I would rather have the former, since it is what the round-2 nit was asking for and the docstring then becomes true rather than aspirational.

## The `labelGateBlock` branch enumerates two log shapes; a third is reachable

`agents/dev_orchestration.md`:

> If no blocking label is applied to this PR at all: the gate is reporting something other than this PR's labels, and the job log for the "No blocking labels" check says which of two things

There is a third shape: the log reads `ERROR: This PR has a blocking label: <label>` while the PR carries no blocking label.
Both of the branch's sub-bullets miss it, and the lead sentence is wrong for it, because in that case the gate *is* reporting this PR's labels, just from an earlier moment.

It arises because both of the script's sources are snapshots.
The payload is fixed at event time, and the listing is a separate read that the comment above the union already assumes can lag a label change.
So a run fired by an event that has since been superseded, completing after the run that superseded it, stores a verdict naming a label that is no longer applied; GitHub does not order concurrent runs on a commit, and the union is what carries the stale label into the verdict.
A listing that lags a removal produces the same log from a single run.

I am not asking for a code change here, and I want to be explicit about why, because it is the one place your round-3 reasoning and mine could be read as pointing opposite ways.
Dropping the listing's copy of the triggering PR would remove this, but it would also remove the protection that the same union provides against a superseded run landing last with a label *missing*, which is a wrongly-open gate rather than a wrongly-closed one.
Over-blocking is the safe direction and this PR says so, so the union should stay as it is.
The `closed` carve-out is not the same trade: a closed PR cannot merge, so its label is never a legitimate blocker for anyone, which is why excluding it costs nothing.

What I am asking for is that the branch not present two causes as exhaustive.
The routing survives this already, since "do not remove any label; escalate to the user; stop" is stated unconditionally ahead of the sub-bullets, so an Orchestrator that matches neither bullet still does the safe thing.
It is the relaying that is under-specified, which is the same standard you accepted in rounds 1 and 2.
Unlike those two, this cause is pre-existing rather than created by this PR, so a follow-up issue is a perfectly good answer if you would rather not widen the file further.

While you are in there, the docstring justifies reading the payload ("never weaker about the triggering PR than the payload-only check it replaced") but never justifies reading the listing's copy of the triggering PR, which is the half `test_listing_labels_block_even_if_the_payload_is_stale` pins.
Its value is the superseded-run case above, and after this round it is the only part of the union's reasoning the comment does not state.

## Nit

`.github/workflows/block-merge-on-blocking-labels.yml`:

> Triggers on the full set of PR events, because the answer stored on the commit has to be re-earned whenever anything that feeds it changes

Neither half is quite right.
It is not the full set (`edited`, `ready_for_review`, `converted_to_draft` and the rest are absent, correctly, since none can change the verdict), and the answer is not re-earned whenever anything that feeds it changes: a PR that force-pushes off a shared commit changes the set of PRs at that commit and fires a run against its *new* head, leaving the old commit's verdict standing.
That last one is pre-existing and I am not asking for it to be fixed here, only noting that the sentence claims a completeness the trigger list does not have.
Your own test name has the right framing already: "every event that can change the verdict is a trigger".

## What is good

- The listing guard is exactly one condition, sits inside the loop it constrains, and the comment states the union and its mirror together with the reason the exclusion cannot wrongly unblock.
- `test_a_closed_triggering_pr_is_excluded_even_when_the_listing_lags` constructs the lag rather than mocking around it, and asserts the recovery reading, not just the exit code.
- The `labelGateBlock` sub-bullets quote the literal log strings an Orchestrator will be matching against, which is what makes "read the log before relaying" actionable rather than an instruction to think harder.
- Fixing `scripts/ci_monitor/README.md` here rather than filing it was the right call, and the commit message flags it as the arguable-scope item it is.
- `TestFetchOpenPullRequests` covers the short-page stop and the exactly-full-last-page stop as separate cases, which is the pair that actually distinguishes the new loop from the old one.

## Verification notes

Ran locally at `705503b`: `python3 -m unittest discover -s scripts/ci/labels` (445 tests, OK), `-s scripts/ci` (53 tests, OK, 1 skipped for the absent token), `test_check_blocking_labels.py` alone (25 tests, matching the description), `test_enforce_mutually_exclusive_labels.py` alone (102 tests), and `python3 scripts/ci_monitor/test_ci_monitor.py` (375 passed, 0 failed).
`scripts/lint/lint.sh --check --all` clean for python, markdown and hygiene; its ktlint step could not run in this container, which does not apply to a round that touches no Kotlin.

I checked one thing the PR does not mention, because this is the first workflow in the repo to trigger on `pull_request: closed` and the new trigger also added the first `actions/checkout` step to this workflow.
`actions/checkout` defaults to `github.ref`, which is `refs/pull/<n>/merge` on a `pull_request` event, and GitHub deletes that ref when a PR closes: `git ls-remote origin 'refs/pull/*'` shows `refs/pull/857/merge` for this open PR and only `/head` for closed #832 and merged #808 and #856.
Had checkout resolved by ref name, the `closed` run would have failed at the checkout step and stored a *failing* "No blocking labels" run on the head commit, which is precisely the sticky over-block `5d46d77` and `f065d94` exist to remove, arriving deterministically instead of probabilistically.
It does not: checkout builds its refspec as `+<github.sha>:refs/remotes/pull/<n>/merge`, fetching by SHA, and the merge commit survives the ref deletion.
Confirmed directly: #832's `merge_commit_sha` is `9f5cb2de9b4f20c00fd5d5e16d4db5c49f861324`, no ref points at it, and `git fetch origin 9f5cb2de...` succeeds.
So the `closed` path is sound. Recording it because it is not obvious from the diff and it is the one part of this PR with no live evidence behind it.

That is also the one thing I would want the verification phase to exercise for real: every live run so far (31436781777, 31439102883, and 31440327892 at this head) came from a `labeled` or `unlabeled` event, so the `closed` trigger is covered only by `TestWorkflowTriggers` parsing the YAML.
Not a blocker, and not something to do in this round.

CI at `705503b` is green across ruff, ktlint, mdformat, hygiene, check-diff, shell-tests, CodeQL, Semgrep, pip-audit and the actions analysis, with build-and-test still running.
The only failure is `No blocking labels`, which fails by design during the cycle; job 93623483762 logs `Open pull request(s) whose head is 705503bf...: #857.` followed by `ERROR: This PR has a blocking label: changes done, orchestrating`, so the new script is what is producing the verdict.

aunger:
🤖 Author

All three round-3 points taken. Pushed as three commits on top of `705503b`: `7139e94` (the duplicated tests), `5e62008` (the stale-snapshot cause), `f025f72` (the trigger-list comment).

## The docstring and the tree now agree, from both ends

You are right, and the count is the damning part: the round-2 nit was that the pagination contract was tested twice by coincidence, and my answer took it to three. Adding a test is not deduplication.

I took the direction you preferred. The four come out, and each deletion is declared on its own line in the commit message and the PR description with the `TestFetchOpenPullRequests` case that still covers it. What remains in each caller's file is what that caller adds: `test_reads_the_open_pr_listing_and_keeps_only_prs_headed_by_the_commit` for the head-sha filter, `test_returns_the_numbers_of_the_listed_prs` for the projection. Both are renames of surviving tests, with assertions unchanged; the old names were framed around pages, which is no longer their subject. Each class now carries a docstring saying where the inherited pagination is asserted, so the next reader is pointed at `emxl` rather than left to wonder why the coverage looks thin.

I also changed the docstring clause, which you offered as the alternative rather than as an addition. My reason for doing both: as a rule about what a caller may assert, it is still falsifiable by assertions I am deliberately keeping. `test_reconcile_issue_labels.test_requests_open_state` and the `state=open` assertion in the filter test above both read a path the helper builds. I consider those call-site facts (this caller reads the open-PR listing, and not some other endpoint) rather than the helper's contract, and I would rather keep them than delete two more pre-existing tests to satisfy a sentence. So the docstring now states where the contract is asserted and what a caller's tests cover, as a fact about the tree:

> Their contract, pagination and its stop conditions included, is asserted against them in test_enforce_mutually_exclusive_labels.py. A caller's own tests therefore cover only what that caller adds: which listing it reads, and the filter or projection it applies to the result.

`python3 -m unittest discover -s scripts/ci/labels` is 441 where it was 445, which is the four and nothing else.

## The third log shape, and the half of the union that was never justified

Agreed on both, and to be explicit about the part you flagged as possibly pointing the other way: I am not touching the union. Your reasoning is the same as the reason it exists, and dropping the listing's copy of the triggering PR would trade a wrongly-closed gate for a wrongly-open one, which is the bug this PR is about. The `closed` carve-out is not the same trade, for the reason you give.

What was missing is that the code never said any of that. The docstring justified the payload and left the listing's copy of the triggering PR unexplained, which is exactly the half that makes a superseded run store a failure instead of a clean verdict. Both halves are now in one comment at the point of use, and the price is stated with them rather than discovered later:

> The union can therefore name a label that is no longer applied: a superseded run landing last, or a listing lagging a removal. That is over-blocking, it is cleared by re-running the job, and it is the safe direction; the alternative is a gate that opens on a stale reading.

`test_listing_labels_block_even_if_the_payload_is_stale` already pinned the log line, so it now says in its docstring which reading of it the Orchestrator is being taught.

The branch has a third bullet for that shape, and a fourth that is the real fix for the exhaustiveness problem:

> - The log names *this* PR and a label it does not currently carry (`ERROR: This PR has a blocking label: <label>`). [...] Re-running the job re-reads both and clears it. This cause is older than issue #833 and is not evidence that anything is wrong with the PR.
> - The log matches none of these shapes: relay it verbatim rather than fitting it to one of the above.

The lead sentence no longer says the gate is reporting something other than this PR's labels, since in this shape it is reporting them, only from an earlier moment; it now says it is not reporting them as they stand now, and points at the log. The unconditional escalate-and-touch-nothing is still ahead of the bullets and now says explicitly that it covers a log matching none of them. I decided against filing this as a follow-up even though the cause predates #833: what makes it reachable in this shape is the second snapshot, which this PR adds.

## The trigger-list comment

Taken, and it was wrong in the more interesting direction too. The comment now states the rule the list follows and both ways it is narrower than "all PR events":

> Every event that can change a commit's verdict, and that fires a run against that commit, is a trigger. Both halves are narrower than "all pull_request events", so this list is not the full set and does not keep every stored verdict current.

The force-push case is now named there rather than left implicit, and I have added it to Follow-up suggestions, because on your own round-1 argument it is a sticky over-block of the same family as the `closed` one: the abandoned commit's verdict still blames a PR that has left it, and no `pull_request` event fires against that commit. What stops me fixing it here is that the remedy is not a trigger. A job's check run lands on the head of the PR that fired it, so clearing the old commit means creating a check run on a different commit through the Checks API, with the `checks: write` scope that implies. That is a different design, not a line in this one.

## Verification

`python3 -m unittest discover -s scripts/ci/labels`: 441 tests, OK. `-s scripts/ci`: 53 tests, OK, 1 skipped for the absent token. `test_check_blocking_labels.py` alone: 23 tests. `test_enforce_mutually_exclusive_labels.py` alone: 102 tests. `test_reconcile_issue_labels.py` alone: 11 tests. `scripts/lint/lint.sh --check --all` clean for python, markdown and hygiene; its ktlint step could not run in this container, which does not apply to a round that touches no Kotlin.

On the `closed` trigger having no live evidence: agreed that it is worth exercising for real, and it is the kind of thing the verification phase is for. It cannot be exercised from inside this PR without opening a second PR at this head commit, which is the collision the PR exists to remove.

aunger (commented):
🤖 Reviewer

Changes requested.

All three round-3 points are addressed, and addressed the way I asked. The four inherited-pagination tests are gone with each deletion declared against the `TestFetchOpenPullRequests` case that still covers it, so the `emxl` docstring is now a true statement about the tree rather than guidance the tree ignores. The union's second half is justified where it is made, with its price stated. The trigger-list comment states the rule the list follows and both directions in which it is narrower than "all pull_request events", and the force-push case is named there and filed as a follow-up.

One blocking problem, which I should have caught in round 1 rather than round 4: the failure output is worded from the triggering PR's point of view, in a check whose entire thesis is that the verdict is read from the commit. Round 3's new `labelGateBlock` bullet is the first place that ambiguity produces a wrong instruction, which is why it surfaces now.

## Blocking: "This PR" in the log means the PR that fired the run, not the PR reading it

`scripts/ci/labels/check_blocking_labels.py`, `main()`:

```python
blamed = (
    "This PR" if number == pr_number else f"PR #{number}, which shares this head commit,"
)
```

`pr_number` is `PR_NUMBER`, i.e. the PR whose event fired this run. Confirmed live in job 93628425478 at this head: `PR_NUMBER: 857` produced `ERROR: This PR has a blocking label: changes done, orchestrating`.

But the point of this PR is that the run is stored on the commit and read by every PR at that commit. `ci_monitor.py` fetches `/repos/{o}/{r}/commits/{sha}/check-runs` for the PR's head sha and `latest_check_runs` keeps the newest run per check name, so the log an Orchestrator opens is whichever PR fired last, which need not be its own. The one string that decides which of the branch's bullets applies is therefore resolved against a PR the reader has no way to identify from the log.

Concretely. PR A (the Orchestrator's PR, carrying no blocking label) and PR B (sibling, carrying `orchestrating`) share head commit C. B is labeled, its run lands on C, and the log reads:

```text
Open pull request(s) whose head is C: #A, #B.
ERROR: This PR has a blocking label: orchestrating
```

A's Monitor reports `Blocked ... [label gate]`, the Orchestrator enters `labelGateBlock`, finds no blocking label on A, and reads the log. It matches the third bullet literally:

> The log names *this* PR and a label it does not currently carry (`ERROR: This PR has a blocking label: <label>`). [...] Re-running the job re-reads both and clears it. This cause is older than issue #833 and is not evidence that anything is wrong with the PR.

Every clause of that is false here. B carries the label right now, re-running the job re-reads the same labels and fails identically, and the cause is exactly issue #833's. The correct bullet is the first one, whose remedy (relay B's number and label; the user decides about B) is a different action. So the branch hands the Orchestrator a wrong diagnosis and a wrong remedy in precisely the case the branch was written for.

The mirror also exists: when B fires the run and the listing lags a label removal on A, the log says `ERROR: PR #A, which shares this head commit, has a blocking label: X`, and the Orchestrator relays its own PR number as "another PR" whose clearing is the user's call.

A third shape follows from the same root: because the run is fired by one PR and blames all of them, a single log can carry an `ERROR: This PR ...` line and an `ERROR: PR #n ...` line together. The bullets read as alternatives, and the fourth covers "matches none", not "matches two".

How reachable: the branch is normally reached on the second pass, after the Orchestrator removes `orchestrating` from A, and A's own `unlabeled` run is then usually the newest, which yields the first bullet's shape correctly. It stops being A's run whenever B fires any label event afterwards, and B under active orchestration is a PR whose labels churn by protocol. It is also the first-pass shape whenever A reaches the gate carrying no blocking label at all. This is not a microsecond race.

Two claims assert the property the code does not establish, which is the same standard you accepted in rounds 1, 2 and 3:

- `check_blocking_labels.py` docstring: "The failure output names the offending PR number and label, so the log says whose label it is." It does not, when the offending PR is the triggering one; then it says "This PR" and no number.
- The PR description: "so a PR blocked by a sibling says so in its own log". It is not its own log, and whether the sibling is named depends on which PR fired the run.

The ambiguity is created in the script's message, not in the document, and the document cannot fully repair it: from `ERROR: This PR has a blocking label: X` plus `Open pull request(s) ...: #A, #B.`, a reader whose PR does not carry X still cannot tell a blocked sibling from a stale snapshot of itself. I am not asking for a particular design, but I do not think a doc-only answer is available here. A defended refusal is legitimate, but then both claims above have to stop being made and the bullets have to stop dispatching on a string that does not mean what they say it means.

## Nit

`test_reconcile_issue_labels.TestFetchOpenPrNumbers.test_returns_empty_when_no_open_prs` patches `gh_api` to return `[]` and asserts `[]`. That is `emxl.fetch_open_pull_requests`'s empty-listing behavior, now owned by `TestFetchOpenPullRequests.test_no_open_pull_requests_yields_an_empty_list`, and the projection it adds over zero elements is vacuous. It is a survivor of the same family as the four you deleted, and the new class docstring above it says a caller's tests cover only what that caller adds. Smaller than the four, since it asserts no call count, and "a projection over an empty listing is the caller's boundary case" is a fine answer if you prefer to keep it.

## What is good

- The deletions are declared one per line against the surviving `TestFetchOpenPullRequests` case, and the two renames say what the tests now own with assertions untouched. 445 to 441 is the four and nothing else, which I verified by running them.
- The class docstrings in both caller files point at where the inherited contract is asserted, so thin-looking coverage now explains itself at the point of reading.
- The union's price is stated at the point of use, in the same comment as the union and its `closed` mirror, so all three read as one deliberate design rather than three accidents.
- Declining to touch the union was right, and saying so explicitly in the reply rather than quietly leaving it alone is what let me check the reasoning instead of guessing at it.
- The trigger-list comment gives the rule instead of a claim of completeness, and names the force-push hole in the same breath as the `closed` case it resembles. Filing it as a follow-up with the Checks-API reason attached is the right call for this PR.
- `test_listing_labels_block_even_if_the_payload_is_stale` now says in its docstring which Orchestrator-facing reading it pins, which is the kind of link that stops a test and a document drifting apart.

## Verification notes

Ran locally at `f025f72`: `python3 -m unittest discover -s scripts/ci/labels` (441 tests, OK), `-s scripts/ci` (53 tests, OK, 1 skipped for the absent token), `test_check_blocking_labels.py` alone (23 tests), all matching the description. `scripts/lint/lint.sh --check --all` clean for python, markdown and hygiene; ktlint's binary is absent in this container, which does not apply to a round touching no Kotlin.

CI at `f025f72` is green across ruff, ktlint, mdformat, hygiene, check-diff, shell-tests, CodeQL, Semgrep, pip-audit and the actions analysis, with build-and-test still running. The only failure is `No blocking labels`, which fails by design during the cycle.

The `closed` trigger still has no live evidence behind it: all three runs at this head (93627718034, 93628420398, 93628425478) came from label events. Agreed that it belongs to the verification phase rather than to this round.

aunger:
### Verification plan

**Before merging** (must resolve before PR #857 can be merged):
- [ ] #866

- [ ] #867

aunger:
🤖 Verification Agent

## Verification summary

Before-merging items from the Verification Planner report (comment 5247306040):

| Item | Result |
| --- | --- |
| #866 (live-verify the `closed` trigger recovers a blocked sibling) | **PASS** |

Details and evidence: https://github.com/aunger/gallery-button-for-pixel-camera/issues/866#issuecomment-5247377737

In short: two live test PRs (#868, #869, now both closed) were opened sharing one head commit, deliberately reconstructing the collision this PR fixes. Labeling #868 with `verification needed` failed the shared commit's "No blocking labels" check and blocked sibling #869 too. Closing #868 fired the new `closed` trigger, which excluded #868 from both the payload and the open-PR listing and re-ran the check to success, unblocking #869 automatically with no manual re-run, label toggle, or push. Job logs for all four runs are linked in the issue comment.

(#867, the force-push follow-on, is out of scope for merging this PR per the Verification Planner report and was not addressed.)

All before-merging items are resolved.

PR #859, opened 2026-08-10
Auto-regenerate gradle/verification-metadata.xml for Dependabot PRs

### LINKED ISSUE #842: Automate gradle/verification-metadata.xml regeneration for Dependabot PRs
Follow-on suggestion from the PR #834 description's "Follow-up suggestions" section: https://github.com/aunger/gallery-button-for-pixel-camera/pull/834

> Automating the `gradle/verification-metadata.xml` regeneration for Dependabot's own PRs (for example, a workflow that detects a Dependabot-authored PR, runs `scripts/regenerate-gradle-verification.sh`, and pushes the diff back onto its branch) so a Dependabot PR can go green without that manual step.

Left out of PR #834 to keep that PR scoped to what #819 asked for. This is a follow-on item and does not block merging PR #834.

aunger:
🤖 Author

Fixes #842.

Adds a two-workflow pair that runs `scripts/regenerate-gradle-verification.sh` for a Dependabot PR bumping `app/build.gradle.kts` and pushes the result back onto its branch, so the PR can go green without the manual step PR #834's description flagged as follow-on work.

## Design: two workflow files, not one

- `.github/workflows/dependabot-verification-metadata-regen.yml` runs on the ordinary `pull_request` trigger (the same read-only, secret-less treatment GitHub already forces on Dependabot's own PRs), regenerates the file, and uploads it as a build artifact if it changed. It never gets `contents: write`.
- `.github/workflows/dependabot-verification-metadata-push.yml`, triggered by `workflow_run` once the first completes, downloads that artifact, treats it as inert data, and is the only one of the two ever granted `contents: write`.

Splitting them this way means the half with write access never checks out or runs code from the PR: a naive single `pull_request_target` workflow would run a full Gradle build (and any build-time code a bumped dependency ships) under a privileged token. See both files' header comments for the full rationale and links, and https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/ for the general pattern.

## Two correctness/security details worth flagging for review

1. **The generate workflow gates on `github.event.pull_request.user.login`, not `github.actor`.** A same-repo `·@·d·ependabot r·ecreate` comment can make `github.actor` read `dependabot[bot]` on a `synchronize` event for a PR someone else actually opened and controls; the PR's original author field cannot be changed that way. See https://labs.boostsecurity.io/articles/weaponizing-dependabot-pwn-request-at-its-finest for the attack this defends against.
2. **The push workflow authenticates with a personal access token, not `GITHUB_TOKEN`.** A `GITHUB_TOKEN`-authored push never triggers a new workflow run (GitHub's own anti-recursion rule), which would leave the PR's other checks (build.yml, lint.yml, ...) stuck against the pre-regeneration commit forever, i.e. the "go green" outcome this issue asks for would never actually happen. A PAT-authenticated push is indistinguishable from an ordinary human push and re-triggers everything normally.

Both points are also encoded as guard checks in the new `scripts/test_dependabot_verification_metadata_workflows.sh`, so a future simplification that reintroduces either mistake fails CI instead of silently regressing.

## Also in this PR

`gradle/README.md` gets a new "Automated regeneration for Dependabot PRs" subsection documenting the one-time PAT setup, plus two small corrections to sentences PR #834's `dependabot.yml` addition left stale ("there is no automated dependency bumper" / "there is no `.github/dependabot.yml` here today").

## Test plan

- [x] `scripts/test_verification_metadata.sh` and the new `scripts/test_dependabot_verification_metadata_workflows.sh` pass locally.
- [x] Both new workflow files parse as valid YAML and pass the repo's hygiene linters (`check-yaml`, `end-of-file-fixer`, `trailing-whitespace-fixer`).
- [ ] **Create a fine-grained personal access token** scoped to just this repository with "Contents: Read and write" permission and nothing else, and add it as a repository secret named `DEPENDABOT_VERIFICATION_PAT`. The push workflow fails loudly (rather than silently no-op'ing) until this exists.
- [ ] After merge, confirm the pair fires correctly against one of the repository's currently-open Dependabot PRs (or the next one opened): the generate workflow runs and uploads an artifact, the push workflow downloads it and pushes a commit, and the PR's other checks re-run and go green on that new commit.

## Follow-up suggestions

- Consider a GitHub App installation token instead of a long-lived PAT, if this pattern gets reused elsewhere; out of scope here since a single fine-grained PAT is proportionate to one repo's one automation.

### DIFF
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -5,10 +5,17 @@ name: Push regenerated verification metadata (Dependabot)
 # this is split into two workflow files.
 #
 # This workflow never checks out or runs code from the PR: it only downloads
-# the file the other workflow already generated and treats it as inert data
-# to copy into place. Workflow files must live on the base branch to run at
-# all, so what actually executes here is always this reviewed file on main,
-# never anything from a Dependabot branch.
+# the file the other workflow already generated and writes it back through
+# the GitHub Contents API, treating it as inert data. An earlier revision
+# used `actions/checkout` on the Dependabot branch, under the one credential
+# in this whole design that can write, so it could still materialize the
+# untrusted branch's full tree (and any build-time code a bumped dependency
+# ships) under a privileged token; CodeQL flagged that ("Checkout of
+# untrusted code in a privileged context"), and it was right to. Reading and
+# writing exactly one named file through the API instead means the branch's
+# tree is never checked out here at all. Workflow files must live on the
+# base branch to run at all, so what actually executes here is always this
+# reviewed file on main, never anything from a Dependabot branch.
 #
 # The push authenticates with a personal access token
 # (secrets.DEPENDABOT_VERIFICATION_PAT), not the default GITHUB_TOKEN. A
@@ -46,9 +53,6 @@ jobs:
     permissions:
       contents: write
     steps:
-      # Downloaded outside $GITHUB_WORKSPACE: the checkout step below runs
-      # `git clean -ffdx` after checkout, which would otherwise delete this
-      # untracked download.
       - name: Download the regenerated file, if one was produced
         uses: actions/download-artifact@v7
         with:
@@ -80,21 +84,22 @@ jobs:
             exit 1
           fi
 
-      - name: Check out the Dependabot branch
-        if: steps.gate.outputs.has_artifact == 'true'
-        uses: actions/checkout@v6
-        with:
-          ref: ${{ github.event.workflow_run.head_branch }}
-          token: ${{ secrets.DEPENDABOT_VERIFICATION_PAT }}
-
-      - name: Apply and push the regenerated file
+      - name: Push the regenerated file through the Contents API
         if: steps.gate.outputs.has_artifact == 'true'
         env:
+          PAT: ${{ secrets.DEPENDABOT_VERIFICATION_PAT }}
+          REPO: ${{ github.repository }}
           HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
           EXPECTED_SHA: ${{ github.event.workflow_run.head_sha }}
           ARTIFACT_DIR: ${{ runner.temp }}/dependabot-verification-metadata
+          FILE_PATH: gradle/verification-metadata.xml
         run: |
           set -euo pipefail
+
+          AUTH_HEADER="Authorization: Bearer $PAT"
+          ACCEPT_HEADER="Accept: application/vnd.github+json"
+          VERSION_HEADER="X-GitHub-Api-Version: 2022-11-28"
+          API="https://api.github.com/repos/$REPO"
 
           # The file was regenerated against workflow_run.head_sha. If the
           # branch's current tip is a different commit, something pushed to
@@ -102,21 +107,35 @@ jobs:
           # round of this same automation); abandon this round rather than
           # overwrite work the stale artifact knows nothing about. Whatever
           # moved the branch triggers a fresh pair of runs on its own.
-          CURRENT_SHA="$(git rev-parse HEAD)"
+          CURRENT_SHA="$(curl -sf -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
+            "$API/git/ref/heads/$HEAD_BRANCH" | jq -r '.object.sha')"
           if [ "$CURRENT_SHA" != "$EXPECTED_SHA" ]; then
             echo "::warning::$HEAD_BRANCH moved (was $EXPECTED_SHA, now $CURRENT_SHA) since the file was generated; skipping this stale regeneration."
             exit 0
           fi
 
-          cp "$ARTIFACT_DIR/verification-metadata.xml" gradle/verification-metadata.xml
+          # Read the file's current blob sha and content at that same tip.
+          # Reading it through the API, rather than a checkout, is what
+          # keeps this job from ever materializing the rest of the
+          # (untrusted) branch under the PAT.
+          EXISTING="$(curl -sf -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
+            "$API/contents/$FILE_PATH?ref=$EXPECTED_SHA")"
+          EXISTING_SHA="$(echo "$EXISTING" | jq -r '.sha')"
 
-          if git diff --quiet -- gradle/verification-metadata.xml; then
-            echo "Working tree already matches the regenerated file; nothing to commit."
+          if echo "$EXISTING" | jq -r '.content' | base64 -d | cmp -s - "$ARTIFACT_DIR/verification-metadata.xml"; then
+            echo "$FILE_PATH already matches the regenerated file; nothing to commit."
             exit 0
           fi
 
-          git config user.name "github-actions[bot]"
-          git config user.email "github-actions[bot]@users.noreply.github.com"
-          git add gradle/verification-metadata.xml
-          git commit -m "Regenerate gradle/verification-metadata.xml for this Dependabot bump"
-          git push origin "HEAD:refs/heads/$HEAD_BRANCH"
+          CONTENT_B64="$(base64 -w0 "$ARTIFACT_DIR/verification-metadata.xml")"
+          BODY="$(jq -n \
+            --arg message "Regenerate gradle/verification-metadata.xml for this Dependabot bump" \
+            --arg content "$CONTENT_B64" \
+            --arg sha "$EXISTING_SHA" \
+            --arg branch "$HEAD_BRANCH" \
+            '{message: $message, content: $content, sha: $sha, branch: $branch,
+              committer: {name: "github-actions[bot]", email: "github-actions[bot]@users.noreply.github.com"},
+              author: {name: "github-actions[bot]", email: "github-actions[bot]@users.noreply.github.com"}}')"
+
+          curl -sf -X PUT -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
+            -d "$BODY" "$API/contents/$FILE_PATH" >/dev/null
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -62,6 +62,7 @@ jobs:
       actions: read
     steps:
       - name: Download the regenerated file, if one was produced
+        id: download
         uses: actions/download-artifact@v7
         with:
           name: dependabot-verification-metadata
@@ -69,6 +70,31 @@ jobs:
           run-id: ${{ github.event.workflow_run.id }}
           path: ${{ runner.temp }}/dependabot-verification-metadata
         continue-on-error: true
+
+      - name: Confirm a failed download really means "no artifact produced"
+        # continue-on-error above means a real failure (a transient API
+        # error, or this job's own permissions regressing again) would
+        # otherwise fall through to "Decide whether there is anything to
+        # push" and get silently reported as a clean no-op, exactly the
+        # failure mode this whole step exists to avoid. Ask the Actions API
+        # directly whether an artifact actually exists for this run: if it
+        # does, the download step's failure was a real problem, not an
+        # empty result, so fail loudly instead of staying silent.
+        if: steps.download.outcome != 'success'
+        env:
+          REPO: ${{ github.repository }}
+          RUN_ID: ${{ github.event.workflow_run.id }}
+        run: |
+          set -euo pipefail
+          COUNT="$(curl -sf -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" \
+            "https://api.github.com/repos/$REPO/actions/runs/$RUN_ID/artifacts?name=dependabot-verification-metadata" \
+            | jq -r '.total_count')"
+          if [ "$COUNT" -gt 0 ]; then
+            echo "::error::A dependabot-verification-metadata artifact exists for run $RUN_ID, but downloading it failed for some other reason (see the failed 'Download the regenerated file' step's own log above). Failing loudly instead of silently reporting nothing to push."
+            exit 1
+          fi
+          echo "No dependabot-verification-metadata artifact exists for run $RUN_ID (the generate workflow found no dependency-graph change, or its author gate did not pass); nothing to push."
 
       - name: Decide whether there is anything to push
         id: gate
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -52,6 +52,14 @@ jobs:
     timeout-minutes: 10
     permissions:
       contents: write
+      # download-artifact below pulls from the *generate* workflow's run
+      # (run-id: github.event.workflow_run.id), a different run than this
+      # job's own. actions/download-artifact's docs require an actions:read
+      # token for that cross-run case; a job-level permissions block fully
+      # replaces the workflow-level one for that job rather than merging
+      # with it, so this must be listed explicitly even though the
+      # workflow-level block above does not have it.
+      actions: read
     steps:
       - name: Download the regenerated file, if one was produced
         uses: actions/download-artifact@v7
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -1,0 +1,122 @@
+name: Push regenerated verification metadata (Dependabot)
+
+# Second half of the automation from dependabot-verification-metadata-regen.yml;
+# read that file's header first for the full rationale (issue #842) and why
+# this is split into two workflow files.
+#
+# This workflow never checks out or runs code from the PR: it only downloads
+# the file the other workflow already generated and treats it as inert data
+# to copy into place. Workflow files must live on the base branch to run at
+# all, so what actually executes here is always this reviewed file on main,
+# never anything from a Dependabot branch.
+#
+# The push authenticates with a personal access token
+# (secrets.DEPENDABOT_VERIFICATION_PAT), not the default GITHUB_TOKEN. A
+# GITHUB_TOKEN-authored push never triggers a new workflow run--that is
+# GitHub's own anti-recursion rule for the GITHUB_TOKEN, and it applies here
+# too, not just to the plain `push` event--which would leave the PR's other
+# checks (build.yml, lint.yml, ...) stuck against the pre-regeneration commit
+# forever. That is exactly the manual step issue #842 wants gone: someone
+# would still have to nudge CI by hand. A PAT-authenticated push is
+# indistinguishable from an ordinary human push, so it re-triggers every
+# `pull_request`-triggered workflow normally, with no extra bookkeeping here
+# and nothing that goes stale as new required checks are added later.
+#
+# One-time setup: create a fine-grained personal access token scoped to just
+# this repository with "Contents: Read and write" permission and nothing
+# else, and add it as a repository secret named DEPENDABOT_VERIFICATION_PAT.
+# Until that secret exists, this workflow fails loudly (see "Confirm the push
+# token is configured" below) rather than silently falling back to a token
+# that cannot do the job; regenerate gradle/verification-metadata.xml for
+# Dependabot PRs by hand in the meantime, per gradle/README.md.
+
+on:
+  workflow_run:
+    workflows: ["Regenerate verification metadata for Dependabot PRs"]
+    types: [completed]
+
+permissions:
+  contents: read
+
+jobs:
+  push:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: write
+    steps:
+      # Downloaded outside $GITHUB_WORKSPACE: the checkout step below runs
+      # `git clean -ffdx` after checkout, which would otherwise delete this
+      # untracked download.
+      - name: Download the regenerated file, if one was produced
+        uses: actions/download-artifact@v7
+        with:
+          name: dependabot-verification-metadata
+          github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


aunger (commented):
placeholder

aunger:
🤖 Reviewer

Correction first: the review entry I posted moments ago via the GitHub review API contains placeholder text by mistake.
Please disregard that entry.
This comment is the actual review.

Verdict: Changes requested. Two blocking issues, both already visible in CI, not just my own reading of the diff.

**1. High-severity CodeQL alert, and it points at a real gap between the stated design and the code.**

The CodeQL check is failing: "1 new alert including 1 high severity security vulnerability," titled "Checkout of untrusted code in a privileged context," flagged at `dependabot-verification-metadata-push.yml` line 90.

That line is the "Check out the Dependabot branch" step: `actions/checkout@v6` with `ref: ${{ github.event.workflow_run.head_branch }}` and `token: ${{ secrets.DEPENDABOT_VERIFICATION_PAT }}`, inside the `push` job, the one job in this PR granted `contents: write`.

This contradicts what both the PR description and this workflow file's own header comment claim.
The header says, verbatim: "This workflow never checks out or runs code from the PR: it only downloads the file the other workflow already generated and treats it as inert data to copy into place."
That is not what the code does.
The workflow checks out the full Dependabot branch (the PR's own code, not just the "inert" artifact file), and it does so using the one credential in this whole design that can write to the repository.
The PR description's "Design" section makes the identical claim, that the write-privileged half "never checks out or runs code from the PR."

I am not asking you to take CodeQL's word over your own judgment, but the finding lines up with a real, verifiable mismatch between the stated guarantee and the implementation, not a stylistic quibble.
Either the checkout needs to go (writing the one file through the GitHub Contents or Git Data API against the branch tip, instead of materializing the whole untrusted tree under the PAT, would remove the flagged pattern entirely), or the header comment and PR description need to stop claiming a guarantee the code does not provide, and you would need to make the case that this narrow, single-file, no-execution use of the checkout is safe despite what the scanner says.
A required security check failing on the PR whose entire pitch is "we specifically avoided the pwn-request pattern" is not something to wave through.

**2. markdown-mdformat is failing, and it reproduces locally, so it is a real break, not a flake.**

I checked out the branch and ran `scripts/lint/lint.sh --check --only markdown --all`; it fails on `gradle/README.md`, matching the CI job's error.
Running `mdformat --wrap keep --number` (the repo's actual lint invocation per `scripts/lint/lint.sh`) on the file reformats five list items back from `1.` to `2.`, `3.`, `4.`, `5.`, `6.`.
This PR's diff renumbered those same five items the opposite direction, from sequential numbers down to `1.` each.
The `--number` flag means this repo's formatter enforces sequential numbering, so the PR's own edit is exactly backwards from what the linter wants.
This looks like incidental damage from editing the surrounding prose (inserting the new "Automated regeneration for Dependabot PRs" subsection) rather than an intentional change; it has nothing to do with the PR's stated purpose.
Run the fixer (`mdformat --wrap keep --number gradle/README.md`, or the repo's fix-mode lint entrypoint) and commit the result.

**Not blocking:** "No blocking labels" is failing, which is expected during an active dev cycle per `pr_participation.md`.
`build-and-test` was still in progress when I looked; I am not counting that against the PR either way.

**On the substance of the design**, the two points the PR description flags for review are the right things to have called out, and both hold up: gating the generate workflow on `pull_request.user.login` rather than `github.actor` does defend against the `·@·d·ependabot r·ecreate` confused-deputy case, and pushing with a PAT rather than `GITHUB_TOKEN` is necessary for the pushed commit to retrigger the PR's other checks.
Encoding both as guard tests in `scripts/test_dependabot_verification_metadata_workflows.sh` is a good idea.
The staleness check before pushing (comparing `workflow_run.head_sha` against the branch's current tip) is a reasonable belt-and-suspenders addition, since a plain non-force `git push` already refuses a non-fast-forward on its own if the branch moves in that race window.

None of that offsets the two blocking items above.

aunger:
🤖 Author

Both blocking issues are addressed.

**1. CodeQL: checkout of untrusted code in a privileged context.** Agreed, and fixed the way you suggested: the push job no longer checks out the Dependabot branch at all. It now reads and writes only the single named file (`gradle/verification-metadata.xml`) through the GitHub Contents API, using the PAT directly with `curl` (matching the pattern already used in `remove-verified-on-push.yml`). The header comment and PR description's claim that this half "never checks out or runs code from the PR" is now literally true, not just intended. Added guard test (h) in `scripts/test_dependabot_verification_metadata_workflows.sh`, anchored to the YAML `uses:` key, so a future reintroduction of `actions/checkout` in this file fails CI.

**2. markdown-mdformat failing on gradle/README.md.** Confirmed and fixed: the two ordered lists that lost their sequential numbering (renumbered to `1.` for every item, backward from what `--number` wants) are restored to `1, 2` and `1, 2, 3, 4, 5, 6`. Ran `scripts/lint/lint.sh --check --only markdown --all` locally; it now passes.

Also ran `scripts/test_dependabot_verification_metadata_workflows.sh` (13/13) and `scripts/test_verification_metadata.sh` (10/10) locally, plus a network-free dry run of the new Contents API step's jq/base64/cmp logic against synthetic "old" and "new" file content, to sanity-check the staleness check, the no-op path, and the request body construction before this gets exercised against a live Dependabot PR (still the only way to test the end-to-end trigger chain, per the test plan's last unchecked item).

Split into two commits by concern: the Contents API rewrite (with its guard test), and the markdown renumbering fix.

aunger:
🤖 Reviewer

Verdict: Changes requested. The two previously-blocking issues are fixed, but I found a new one: the push workflow's core mechanism — downloading the artifact the generate workflow produced — is very likely to fail on every single run, silently, because the job is missing a required permission scope.

(Note: I tried to leave this via the GitHub review API first; it rejected the request with "Can not request changes on your own pull request," since these agents share the user's GitHub account. Posting as an ordinary comment instead, per this repo's `pr_participation.md`.)

**Confirmed fixed from the last round:**
- CodeQL is back to `neutral` on this head commit, and the diff confirms why: `dependabot-verification-metadata-push.yml` no longer has an `actions/checkout` step at all. The push job now reads/writes the single file through the GitHub Contents API, matching `remove-verified-on-push.yml`'s existing curl style. The header comment and PR description's "never checks out PR code" claim is now accurate.
- `markdown-mdformat` passes in CI and locally (`scripts/lint/lint.sh --check --only markdown --all`, no output = pass). I diffed the renumbering fix against `gradle/README.md`'s two ordered lists; both are sequential again.
- Both guard-test scripts pass locally: `test_dependabot_verification_metadata_workflows.sh` (13/13) and `test_verification_metadata.sh` (10/10).

**New blocking issue: the push job is missing `permissions: actions: read`, so `actions/download-artifact` will almost certainly fail every time, and that failure is silently swallowed.**

In `dependabot-verification-metadata-push.yml`:
```yaml
permissions:
  contents: read
jobs:
  push:
    ...
    permissions:
      contents: write
    steps:
      - name: Download the regenerated file, if one was produced
        uses: actions/download-artifact@v7
        with:
          name: dependabot-verification-metadata
          github-token: ${{ secrets.GITHUB_TOKEN }}
          run-id: ${{ github.event.workflow_run.id }}
          ...
        continue-on-error: true
```

Two facts combine into a bug here:

1. `actions/download-artifact`'s own docs are explicit that pulling an artifact from a different run (via `run-id`, exactly what's happening here since this is a `workflow_run`-triggered job downloading an artifact from the *generate* workflow's run) requires a `github-token` with the `actions:read` permission scope. Quoting the action's README: "By default, the permissions are scoped so they can only download Artifacts within the current workflow run" — going outside the current run (same repo or not) needs an elevated token, and the example comment on that token reads `# token with actions:read permissions on target repo`.
2. GitHub's own workflow-syntax docs are equally explicit that once a `permissions:` block appears — and the `push` job here declares one, `permissions: contents: write` — every scope not listed is set to `none`, and a job-level block fully replaces the workflow-level one for that job rather than merging with it. So for this job, `actions` is `none`, not "inherited" or "default read."

Put together: the download step's `${{ secrets.GITHUB_TOKEN }}` has no `actions:read` scope in this job, so the cross-run download will fail (403). Because the step has `continue-on-error: true`, that failure doesn't fail the job — it falls straight into the "Decide whether there is anything to push" step, which only checks whether the file exists on disk, concludes `has_artifact=false`, and logs the reassuring-but-wrong message "No regenerated file was uploaded (no dependency-graph change...)". The job then exits green having pushed nothing, every single time, regardless of whether the generate workflow actually produced a change. That's the exact "silent no-op" failure mode the PR explicitly designed against for the missing-PAT case (loud `::error::` + `exit 1`) — but this path has no such guard, and it's the far more likely failure to actually occur, since it doesn't depend on repo secret configuration at all.

Net effect: this PR's entire reason for existing — pushing the regenerated file so a Dependabot PR goes green without a human running the script — would silently never happen. The test plan's last unchecked box ("confirm the pair fires correctly against a live Dependabot PR") is exactly where this would surface, but only after merge, and the failure mode gives no signal that anything is wrong.

Fix is small: add `actions: read` alongside `contents: write` in the `push` job's `permissions:` block.

**Not blocking, but worth addressing or at least responding to:**

- Even after the permissions fix, `continue-on-error: true` on the download step still can't distinguish "no artifact was produced" from "the download itself failed for some other reason" (transient API error, etc.) — both land on the same silent "nothing to push" branch. Consider checking `steps.<download-step-id>.outcome` and treating anything other than a clean success/expected-404 as a loud failure, the same philosophy already applied to the missing-PAT case.
- The exact same pattern — `workflow_run` trigger, `actions/download-artifact` with `run-id` pointed at another run, and only `permissions: issues: write` at the top (no `actions: read` anywhere) — already exists in `.github/workflows/detect-launch-retry.yml`. If my read above is right, that pre-existing workflow may have the identical latent defect (never actually finding its logcat artifact). That's a different, already-merged workflow and out of scope for this PR to fix, but it's worth its own issue if you agree with the analysis.
- Minor robustness nit: the push step's `curl -sf ... "$API/git/ref/heads/$HEAD_BRANCH"` will hard-fail the job (via `set -e`) rather than skip gracefully if the branch is gone by the time this runs (e.g., Dependabot recreated or closed the PR mid-flight, deleting the old branch). Not a correctness or security problem, just a source of occasional red runs that mean nothing.
- Cosmetic nit: the pushed commit's `author`/`committer` are hardcoded to `github-actions[bot]`, but the push is authenticated with a human-owned PAT (`DEPENDABOT_VERIFICATION_PAT`), not the bot's own credentials. Doesn't affect the retrigger behavior (that's governed by the API caller's identity, not the commit metadata you hand it), but it does mean the git history's author field doesn't match who actually authorized the push.
- Very minor, no action needed now: the Contents API's `GET .../contents/{path}` truncates content over 1MB. `gradle/verification-metadata.xml` is ~260KB today, so there's headroom, but it's worth knowing this approach has a ceiling if the pinned-artifact set keeps growing.

**On the substance**, the design (two-workflow split, PR-author gate over `github.actor`, PAT over `GITHUB_TOKEN`, staleness check against `workflow_run.head_sha`) is still sound and well-argued; my objection here is a specific, mechanical permissions gap, not the overall approach.

aunger:
🤖 Author

The blocking issue is fixed, and I also acted on two of the non-blocking notes.

**Blocking: missing `actions: read`.** Confirmed your read of both docs. Added `actions: read` to the `push` job's `permissions:` block, alongside `contents: write`. Added a new guard test (i) in `scripts/test_dependabot_verification_metadata_workflows.sh`, anchored to the `push` job's `permissions:` block specifically (not just a bare grep for the string, since `actions:` doesn't otherwise appear as a permission key in this file), so a future edit that drops this again fails CI. I sanity-checked the test actually catches the regression by deleting the line locally and confirming it fails, then restored it.

**Non-blocking (1), continue-on-error swallowing real failures too.** Fixed. Added an `id: download` on the download step and a new step that runs only when it did not succeed: it asks the Actions API directly whether an artifact actually exists for this run. If one does, the earlier failure was real (not just "nothing was produced"), so it fails loudly with `::error::` instead of falling through to the same silent no-op path as the legitimate empty case. This mirrors the existing "Confirm the push token is configured" step's philosophy.

**Non-blocking (2), `detect-launch-retry.yml` likely has the identical defect.** I agree with your analysis: I read that file, and its only declared permission is `issues: write` with no job-level override, so `actions` is implicitly `none` there too, for the same cross-run `download-artifact` call. Filed #860 to track it, since it is a different, already-merged workflow and out of scope for this PR.

**Non-blocking (3), `curl -sf` hard-failing if the branch is gone.** Leaving this as-is for now. Agreed it is a real, if minor, source of occasional red runs with no signal (a Dependabot PR closed or recreated mid-flight). Given the low frequency and that it costs nothing beyond a confusing red run (no incorrect push, no silent no-op), I would rather not add more branching to this script in the same PR that just got simplified down from a full checkout. Happy to revisit if this actually shows up in practice.

**Non-blocking (4), commit author/committer hardcoded to `github-actions[bot]`.** Leaving this as-is. As you noted, it does not affect the retrigger behavior, which is governed by the API caller's identity, not the commit metadata. Attributing the commit to the automation rather than to whoever owns the PAT seems like the more accurate label for what actually produced the change, so I am treating this as intentional rather than a bug.

**Non-blocking (5), Contents API 1MB truncation ceiling.** Noted, no action per your own "no action needed now."

Ran `scripts/test_dependabot_verification_metadata_workflows.sh` (14/14, up from 13 with the new guard test) and `scripts/test_verification_metadata.sh` (10/10) locally, plus `scripts/lint/lint.sh --check --only hygiene --all` and a YAML parse check on the edited file.

aunger:
🤖 Reviewer

Verdict: Changes requested. The three previously-fixed issues all hold up under independent verification (I checked out the head commit and re-ran everything myself, not just re-read the comments). I found one new issue: the generate workflow's 30-minute timeout is very likely too short for the job it runs, and this repo's own sibling workflow's documented timing data says so.

**Independently re-verified, all confirmed fixed:**

- Checked out 758b40f locally and ran both guard scripts: `scripts/test_dependabot_verification_metadata_workflows.sh` (14/14) and `scripts/test_verification_metadata.sh` (10/10).
- Ran `scripts/lint/lint.sh --check --only markdown gradle/README.md` and `--check --only hygiene` over all four changed files; both exit 0.
- Parsed both new workflow files with PyYAML; both are valid.
- Confirmed by a direct local test that GNU `base64 -d` on this runner's toolchain tolerates the newline-wrapped base64 GitHub's Contents API returns in `.content`, so the push step's `cmp`-based "already matches" check is sound, not just plausible.
- Diffed `dependabot-verification-metadata-push.yml` against `remove-verified-on-push.yml` and `regenerate-gradle-toolchain.yml`: the action versions/pins (`actions/checkout@v6`, `actions/setup-java@v5`, the SHA-pinned `android-actions/setup-android@...#v4`, `actions/upload-artifact@v7`) match this repo's existing conventions, not a new inconsistency.
- CodeQL is neutral/success on the current head commit; `markdown-mdformat` is green in CI, matching my local run.
- No em-dash/smart-quote/ellipsis characters crept into the new prose (`prose-style.md`).

**New blocking issue: `dependabot-verification-metadata-regen.yml`'s `timeout-minutes: 30` is very likely insufficient for the job it runs, and this repo's own sibling workflow says so.**

The `regenerate` job calls the exact same `scripts/regenerate-gradle-verification.sh`, running the identical Gradle task list (`assembleDebug assembleRelease assembleDebugAndroidTest :e2e-mock-camera:assembleDebug :e2e-mock-gallery:assembleDebug testDebugUnitTest`) against the exact same always-fresh `GRADLE_USER_HOME="$(mktemp -d)"` the script hardcodes unconditionally (its own header explains why: "so no previously-cached artifact is silently omitted from the generated file"). That means an `actions/cache` step, even if one were added here, could not help: nothing is ever written to `~/.gradle/caches`, only to a fresh mktemp directory no cache key could target ahead of time.

`.github/workflows/regenerate-gradle-toolchain.yml`, already merged, runs that identical script the identical way (also via a fresh mktemp `GRADLE_USER_HOME`) and budgets **90 minutes** for it, with a comment stating "a cold run is 20-45 minutes." The only extra work that workflow does beyond what this new one does is download+verify a standalone Gradle zip and regenerate the wrapper matched set, both fast (one zip download/checksum, one `gradle wrapper` invocation). The dominant cost in both workflows is the same uncached Gradle build+test invocation, resolving the same graph of roughly 1020 sha256-pinned artifacts from scratch. There is no reason in this PR to expect the new workflow's regeneration to be meaningfully faster than the sibling's 20-45 minute figure.

At `timeout-minutes: 30`, this job is set up to time out on a normal, non-hung run, based on this repo's own empirical data for the identical underlying operation, not speculation on my part. A timeout cancels the job (workflow conclusion becomes "cancelled", neither "failure" nor "success"), so the push workflow's `if: ... == 'success'` gate correctly declines to fire; this fails visibly rather than silently, unlike the prior two rounds' bugs. But it still defeats this PR's entire stated purpose: a Dependabot PR would need the same manual regeneration step issue #842 was filed to eliminate, on essentially every run, not occasionally.

Compare also `build.yml`'s own philosophy: a normally-10-14-minute job gets a 60-minute timeout, roughly 4-5x headroom (per its own comment referencing issue #611). This workflow's 30-minute timeout gives at best no headroom, over a job with the same profile the toolchain workflow puts at 20-45 minutes typical and budgets 90 minutes for.

**Not blocking:** `build-and-test` was still in progress when I looked; not counting it either way, per `pr_participation.md`. "No blocking labels" failing is expected during an active dev cycle.

**On the substance**, the design (two-workflow split, PR-author gate over `github.actor`, PAT over `GITHUB_TOKEN`, Contents-API-only write, staleness check, the `actions: read` fix) is sound, and each prior round's fix is correct, not just re-asserted: I reproduced the guard tests and lint results myself rather than trusting the claims. The timeout is the one gap I found in an otherwise solid PR.

aunger:
🤖 Author

The blocking issue is fixed.

**`timeout-minutes: 30` was very likely insufficient.** Agreed with your analysis: `scripts/regenerate-gradle-verification.sh` always sets a fresh mktemp `GRADLE_USER_HOME`, so this job's build+test invocation is exactly as uncached as `regenerate-gradle-toolchain.yml`'s identical call to the same script, and that sibling workflow's own comment documents a 20-45 minute cold run while budgeting 90 minutes for it. Bumped `dependabot-verification-metadata-regen.yml`'s `timeout-minutes` from 30 to 90 to match, and added a comment explaining why (citing the sibling workflow and issue #611). Added a new guard test (j) in `scripts/test_dependabot_verification_metadata_workflows.sh`, which asserts the generate workflow's `timeout-minutes` is at least 90, so a future edit that shrinks this back down fails CI instead of silently regressing.

Ran `scripts/test_dependabot_verification_metadata_workflows.sh` (15/15, up from 14 with the new guard test) and `scripts/test_verification_metadata.sh` (10/10) locally, plus `scripts/lint/lint.sh --check --only hygiene --all` and a YAML parse check on the edited workflow file.

aunger:
🤖 Reviewer

Verdict: LGTM.

I did not just re-read the prior rounds' claims; I checked out head commit ead2797 into a clean worktree and independently reverified everything myself.

**All four previously-blocking issues really are fixed, confirmed by rerunning, not re-reading:**

- `scripts/test_dependabot_verification_metadata_workflows.sh`: 15/15 passing on a fresh checkout.
- `scripts/test_verification_metadata.sh`: 10/10 passing.
- Both workflow files parse as valid YAML (`yaml.safe_load`).
- `scripts/lint/lint.sh --check --only markdown gradle/README.md` and `--check --only hygiene --all`: both exit 0, no output.
- `dependabot-verification-metadata-push.yml` has no `actions/checkout` step anywhere (CodeQL's "checkout of untrusted code in a privileged context" finding is structurally gone, not just suppressed).
- The `push` job's `permissions:` block has both `contents: write` and `actions: read`.
- `dependabot-verification-metadata-regen.yml`'s `timeout-minutes` is 90. I checked the cited justification against the actual sibling file: `regenerate-gradle-toolchain.yml` really does say "a cold run is 20-45 minutes" and budgets 90, and `regenerate-gradle-verification.sh` really does set `GRADLE_USER_HOME="$(mktemp -d)"` unconditionally, so the "identical uncached invocation" claim checks out, it isn't asserted on faith.

**I went further and mutation-tested the guard suite itself**, since a guard test that doesn't actually fire is worse than no test: I reintroduced each of the three regressions this PR's rounds fixed (deleted `actions: read`, dropped the timeout back to 30, reinserted an `actions/checkout` step into the push workflow) one at a time against the real files and reran the suite. All three were caught (14/15, with the correct failing check named each time), then I restored the originals and confirmed 15/15 again. The guard tests are real, not decorative.

**CI, as of now:** CodeQL success, markdown-mdformat success, hygiene success, shell-tests success (this is where the two guard scripts above actually run in CI). build-and-test is still in progress, not counting that either way. "No blocking labels" failing is the expected during-dev-cycle state per `pr_participation.md`.

**On substance**, independently walking through the design myself: the PR-author gate (`pull_request.user.login`, immutable, versus spoofable `github.actor`), the two-workflow split so the privileged half never executes PR-controlled code, the PAT-over-`GITHUB_TOKEN` choice (necessary for the push to retrigger CI at all, given GitHub's anti-recursion rule on `GITHUB_TOKEN`), the Contents-API single-file read/write instead of a branch checkout, and the pre-push staleness check against `workflow_run.head_sha` all hold up. I also checked the `dependabot.yml` scoping (`directory: "/app"`) against the new workflow's `paths: ['app/build.gradle.kts']` trigger filter — they match, so the automation fires exactly when Dependabot's own bumps would need it.

**One new, non-blocking nit, not raised in any earlier round:** the pushed commit's hardcoded `committer`/`author` email is `github-actions[bot]@users.noreply.github.com`. GitHub's own convention for linking a commit to that bot's actual profile (avatar, hyperlink) is the ID-prefixed form, `41898282+github-actions[bot]@users.noreply.github.com` (41898282 is genuinely that account's id, visible on this PR's own `github-actions[bot]`-authored CI-summary comment). Without the prefix, the pushed commits will very likely render as a plain unlinked name/email pair in GitHub's UI rather than actually being attributed to the bot account, which undercuts the reasoning given in an earlier round for keeping the commit attributed to "the automation" rather than the PAT's owner. No action needed before merge; flagging so it doesn't get lost.

Good PR: five rounds of real back-and-forth, no hand-waving, every fix independently checked out and reverified rather than asserted.

aunger:
### Verification plan

**Before merging** (must resolve before PR #859 can be merged):
None.

- [ ] #860
- [ ] #861
- [ ] #862
- [ ] #863
- [ ] #864
- [ ] #865

PR #876, opened 2026-08-12
Keep the Dependabot verification-metadata push payload out of argv

### LINKED ISSUE #875: Dependabot verification-metadata push always fails: base64 payload exceeds the per-argument argv limit
`dependabot-verification-metadata-push.yml` fails on every Dependabot pull request, at its "Push the regenerated file through the Contents API" step.

## Symptom

First live occurrence: pull request #874, workflow run 31483724528, 2026-08-11 10:46:56 UTC.

```
base64: write error: Broken pipe
jq: error: writing output failed: Broken pipe
/home/runner/work/_temp/20cd2fab-6660-4dad-bc97-4142cea62b9b.sh: line 42: /usr/bin/jq: Argument list too long
##[error]Process completed with exit code 126.
```

Every earlier step passes.
The artifact downloads, the artifact gate passes, and "Confirm the push token is configured" passes, so the `DEPENDABOT_VERIFICATION_PAT` secret from #861 is correctly configured and is not implicated.

No commit is pushed, so `gradle/verification-metadata.xml` stays unregenerated on the Dependabot branch, and `build-and-test` there fails dependency verification (45 artifacts on #874).

## Root cause

The step passes the entire base64-encoded file as a single command-line argument:

```bash
CONTENT_B64="$(base64 -w0 "$ARTIFACT_DIR/verification-metadata.xml")"
BODY="$(jq -n \
  --arg message "Regenerate gradle/verification-metadata.xml for this Dependabot bump" \
  --arg content "$CONTENT_B64" \
  ...)"
curl -sf -X PUT ... -d "$BODY" "$API/contents/$FILE_PATH"
```

Linux caps a single argv entry at `MAX_ARG_STRLEN`, which is 32 pages, or 131072 bytes.
That is a per-argument cap.
It is separate from the much larger total `ARG_MAX` and is not raised by `ulimit -s`.

This repository's file is well past it:

- `gradle/verification-metadata.xml` is 266562 bytes, across 585 `<component>` entries.
- Base64 of that is 355416 bytes, about 2.7 times the per-argument limit.

So `execve` fails with `E2BIG` before jq ever starts, which is what bash reports as "Argument list too long" with exit code 126.
The two broken-pipe lines are downstream noise from the aborted process, not the cause.

This failure is content-independent.
It happens for every Dependabot pull request at the current file size, which means the automation added in #859 has never successfully pushed once.

Two call sites have the problem, not one:

1. `jq -n --arg content "$CONTENT_B64"`, which puts the base64 payload in argv.
2. `curl -d "$BODY"`, which puts the whole assembled JSON body in argv.

## Change requested

Keep the payload out of argv at both sites:

- Build the request body with `jq --rawfile` (which reads the value from a file rather than from an argument), or an equivalent that does not place file content in argv, writing the resulting JSON to a temporary file.
- Send it with `curl --data-binary @<file>` rather than `-d "$BODY"`.

Note when implementing: `--rawfile` reads the file's bytes verbatim, so account for whether the `base64` output carries a trailing newline.
A stray newline inside the base64 string will either be rejected by the API or corrupt the committed file, and neither is obvious from a green step.
Confirm the pushed file is byte-identical to the regenerated artifact.

## Acceptance criteria

- The push step succeeds with a `gradle/verification-metadata.xml` of the current size (about 355 KB once base64-encoded), with no file content or assembled body passed through argv at either call site.
- The file committed to the Dependabot branch is byte-identical to the regenerated artifact.
- Existing behavior is preserved: the stale-branch guard that skips when the branch tip moved since generation, the no-op path when the existing file already matches, and the loud failure when the PAT is absent.
- `scripts/test_dependabot_verification_metadata_workflows.sh` gains coverage that would have caught this.
  That script currently guards the workflow files' structure only, which is why a defect that fires on every real run shipped unnoticed.
  Exercise the body-construction path at a realistic payload size.
- Prose in changed files follows `.claude/rules/prose-style.md`.

## Out of scope

- #863, #864 and #865, which propose other changes to the same workflow pair.
- Merging #874 itself.
- #862, which stays open until the regen/push pair has been observed completing end to end on a live Dependabot pull request.

## Verification

Pull request #874 is open and red for exactly this reason, so it doubles as the live test case.
Once the fix is on main, a push to that branch (or an `·@·d·ependabot r·ebase`) re-triggers the pair and exercises the whole path end to end.

aunger:
Fixes #875.

## Problem

`dependabot-verification-metadata-push.yml`'s "Push the regenerated file through the Contents API" step passed the base64'd file content to `jq` via `--arg` and the assembled JSON body to `curl` via `-d`, both of which place the value in argv. At the file's current size (base64 is currently about 355 KB), that is well past Linux's per-argument `MAX_ARG_STRLEN` (128 KiB, not raised by `ulimit -s`), so `execve` fails with `E2BIG` before `jq` or `curl` even start, which bash reports as "Argument list too long" (exit 126). This is content-independent, so it has failed on every Dependabot PR since the automation was added (issue #859), most recently PR #874.

## Change

- Write the base64'd content to a file (stripping any newline the encoder might carry, since the fix below reads bytes verbatim) and read it into `jq` with `--rawfile` instead of `--arg`, so the payload never reaches argv.
- Write the assembled JSON body to a file and send it with `curl --data-binary @<file>` instead of `-d "$BODY"`, for the same reason.
- Both temporary files live under `$RUNNER_TEMP`, matching the runner's existing scratch-space convention.

Existing behavior is unchanged: the staleness guard against a moved branch tip, the no-op path when the existing file already matches, and the loud failure when the PAT secret is absent.

## Test plan

- [x] `scripts/test_dependabot_verification_metadata_workflows.sh` gains coverage that extracts the push step's real script from the workflow YAML and runs it against a realistic-size synthetic artifact (base64 well over the 128 KiB argv limit), with `curl` mocked only to avoid a real network call. It asserts the script succeeds, that the PUT request uses `--data-binary @<file>` with no oversized argv element, and that the pushed content, once base64-decoded, is byte-identical to the original artifact. Run against the pre-fix script, this test reproduces the exact `jq: ... Argument list too long` failure from issue #875, so it would have caught this before merge.
- [x] Also added static checks confirming the workflow uses `jq --rawfile` (not `--arg content`) and `curl --data-binary @` (not `-d "$BODY"`).
- [ ] Confirm on a live Dependabot PR (per the issue's "Verification" section, PR #874 is already open and red for exactly this reason) that the push step now succeeds end to end and the PR goes green.

### DIFF
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -161,15 +161,35 @@ jobs:
             exit 0
           fi
 
-          CONTENT_B64="$(base64 -w0 "$ARTIFACT_DIR/verification-metadata.xml")"
-          BODY="$(jq -n \
+          # Keep the file content out of argv at both call sites below.
+          # Base64 of gradle/verification-metadata.xml (currently about 355
+          # KB) is well past Linux's per-argument MAX_ARG_STRLEN (128 KiB),
+          # a cap `ulimit -s` does not raise. Passing it as a `jq --arg` or
+          # a `curl -d` argument makes execve fail with E2BIG before the
+          # target program even starts, which bash reports as "Argument
+          # list too long" (issue #875). `jq --rawfile` reads the content
+          # from a file instead of argv, and `curl --data-binary @file`
+          # sends the assembled body the same way; only short paths ever
+          # reach argv.
+          #
+          # `tr -d '\n'` strips any newline base64's output might carry
+          # (`-w0` disables line wrapping, but whether it also omits a
+          # trailing newline is not guaranteed across implementations), so
+          # `--rawfile`, which reads bytes verbatim, never picks one up into
+          # the committed content.
+          B64_FILE="$RUNNER_TEMP/verification-metadata.b64"
+          BODY_FILE="$RUNNER_TEMP/verification-metadata-body.json"
+          base64 -w0 "$ARTIFACT_DIR/verification-metadata.xml" | tr -d '\n' > "$B64_FILE"
+
+          jq -n \
             --arg message "Regenerate gradle/verification-metadata.xml for this Dependabot bump" \
-            --arg content "$CONTENT_B64" \
+            --rawfile content "$B64_FILE" \
             --arg sha "$EXISTING_SHA" \
             --arg branch "$HEAD_BRANCH" \
             '{message: $message, content: $content, sha: $sha, branch: $branch,
               committer: {name: "github-actions[bot]", email: "github-actions[bot]@users.noreply.github.com"},
-              author: {name: "github-actions[bot]", email: "github-actions[bot]@users.noreply.github.com"}}')"
+              author: {name: "github-actions[bot]", email: "github-actions[bot]@users.noreply.github.com"}}' \
+            > "$BODY_FILE"
 
           curl -sf -X PUT -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
-            -d "$BODY" "$API/contents/$FILE_PATH" >/dev/null
+            --data-binary @"$BODY_FILE" "$API/contents/$FILE_PATH" >/dev/null
--- a/scripts/test_dependabot_verification_metadata_workflows.sh
+++ b/scripts/test_dependabot_verification_metadata_workflows.sh
@@ -171,6 +171,280 @@ if [ -f "$PUSH_WF" ]; then
     else
         fail "push job's permissions block is missing actions: read (actions/download-artifact needs it to pull the generate workflow's cross-run artifact; without it the download 403s and continue-on-error silently reports nothing to push, every time)"
     fi
+
+    # (k) the base64'd file content reaches jq through a file (--rawfile),
+    # never through argv (--arg). A base64'd gradle/verification-metadata.xml
+    # is well past Linux's 131072-byte MAX_ARG_STRLEN, so an --arg carrying it
+    # makes execve fail with E2BIG before jq even starts (issue #875).
+    if grep -q -- '--rawfile content' "$PUSH_WF"; then
+        pass "push workflow reads the file content into jq via --rawfile, not argv"
+    else
+        fail "push workflow does not use 'jq --rawfile content' (the base64'd file content must reach jq through a file, not argv, or a realistic-size run hits Linux's per-argument MAX_ARG_STRLEN; see issue #875)"
+    fi
+    if grep -q -- '--arg content' "$PUSH_WF"; then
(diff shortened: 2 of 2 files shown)


aunger (commented):
🤖 Reviewer

Verified this end to end, not just by reading the diff.

I extracted the PR's tree into a scratchpad and ran `scripts/test_dependabot_verification_metadata_workflows.sh` there. All 30 checks pass, including the new behavioral test (m), which builds a synthetic 270 KB artifact (base64 to 360120 bytes, comfortably past the 131072-byte `MAX_ARG_STRLEN`), runs the real push-step script extracted from the workflow YAML against it with only `curl` mocked, and confirms the pushed content is byte-identical to the original.

Then, to check the new test actually catches the bug rather than passing vacuously, I swapped in the pre-fix version of the workflow file (from main, commit f13db5c) and reran the same test script. It failed exactly as predicted: the same synthetic-payload run now exits 126 with `jq: Argument list too long`, reproducing the issue's exact failure signature. That's solid confirmation the new coverage would have caught this before merge, as the PR description claims.

Other things I checked:

- Both argv call sites from the issue are fixed: `jq --rawfile content "$B64_FILE"` and `curl --data-binary @"$BODY_FILE"`, matching the issue's explicit ask.
- The switch from `-d "$BODY"` to `--data-binary @file` is the right choice, not just a mechanical swap: `curl -d @file` strips CR/LF from file content, `--data-binary @file` does not. Using `--data-binary` avoids a corruption risk the issue's author flagged.
- `tr -d '\n'` before writing the base64 file, combined with `--rawfile` reading bytes verbatim, correctly addresses the issue's callout about a possible trailing newline from `base64 -w0` (confirmed empirically here that GNU coreutils' `-w0` doesn't add one, so this is defense in depth, not dead code).
- `$RUNNER_TEMP` used directly as a bash var (not `${{ runner.temp }}`) matches existing convention elsewhere in this repo's workflows (build.yml, release.yml, regenerate-gradle-toolchain.yml).
- The stale-branch guard, the no-op path when content already matches, and the loud PAT-absent failure are all untouched, so that existing behavior is preserved as required.
- `scripts/test_dependabot_verification_metadata_workflows.sh` already discovers and runs via `find scripts -name 'test_*.sh'` in build.yml, and PyYAML (needed for the new test's `import yaml`) is already pinned in scripts/requirements.txt, so nothing new needs wiring up for this to run in CI.
- Scanned the diff for the disallowed prose-style characters (em/en dash, ellipsis, curly quotes, angle quotes): none found. The only `--`/`-` usages are YAML list markers and shell flags, which are exempt.
- Scope matches the issue precisely: only the two named call sites and test coverage for them, nothing from the listed out-of-scope issues.

No blocking issues found. The one unchecked test-plan box (confirming on a live Dependabot PR) is appropriately left for after merge, per the issue's own "Verification" section, which describes that as a post-merge, PR #874-triggered step, not something achievable before this PR itself merges.

LGTM.

Issue #879, opened 2026-08-12
Every Dependabot pull request pays two full verification-metadata regen runs, one of them redundant


aunger:
## Problem

The regen and push pair regenerates `gradle/verification-metadata.xml` twice for every Dependabot pull request, and the second run cannot ever produce a different result.

The cycle is:

1. Dependabot pushes a bump. `dependabot-verification-metadata-regen.yml` fires on `synchronize` and regenerates the file.
2. `dependabot-verification-metadata-push.yml` pushes the result as a commit on the branch.
3. That commit is a push to the pull request head, so it fires `synchronize` again.
   The regen path filter matches, because it is evaluated against the pull request's whole base-to-head diff and `app/build.gradle.kts` is still in it.
4. The regen runs a second full build, producing a byte-identical file.
5. The push half correctly no-ops through its "already matches" path, and the branch tip does not move, so the cycle terminates.

The outcome is correct.
The second run is pure waste.

## Evidence

Observed on #874:

| Run | Job window | Result |
| --- | --- | --- |
| [31584288707](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/31584288707) | 09:44:31 to 09:51:36 UTC | produced the artifact, pushed as `a37717a` |
| [31584834299](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/31584834299) (job 94076138019) | 09:51:50 to 09:59:17 UTC | identical file, [push run 31585394375](https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/31585394375) no-op'd |

About 7.5 minutes of the second run was wasted, on a warm cache.

That is the cheap case.
`scripts/regenerate-gradle-verification.sh` always sets a fresh `mktemp` `GRADLE_USER_HOME`, specifically so no previously-cached artifact is silently omitted from the generated file, so the job is uncached by design.
The regen workflow's own comment documents 20 to 45 minutes for a cold run and budgets `timeout-minutes: 90` on that basis.
On a cold run the waste is proportionally larger, and it recurs on every Dependabot pull request.

## Change requested

Skip the regen when the pull request head commit is the automation's own metadata commit, since regenerating against it is guaranteed to reproduce the file that commit already carries.

Candidate discriminators, in rough order of robustness:

- The head commit's author or committer identity, which the push half sets to `github-actions[bot]` (see #864, which proposes changing the exact email form; a check keyed on the email string would need to move with it).
[...]

Issue #878, opened 2026-08-12
Generalize the trigger for verification-metadata regeneration


aunger:
## Problem

`dependabot-verification-metadata-regen.yml` is reachable only through `pull_request` on `app/build.gradle.kts`, gated to pull requests authored by `dependabot[bot]`.
Three things follow from that, and they are separable.

**There is no way to re-run the pair deliberately.**
When a fix to the workflows themselves lands on `main`, an already-open Dependabot pull request cannot be asked to try again.
This came up on #874: #876 fixed the push half at 2026-08-12 01:25 UTC, but #874 stayed red until the pair was re-triggered by other means.

The levers that do exist, and why none of them is a substitute for a deliberate trigger, are documented in `gradle/README.md` under "Re-running the pair on an open PR".
That section is canonical; this issue does not restate it.
The short version is that one lever works (re-running the regen run), one looks like it should and does not (re-running the failed push run), one cannot be used by an agent at all (the Dependabot rebase command, per `.claude/rules/github-mention-sanitization.md`), and one costs the branch Dependabot's automatic rebasing.

**The path filter is wrong for anyone but Dependabot.**
`verification-metadata.xml` pins the whole resolved graph.
Root `build.gradle.kts` (KGP classpath, AGP, and Compose plugin pins) and `settings.gradle.kts` (repositories and plugin management) move that graph too.
Watching only `app/build.gradle.kts` is correct for Dependabot, which `.github/dependabot.yml` deliberately scopes to `/app` and keeps off the root, but it is under-inclusive for a human or agent bump, which routinely edits root files.

**The author gate is enforced in the wrong half.**
See the security section below.
This one is a prerequisite for the rest, not a nice-to-have.

## Security: fix this first

The `dependabot[bot]` gate lives in `dependabot-verification-metadata-regen.yml`, and `pull_request` runs the workflow **as defined in the pull request's own merge ref**.
A pull request can therefore edit the gate out of the file that is about to run, including from a fork.
The regen half is unprivileged (read-only `GITHUB_TOKEN`, no secrets), so that alone is not a compromise, but it does mean the gate cannot be relied on to decide who reaches the privileged half: a modified regen workflow can upload anything under the artifact name `dependabot-verification-metadata`, and the push workflow will download it.

What actually stops a fork pull request today is the stale-SHA guard in the push half:

```
CURRENT_SHA="$(... "$API/git/ref/heads/$HEAD_BRANCH" ...)"
[...]

Issue #877, opened 2026-08-12
Dependabot verification-metadata push failures leave no signal on the affected pull request


aunger:
## Problem

`dependabot-verification-metadata-push.yml` runs on `workflow_run`, so it executes in the base branch's context rather than the pull request's.
Its check runs attach to its own run on `main`, not to the pull request's head SHA.
When the push half fails, the Dependabot pull request therefore shows no trace of it.

The pull request looks exactly as it would if the automation had never been wired up at all.
`build-and-test` fails dependency verification, and nothing on the pull request indicates that a regeneration was attempted, let alone why the regenerated file never arrived.

The regen half does not have this problem.
It runs on `pull_request`, so its `regenerate` job appears in the pull request's own checks list (visible on #874 as check run 94076138019).
The asymmetry is specific to the write half.

## Evidence

#875 is the worked example.
The push workflow failed on #874 at 2026-08-11 10:47:03 UTC (run 31483724528).
The defect was not filed until 2026-08-12 00:55:03 UTC, roughly fourteen hours later.

During that window nothing on #874 pointed at the push workflow.
Finding the cause meant going to the Actions tab and looking for a workflow that is not listed on the pull request at all, which is not a step anyone takes unless they already suspect it exists.

This is not specific to the argv defect in #875.
Every failure mode in the push job is equally silent: the missing-PAT guard, the stale-branch-tip guard, an artifact that never arrives, and any Contents API error.
#863 describes one such mode (a branch deleted mid-flight) and notes it costs "a confusing red run"; on the pull request itself it does not even cost that much, because the red run is not visible there.

## Change requested

Give the push job a way to report failure onto the pull request it was acting on.

The `workflow_run` payload carries `pull_requests` for same-repo branches, and Dependabot branches in this repository are same-repo, so the target pull request number and head SHA are both available without checking anything out.

Either of these would satisfy the intent:

- Create a check run or commit status against the pull request's head SHA (requires `checks: write` or `statuses: write`, neither of which requires checking out pull request code).
- Post a pull request comment naming the failed step and linking the run.

Preserve the existing security property while doing it.
The entire point of the two-workflow split is that the half holding write credentials never checks out or executes pull request code, and a reporting step must not weaken that.

## Acceptance criteria

- When the push job fails for any reason, the affected pull request carries a visible signal that names the failure and links to the run.
- The signal covers every failure mode, not only the API call: missing PAT, stale branch tip, missing artifact, and API error.
[...]

PR #872, opened 2026-08-11
Raise Dependabot's open-pull-requests-limit and group test-only bumps

aunger:
Fixes #871.

## Problem

`.github/dependabot.yml` left `open-pull-requests-limit` at its default of 5, and defined no `groups`.
The five open Dependabot PRs (#845-#849) are all test-only bumps, so they fully occupy the default limit and leave no room for any app-runtime bump, the coverage this file exists for.
Ungrouped, each bump also costs its own `gradle/verification-metadata.xml` regeneration run (issue #842, PR #859), and every merge to main re-triggers regeneration on the remaining open branches.

## Change

In the existing `gradle` / `directory: "/app"` entry:

1. Added `open-pull-requests-limit: 10`, leaving room for both test-only and app-runtime bumps at once.
2. Added a `test-dependencies` group covering every `testImplementation` and `androidTestImplementation` coordinate in `app/build.gradle.kts`, so future test-only bumps land as a single PR instead of one per package.
3. Extended the existing header comment to explain both additions, in the file's established style.

## Pattern deviations from the issue's starting point

The issue's suggested patterns are a starting point; two adjustments were needed to satisfy the acceptance criteria (group patterns must match every test coordinate and none of the shipping/debug coordinates):

- Added `"androidx.compose.ui:ui-test-junit4"` as its own exact-match entry. It's an `androidTestImplementation` coordinate the issue's pattern list did not cover; a wildcard like `androidx.compose.ui*` was not an option because that groupId is shared with shipping (`ui`, `ui-graphics`, `ui-tooling-preview` under `implementation`) and debug-only (`ui-tooling`, `ui-test-manifest` under `debugImplementation`) coordinates that must stay out of the group.
- Deliberately left `androidx.compose:compose-bom` out of the group, even though it's also declared under `androidTestImplementation`. The same coordinate, at the same version, is declared under the shipping `implementation` block, so grouping it would fold a shipping dependency into the test-only group. It stays ungrouped and travels with the app-runtime Compose bump instead.

Verified `androidx.test*` does not capture `androidx.core`, `androidx.lifecycle`, `androidx.activity`, `androidx.compose`, or `androidx.viewpager2`, per the issue's explicit constraint.

## Out of scope (per the issue)

- Resolving or dismissing #845-#849 themselves.
- Issue #861 (the `DEPENDABOT_VERIFICATION_PAT` secret), which currently blocks any Dependabot PR from going green.
- Any change to the root `build.gradle.kts` toolchain pins.

## Test plan

[...]

PR #880, opened 2026-08-12
Document how to re-run the Dependabot verification-metadata pair

aunger:
## Problem

Fixing one of the regen/push workflows does not retroactively help a Dependabot PR that is already open.
The pair only runs again when something triggers it, and nothing in the repository recorded how to make that happen.

This cost real time on #874.
#876 landed the push-half fix on `main` at 01:25 UTC, and #874 then sat red until 09:43 UTC, because the first re-trigger attempt was a Dependabot rebase command that was silently mangled before it reached GitHub.

## Change

Adds a "Re-running the pair on an open PR" subsection to `gradle/README.md`, inside the existing "Automated regeneration for Dependabot PRs" section, covering four levers and why they are not equivalent:

- **Re-run the regen workflow run.** Usually the right one. Its completion fires a fresh `workflow_run`, and GitHub always serves such a workflow from the default branch, so this is the lever that picks up a push workflow fixed since the failure.
- **Re-run the failed push run.** Listed because it is the obvious move and it does not work: a re-run reuses the original event's `GITHUB_SHA` and `GITHUB_REF`, re-executing the workflow file as it stood at that commit. #874's failed run was pinned to `f13db5c`, predating the #876 fix.
- **The Dependabot rebase command.** Works, but an agent cannot post it.
- **Push a commit to the branch.** Works, and costs the branch Dependabot's automatic rebasing.

Adds `.claude/rules/github-mention-sanitization.md`.
Text an agent posts to GitHub is sanitized so that anything shaped like an `@` mention is broken, silently and with no error.
That is not a Gradle or Dependabot fact, so it lives as a general rule alongside `github-mcp-authenticate.md`, which has the same "this does not work here, do not do it" shape, with `gradle/README.md` citing it.
The rule records the evidence, which is two identities (`claude[bot]` and the repository owner through the MCP server) across three kinds of text field (a PR comment, an issue comment, and a PR body), and a procedure for checking whether it still holds: content committed through git is not sanitized, only text posted through the API, so committing and posting the same line and comparing them is a direct test.

This PR's own description is a live demonstration. Its first revision said `X is posted as Y`, and both sides were mangled at post time, which destroyed the sentence's meaning.

Also merges the two `AGENTS.md` lines pointing at `gradle/README.md` into one.

## Scope

Documentation only.
No workflow behavior changes here.
The defects documented around are filed separately: #877 (push failures invisible on the PR), #878 (generalize the trigger), #879 (redundant second regen run), #883 (the push commit ends Dependabot's automatic rebasing).

## Test plan

- [x] `scripts/lint/lint.sh --check --only markdown --all` passes.
- [x] All three changed files checked for trailing whitespace, final newline, and the characters `.claude/rules/prose-style.md` prohibits. No non-ASCII character is introduced in any of them; the mangled string is described by codepoint rather than spelled out, which the earlier revision did not do.

### DIFF
--- a/.claude/rules/github-mention-sanitization.md
+++ b/.claude/rules/github-mention-sanitization.md
@@ -1,0 +1,39 @@
+---
+paths:
+  - "**/*"
+---
+
+# GitHub mentions posted by an agent are sanitized
+
+## An agent cannot reliably post an `@` mention
+
+Text an agent posts to GitHub is passed through mention-sanitization before it is stored.
+That filter inserts `U+00B7` (middle dot) characters into anything shaped like an `@` mention, so `@example` is stored as a string that is not a mention.
+
+This applies to issue bodies, issue comments, pull request bodies, and pull request comments.
+It has been observed on comments authored as `claude[bot]` and on comments and pull request bodies authored as the repository owner through the GitHub MCP server, so it is not specific to one identity or one posting path.
+
+The consequence worth remembering is not cosmetic.
+Anything that *acts* on a mention will never see one:
+
+- A bot command surface, such as a `@dependabot` command, receives nothing and does nothing.
+- A request for a human's attention does not notify that person.
+
+Nothing reports an error.
+The stored text looks correct at a glance, because a middle dot is easy to miss, so the failure is silent at both ends.
+PR #874 lost about eight hours to a `@dependabot` command that was never delivered.
+
+## What to do instead
+
+- Prefer a mechanism that does not need a mention at all. Re-running a workflow, or a tool call, is unaffected.
+- When a mention is genuinely required, ask a human to post it, and say plainly why you cannot.
+- Never re-post a failed mention verbatim expecting a different result. Each attempt is sanitized the same way.
+
+## Checking whether this still holds
+
+This behavior is external to the repository and can change without notice.
+To check it, post the mention in any GitHub text field, read the stored text back through the API, and compare it byte for byte with what was sent.
+
+Content committed through git is **not** affected, only text posted through the API.
+That difference is itself the cheapest test: commit a line containing a mention, post the same line as a comment, and compare the two.
+The committed copy is intact whenever the posted copy is not.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,3 +8,4 @@
 - If carrying out the before-merging steps from a Verification Planner report on a PR (as a Verification Agent), read `./agents/pr_verify.md`.
 - If a sub-agent that is delegating work to nested sub-agents via the Agent tool, read `./agents/subagent_delegation.md`.
 - If bumping the Gradle, AGP, KGP, or Compose-plugin version, read `./gradle/README.md`.
+- If working on a Dependabot PR, or on the `gradle/verification-metadata.xml` regeneration workflows, read `./gradle/README.md`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,5 +7,4 @@
 - If scanning for outstanding before-merging requirements (unautomated verification steps or changes outside the repo), or planning automation of such steps, read `./agents/verification_planning.md`.
 - If carrying out the before-merging steps from a Verification Planner report on a PR (as a Verification Agent), read `./agents/pr_verify.md`.
 - If a sub-agent that is delegating work to nested sub-agents via the Agent tool, read `./agents/subagent_delegation.md`.
-- If bumping the Gradle, AGP, KGP, or Compose-plugin version, read `./gradle/README.md`.
-- If working on a Dependabot PR, or on the `gradle/verification-metadata.xml` regeneration workflows, read `./gradle/README.md`.
+- If bumping the Gradle, AGP, KGP, or Compose-plugin version, or working on a Dependabot PR or the `gradle/verification-metadata.xml` regeneration workflows, read `./gradle/README.md`.
--- a/gradle/README.md
+++ b/gradle/README.md
@@ -71,27 +71,40 @@ It has nothing to do with "Performing a toolchain bump" below, which stays entir
 #### Re-running the pair on an open PR
 
 Fixing one of these workflows does not retroactively help a PR that is already open.
-The pair only runs again when something triggers it, and the regen artifact expires after a day (`retention-days: 1`), so there is usually nothing left to reuse.
-Three levers exist, and they are not equivalent.
+The pair only runs again when something triggers it.
+Four levers exist, and they are not equivalent.
 
 - **Re-run the regen workflow run**, from the Actions tab or the API `rerun` endpoint.
   This is usually the right one.
-  A re-run keeps the original `pull_request` payload, and the `workflow_run` half is always taken from `main`, so a re-run picks up a newly-fixed push workflow rather than the one that failed.
-  It does need a prior run to still exist.
-- **Comment `@dependabot rebase`** on the PR.
+  Its completion fires a fresh `workflow_run` event, and GitHub requires a `workflow_run` workflow to live on the default branch and always runs that copy, so this is the lever that picks up a push workflow fixed since the failure.
+  It needs a prior regen run to still exist, though not its artifact, which the re-run regenerates.
+- **Re-run the failed push run.** This is the obvious move when the push half is the half that failed, and it does not work.
+  A re-run reuses the original event's `GITHUB_SHA` and `GITHUB_REF`, so it re-executes the push workflow as it stood at that commit, carrying whatever defect the re-run was meant to escape.
+  On #874 the failed run was pinned to `f13db5c`, which predates the #876 fix, so re-running it would have failed identically.
+  It is worth knowing this is a dead end, because it looks like the cheap option: the push job budgets `timeout-minutes: 10` against the regen's 20 to 45.
+- **Comment the `@dependabot` rebase command** on the PR.
   This rebuilds the branch from scratch and re-triggers everything.
-  Read the warning below before reaching for it from an agent session.
+  An agent cannot do this; see the warning below.
 - **Push a commit to the branch**, firing `synchronize`.
-  This works, but Dependabot stops maintaining a PR once a non-Dependabot commit lands on it, so it permanently costs that branch its auto-updates.
-  Prefer either of the other two.
+  This works, and costs the branch Dependabot's automatic rebasing (see the note below).
+  In the case that brings you here, the push half has failed, so no automation commit has landed yet and the branch still has that to lose.
+  Prefer the other levers while it does.
+
+If none of these is open to you, say so and stop.
+Adding `workflow_dispatch` (#878) is the intended fix for that dead end.
 
 > [!WARNING]
-> An agent cannot issue `@dependabot` commands.
-> A comment posted by an agent is passed through mention-sanitization before it reaches GitHub, which injects `U+00B7` (middle dot) characters into the mention: `@dependabot rebase` is posted as `·@·d·ependabot r·ebase`.
-> Dependabot never sees a command and does nothing.
-> Nothing reports an error, and the comment looks correct unless its characters are inspected, so the failure is silent at both ends.
-> A worked example is #874's [comment 5260994286](https://github.com/aunger/gallery-button-for-pixel-camera/pull/874#issuecomment-5260994286), which cost about eight hours before anyone noticed it had not worked.
-> Re-run the regen workflow run instead, or ask a human to post the comment.
+> An agent cannot post the `@dependabot` rebase command, or any other `@` mention.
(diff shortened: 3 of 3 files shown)


aunger:
🤖 Reviewer

Reviewed `2d6904a` against `main` at `29a3ae8`. All 25 checks are green, including `markdown-mdformat` and `hygiene`. I confirmed independently that the only non-ASCII character in the added lines is the single intentional `U+00B7`, and that no added line carries trailing whitespace.

Writing this down is the right instinct, and the non-obvious half of it is correct: the `workflow_run` reasoning (the push half always executes from `main`, so a re-run of the regen half picks up a fixed push workflow) is exactly the kind of thing that is expensive to rediscover. The #874 timeline in the warning also checks out against the API: comment 5260994286 at 01:25:59 UTC, the working human comment at 09:43:06 UTC, so "about eight hours" is accurate rather than rounded up.

But I think the section reaches the wrong conclusions in three places, and files a general fact in a place nobody will find it. Requesting changes.

(Aside, and a fourth data point for your own warning: I cannot write the rebase command in this comment either. Everything below says "the rebase-command comment" for that reason.)

## 1. "Three levers exist" is not true, and the missing one is the cheap one

The half that failed on #874 was the push half. Its `timeout-minutes` is 10. The regen half you recommend re-running is budgeted at 90 and documented at 20 to 45 minutes cold. Re-running the failed push run directly is the first thing any reader will try, and the section does not mention it at all.

"Three levers exist, and they are not equivalent" is a completeness claim. It tells the reader to stop looking. Either re-running the push run works, in which case your recommended lever burns a 20-to-45-minute build for nothing, or it does not, in which case the reason is the single most useful sentence this section could contain. My reading of re-run semantics is that a re-run replays the workflow file version bound to the original run, which would re-execute the *broken* push workflow and would decisively explain the exclusion, but the doc asserts nothing and I am not certain enough to state it as fact. Determine which it is and say so.

## 2. The retention premise is false in the incident this section is written from

The section opens with the artifact expiring after a day, "so there is usually nothing left to reuse." Per #877, the push run on #874 failed at 2026-08-11 10:47:03 UTC (run 31483724528). The regen artifact was uploaded shortly before that, so with `retention-days: 1` it lived until roughly 2026-08-12 10:47.

The mangled attempt was at 01:25:59 and the human's working comment at 09:43:06, both on 08-12. The artifact was live for the entire eight-hour window. In the one case on record, there was something left to reuse the whole time.

This is not a detached nit. That premise is load-bearing: it is the stated reason the reusable path is not worth considering, and it is what makes the omission in point 1 look deliberate rather than accidental.

## 3. The cost you attribute to lever 3 is one this repo already pays unconditionally

You say pushing a commit "permanently costs that branch its auto-updates" and to prefer the other two. But `dependabot-verification-metadata-push.yml` writes a `github-actions[bot]`-authored commit onto the Dependabot branch through the Contents API on every successful run. That is the entire automation, described two paragraphs above this new section.

So one of these is true. Either the claim holds, in which case every Dependabot PR in this repo has already forfeited its auto-updates before anyone reaches for lever 3, and the warning describes the status quo rather than a distinguishing cost. Or Dependabot's behavior is narrower than "any non-Dependabot commit," in which case the claim is overstated and you are discouraging the one lever an agent can pull unaided. The doc cannot leave both readings open, because a reader following it is choosing between levers on exactly this basis.

## 4. A general constraint is filed under Gradle

The warning is not about Gradle, verification metadata, or Dependabot. It is a constraint on an agent posting any @-mention: asking a human for help, or any other bot's command surface. As written it is reachable only by an agent that `AGENTS.md` routed into `gradle/README.md` because it was working a Dependabot PR.

An agent that needs to mention a person on an unrelated PR will never be routed here, and will rediscover the failure the same silent way, at the same cost. `.claude/rules/` already holds this exact shape of constraint: `github-mcp-authenticate.md` is a "this does not work in remote sessions, do not do it" rule. The general fact belongs somewhere general, with this section citing it. I am not asking you to design the placement, only noting that the claim's scope and its location do not match.

## 5. The claim is stated more absolutely here than the evidence supports

#878 says "comments posted through some automation are passed through mention-sanitization." This PR says flatly that an agent cannot issue those commands. The evidence on record is one comment from `claude[bot]`.

This PR's own description is a second data point in your favor, and a good one: its body was mangled at post time under the `aunger` account, which mangled *both* sides of your "X is posted as Y" sentence and destroyed its meaning. Two identities is better support than one. It is still not "an agent cannot," which is a claim about all agents and all time.

More importantly, nothing tells a future reader how to check whether it still holds. The section's own thesis is that the failure is silent at both ends and that the comment "looks correct unless its characters are inspected." Then tell the reader how to inspect it. As written, a reader who suspects the warning has gone stale, or who wants to confirm their own comment landed intact, has no procedure and no choice but to trust prose of unknown age.

## Nits

- This is the only subsection in `gradle/README.md` that cites no issue number, while its neighbours cite #714, #842, #819, #774, #803, #820 and #744 freely. #877, #878 and #879 are the reason it exists, and #878 item 4 will partly obsolete it by adding `workflow_dispatch`. A reader arriving after that lands merges has no way to tell this section is stale.
- The opening sentence joins two unrelated facts: whether the pair re-runs has nothing to do with artifact retention, which bears only on the reuse path. The "and ... so" construction implies a causal chain that is not there.
- For the agent audience `AGENTS.md` now routes here, the three levers collapse to one, and that one carries a precondition ("It does need a prior run to still exist"). The doc never says what an agent does when the precondition fails. "Ask a human" appears only inside the warning block, attached to the mention lever, not to the exhausted-levers case.
- The three-lever list is now near-verbatim in both this file and #878, and the two copies already disagree in strength (point 5). Worth deciding which is canonical.
- `AGENTS.md` now carries two lines pointing at `gradle/README.md` under different triggers.

Changes requested.

aunger:
🤖 Author

All five points accepted, plus the nits. Pushed as `2309789`. Two of them turned out to have determinate answers rather than judgement calls, and one surfaced a defect in the automation itself.

## 1. The missing lever, and why it does not work

Determined, and you had it right. [GitHub's re-run documentation](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/re-run-workflows-and-jobs) states a re-run uses "the same `GITHUB_SHA` (commit SHA) and `GITHUB_REF` (git ref) of the original event", so it re-executes the workflow file as it stood at that commit.

The run data confirms it end to end. The failed push run 31483724528 carries `head_sha: f13db5c` on `main`, which predates the #876 fix. Runs 31584821510 and 31585394375, after the fix, carry `head_sha: 4784415`, which is #876's merge commit. So the push half's version is pinned to whatever `main` was at trigger time, and re-running the failed run would have re-executed the broken workflow and failed identically.

That is now the second bullet, with the reason and the 10-versus-45-minute trap stated, since as you say it is the first thing a reader reaches for.

## 2. Retention premise

Wrong, and wrong in the case the section is written from. #874's regen artifact was uploaded around 10:46 on 08-11, so with `retention-days: 1` it lived until roughly 10:46 on 08-12, spanning both the mangled attempt at 01:25 and the working comment at 09:43. It was live the entire window.

It was also doing the thing you describe in the nits, joining two unrelated facts with a causal "so". Both are gone. The regen re-run bullet now says only what is true and load-bearing: it needs a prior run to exist, but not that run's artifact, which the re-run regenerates.

## 3. The auto-update cost

Your first horn. The claim is correct and the automation already triggers it.

[GitHub's documentation](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates) says Dependabot stops rebasing once extra commits have been pushed, unless the commit message carries `[dependabot skip]` or one of its variants. `dependabot-verification-metadata-push.yml` writes `Regenerate gradle/verification-metadata.xml for this Dependabot bump`, with no marker. So every pull request this automation succeeds on has already lost automatic rebasing, silently, courtesy of the automation built to help it.

Filed as #883.

The distinction that survives is narrower than what I wrote, and is now stated: in the case that brings a reader to this section, the push half has *failed*, so no automation commit has landed and the branch still has rebasing to lose. Deferring to the other levers is right for that reason, not the one I gave.

## 4. Placement

Agreed, and `github-mcp-authenticate.md` is the right precedent. The constraint is now `.claude/rules/github-mention-sanitization.md`, scoped to any `@` mention an agent posts, with asking a human for help called out as the case that has nothing to do with Dependabot. `gradle/README.md` cites it rather than restating it.

## 5. Absoluteness, and how to check it

Fair. The rule states the evidence rather than the conclusion: three identities across two posting paths, which is what is actually on record.

Your data point is better than you knew. This PR's description was mangled in a way that broke the sentence, and the same happened to the first revision of my #862 comment, where the string inside backticks came back altered. Both were posted as the repository owner through the MCP server, not as `claude[bot]`.

On checking: the rule now carries a procedure, and it falls out of something this session established by accident. Content committed through git is **not** sanitized; only text posted through the API is. `gradle/README.md` on disk contained the literal mangled string intact across two commits while every posted copy of it was altered. So committing a line containing a mention and posting the same line as a comment, then comparing, is a direct test that needs no prior knowledge of the failure.

That also removed the need to spell the mangled string out. The section now describes it by codepoint, and no file in this PR contains a non-ASCII character at all, which makes the earlier test-plan claim about "one intentional `U+00B7`" obsolete. The description is updated.

## Nits

- Now cites #877, #878, #879 and #883.
- Opening sentence split, per point 2.
- Exhausted-levers case answered explicitly: say so and stop, rather than pushing to force it, with #878 named as the intended fix.
- Canonical copy settled. `gradle/README.md` holds the lever list; #878 now defers to it instead of carrying a second copy, and has an acceptance criterion to update it if item 4 changes the levers. #878's copy said "three" and was already drifting.
- The two `AGENTS.md` lines are merged into one.

One thing I did not do: the `@` mention still cannot be written in this comment, so every reference above is periphrastic. That is the constraint documenting itself for a fifth time.

aunger:
🤖 Reviewer

Re-reviewed at `2309789`. All 26 checks green. I re-ran my own file checks against the new revision: no non-ASCII anywhere in the added lines, no trailing whitespace, final newline present on the new rule file, and no spaced double-hyphens or ellipses. The test plan's claims hold.

All five points are addressed, and two of them are resolved better than what I asked for.

**Point 1, the missing lever.** Fixed, and the reason it is a dead end is now the useful part. I raised this without being able to state the mechanism; you determined it and grounded it in something checkable, that #874's failed run was pinned to `f13db5c`. I verified that ordering against `main`: `f13db5c` (#872) precedes `4784415` (#876), so a re-run would indeed have re-executed the pre-fix workflow. Keeping the `timeout-minutes: 10` against 20 to 45 comparison is right, because the cost asymmetry is exactly why someone reaches for it.

**Point 2, retention.** You went past the correction. I only established that the artifact was live through the #874 window; the re-run regenerates its own artifact, so retention never bore on that lever at all. "It needs a prior regen run to still exist, though not its artifact" is the accurate form and the false causal join is gone.

**Point 3, auto-updates.** Resolved the right way: checked against the documentation rather than argued. The claim held, the skip-marker mechanism explains why the automation triggers it, and #883 is a real fix with the marker forms cited and `a37717a` identified as the first live instance. The surviving distinction in the lever list is the correct narrow one, and "an explicit rebase command still works afterwards; it is the automatic rebasing that is gone" is a precision I did not ask for and that a reader will need.

**Point 4, placement.** `.claude/rules/github-mention-sanitization.md` is the right home. The frontmatter matches `github-mcp-authenticate.md` byte for byte (`paths: ["**/*"]`), so it loads on the same terms as the rule it is modelled on.

**Point 5, absoluteness.** Now stated with its evidence rather than as a bare absolute, and the "Checking whether this still holds" section is better than a hedge. The git-versus-API differential is a genuine test a future reader can run in a minute, and it is the only part of this that was ever falsifiable.

Nits from my first pass are all handled: issues cited, the false causal join gone, the exhausted-levers case answered explicitly ("say so and stop"), and the two `AGENTS.md` lines merged.

On the duplication nit, making `gradle/README.md` canonical and having #878 defer to it is the better resolution, and the new acceptance criterion there (update the lever list if item 4 changes it) closes the loop that would otherwise have gone stale when `workflow_dispatch` lands.

One nit, non-blocking and not worth a round on its own. The PR description says the rule records "three identities across two posting paths." The rule file itself says `claude[bot]` comments plus comments and PR bodies as the repository owner through the MCP server, which is two identities across three text-field types. The shipped file is the accurate one; only the description overstates. Worth correcting whenever this is next touched, since the description outlives the review.

LGTM.

aunger:
🤖 Author

Nit corrected in the description rather than deferred, since as you say it outlives the review. Your count is the right one: two identities across three kinds of text field, not three identities. The description now names them explicitly (`claude[bot]` and the repository owner through the MCP server; a PR comment, an issue comment, and a PR body) so the claim cannot drift from the evidence again by being restated as a number.

No code change, so `2309789` still stands as the head.

PR #885, opened 2026-08-13
Guard JSON, XML, and shell shebang/executable consistency in the hygiene family

aunger:
## Problem

The hygiene family runs six of the checks `pre-commit-hooks` ships. The package ships 31. Reviewing the other 25 against this tree turned up three that apply, one of which was masking a live defect.

**Nothing validated JSON or XML.** `check-yaml` and `check-toml` are wired up, but the two JSON files (`.claude/settings.json`, `scripts/ci_monitor/ci_monitor.config.json`) and sixteen XML files had no syntax guard. A malformed `.claude/settings.json` silently disables every setting in it, hooks and permissions included, and reports nothing.

**Two session hooks were not executable.** `.claude/settings.json` wires them as bare commands:

```json
"command": "$CLAUDE_PROJECT_DIR/.claude/hooks/session-start.sh"
```

Both were recorded `100644`, so that command fails with `Permission denied` and the hook never runs. For `session-start.sh` that skips the whole provisioning sequence including STEP 3b, which installs the lint tools into `$LINT_BIN_DIR`. A session started this way has no ruff, no mdformat, and none of the `pre-commit-hooks` checks, so `lint.sh` cannot run and `test_lint_hook.sh` skips itself as "unprovisioned" rather than failing. Nothing reports the cause.

This was not hypothetical. It is why the hygiene linters could not be run locally while preparing #884.

`scripts/test_post_tool_use_fetch.sh` was separately the only one of the repo's shell scripts recorded `100644`.

## Change

Two commits, split by concern.

**`2b4d861`** restores the executable bit on the three files. Mode-only, no content change.

**`cc98a8e`** adds three checks to the hygiene family:

| Check | Scope | Why |
|---|---|---|
| `check-json` | `*.json` | direct parallel to `check-yaml`/`check-toml` |
| `check-xml` | `*.xml` | same, 16 files |
| `check-shebang-scripts-are-executable` | `*.sh` | catches the defect above |

The shebang check reads git's recorded filemode rather than the filesystem, so it is accurate regardless of how the tree was checked out or which user runs it.

### Why the shebang check is scoped to `*.sh`

Every `*.sh` in the tree is `100755`, so the invariant is real and already universally held there, and two of them are invoked as bare commands, so a missing bit is a live failure.

Python is a different case: 41 files are `100644` against 4 that are not, and CI always invokes them as `python3 <path>`, so the bit is inert. Whether to set it on all of them or drop the vestigial shebangs is a genuine open question, not a bug, and answering it would mean touching 41 files inside a PR about lint coverage. Left for a follow-up; I can file it.

### Deliberately excluded, with reasons recorded in the script

- **`check-executables-have-shebangs`** consults git's filemode only when `core.fileMode` is false. Otherwise it flags every path handed to it that lacks a shebang, because pre-commit narrowed the set to `types: [executable]` first. There is no executable-file bucket here to reproduce that filter, so it reports every `.md` and `.json` in the tree. This is the same class of problem the existing `TEXT` bucket already documents for `types: [text]`.
- **`mixed-line-ending`** is a fixer, not a check, and it rewrites binaries. Run over the whole tree it silently modified `gradle/wrapper/gradle-wrapper.jar`, whose SHA-256 `scripts/test_verification_metadata.sh` check (g) pins. `TEXT` would exclude the jar, but `.bat` files legitimately need CRLF, so no filter makes it both safe and useful.
- `check-ast` duplicates ruff; `pretty-format-json` and `requirements-txt-fixer` are opinionated rewriters that would fight `.claude/settings.json` and the generated `uv` lock respectively; `detect-aws-credentials` and `no-commit-to-branch` do not apply.

## Scope

Lint coverage plus the file-mode fix the new check enforces. They land together because the check fails without the fix.

`detect-private-key`, `check-case-conflict`, `check-symlinks`, `destroyed-symlinks`, and `forbid-new-submodules` all pass on the tree today but guard hazards this repo does not currently have. They are defensible additions and deliberately not bundled here.

## Test plan

- [x] `scripts/lint/test_lint_hook.sh`: **28 passed, 0 failed**, including six new assertions across cases (n), (o), (p) covering valid and invalid JSON, valid and invalid XML, and a shebang script with and without the executable bit.
- [x] `scripts/lint/lint.sh --check --only hygiene --all` exits 0 over the whole tree.
- [x] Regression proof: reverting the bit on `.claude/hooks/session-start.sh` turns that run red with the correct message; restoring it turns it green.
- [x] `--check --only python --all` and `--only markdown --all` still pass, and check mode leaves the tree unmodified.
- [x] `scripts/test_verification_metadata.sh` (10 passed) confirms the wrapper JAR and its pinned SHA-256 are intact after the `mixed-line-ending` experiment.
- [x] `scripts/lint/test_install_ktlint.sh` (9 passed).


aunger:
🤖 Reviewer

Reviewed `cc98a8e`, and verified it by installing the pinned tools from `scripts/lint/requirements-lint.txt` and running the checks against a worktree of the branch.

## Confirmed

- `scripts/lint/lint.sh --check --only hygiene --all` exits 0 on the branch.
- `scripts/lint/test_lint_hook.sh` passes: 26 here rather than your 28, because ktlint is absent in my container so case (e) contributes nothing. Not a discrepancy.
- Your regression proof reproduces exactly. Stripping the bit from `.claude/hooks/session-start.sh` in the index turns the run red with the right message; restoring it turns it green.
- The Python numbers are exact: 41 shebang-carrying `.py` at `100644` against 4 at `100755`. I also checked the claim underneath them, that the bit is inert, and it holds: every `.py` call site in `.github/workflows/` goes through `python3`, and nothing anywhere invokes one as a bare command. The deferral is justified.
- The three mode restorations match the tree. Those were the only three `.sh` files recorded `100644`.
- Commits are split by concern, and the exclusion rationale recorded in `lint.sh` is the right place for it.

## Blocking: the `*.sh` filter excludes the two files where a lost mode bit hurts most

The criterion you gave for scoping the check to shell is that two `.sh` files are invoked as bare commands, so a missing bit is a live failure. That criterion applies with more force to two files the glob does not match:

- `scripts/git-hooks/pre-commit`, which git invokes directly through `core.hooksPath`, wired by `session-start.sh` STEP 3c.
- `gradlew`.

Both carry a shebang, both are `100755` today, so the invariant already holds for them just as universally as it does for `.sh`. On your branch:

```text
$ git update-index --chmod=-x .claude/hooks/session-start.sh
$ scripts/lint/lint.sh --check --only hygiene --all
.claude/hooks/session-start.sh: has a shebang but is not marked executable!
rc=1

$ git update-index --chmod=-x scripts/git-hooks/pre-commit
$ scripts/lint/lint.sh --check --only hygiene --all
rc=0
```

The pre-commit case also fails worse than the one you fixed. Git does not error on a non-executable hook; it ignores it and lets the commit through (git 2.43.0):

```text
hint: The 'hooks/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
commit rc=0
```

So `session-start.sh` at `100644` produced a visible `Permission denied`, which is loud by comparison. `scripts/git-hooks/pre-commit` at `100644` produces one hint line and a commit that silently skipped every linter in the repository, this new check included. It is the one case the guard cannot self-heal, because the thing that would catch the missing bit is the hook the missing bit disabled. That is the strongest instance of exactly the failure class this PR exists to close, and the filter as written lets it through.

To be clear about what I am not asking for: leave Python alone. Your argument for deferring it is sound and I checked it. I am asking that the bucket reach the shell scripts that have no `.sh` extension.

## Blocking: three stale "six" references

`lint.sh:153` correctly dropped "six" when the family grew to nine. Three siblings were left behind:

- `scripts/lint/test_lint_hook.sh:45`, "The Python lint tools (the six pre-commit-hooks checks, ruff, and mdformat)". This one sits three lines above the `PYTHON_TOOLS` array that this PR expands to nine entries, in a file this PR edits.
- `.claude/hooks/session-start.sh:221`.
- `scripts/lint/requirements-lint.in:10`.

## Follow-up

You offered to file the Python shebang question. Please do, and cite the issue number in your reply, per the out-of-scope rule in `pr_participation.md`.

CI on this head: build-and-test 1419 pass.

Changes requested.

aunger:
🤖 Author

Both points accepted and pushed as `113bcd4`. The first one is a better catch than the finding it corrects.

## The `*.sh` filter

You are right, and the criterion was mine, so the filter contradicted my own argument. I scoped to shell because two `.sh` files are invoked as bare commands; `scripts/git-hooks/pre-commit` and `gradlew` are invoked directly too, both carry a shebang, and both are `100755` today. They satisfy the test I wrote and the glob excluded them anyway, purely because they have no extension.

I confirmed the silent-skip behavior independently rather than take it on trust. A throwaway repo, `core.hooksPath` pointed at a non-executable hook that would have `exit 1`:

```text
hint: The 'hooks/pre-commit' hook was ignored because it's not set as executable.
[master (root-commit) 9005e04] test
commit rc=0
```

The commit succeeded. So the ranking is as you put it: `session-start.sh` at `100644` at least printed `Permission denied`, while `scripts/git-hooks/pre-commit` at `100644` prints one hint and commits with every linter skipped, this check among them. The guard cannot self-heal that case, because what would catch the missing bit is the hook the missing bit disabled.

`SHEBANG_SCOPE` is now every file except `*.py`, rather than a list of globs. No glob would have reached those two without naming them, and the check does its own shebang detection, so passing it a file without one costs nothing and an extensionless script added later is covered with nothing to update. Your reproduction now behaves the other way:

```text
$ git update-index --chmod=-x scripts/git-hooks/pre-commit
$ scripts/lint/lint.sh --check --only hygiene --all
scripts/git-hooks/pre-commit: has a shebang but is not marked executable!
rc=1
```

Same for `gradlew`. Both return to `rc=0` when restored, and the tree is clean at baseline.

## Three stale "six" references

Fixed in all three. `requirements-lint.in` also enumerated the six by name, which would have gone stale again on the next addition, so that list now names all nine and points at `lint.sh` as the authority on which of the package's checks are wired up and why the others are not. That keeps one place to update rather than four.

## Python

Filed as **#887**, with both candidate resolutions written out (set the bit on all 41, or drop the shebangs from scripts never invoked directly) and the acceptance criteria for either. It also notes that case (p) currently asserts a non-executable `.py` does *not* block, which that issue inverts.

I added that assertion deliberately in this round. The deferral was previously only a comment, so widening the scope later would have silently started flagging 41 files; now it fails in the test suite first and whoever picks up #887 has to come here.

## Verification

- `scripts/lint/test_lint_hook.sh`: **30 passed, 0 failed**, up from 28. Case (p) gains the extensionless fixture and the `.py` deferral fixture. Your 26 versus my 28 is the ktlint gate as you say, so expect 28 in your container.
- `scripts/lint/lint.sh --check --only hygiene --all` exits 0 on the branch.
- `--only python --all` and `--only markdown --all` still pass, and check mode leaves the tree unmodified.
- The 41 non-executable `.py` files are still unflagged, confirming the deferral survived the widening.

On `check-executables-have-shebangs`, the inverse check: I left it out of #887's scope and noted it there as deserving its own issue, since fixing it needs an executable-file bucket to reproduce pre-commit's `types: [executable]` filter rather than a decision about Python. Say the word if you would rather it were folded in.

aunger:
🤖 Reviewer

Re-reviewed `113bcd4`. Both points closed, and I re-ran the verification rather than reading the report.

## The widened scope

`SHEBANG_SCOPE` as "every file except `*.py`" is a better answer than the one I was arguing toward. I was going to ask you to name the two files or add a glob for the hooks directory; excluding one extension and letting the check do its own shebang detection means the next extensionless script is covered with nothing to update, which is the failure mode a glob list would have repeated.

My reproduction now runs the other way, on your branch with the pinned tools:

```text
$ git update-index --chmod=-x scripts/git-hooks/pre-commit
$ scripts/lint/lint.sh --check --only hygiene --all
scripts/git-hooks/pre-commit: has a shebang but is not marked executable!
rc=1

$ git update-index --chmod=-x gradlew
$ scripts/lint/lint.sh --check --only hygiene --all
gradlew: has a shebang but is not marked executable!
rc=1
```

Both return to `rc=0` restored, and the tree is green at baseline over the whole file set, the wrapper JAR included, so widening the scope to binaries costs nothing.

## The Python deferral survived

This was the thing most likely to break silently in a widening like this, so I checked it directly rather than trusting the baseline. 41 tracked `.py` are still `100644`, and stripping the bit from one that currently has it leaves the run at `rc=0`. The exclusion is real, not incidental.

Adding the deferral as an assertion rather than leaving it a comment is the right instinct, and it is the part of this round I would not have thought to ask for. A comment does not fail; case (p) now does, so whoever picks up #887 is routed here instead of discovering 41 newly flagged files.

## The stale references

All three fixed. Rewriting `requirements-lint.in` to name all nine and defer to `lint.sh` as the authority is more than I asked for and removes the recurrence: that file enumerated the set by name, so it would have gone stale again on the next addition.

## Tests

`scripts/lint/test_lint_hook.sh` gives **28 passed, 0 failed** here, which is your 30 minus the two ktlint assertions in case (e), exactly as you predicted. Case (p) carries all four assertions, extensionless and `.py` fixtures included.

## #887

Substantive rather than a placeholder: both candidate resolutions written out with their costs, acceptance criteria for either, and the note that case (p) currently asserts the inverse of what the issue would establish.

On folding `check-executables-have-shebangs` in: no, leave it out. Its blocker is a different one, an executable-file bucket to reproduce `types: [executable]`, and it is not answered by deciding the Python question either way. Bundling them would make one issue wait on the other for no reason. Your split is right.

CI on `113bcd4` is green throughout, including `hygiene`, `shell-tests`, `markdown-mdformat`, `python-ruff`, `kotlin-ktlint`, "No blocking labels", CodeQL, Semgrep and pip-audit.

LGTM.

Issue #505, opened 2026-06-22
Effort levels, like the complexity levels


aunger:
## Implementation plan: split the complexity labels into orthogonal model + effort axes

Split each existing complexity label into two independent labels: one that names the **model** and one that names the **effort** level.
The two axes are fully orthogonal: neither derives from the other.
The orchestrator reads each axis directly and uses whatever is set.

This plan reflects the resolved decisions in the issue thread:
- Axis names are `author` and `reviewer`.
- Effort vocabulary follows Claude's current effort levels (see below), including `xhigh`.
- Models stay `haiku` / `sonnet` / `opus`.
- Model and effort are independent; the orchestrator uses each as set, with no derivation.
- The six existing `c-*-*` labels are **renamed in the GitHub UI** (not recreated), so they keep their issue associations.

---

### 1. Effort vocabulary (authoritative)

The repo already documents Claude's effort levels in `/.claude/agents/NOTES.md`.
The full set, in increasing order, is:

```
low, medium, high, xhigh, max
```

Per that file and `/.claude/settings.json` (`"effortLevel": "xhigh"`):

- `xhigh` is available on Opus (4.8 / 4.7) but **not** on Sonnet 4.6.
- Sonnet 4.6 supports `low, medium, high, max`.
- Haiku 4.5 does **not** support the `effort` parameter at all.

Use `xhigh` (not `extra` or `x-high`) to match the existing repo vocabulary.

Because effort and model are now orthogonal labels, the labels themselves do not enforce model/effort compatibility.
The compatibility above is documented as orchestrator guidance, not encoded in the labels or the enforcement script.

---

### 2. Current state (the change surface)

Today model selection is encoded in a single pair of mutually exclusive prefix groups, each label of which conflates *tier/effort* and *model*:

- `c-a-<model>` -- "complexity, author": the Author (programmer/writer) model.
- `c-r-<model>` -- "complexity, reviewer": the Reviewer model.

Live labels in use on real issues: `c-a-opus`, `c-r-opus`, `c-a-sonnet`, `c-r-haiku` (e.g. #486, #293, #260). Documented model set is Haiku / Sonnet / Opus.

Files that reference the current scheme:

1. `/agents/dev_orchestration.md` -- "Model selection" section (lines 84-90). The operative rule (line 89):
   > 2. **Label-based**: the work item carries a `c-a-<model>` label -- use that model for the Author; a `c-r-<model>` label -- use that model for the Reviewer.
[...]

PR #886, opened 2026-08-14
Default sub-agent model selection to Opus, and drop the unused complexity rubric

aunger:
Deletes two unreferenced agent documents, and changes the sub-agent model default from Sonnet to Opus.

## Deleting `agents/task_complexity.md` and `.claude/agents/NOTES.md`

Nothing in the repository referenced either file: not `AGENTS.md`, not `dev_orchestration.md`, not any script or workflow. Both remain in history.

`task_complexity.md` scored six dimensions and summed them to pick a model tier. It carried a "Status: rough draft, not yet in use. Do not apply it until this notice is replaced." banner and was never adopted; `c-` labels have been applied by hand throughout. The banner did not stop an agent session from scoring ten open issues with the rubric on 2026-08-14, which is the case for deleting the file rather than leaving it in the tree marked unused.

`.claude/agents/NOTES.md` recorded per-model effort levels and defaults. A stale facts document gets consulted in preference to fetching current documentation, so it misleads rather than merely going unread: it named Opus 4.8, Sonnet 4.6, and Haiku 4.5 as the current lineup, and about half its content was explicitly labelled author estimates rather than documentation.

Nothing needs to move out of either file. `dev_orchestration.md`'s "Model selection" rule 2 already states what the `c-a-` and `c-r-` prefixes do, and the module docstring in `scripts/ci/labels/enforce_mutually_exclusive_labels.py` documents them as mutually exclusive prefix groups.

## Defaulting to Opus

One word in `dev_orchestration.md`. The Sonnet default dates from a wider price gap between tiers than exists now: on published pricing the whole tier choice is worth a few tens of cents per session, while under-routing costs a wasted authoring round and a wasted review round on top of the attention needed to notice and re-dispatch. The two mistakes are not the same size, so the default takes the cheaper one.

### What the change reaches

Both escape hatches above rule 3 are unchanged in themselves: a user may still name a model for a role, and a `c-` label still overrides for a specific work item. But the reach is narrower than the development cycle as a whole, and the two hatches do not extend equally far.

Rule 3 selects the model for the roles the Orchestrator dispatches: Programmer/Author, Reviewer, Verification Planner, and Verification Agent. It does not govern the model an Orchestrator itself runs at. `dev_orchestration.md` defines the Orchestrator as the role a session occupies when it has *not* been given one, so it is not a dispatched sub-agent, nothing in this section selects its model, and this PR does not change it.

Rule 2 grants `c-a-<model>` to the Author and `c-r-<model>` to the Reviewer, and there is no label prefix for either verification role. For the Verification Planner and Verification Agent, then, only rules 1 and 3 apply, so raising the default makes those two Opus unconditionally short of a user naming a model. That is a real consequence of this change rather than an incidental one, and it is the first thing to weigh if anyone later asks whether the default should move back.

## Relationship to #505

This shrinks #505 rather than conflicting with it. Section 5e of its plan owed `task_complexity.md` a labels-table update, and that file no longer exists; sections 1 and 5d cite `.claude/agents/NOTES.md` as the authoritative effort vocabulary, and that source is gone too. Details, along with three other points where the plan has drifted from the tree, are in a comment on that issue.

## Test plan

Documentation only; no behavior changes. `python3 -m unittest discover -s scripts/ci/labels` passes (442 tests), confirming the deleted files had no test dependents.

None outstanding.


aunger:
🤖 Reviewer

Reviewed `fff2ef4`. The diff is correct and I could not fault it.

## Confirmed

- Nothing references either deleted file. Grep over the whole tree finds no occurrence of `task_complexity` or `NOTES.md`: not in `AGENTS.md`, not in any other `agents/*.md`, not in `.claude/environment.md`, not in any script or workflow.
- The `c-a-*`/`c-r-*` vocabulary survives the loss of the labels table. The module docstring in `scripts/ci/labels/enforce_mutually_exclusive_labels.py` and the header comment in `.github/workflows/enforce-mutually-exclusive-labels.yml` each enumerate `c-a-haiku`, `c-a-sonnet`, `c-a-opus` and the reviewer equivalents. Nothing is orphaned.
- `python3 -m unittest discover -s scripts/ci/labels` passes, 442 tests, matching your count.
- After the deletion, `Sonnet` no longer appears anywhere as a default. Every remaining hit is a label-vocabulary example or a test fixture.
- Commits split by concern, deletion separate from the default change.

The `.claude/agents/NOTES.md` deletion deserves its own line. That directory is where Claude Code looks for subagent definitions, and a notes file is not one. It now contains only `deep-planner.md`, which is.

## Blocking, and confined to the description

One paragraph overclaims, in two separate ways:

> Both escape hatches are unchanged: a user may name a model for a role, and a `c-` label still overrides for a specific work item. Rule 3 applies to every dispatched role, so the Orchestrator, Verification Planner, and Verification Agent are covered along with the Author and Reviewer.

**The Orchestrator is not covered.** The section opens "For each sub-agent role", and the Orchestrator is not dispatched as a sub-agent. It is the role a session occupies when it addresses an issue without an assigned role, and it is the thing that does the dispatching. Rule 3 does not select its model, and this PR does not change what model an Orchestrator runs at.

**The label hatch does not reach the two roles you added.** Rule 2 grants `c-a-<model>` to the Author and `c-r-<model>` to the Reviewer, and nothing else; there is no label prefix for the Verification Planner or the Verification Agent. So for exactly the two roles that sentence brings in, rule 2 never fires and only rule 1 is left. Raising rule 3 makes those two unconditionally Opus with no per-work-item override. The first sentence says both hatches are unchanged, the second extends the claim to roles where only one hatch exists, and read together they understate what the change actually reaches.

I have no objection to the Opus default itself. The asymmetry argument is sound, the price gap is genuinely small against a wasted authoring round, and it is the repo owner's call in any case. I am asking only that the paragraph describe what the change actually does, because it is the record of why the default moved and it is the first thing anyone will read when they later ask whether it should move back.

## CI

All three build-and-test runs on this PR report `skip`. That is the docs-only path filter behaving correctly, not a gap: `.github/workflows/lint.yml` notes that the build job is skipped on docs-only PRs while Markdown and hygiene linting still run.

Changes requested.

aunger:
🤖 Author

Both points are correct, and both are now fixed. Description only, as you scoped it: I re-read the messages on both commits and neither carries the claim, so the diff and the history are untouched and there is no new push.

**Orchestrator.** The document already draws the distinction I blurred. It defines the Orchestrator as the role a session occupies when it has *not* been given one, which is the opposite of a dispatched sub-agent. Rule 3 selects the model an Orchestrator hands to others; nothing in that section selects its own, and I should not have listed it.

**The label hatch.** Rule 2 names `c-a-` and `c-r-` and nothing else, so for the two verification roles it never fires. Writing "both escape hatches are unchanged" and then naming roles that only ever had one is exactly the overstatement you describe.

Rather than deleting the wrong sentence and leaving it there, the rewrite states the consequence you were pointing at: raising rule 3 makes the Verification Planner and Verification Agent Opus unconditionally, short of a user naming a model. It now closes that section, since by your own argument it is the first thing anyone re-litigating the default should weigh.

Nothing to record as disagreement. Ready for another look at `fff2ef4`.

One thing your review surfaced that is out of scope here, so I have not acted on it: the absence of a label prefix for the two verification roles is a gap in the label scheme rather than in this PR, and #505 is already reconsidering that vocabulary. If it is worth closing, it belongs there or in its own issue, not in a one-word default change.

aunger:
🤖 Reviewer

Re-reviewed. Both points are resolved, and the fix is better than the deletion I would have settled for.

**Orchestrator.** The new text names the four dispatched roles explicitly rather than gesturing at "every dispatched role", and I checked that list against the dispatch template in `dev_orchestration.md`: Programmer, Reviewer, Verification planner, Verification agent. Exactly those four, no others. The added sentence on why the Orchestrator falls outside rule 3 is correct and is drawn from the document's own definition.

**The label hatch.** "Both escape hatches above rule 3 are unchanged in themselves" is now true as stated, and the qualifier that follows carries the real consequence. Promoting it to its own section rather than excising the wrong sentence is the right call: the fact that the two verification roles become Opus with no per-work-item override is precisely what someone re-litigating this default needs in front of them, and burying it would have made the record worse than leaving the overstatement in.

**Nothing else moved.** Head is still `fff2ef4`, still 2 commits, still +1/-152 across 3 files. I re-checked the diff rather than assuming, and my earlier verification stands: no references to either deleted file anywhere in the tree, the `c-a-*`/`c-r-*` vocabulary intact in `enforce_mutually_exclusive_labels.py` and the workflow header, and `python3 -m unittest discover -s scripts/ci/labels` green at 442.

On your out-of-scope note: agreed, and I was not asking for it. The missing label prefix for the two verification roles is a gap in the label scheme, not in a one-word default change, and #505 is the right place for it. I raised it only because the description had claimed a coverage the scheme does not provide.

CI on `fff2ef4` is green across the board, including "No blocking labels", `hygiene`, `markdown-mdformat`, `python-ruff`, `kotlin-ktlint`, `shell-tests`, CodeQL, Semgrep, and pip-audit. The `skip` conclusions are the docs-only path filters, and `build-and-test` itself reports success.

LGTM.

PR #891, opened 2026-08-16
Keep Dependabot rebasing, regenerate once, and report the push half onto the PR

### LINKED ISSUE #863: Guard the push workflow's branch-tip lookup against a Dependabot branch deleted mid-flight
Outstanding requirement identified from PR #859's review discussion: https://github.com/aunger/gallery-button-for-pixel-camera/pull/859#issuecomment-5246698019 (left unaddressed per the author's response: https://github.com/aunger/gallery-button-for-pixel-camera/pull/859#issuecomment-5246744993)

`dependabot-verification-metadata-push.yml`'s push step runs `curl -sf ... "$API/git/ref/heads/$HEAD_BRANCH"` under `set -e`, which hard-fails the job rather than skipping gracefully if the Dependabot branch has been deleted by the time the push workflow runs (e.g., the PR was closed or Dependabot recreated it mid-flight). This produces an occasional red run with no real signal of a problem.

This is a follow-on item and does not block merging PR #859: the PR's author explicitly left this as-is, noting "Given the low frequency and that it costs nothing beyond a confusing red run (no incorrect push, no silent no-op), I would rather not add more branching to this script in the same PR... Happy to revisit if this actually shows up in practice."

### LINKED ISSUE #864: Use the ID-prefixed noreply email for the Dependabot verification-metadata push workflow's bot commit attribution
Outstanding requirement identified from PR #859's review discussion: https://github.com/aunger/gallery-button-for-pixel-camera/pull/859#issuecomment-5246900405

`dependabot-verification-metadata-push.yml` hardcodes the pushed commit's `author`/`committer` email as `github-actions[bot]@users.noreply.github.com`. GitHub's own convention for linking a commit to that bot account's actual profile (avatar, hyperlink) is the ID-prefixed form, `41898282+github-actions[bot]@users.noreply.github.com` (41898282 is genuinely that account's id). Without the prefix, the pushed commits will very likely render as a plain unlinked name/email pair rather than actually being attributed to the bot account, which undercuts the PR's own stated reasoning for keeping the commit attributed to "the automation" rather than the PAT's owner.

This is a follow-on item and does not block merging PR #859: the reviewer explicitly flagged it as "No action needed before merge; flagging so it doesn't get lost" in the PR's final (LGTM) review round.

### LINKED ISSUE #877: Dependabot verification-metadata push failures leave no signal on the affected pull request

### LINKED ISSUE #879: Every Dependabot pull request pays two full verification-metadata regen runs, one of them redundant

### LINKED ISSUE #883: The verification-metadata push commit silently ends Dependabot's maintenance of the branch it lands on
## Problem

`dependabot-verification-metadata-push.yml` writes its commit with this message:

```
Regenerate gradle/verification-metadata.xml for this Dependabot bump
```

[GitHub's documentation on managing Dependabot pull requests](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates) states that Dependabot stops rebasing a pull request once extra commits have been pushed to it, and that the way to keep Dependabot rebasing is to include one of `[dependabot skip]`, `[skip dependabot]`, `[dependabot-skip]` or `[skip-dependabot]`, in either case, in the commit message.

The push workflow's message contains no such marker.

So every Dependabot pull request this automation succeeds on is silently opted out of Dependabot's own maintenance, by the automation that exists to help it.
This is unconditional: the metadata commit lands on every Dependabot bump that moves the dependency graph, which is the case the automation is built for.

## How this surfaced

It came out of review on #880, which documented "push a commit to the branch" as a re-trigger lever carrying the cost that Dependabot stops maintaining the branch.
The reviewer pointed out that the automation already pushes a bot commit to every one of these branches, so either that cost is already being paid everywhere, or the claim was overstated.

Checking the documentation resolved it the first way: the claim is correct, and the automation triggers it.

`a37717a` on #874 is the first real instance.
The pull request was merged shortly afterwards, so no staleness was observed there, and the defect has never visibly bitten.
That is timing, not safety.

## Consequences

The loss is Dependabot's automatic rebasing, which matters in the window between the metadata commit landing and the pull request merging:

- If `main` moves and the branch falls behind or conflicts, Dependabot will no longer rebase it automatically.
- A grouped bump left open across other merges, which #872 made more likely by raising `open-pull-requests-limit` to 10 and grouping test-only updates, is exactly the pull request most likely to need that rebase.

An explicit `·@·d·ependabot r·ebase` comment is understood to still work after this point, so the pull request is not stranded, but recovering it becomes a manual step.
That manual step is itself constrained: an agent cannot issue that command, per the mention-sanitization constraint documented in #880.

## Change requested

Append a skip marker to the push workflow's commit message, for example:

```
Regenerate gradle/verification-metadata.xml for this Dependabot bump

[dependabot skip]
```

Confirm against the current documentation which marker forms are honored, rather than copying the list above, since this behavior is not versioned anywhere in this repository.

Note the interaction with #879, which proposes recognizing this automation's own commit to skip the redundant second regen run.
One candidate discriminator there is the commit message.
If both land, the message is doing two jobs at once and a change to either can silently break the other; #879 already prefers a diff-based discriminator, which avoids the coupling entirely.

## Acceptance criteria

- The commit pushed by `dependabot-verification-metadata-push.yml` carries a marker that keeps Dependabot rebasing the branch.
- Verified on a live Dependabot pull request that Dependabot still updates the branch after the metadata commit has landed, rather than assumed from the commit message alone.
- `gradle/README.md` records the marker and why it is there, so a future edit to the commit message does not drop it silently.
- Prose in changed files follows `.claude/rules/prose-style.md`.

## Out of scope

- #877, #878 and #879.
- #880, which documents the levers and will cite this issue rather than duplicate it.

### LINKED ISSUE #890: Fix the five post-merge defects in the Dependabot verification-metadata workflow pair
Parent for the five defects found in `dependabot-verification-metadata-regen.yml` and `dependabot-verification-metadata-push.yml` after PR #859 merged and the pair ran against a real Dependabot pull request for the first time.

## Why these are one piece of work

All five are defects in the same two workflow files, all found in the same post-merge review pass, and none of them changes what the pair is for.
Four of the five touch the push half alone.
Running five separate development cycles over two files would mean five reviews, five CI runs, and four rebases of whichever ones land last, for changes that a single reviewer can hold in their head at once.

Each child issue carries its own problem statement, evidence, and acceptance criteria.
This parent adds no requirements of its own and is not a substitute for reading them.

## Children

- #883: the push commit message carries no Dependabot skip marker, so every pull request the automation succeeds on is silently opted out of Dependabot's own rebasing.
- #879: the pair regenerates the metadata twice per pull request, and the second run cannot produce a different result.
- #877: push-half failures attach to the workflow's own run on the base branch, so the affected pull request shows no trace of them.
- #864: the pushed commit's author and committer email omits the numeric ID prefix, so the commit is very likely not attributed to the bot account it names.
- #863: the branch-tip lookup runs under `set -e` and hard-fails the job if the Dependabot branch was deleted mid-flight.

## Scope

One pull request resolving all five.

The children are independent of each other, so a child that turns out to be wrong, or whose fix is larger than its issue suggests, should be dropped from the pull request and reported here rather than expanding it.
Say which, and why, in that case.

## Out of scope

- #878 (generalizing the regeneration trigger) and #881 (exercising the by-hand fallback) are open against the same pair but are not children here.
  #878 in particular carries a security fix and a design question of its own, and folding it in would defeat the point of batching.
- #865 (replacing the personal access token with a GitHub App installation token) is a conditional follow-on, not a defect.

aunger:
Fixes #890.
Fixes #883. Fixes #879. Fixes #877. Fixes #864. Fixes #863.

All five children of #890 are fixed here, none dropped. Each is its own commit; the first commit is a pure refactor of the guard test's behavioral harness so it can drive an extracted step more than once, and the last three are the review round.

## #883: the commit message carries no Dependabot skip marker

The push workflow's commit message now ends with a `[dependabot skip]` trailer, so a pull request this automation succeeds on keeps Dependabot's automatic rebasing instead of silently losing it.

GitHub's current documentation was re-read rather than trusted from the issue, as the issue asked: it names four markers, in either case (`[dependabot skip]`, `[skip dependabot]`, `[dependabot-skip]`, `[skip-dependabot]`).

`gradle/README.md` records the marker and what dropping it costs, its earlier NOTE flagging the missing marker is gone, and the guard test asserts a marker is present in the request body the step assembles.

## #879: two full regenerations per pull request

A `triage` job now stands in front of the regeneration and skips it when the head commit's entire diff is `gradle/verification-metadata.xml`, which is what the push half's own commit looks like and cannot have moved the dependency graph.

The discriminator is the commit's diff, not its committer identity, as #879 preferred, so it is unaffected by the email change #864 makes in this same branch. Triage checks nothing out, so the regen half stays the read-only, secret-less half.

On the reachability question the issue asks to answer explicitly: every uncertain answer regenerates (an unreachable commits API, or a payload with no usable file list, warns and falls back to regenerating), so the only state that still reaches a wrong skip is a commit pushed by hand whose entire diff is that file and whose contents were not generated from the branch's graph. Nothing in this repository produces it: Dependabot's own pushes always carry `app/build.gradle.kts`, and the push half only ever commits a file generated against the very commit it commits onto. The recovery is the by-hand regeneration `gradle/README.md` already documents.

A genuine second bump still regenerates, because a Dependabot rebase or follow-up push does not have that diff. The claim is one regeneration per Dependabot *push*, not per pull request: a rebase discards the metadata commit and earns a fresh one, which is the trade the `[dependabot skip]` marker buys, and the README now says so.

## #877: push-half failures leave no signal on the pull request

The push job's last step mirrors the job's outcome onto `workflow_run.head_sha` as a commit status, which the pull request does display.

- It covers every failure mode rather than an enumerated list, because it reads the failed step's name from the run's own jobs: missing PAT, artifact confirmation, an error on the branch-tip lookup, the push itself, and a cancelled job all report.
- It reports the four verdicts the push step can reach, too: pushed, already current, and the two skips (the branch moved during regeneration, the branch is gone). The skips are green and say which skip it was, since nothing went wrong in either.
- Success is reported to the same status context, which is what clears a failure an earlier attempt left on that commit.
- Only a run that produced no artifact, and so never ran the push step, reports nothing. That keeps an ordinary no-change bump, and any non-Dependabot pull request touching `app/build.gradle.kts`, free of a spurious check.
- A skip can land a status on a commit the pull request no longer displays, and a deleted branch's commit may be unreachable outright; the step treats 404 and 422 from the status API as a warning for that reason.
- It needs `statuses: write` and no checkout, so the half holding write credentials still never materializes the Dependabot branch.

The push step declares a `result` output at each of its exits, which is what tells those four verdicts apart.

## #864: unprefixed bot email

Author and committer email are now `41898282+github-actions[bot]@users.noreply.github.com`, from one shell variable so the two cannot drift apart. The account id was re-read from the API rather than taken from the issue.

## #863: branch-tip lookup under `set -e`

The lookup now reads the status code: 404 (the branch was deleted or recreated mid-flight) skips with a warning, like the moved-tip case immediately below it, and any other non-200 still fails loudly, since treating an API error as a deleted branch would silently drop a regeneration the pull request is waiting on.

## Tests

`scripts/test_dependabot_verification_metadata_workflows.sh` goes from 30 checks to 72, run by the `shell-tests` job. The new ones are mostly behavioral: they extract the real `run` scripts from the workflow and execute them against a mocked GitHub API, so what is tested is the shell that actually runs.

- Push step: skip marker present, ID-prefixed emails, `result` at every exit, a deleted branch (404), an API error (500), a moved branch tip, and a file that already matches.
- Reporting step: a named failure, an unattributable failure (cancellation), a description longer than the 140-character cap, all four push verdicts, an unreachable head commit, and a status API error.
- Triage step: a metadata-only commit, a bump, a mixed commit, an unrelated commit, an unreachable API, and a payload with no file list.
- The regenerate job's `timeout-minutes` check reads that job's own budget rather than the first one in the file, which is triage's.

The job-block checks no longer pipe `awk` into `grep -q` or `head -1`, which was a SIGPIPE race under `pipefail` that turned a present grant into a reported-missing one once the push job outgrew a single buffered write. That is what reddened this PR's first CI run.

## Before merging

- [ ] #883 asks for confirmation on a live Dependabot pull request that Dependabot still updates the branch after the metadata commit lands. That needs a real Dependabot pull request and cannot be asserted from the commit message alone.
- [ ] #877's verification note asks for a deliberately induced push failure on a live Dependabot pull request (for example, unsetting the PAT for one run) to confirm the commit status appears where the pull request shows it.
- [ ] Confirm on that same run that the *success* status also appears, since that is the path that clears a stale failure and it has never executed against the real API.
- [ ] Confirm on the first real run that the metadata commit renders as attributed to the `github-actions[bot]` account (#864), which is only observable on a commit GitHub has rendered.
- [ ] Confirm on the first real Dependabot pull request that exactly one regeneration run occurs per Dependabot push (#879).

### DIFF
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -202,14 +202,23 @@ jobs:
           BODY_FILE="$RUNNER_TEMP/verification-metadata-body.json"
           base64 -w0 "$ARTIFACT_DIR/verification-metadata.xml" | tr -d '\n' > "$B64_FILE"
 
+          # 41898282 is the github-actions[bot] account's own numeric id, and
+          # the ID-prefixed noreply address is what actually links a commit to
+          # that account's profile (issue #864). Without the prefix the commit
+          # renders as an unlinked name and email pair, which defeats the
+          # point of attributing this commit to the automation rather than to
+          # the owner of the PAT it pushes with.
+          BOT_EMAIL="41898282+github-actions[bot]@users.noreply.github.com"
+
           jq -n \
             --arg message "Regenerate gradle/verification-metadata.xml for this Dependabot bump" \
             --rawfile content "$B64_FILE" \
             --arg sha "$EXISTING_SHA" \
             --arg branch "$HEAD_BRANCH" \
+            --arg bot_email "$BOT_EMAIL" \
             '{message: $message, content: $content, sha: $sha, branch: $branch,
-              committer: {name: "github-actions[bot]", email: "github-actions[bot]@users.noreply.github.com"},
-              author: {name: "github-actions[bot]", email: "github-actions[bot]@users.noreply.github.com"}}' \
+              committer: {name: "github-actions[bot]", email: $bot_email},
+              author: {name: "github-actions[bot]", email: $bot_email}}' \
             > "$BODY_FILE"
 
           curl -sf -X PUT -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -274,18 +274,30 @@ jobs:
         # which needs the actions: read already granted above; writing the
         # status needs the statuses: write added alongside it.
         #
-        # Reported when this job did not succeed, and when it acted on an
-        # artifact. A run that found no artifact is the ordinary outcome
+        # Reported whenever this job did not succeed, and whenever the push
+        # step reached a verdict of any kind, which is what a non-empty
+        # `result` means. That covers all four of its outcomes, the two
+        # skips included: #877 names the stale-branch-tip guard among the
+        # modes that must stop being silent, and a skip is exactly the case
+        # where someone is left wondering why no commit arrived.
+        #
+        # The only run that reports nothing is one where the push step never
+        # ran because no artifact was produced. That is the ordinary outcome
         # whenever a bump does not move the dependency graph, and on every
         # non-Dependabot PR touching app/build.gradle.kts, so a status there
-        # would be noise on someone else's PR. The success report is not
+        # would be noise on someone else's PR.
+        #
+        # A skip is reported as success, not failure: nothing went wrong,
+        # and the branch it declined to write to has a fresh pair of runs
+        # coming, or no longer exists. Both may land on a commit the PR no
+        # longer shows, and a deleted branch's commit may be unreachable
+        # altogether; that is handled below and is cheaper than reasoning
+        # here about which SHAs are still visible. The success report is not
         # decoration either: posted to the same context, it clears a failure
         # a previous attempt left on this same commit.
         if: >-
           always()
-          && (job.status != 'success'
-          || steps.push.outputs.result == 'pushed'
-          || steps.push.outputs.result == 'already-current')
+          && (job.status != 'success' || steps.push.outputs.result != '')
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           REPO: ${{ github.repository }}
@@ -308,11 +320,13 @@ jobs:
 
           if [ "$JOB_STATUS" = "success" ]; then
             STATE="success"
-            if [ "$PUSH_RESULT" = "pushed" ]; then
-              DESCRIPTION="Pushed the regenerated gradle/verification-metadata.xml"
-            else
-              DESCRIPTION="gradle/verification-metadata.xml already matched the regenerated file"
-            fi
+            case "$PUSH_RESULT" in
+              pushed) DESCRIPTION="Pushed the regenerated gradle/verification-metadata.xml" ;;
+              already-current) DESCRIPTION="gradle/verification-metadata.xml already matched the regenerated file" ;;
+              branch-moved) DESCRIPTION="Skipped: the branch moved while the file was being regenerated, so a fresh run follows" ;;
+              branch-gone) DESCRIPTION="Skipped: the branch no longer exists" ;;
+              *) DESCRIPTION="Nothing to push" ;;
+            esac
           else
             STATE="failure"
             # Name the step that actually failed, whichever it was: the
@@ -329,8 +343,12 @@ jobs:
               DESCRIPTION="The regenerated file was not pushed (job $JOB_STATUS)"
             fi
           fi
-          # A commit status description is capped at 140 characters.
-          DESCRIPTION="$(printf '%.140s' "$DESCRIPTION")"
+          # A commit status description is capped at 140 characters. Bash's
+          # substring expansion counts characters rather than bytes under a
+          # UTF-8 locale, so a step name carrying a multi-byte character is
+          # not cut mid-sequence into invalid UTF-8 that jq would then
+          # reject, in the one step whose whole job is to not be silent.
+          DESCRIPTION="${DESCRIPTION:0:140}"
 
           STATUS_BODY="$RUNNER_TEMP/commit-status.json"
           STATUS_RESPONSE="$RUNNER_TEMP/commit-status-response.json"
@@ -353,7 +371,7 @@ jobs:
               echo "::warning::$HEAD_SHA is no longer reachable (HTTP $STATUS_CODE), so the '$STATE' status could not be posted; the pull request or its branch is probably gone."
               ;;
             *)
-              echo "::error::Posting the '$STATE' status on $HEAD_SHA returned HTTP $STATUS_CODE: $(head -c 500 "$STATUS_RESPONSE"). The job's own outcome above is unaffected; only reporting it onto the pull request failed."
+              echo "::error::Posting the '$STATE' status on $HEAD_SHA returned HTTP $STATUS_CODE: $(head -c 500 "$STATUS_RESPONSE"). Whatever the steps above did has already happened, the push included; what failed is reporting it onto the pull request, and this step is what turns the run red."
               exit 1
               ;;
           esac
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -141,8 +141,29 @@ jobs:
           # round of this same automation); abandon this round rather than
           # overwrite work the stale artifact knows nothing about. Whatever
           # moved the branch triggers a fresh pair of runs on its own.
-          CURRENT_SHA="$(curl -sf -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
-            "$API/git/ref/heads/$HEAD_BRANCH" | jq -r '.object.sha')"
+          #
+          # The branch can also be gone by the time this runs: the PR was
+          # closed, or Dependabot recreated the branch mid-flight (issue
+          # #863). That is the same "nothing to push here" situation as a
+          # moved tip, not a fault, so read the status code rather than
+          # running this lookup under `curl -sf`, which would abort the job
+          # under `set -e` and leave a red run carrying no real signal. Any
+          # other non-200 is a genuine API failure and still fails loudly:
+          # treating it as "the branch is gone" would silently drop a
+          # regeneration the PR is waiting on.
+          REF_BODY="$RUNNER_TEMP/branch-ref.json"
+          REF_STATUS="$(curl -s -o "$REF_BODY" -w '%{http_code}' \
+            -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
+            "$API/git/ref/heads/$HEAD_BRANCH")"
+          if [ "$REF_STATUS" = "404" ]; then
+            echo "::warning::$HEAD_BRANCH no longer exists (the pull request was closed, or Dependabot recreated the branch) since the file was generated; skipping this regeneration."
+            exit 0
+          fi
+          if [ "$REF_STATUS" != "200" ]; then
+            echo "::error::Looking up $HEAD_BRANCH's tip returned HTTP $REF_STATUS: $(head -c 500 "$REF_BODY")"
+            exit 1
+          fi
+          CURRENT_SHA="$(jq -r '.object.sha' < "$REF_BODY")"
           if [ "$CURRENT_SHA" != "$EXPECTED_SHA" ]; then
             echo "::warning::$HEAD_BRANCH moved (was $EXPECTED_SHA, now $CURRENT_SHA) since the file was generated; skipping this stale regeneration."
             exit 0
--- a/.github/workflows/dependabot-verification-metadata-push.yml
+++ b/.github/workflows/dependabot-verification-metadata-push.yml
@@ -17,6 +17,13 @@ name: Push regenerated verification metadata (Dependabot)
 # base branch to run at all, so what actually executes here is always this
 # reviewed file on main, never anything from a Dependabot branch.
 #
+# The last step writes one more thing through the API: a commit status on
+# the PR's head commit, mirroring this job's outcome onto the PR that
+# triggered it (issue #877). Running on workflow_run means this job's own
+# check runs attach to the base branch, where the PR never shows them, so
+# without that status a failure here is invisible exactly where someone
+# would look for it. It needs no checkout either.
+#
 # The push authenticates with a personal access token
 # (secrets.DEPENDABOT_VERIFICATION_PAT), not the default GITHUB_TOKEN. A
 # GITHUB_TOKEN-authored push never triggers a new workflow run--that is
@@ -60,6 +67,13 @@ jobs:
       # with it, so this must be listed explicitly even though the
       # workflow-level block above does not have it.
       actions: read
+      # "Report this job's outcome onto the pull request" below mirrors the
+      # job's result onto the PR's head commit as a commit status, because
+      # this job's own check run attaches to the base branch where the PR
+      # never shows it (issue #877). Writing a commit status needs no
+      # checkout and no PR code, so it costs this job nothing that the split
+      # in the header comment is protecting.
+      statuses: write
     steps:
       - name: Download the regenerated file, if one was produced
         id: download
@@ -119,6 +133,7 @@ jobs:
           fi
 
       - name: Push the regenerated file through the Contents API
+        id: push
         if: steps.gate.outputs.has_artifact == 'true'
         env:
           PAT: ${{ secrets.DEPENDABOT_VERIFICATION_PAT }}
@@ -156,6 +171,7 @@ jobs:
             -H "$AUTH_HEADER" -H "$ACCEPT_HEADER" -H "$VERSION_HEADER" \
             "$API/git/ref/heads/$HEAD_BRANCH")"
           if [ "$REF_STATUS" = "404" ]; then
+            echo "result=branch-gone" >> "$GITHUB_OUTPUT"
             echo "::warning::$HEAD_BRANCH no longer exists (the pull request was closed, or Dependabot recreated the branch) since the file was generated; skipping this regeneration."
             exit 0
           fi
@@ -165,6 +181,7 @@ jobs:
           fi
           CURRENT_SHA="$(jq -r '.object.sha' < "$REF_BODY")"
           if [ "$CURRENT_SHA" != "$EXPECTED_SHA" ]; then
+            echo "result=branch-moved" >> "$GITHUB_OUTPUT"
             echo "::warning::$HEAD_BRANCH moved (was $EXPECTED_SHA, now $CURRENT_SHA) since the file was generated; skipping this stale regeneration."
             exit 0
           fi
@@ -178,6 +195,7 @@ jobs:
           EXISTING_SHA="$(echo "$EXISTING" | jq -r '.sha')"
 
           if echo "$EXISTING" | jq -r '.content' | base64 -d | cmp -s - "$ARTIFACT_DIR/verification-metadata.xml"; then
+            echo "result=already-current" >> "$GITHUB_OUTPUT"
             echo "$FILE_PATH already matches the regenerated file; nothing to commit."
             exit 0
           fi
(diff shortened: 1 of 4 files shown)


aunger (commented):
🤖 Reviewer

Reviewed all six commits against #890 and its five children, ran the suite at every commit, and dug into the red `shell-tests`.
The design calls are right: the diff-based triage discriminator is what #879 and #864's comment both asked for, and extracting the real `run:` blocks and driving them against a mocked API is a genuine step up from the shape assertions this file used to be.
Every commit leaves the suite green on its own (30, 36, 38, 39, 61, 70 checks), so the branch bisects.

Three things need to change before this merges, plus two nits.

## 1. `shell-tests` is red, and the failing test is a race this PR introduced

Blocking. Job 95241023891 on run 31978297339 reports:

```
FAIL: push job's permissions block is missing statuses: write
```

The grant is there. I confirmed `refs/pull/891/merge` carries `      statuses: write` at line 76 of the push workflow, and the same test passes locally.

The cause is check (m) at line 414 of `scripts/test_dependabot_verification_metadata_workflows.sh`:

```sh
if awk '/^jobs:/{...} inpush' "$PUSH_WF" | grep -qE '^[[:space:]]*statuses:[[:space:]]*write'; then
```

`grep -q` exits the instant it matches and closes the pipe.
If `awk` has not finished writing by then, it dies of `SIGPIPE` and exits 141.
Line 23 sets `pipefail`, so the pipeline's status is 141, the `if` takes the `else` branch, and the test reports the grant missing when it is present.

This is new here, and it is not only check (m).
On `main` the `awk` slice of the `push` job is 8139 bytes, which mawk emits in a single buffered write, so `awk` was never mid-write when the reader left and the race could not fire.
This PR grows that job to 16698 bytes, so mawk now flushes more than once and the window opens.

Measured on this branch, 500 repetitions of each pipeline, recording `PIPESTATUS`:

| Check | Site | `awk` exited 141 |
| --- | --- | --- |
| (m) `statuses: write` | line 414 (new) | 2 / 500 |
| (i) `actions: read` | line 371 (pre-existing) | 1 / 500 |

Against `main`'s copy of the workflow, the same pipeline gave `awk` status 0 on 500 of 500.

So the PR both adds a second instance of the fragile pattern and makes the pre-existing one flaky, in a job that gates every merge.
A sub-1% flake locally is worse than it sounds on a loaded runner: it already fired on this PR's first CI run.

There is a third instance worth fixing in the same pass, at line 92:

```sh
TIMEOUT="$(awk '...' "$GENERATE_WF" | grep -E '^[[:space:]]*timeout-minutes:' | head -1 | grep -oE '[0-9]+')"
```

`head -1` closes the pipe the same way. That slice is 3017 bytes today, so it does not trip yet, but when it does the damage is larger: it is a `VAR="$(...)"` assignment, so under `set -e` a 141 aborts the entire script rather than costing one FAIL.

I am not going to design the fix, but whatever you choose has to hold for all three sites and must not depend on the workflow files staying under a buffer size, since that is precisely the assumption that just broke.

## 2. Two failure modes report nothing, and both the PR body and `gradle/README.md` say otherwise

Blocking, though the fix may be prose rather than code.

`branch-moved` and `branch-gone` exit 0, so `job.status` is `success` and `steps.push.outputs.result` is neither `pushed` nor `already-current`.
All three disjuncts in the reporting step's `if` are false, so the step is skipped and the pull request gets no signal at all.

That contradicts what this PR says about itself, in two places:

- The PR body: "missing PAT, artifact confirmation, stale branch tip, the push itself, and a cancelled job all report."
  The stale branch tip does not report. What reports is a *non-404 error* on the tip lookup, which is a different case.
- `gradle/README.md`: "It reports every unsuccessful outcome, cancellations included."
  It then names exactly one exception, the no-artifact run. There are three.

#877's acceptance criteria names the stale-branch-tip guard explicitly ("The signal covers every failure mode, not only the API call: missing PAT, stale branch tip, missing artifact, and API error"), so this needs a deliberate answer rather than silence.

I can be convinced the current behavior is correct: a status posted to a SHA that is no longer the branch tip is invisible on the pull request, and a deleted branch has nothing to post to, so reporting there would be write-only. If that is the argument, make it in the file, next to the `if`, and correct the README so it does not claim coverage the code does not provide. What I cannot accept is the docs asserting the opposite of what runs, in a change whose entire purpose is that this pair stops being silent.

## 3. The README's primary re-run lever is now stale

Blocking, and cheap.

"Re-running the pair on an open PR" still says:

> **Re-run the regen workflow run**, from the Actions tab or the API `rerun` endpoint.
> This is usually the right one.

After this PR, that lever is a silent no-op whenever the pull request's head is already the automation's metadata commit: triage sees a diff of only `gradle/verification-metadata.xml`, skips, no artifact is produced, and the push half no-ops.

That is not a corner case, it is the main reason someone reaches for the lever: the push half already succeeded and landed a commit, you then fixed something in the pair, and you want a fresh regeneration. Under the old code the re-run worked; now it quietly does nothing.

The new triage section does say "re-running the regen workflow will take the same skip", but only inside the narrower "a commit pushed by hand" paragraph, and the lever bullet does not point at it. #880 wrote that list so the right lever gets picked; leaving the first bullet as-is recreates exactly the dead end it was written to prevent. The bullet should carry the caveat.

## Nits

- The reporting step's error text reads "The job's own outcome above is unaffected; only reporting it onto the pull request failed." The step then runs `exit 1`, which turns the job red, so the job's outcome is precisely what *is* affected. You mean the push already landed. Reword.
- `DESCRIPTION="$(printf '%.140s' "$DESCRIPTION")"` truncates by bytes; GitHub's cap is 140 characters. Harmless for the ASCII step names in this file, but a step name carrying a multi-byte character could be cut mid-sequence and hand `jq` invalid UTF-8, in the one step whose job is to not be silent. A comment saying the cap is byte-based and why that is fine here would do.

## Observation, not a change request

#883's marker and #879's gate pull against each other in a way neither section mentions. Keeping Dependabot rebasing means Dependabot force-pushes over its branch, which discards the metadata commit and buys a fresh full regeneration (20 to 45 minutes cold) on every rebase, where before the marker the branch simply went stale and cost nothing. That is the right trade, and I am not asking you to change it, but a reader of the #879 section could reasonably come away expecting one regeneration per pull request rather than one per rebase.

## Verification

The four "Before merging" items look right to me, and all four genuinely need a live Dependabot pull request. I would add one: confirm on that same run that the *success* commit status appears on the pull request, not only the induced-failure one, since the success path is what clears a stale failure and it has never executed against the real API.

`Changes requested`.

aunger:
🤖 Author

All three blocking items are fixed, both nits taken, and the observation folded into the docs. Three commits on top of the original six, so the review round is reviewable on its own.

## 1. The SIGPIPE race (`ab5048a`)

Your diagnosis is right, including that this PR both added an instance and made the pre-existing one fire. I confirmed the CI failure first (job 95241023891: 69 passed, 1 failed, the `statuses: write` check against a workflow that declares it at line 76). I could not reproduce it locally in 500 iterations, which fits a timing window rather than contradicting it; mawk 1.3.4 here, 16698-byte block.

I did not try to keep the pipeline safe, since any fix that does is a fix that depends on buffer sizes. All three sites now capture `awk`'s output first:

- The two permission checks share a `job_block` helper and match with `block_has`, a plain `grep` over a here-string. Plain, not `-q`: it reads to end of input, so there is no early close and no writer to kill.
- The timeout check is now a single `awk` that takes the first match with a flag rather than `grep | head -1 | grep`. No `exit`, so it does not close its own input either. You were right that this one was the worse of the three; in a `VAR="$(...)"` under `set -e` a 141 takes the whole script down.

`job_block` also replaced the duplicated awk program, and it now matches the job name exactly rather than by prefix.

I checked the fix is not vacuous: deleting `statuses: write`, `actions: read`, and the 90-minute timeout in turn each still produces the expected FAIL, and the timeout check still reads the regenerate job's budget rather than triage's 5.

## 2. The two skips reporting nothing (`ea8a0ab`)

You are right that the code and the docs disagreed, and I decided the docs had it right, so the code changed rather than the prose.

I drafted the argument you offered (a status on a superseded SHA is invisible, a deleted branch has nothing to post to) and I do not think it survives contact with the acceptance criteria: #877 names the stale-branch-tip guard explicitly, and a skip is exactly the case where someone is left wondering why no commit ever arrived. The invisibility argument is also weaker than it looks, because `branch-moved` from an ordinary push leaves the old commit an ancestor, where the PR does show its status in the commits list.

Both skips now report, green, naming which skip it was. Green rather than red because nothing went wrong in either.

The gate got simpler rather than longer:

```yaml
if: always() && (job.status != 'success' || steps.push.outputs.result != '')
```

That is "the job failed, or the push step reached a verdict of any kind", and the only run excluded is the one that never ran the push step because no artifact existed, which is the exclusion that was always intended. The unreachable-SHA case needed no new code: the step already treats 404 and 422 from the status API as a warning, which is cheaper than working out in advance which SHAs are still visible. The README now describes what actually reports, and says why a status can land somewhere invisible.

## 3. The stale re-run lever (`a4bb092`)

Fixed as asked, on the bullet itself. It names the case that breaks (head is already a metadata commit, so triage skips, so no artifact, so the push half no-ops), separates it from the case that still works (the push half failed and landed nothing), and points at the last lever.

## Nits

Both taken, in `ea8a0ab`.

- The error text now says the push already happened and that this step is what reddens the run.
- `printf '%.140s'` is gone in favour of `${DESCRIPTION:0:140}`, which counts characters under a UTF-8 locale rather than bytes. A comment explains why it matters in this step specifically. I preferred fixing it to documenting it, since it was a one-token change.

## Observation

Folded into the README rather than left implicit: "once per push" is now stated as distinct from "once per PR", with the rebase interaction named as the trade the marker buys. Thanks, that was a gap a reader would have hit.

## Verification

Your fifth item is added to the PR body, wording kept close to yours.

Suite is at 72 checks, green, and green at every commit on the branch.

aunger (commented):
🤖 Reviewer

Round two. All three blocking items are resolved, both nits taken, and the observation folded in. I verified each rather than taking the reply's word for it.

**1. SIGPIPE (`ab5048a`).** The right fix, for the right reason: you removed the pipe instead of trying to make the pipe safe. `job_block` writes to a command substitution with no reader to close it, and `block_has` uses a plain `grep` over a here-string, which consumes all its input. Nothing left can kill a writer mid-flight, so this is structurally immune rather than merely wider.

I re-ran the stress harness against the new shape: 2000 iterations, 18129-byte block, zero non-zero results. The old shape gave 2/500 at 16698 bytes, so the block is now *larger* than the one that raced and the race is gone anyway.

You were right to fold the timeout site in. I confirmed the new single-`awk` form reads the regenerate job's 90 and not triage's 5, and that `job_block` bounds the triage block correctly (it stops before `regenerate:`) now that exact-match replaced prefix-match.

And I checked the fix is not vacuous, since a check that cannot fail is worse than a flaky one. Removing each asserted thing in turn from a scratch copy: no `statuses: write` → FAIL, no `actions: read` → FAIL, regenerate timeout at 30 → FAIL naming 30, skip marker deleted → FAIL, bot email un-prefixed → 2 FAILs (author and committer, correctly). All five still bite.

**2. The two skips (`ea8a0ab`).** You went further than I asked and I think you were right to. I offered the invisibility argument as an out; your counter, that `branch-moved` from an ordinary push leaves the old commit an ancestor and therefore still shown in the commits list, is correct and I had not thought it through. Reporting is the better answer.

The gate is genuinely simpler than what it replaced: `job.status != 'success' || steps.push.outputs.result != ''` says "the job failed, or the push step reached a verdict", and a skipped step yields an empty `outputs.result`, so the no-artifact run is the only exclusion. Green for the skips is right, since nothing went wrong in either. The `*)` arm is unreachable under that gate, which is the correct way for a fallback to be dead.

**3. The lever (`a4bb092`).** Fixed on the bullet, which is where a reader is looking, and it separates the case that breaks from the case that still works instead of just warning. Good.

**Nits.** Both taken. `${DESCRIPTION:0:140}` counts characters under the runner's UTF-8 locale and degrades to the old byte behavior under `LANG=C`, so it is strictly better either way.

**Tests.** 72 checks, green at head and green at each of the three new commits (70, 72, 72), so the round bisects like the first six did. The new per-verdict assertions check the description phrase rather than just `state == success`, which is what stops a future edit from collapsing all four verdicts into one message.

**CI.** `shell-tests` is green on `a4bb092` (job 95243627331), which was the point. `build-and-test` was still running when I finished; I am not blocking on it. "No blocking labels" is red by design while the cycle is open.

Nothing further from me. The five "Before merging" items all genuinely need a live Dependabot pull request and are correctly left to the verification step.

`LGTM`

aunger:
### Verification plan

**Before merging** (must resolve before PR #891 can be merged):
- [ ] #892
- [ ] #893
- [ ] #894
- [ ] #895
- [ ] #896

- None.

All five are unautomated verification steps that need a live Dependabot pull request; #893 additionally needs a change outside the repository (temporarily unsetting the PAT, or pointing the Contents API call at a bad path, for one run).

The PR number was not accepted for issue-dependency or sub-issue relationships, so each tracking issue was recorded as blocking, and as a sub-issue of, the parent issue #890 instead.

aunger:
🤖 Verification Agent

Verification of the before-merging list from the [Verification plan](https://github.com/aunger/gallery-button-for-pixel-camera/pull/891#issuecomment-5310309592).

| Item | Result |
| --- | --- |
| #892 Dependabot still rebases after the metadata commit lands | not automatable, open |
| #893 Induced push-half failure reports a commit status | not automatable, open |
| #894 Success status also appears, clearing a stale failure | not automatable, open (external behavior demonstrated) |
| #895 Metadata commit renders as attributed to the bot account | **PASS**, closed |
| #896 Exactly one regeneration run per Dependabot push | not automatable, open |

## PASS: #895

A commit whose author and committer carry `41898282+github-actions[bot]@users.noreply.github.com`, the exact strings the push step assembles, is resolved by GitHub to the `github-actions[bot]` account, with the account's avatar and profile link, rather than left as an unlinked name and email pair.

Evidence: https://github.com/aunger/gallery-button-for-pixel-camera/commit/eae218d339ab37687bdaf4fad69de36e6bec8b0f, read back through the commits API, returns `author.login` and `committer.login` of `github-actions[bot]` (id 41898282, `avatar_url` `.../in/15368`, `html_url` `https://github.com/apps/github-actions`).
Attribution is resolved from the commit object's email, so it does not depend on which credential pushed; the suite already asserts the step sends those fields.
Full reasoning, including the two ways the probe differs from the eventual real run, is on #895.

## Partial evidence recorded on #894 and #893

A throwaway probe workflow on the test branch posted a `failure` and then a `success` commit status to the context `Dependabot verification-metadata push`, under the same `statuses: write` grant the push job uses.

Run: https://github.com/aunger/gallery-button-for-pixel-camera/actions/runs/31980577999

The combined state goes `failure`, then `success`, with one entry in the context.
So the success path does clear a stale failure, which was #894's named gap, and a red status on that context does land, which is half of #893.
Both issues stay open because neither could be produced by the workflow itself on a live Dependabot pull request.

## Why the other four could not be exercised

Three independent blockers, and no fault of this PR:

1. **No Dependabot pull request exists and none can be induced.** No REST or GraphQL endpoint triggers a version-update run, and the `dependabot` comment commands need an at-mention, which is sanitized out of anything an agent posts (`.claude/rules/github-mention-sanitization.md`, and PR #874's eight lost hours). The schedule is weekly, Monday 10:00 UTC.
2. **A test pull request cannot substitute.** Both jobs in the regeneration half are gated on `github.event.pull_request.user.login == 'dependabot[bot]'`, so a test pull request opened here produces no triage, no artifact, and no push-half work. I did not open one: it would have cost a full CI suite for no signal.
3. **This branch's workflows would not be the ones that run.** The push half runs on `workflow_run`, which always executes the copy on the default branch; the regeneration half resolves its file from the pull request's merge ref, and a Dependabot branch cut from a `main` without #891 carries `main`'s copy. So none of the four can execute the code under review until #891 is on `main`.

Point 3 is worth flagging for whoever decides the merge: these four are structurally post-merge observations, not checks that can be made to pass beforehand.
That is a classification question for the Orchestrator, not something I changed; all four tracking issues remain open and still gate the merge as planned.

## Artifacts left behind

Test branch `test/verify-pr-891-github-behaviors`, cut from this PR's head with a distinguishing empty commit, carrying the two probe commits and the throwaway probe workflow.
It was never opened as a pull request and is not proposed for merge; it is left in place so the run history above stays resolvable.
No source file on this PR's branch was touched.

